PR #150, opened 2022-06-10
Reduce platform-app queuing to bare minimum, re-implement as counter-factual to Kafka/RabbitMQ

### LINKED ISSUE #142: Execute a job on its cron  and execute jobs via OBAN queue
Implement minimal cron queue and command execution using OBAN

taylordowns2000:
@stuartc , this is not ready for merge but I wanted to talk about the TODO in there, and get your thoughts on the general approach. As it stands it:

1. Runs jobs behind an oban queue
2. Executes cronjobs based on their cron expressions
3. BUT, re: point 2, it just runs 'em... haven't implemented the selection of the correct `run_result` dataclip and want your eyes on how I create an [invocation](https://github.com/OpenFn/Lightning/compare/oban?expand=1#diff-5f15b95df4bb6422dee597117477f1a6f407369f35c78d459f70b7107aa98bb0R37-R42) before going further.

# Next steps

- [x] get @stuartc 's brain on invocations
- [x] implement invocation logic
- [x] write tests

when merged, this closes #142


lib/lightning/jobs/query.ex
```diff
@@ -108,4 +108,17 @@ defmodule Lightning.Jobs.Query do
     initiating_jobs_query(project)
     |> downstream_jobs_query()
   end
+
+  @doc """
+  Returns active jobs with their cron triggers for use in the cron scheduling
+  service.
+  """
+  def for_scheduler do
```
stuartc, line 116: can we name this something less specific, like `enabled_cron_jobs`?

lib/lightning/pipeline.ex
```diff
@@ -11,6 +16,15 @@ defmodule Lightning.Pipeline do
   alias Lightning.Repo
   import Ecto.Query, only: [select: 3]
 
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"event_id" => id, "run_id" => _id}}) do
```
stuartc, line 20: I think we can just drop the `run_id` from this function, only really need the event.

stuartc (commented):
Everything looks ok for now, but I don't see any changes relating to the concerns you brought up. Does everything work as expected? I know I was going on about dataclips making their way into the next cron job (re: my bigger comment above).

PR #151, opened 2022-06-10
API for jobs, runs and projects

stuartc:
JWT Generation added to `Accounts`


config/config.exs
```diff
@@ -21,6 +21,8 @@ config :lightning, LightningWeb.Endpoint,
   pubsub_server: Lightning.PubSub,
   live_view: [signing_salt: "EfrmuOUr"]
 
+config :joken, default_signer: "secret"
```
taylordowns2000, line 24: should the signer and salt be provided by env?

taylordowns2000 (approved):
@stuartc , this looks good from the perspective of the router — do we need to move the `signing_salt` or `default_signer` into an env variable? otherwise good to go.

PR #164, opened 2022-06-30
Adapt Navbar to the Figma design style

### LINKED ISSUE #162: Navigate the web app using a navbar
Edit the navbar to have workflows, run history, dataclips and project settings

The navbar background colour (closest is grey900) should be referenced as 'primary'. 
When an item is selected/hover it should be 'primary--light'

When a user logs in, they should get taken straight into a project.

Each item in the navbar should have a 'hover' and 'active' state so that a user knows which page they are on. 

If you need to get an icon, you can export them from figma by clicking on them, then clicking 'export' (right hand side bar)

https://www.figma.com/file/ddg6ahCGnq2nPYIurpnoHF/OpenFN-High-Fidelity?node-id=1%3A3464

If there's no project, redirect to an empty project page

(note: I will create a separate issue for the project picker functionality)

elias-ba:
Closes #162


elias-ba:
Screenshot from 2022-06-30 09-07-18

lib/lightning/projects.ex
```diff
@@ -145,6 +145,16 @@ defmodule Lightning.Projects do
     |> Repo.all()
   end
 
+  @spec select_project_for_user(user :: User.t()) :: Project.t()
```
stuartc, line 148: This specs return should be `... :: Project.t() | nil`, since they may not have one.

Perhaps it should be named `first_project_for_user` as well, since the name doesn't reflect the `limit` clause.

assets/yarn.lock
```diff
@@ -0,0 +1,146 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
stuartc, line 1: Please remove this file, we're not using yarn on this project. `npm install` only please :+1:

lib/lightning_web/live/components/icon.ex
```diff
@@ -14,6 +14,119 @@ defmodule LightningWeb.Components.Icon do
     """
   end
 
+  def project(assigns) do
```
stuartc, line 17: Is there a reason you're not using the `.outer_svg` helper with these new functions?

lib/lightning_web/live/first_setup_live/superuser.ex
```diff
@@ -66,7 +66,7 @@ defmodule LightningWeb.FirstSetupLive.Superuser do
     if Lightning.Accounts.has_one_superuser?() do
       socket
       |> put_flash(:warn, "Superuser account already exists.")
-      |> push_redirect(to: Routes.dashboard_index_path(socket, :index))
+      |> push_redirect(to: Routes.dashboard_index_path(socket, :show))
```
stuartc, line 69: I think we should be redirecting to the index by default, and _if_ a user has a project, then redirect again to the show page.

I'm getting the sense that the index page's template is a fallback.

stuartc (changes requested):
Cool, made some comments but yeah could you make that change regarding redirecting.

So if no project is selected or the user arrives at `/`, we send them to the dashboard index, that function checks finds the first project and then redirects to `/projects/:id` (show)... If the first project query returns `nil` we render from index a box that says "Sorry, it looks like you don't have any projects."

I'll take a look at the failing tests now..

stuartc:
@elias-ba I've added a commit to make the pipeline test insensitive to the width of the border that `core` makes.

I assume it was failing on your end (at first) was that you aren't running Node 16.15.0. Please run `asdf install` in the root of the project to ensure you have all the correct versions.

lib/lightning_web/live/components/icon.ex
elias-ba: Nope, no specific reason. Changed it.

assets/yarn.lock
elias-ba: Done

PR #166, opened 2022-07-12
Project picker

### LINKED ISSUE #163: Add a project picker to the navbar
As a user that has several projects, I want to see what project I am currently inside in the navbar and switch to another one, so that I can easily switch projects and still know which context I am in. 

Figma view of the project picker open can be found here: https://www.figma.com/file/ddg6ahCGnq2nPYIurpnoHF/OpenFN-High-Fidelity?node-id=1%3A3464

You can use a Pheonix hook to add a value onto the select item (for example a dataurl) - you can then use the javascript API to trigger a redirect. 

(see project diagram/index.js for an example of how this is done)

elias-ba:
Closes #163

For styling purposes, I really think we should consider this: https://petal.build/components/dropdowns

### DIFF
--- a/lib/lightning_web/templates/layout/live.html.heex
+++ b/lib/lightning_web/templates/layout/live.html.heex
@@ -2,17 +2,35 @@
   <div class="flex flex-row h-full">
     <nav class="basis-48 bg-primary text-white" style="width: 300px">
       <div class="flex flex-col h-full">
-        <div class="w-full h-20 bg-primary flex items-center justify-center">
-          <%= live_redirect(to: Routes.dashboard_index_path(@socket, :index)) do %>
-            <img
-              class="h-10 w-10"
-              src={
-                Routes.static_path(@socket || @conn, "/images/logo-white.svg")
-              }
-              alt="OpenFn"
-            />
-          <% end %>
-        </div>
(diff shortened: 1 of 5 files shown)


lib/lightning_web/templates/layout/live.html.heex
```diff
@@ -2,17 +2,35 @@
...
+            <%= live_redirect(class: "px-4 py-3", to: Routes.project_dashboard_index_path(@socket, :show, @project.id)) do %>
+              <img
+                class="h-6 w-6 inline-block align-middle"
+                src={
+                  Routes.static_path(@socket || @conn, "/images/logo-white.svg")
+                }
+                alt="OpenFn"
+              />
+            <% end %>
+            <select
```
elias-ba, line 16: @taylordowns2000 I will need your help on the stylization of the project picker. Here is the part where the select field is. The major problem I faced on it is when the project name is long, the select field goes below the OpenFn logo (which is not corresponding to the design [here](https://www.figma.com/file/ddg6ahCGnq2nPYIurpnoHF/OpenFN-High-Fidelity?node-id=1%3A3464)). Thanks in advance for helping on that.

taylordowns2000:
Functionality not yet there. Almost

PR #169, opened 2022-07-13
add and edit dataclips

taylordowns2000:
@stuartc , ready for your review — works in testing, but i'd love your eyes before merging as it's got all those dataclip effects... they all need a project_id now.


lib/lightning_web/live/dataclip_live/index.html.heex
```diff
@@ -1,52 +1,61 @@
...
-      <div class="grid grid-cols-10 gap-4" id={"dataclip-#{dataclip.id}"}>
-        <div class="col-span-2">
-          <%= dataclip.inserted_at |> Calendar.strftime("%c") %>
-        </div>
-        <div class="col-span-2"><.type_pill dataclip={dataclip} /></div>
-        <div class="col-span-4"><%= dataclip.body %></div>
+      <%= for dataclip <- @page.entries do %>
+        <.tr id={"dataclip-#{dataclip.id}"}>
+          <.td>
+            <%= dataclip.inserted_at |> Calendar.strftime("%c") %>
```
stuartc, line 29: I don't have a solid alternative right now (that isn't cludgy), but in liveview templates function calls should be avoided: https://hexdocs.pm/phoenix_live_view/assigns-eex.html#change-tracking

test/lightning_web/live/dashboard_live_test.exs
```diff
@@ -24,7 +24,7 @@ defmodule LightningWeb.DashboardLiveTest do
 
     test "User is assigned a project, should redirect", %{
       conn: conn,
-      user: user,
+      user: _user,
```
stuartc, line 27: If we're not using the `:user` here, you can safely remove the match.

test/support/fixtures/invocation_fixtures.ex
```diff
@@ -29,7 +32,8 @@ defmodule Lightning.InvocationFixtures do
         Lightning.ProjectsFixtures.project_fixture().id
       end)
       |> Keyword.put_new_lazy(:dataclip_id, fn ->
-        dataclip_fixture().id
+        # TODO: How do we only pass certain attrs here? Just the project_id
+        dataclip_fixture(attrs).id
```
stuartc, line 36: In this case I would split out the creation of the `project_id`.

So `attrs = attrs |> Keyword.put_new_lazy(:project_id, ...)` and then in the subsequent put new lazys take the `project_id` out of the attrs with `dataclip_fixture(project_id: Keyword.get(:project_id)`

lib/lightning_web/live/dataclip_live/edit.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  @impl true
+  def handle_params(params, _url, socket) do
+    {:noreply, apply_action(socket, socket.assigns.live_action, params)}
+  end
+
+  defp apply_action(socket, :edit, %{"id" => id}) do
+    socket
+    |> assign(:page_title, "Edit Dataclip")
+    |> assign(:dataclip, Invocation.get_dataclip!(id))
```
stuartc, line 28: Just spotted a security hole here, the dataclip fetching needs to be scoped by Project.

stuartc (approved):
Looks good to me, I did add a few comments that can be addressed when you feel necessary.

lib/lightning_web/live/dataclip_live/index.html.heex
taylordowns2000: ok, let's leave for now but i hear you.

PR #174, opened 2022-07-15
add user dropdown for profile/logout to header

### LINKED ISSUE #165: Add user icon to the header
As a user that would like to edit my settings, I would like to access my user settings from any page

Add a user icon to the header. When a user clicks on the icon, they should be taken to the user settings page. 
The current icon that's being used for user settings can be used for now

(from my understanding this means adding a user icon to the left of the main_section header. 

The figma file can be found here: https://www.figma.com/file/ddg6ahCGnq2nPYIurpnoHF/OpenFN-High-Fidelity?node-id=1%3A3304

taylordowns2000:
@stuartc , a few questions for you here:

1. I've had to pass `socket` to the header _everywhere_ and that seems like a bit of a pain. Is there any way we could give this component access to the socket everywhere without explicitly passing it in?
2. What's the lowest cost way to pass the current user into the header? (Similar to question 1 above.) I'd like the user icon to show their initials, and maybe later even their photo instead of the grey unknown avatar we see here.
3. In `lib/lightning_web/templates/user_session/new.html.heex` , where we expect a socket to _not_ exist, I pass `socket={nil}` because when I don't pass any `socket` at all, I get a `key :socket not found` error when my `<%= if @socket do %>` code is reached. Any thoughts on the best way to only _sometimes_ pass a socket to this header component?

image

this closes #165


stuartc (commented):
1. Off the cuff, this _is_ what you have to do... since we require the socket to generate URLs. The way we do it for the menu is by calling `<%= render("menu_items.html", assigns) %>`, which you could argue might be an anti-pattern since we're passing in the entire `assigns`. I'm not sure I know a more elegant or easier way to get URLs into that component than passing in `@socket`.
2. `assigns.socket` or `@socket` only exists in LiveView components. Where as our `user_session` templates are rendered regular phoenix controllers. You might be able to (for that template specifically) pass in `@conn` under the `socket` key. While sockets and conns are pretty different, from a URL generation point of view I think they are compatible.

PR #177, opened 2022-07-15
Project Picker Dropdown using Petal

elias-ba:


lib/lightning_web/templates/layout/live.html.heex
```diff
@@ -12,7 +12,36 @@
...
           <% end %>
+          <%= if assigns[:project] do %>
+            <div class="p-2 my-4 mb-4 text-center text-indigo-300 bg-primary-900">
+              <.dropdown
+                id="project-picker"
+                placement="right"
+                label={@project.name}
+                js_lib="live_view_js"
+              >
+                <%= for project <- Lightning.Projects.get_projects_for_user(@current_user) do %>
```
stuartc, line 23: Please set the project list in the LiveView and not in the template, this causes every render to fetch projects... Set it in the liveview `assign` and access it from projects with something like `@user_projects`.

elias-ba: Thanks @stuartc, is there any place (file) that is global enough for assigning the list of projects in order to access to them no matter the menu item clicked ? @taylordowns2000 any idea ?

elias-ba: @stuartc This is now done. I, now, assign the `projects` in the `hooks.ex` file. Please review again and let me know.

PR #188, opened 2022-07-20
Reintroduce layout components to fix diagram overscroll

stuartc:
All pages need to be migrated to this pattern:

```heex
<Layout.page_content>
  <:header>
    <!-- Same header as before -->
    <Layout.header title={@page_title} socket={@socket} />
  </:header>

  <Layout.centered>
  <!-- Main inner content -->
  </Layout.centered>
</Layout.page_content>
```

### DIFF
--- a/lib/lightning_web.ex
+++ b/lib/lightning_web.ex
@@ -107,6 +107,7 @@ defmodule LightningWeb do
       alias PetalComponents.Heroicons
 
       alias LightningWeb.Components
+      alias Components.Layout
       alias Components.MainSection
       alias Components.Settings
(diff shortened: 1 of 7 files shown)


stuartc:
@elias-ba @taylordowns2000 I've reintroduced the layout changes that I had done to just the diagram, that need to be everywhere.

I've done the diagram and jobs pages (see the diff), but since @elias-ba you're going through the other pages right now I'd like to ask if you could ensure the main liveview pages use these new layout components.

The structure is important as it prevents the inner content from expanding the entire container past the viewport height.

PR #200, opened 2022-07-20
Encrypted body for Credentials

stuartc:
Encrypted `body` column for Credentials.

- Refactored and broke apart the state assember, since it was previously
  relying on iodata straight from the db
- Added a `DEPLOYMENT.md` section to the mix docs, I've only covered the
  encryption stuff - need to band together and add more here


stuartc:
@taylordowns2000 there is a slight downturn in the coverage, specifically some stuff I added to the Vault that moans at you (in a constructive way) when you don't have an encryption key in your env. I could disable coverage checks for this file, but I think it should be covered... I'm just not in the mood to write a GenServer test when this branch is imho good to go, so despite the codecov failure, the coverage everywhere else is either unchanged or up from what it was on main.

taylordowns2000 (commented):
@stuartc , this looks great. two questions to discuss before merging:
1. why the `@moduledoc false` changes? particularly in the case of `lib/lightning_web/live/components/icon.ex`, it seems like we'd want these docs. i think my `mix format` deletes all commented code... forcing you to either use proper elixir docs or have no docs at all.
2. this will feel strange, but i wonder if we should "test" that the strings stored in the table are, in fact, encrypted. could we sidestep vault somehow to show that they're encrypted? or try to decrypt them without the proper key? i wonder how we demonstrate the feature at the business-level (if at all) via tests.

stuartc:
> @stuartc , this looks great. two questions to discuss before merging:
> 
> 1. why the `@moduledoc false` changes? particularly in the case of `lib/lightning_web/live/components/icon.ex`, it seems like we'd want these docs. i think my `mix format` deletes all commented code... forcing you to either use proper elixir docs or have no docs at all.

I was adding pages to our mix docs and the number of modules listed is eyewatering, so by saying `@moduledoc false` we can constrain the publicly documented modules to ones that matter the most... 

We could group all of the icon and layout modules into a grouping called 'Layout and Styling` or something, I dunno. Open to ideas, just don't want Icons sitting in the same level of importance as `Pipeline` in our mix docs.
Are you sure mix format is removing comments, it really shouldn't.

> 2. this will feel strange, but i wonder if we should "test" that the strings stored in the table are, in fact, encrypted. could we sidestep vault somehow to show that they're encrypted? or try to decrypt them without the proper key? i wonder how we demonstrate the feature at the business-level (if at all) via tests.

I mean we can. It's easy demonstrate, since I had to rewrite all our state creation queries because of it! Haha, I'll add a test case.

stuartc:
@taylordowns2000 added a test that does a select but casts the credential body as a string. Also threw in a test that should get our code coverage into the green.

taylordowns2000 (approved):

Issue #208, opened 2022-07-21
Issue with the database URL when deploying with docker


amberrignell:
Current: 
- Run docker command (`docker compose run --rm web mix ecto.migrate`)
- Get error: "Invalid hostPort: 4000}"
- Change url in [docker-compose file](https://github.com/OpenFn/Lightning/blob/main/docker-compose.yml#L47) to hardcode the PORT value
- Run docker command again
- Get error: "invalid url "postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT:-5432}/${POSTGRES_DB}", host is not present"
- Hardcode those values and run command again
- Still getting an error: invalid url "postgresql://hello:password@postgres:5432/lightning_dev", host is not present

Expected: 
Run the docker command, navigate to localhost/4000 and see lightning up and running

I think it might be an issue with parsing environment variables and templating, but I'm confused as to why I still get an error when the values are hardcoded ...

NickGowdy:
Hey @amberrignell 

I had similar issues yesterday following your steps. I then went to the contributing section of the readme and that worked straight away. It's unclear to me which steps should be followed for someone cloning the project for the first time.

taylordowns2000:
@NickGowdy , thanks for posting! I know that @elias-ba is working on fixing the `docker-compose.yaml` and the `README.md` right now. Will ping you when this is sorted out :-)

taylordowns2000:
@NickGowdy , quick question: are you running on an Apple M1? We've got a build [issue](https://github.com/OpenFn/Lightning/issues/119) there related to https://github.com/jayjun/rambo/pull/13 - if so, let me know and I can provide a workaround.

NickGowdy:
> @NickGowdy , quick question: are you running on an Apple M1? We've got a build [issue](https://github.com/OpenFn/Lightning/issues/119) there related to [jayjun/rambo#13](https://github.com/jayjun/rambo/pull/13) - if so, let me know and I can provide a workaround.

No, I'm using one of the older MacBook Pros with an Intel CPU.

taylordowns2000:
OK, so we're caught in two minds and maybe you could let us know your preference @NickGowdy ! 🙏 

Context:
1. Right now, the `Dockerfile` builds an image that's optimized for [Dockerhub](https://hub.docker.com/repository/docker/openfn/lightning) and our Kubernetes deployments.
[...]

PR #213, opened 2022-07-22
A Dockerfile for devs; better README

taylordowns2000:
This proposes a new `README` and a `Dockerfile-dev` for building development docker images. @elias-ba , @stuartc , care to kick the tires and let me know what you think?

This addresses #208 , @NickGowdy , and works properly on my linux machine, but is a little complex in that it involves maintaining two different `Dockerfiles`, one for building the released images to Dockerhub and the other for building developer friendly* images. In the future, maybe we could combine them?

Biggest change is the README... in short:
1. if you just want to **run**, use docker
2. if you want to **deploy**, see the deployment file (work in progress)
3. if you want to **contribute**, set up a DB however you like then run it as a mix app (with IEX)

Thoughts?

*Stu, the dev friendly one is just what you'd set up back in April with a few small version and npm changes.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -29,7 +29,8 @@ Note that for secure deployments, it's recommended to use a combination of
 - `PRIMARY_ENCRYPTION_KEY` - a base64 encoded 32 character long string.
   See [Encryption](#encryption).
 - `ADAPTORS_PATH` - where you store your locally installed adaptors
-- `LIGHTNING_LISTEN_ADDRESS`": "0.0.0.0",
+- `LIGHTNING_LISTEN_ADDRESS`" - the address the web server should bind to,
+  defaults to `0.0.0.0`
 - `LOG_LEVEL` - how noisy you want the logs to be (e.g. `debug`, `info`)
 - `MIX_ENV` - your mix env, likely `prod` for deployment
 - `NODE_ENV` - node env, likely `production` for deployment
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -16,10 +16,26 @@ mix lightning.gen_encryption_key
 0bJ9w+hn4ebQrsCaWXuA9JY49fP9kbHmywGd5K7k+/s=
 ```
 
-Copy your key (NOT THIS ONE) and set it as `PRIMARY_ENCRYPTION_KEY` in your
+Copy your key _(NOT THIS ONE)_ and set it as `PRIMARY_ENCRYPTION_KEY` in your
 environment.
 
 ## Environment Variables
 
-- `PRIMARY_ENCRYPTION_KEY`  
-  Base64 encoded 32 character long string. See [Encryption](#encryption).
+Note that for secure deployments, it's recommended to use a combination of
+`secrets` and `configMaps` to generate secure environment variables.
+
+- `SECRET_KEY_BASE` - a secret key used as a base to generate secrets for
+  encrypting and signing data.
+- `PRIMARY_ENCRYPTION_KEY` - a base64 encoded 32 character long string.
+  See [Encryption](#encryption).
+- `ADAPTORS_PATH` - where you store your locally installed adaptors
+- `LIGHTNING_LISTEN_ADDRESS`": "0.0.0.0",
+- `LOG_LEVEL` - how noisy you want the logs to be (e.g. `debug`, `info`)
+- `MIX_ENV` - your mix env, likely `prod` for deployment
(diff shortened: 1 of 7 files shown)


taylordowns2000:
this to address #208

taylordowns2000:
@amberrignell , particularly interested if this makes your docker setup work. Based on Nick's comment it seems like we've also now inadvertently got a minimum Docker version--his setup only _almost_ worked after upgrading Docker.

elias-ba:
@taylordowns2000 this great and it works fine for me. I love the approach in here and I think its a much simpler solution combining all we were discussing since this morning.

elias-ba (commented):
@taylordowns2000 this great and it works fine for me. I love the approach in here and I think its a much simpler solution combining all we were discussing since this morning.

taylordowns2000:
Thanks @elias-ba ... yeah this was a big team effort and I think we did well consider that neither of us are Docker-native devs! Let's see what Amber, Stu, and Nick think. Have a great weekend 👋

DEPLOYMENT.md
```diff
@@ -2,24 +2,40 @@
...
 
-The key is expected to be a randomized set of bytes, 32 long; and Base64
-encoded when setting the environment variable.
+The key is expected to be a randomized set of bytes, 32 long; and Base64 encoded
+when setting the environment variable.
 
-There is a mix task that can generate keys in the correct shape for use
-as an environment variable:
+There is a mix task that can generate keys in the correct shape for use as an
+environment variable:
```
stuartc, line 12: @taylordowns2000 as an aside, what are you ruler numbers in your editor? I use 80 chars, and so does Elias - no real issue here; just noticing some formatting that didn't change the actual copy.

DEPLOYMENT.md
```diff
@@ -2,24 +2,40 @@
...
-
+Note that for secure deployments, it's recommended to use a combination of
+`secrets` and `configMaps` to generate secure environment variables.
+
+- `SECRET_KEY_BASE` - a secret key used as a base to generate secrets for
+  encrypting and signing data.
+- `PRIMARY_ENCRYPTION_KEY` - a base64 encoded 32 character long string.
+  See [Encryption](#encryption).
+- `ADAPTORS_PATH` - where you store your locally installed adaptors
+- `LIGHTNING_LISTEN_ADDRESS`": "0.0.0.0",
```
stuartc, line 32: Should probably add a description here. Something like:

> the address the web server should bind to, defaults to `0.0.0.0`

Assuming that default is correct..

Dockerfile-dev
```diff
@@ -0,0 +1,117 @@
...
+FROM ${BUILDER_IMAGE} as dev
+
+# install build and dev dependencies
+RUN apt-get update -y && apt-get install -y \
+  build-essential curl git inotify-tools
+
+# isntall nodejs using nodesource
+# https://github.com/nodesource/distributions/blob/master/README.md#deb
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+RUN apt-get install -y nodejs
```
stuartc, line 31: @taylordowns2000 I would suggest changing this to the way we do it for our main Dockerfile: https://github.com/OpenFn/Lightning/blob/main/Dockerfile#L31

We don't use nodesource or deb files, we have a script that installs directly from the official binaries - this allows us to install an exact version based on an environment variable... 

It may need to be tweaked a bit to support arm64..

stuartc (approved):
@taylordowns2000 this looks good, left a few notes that may be of consideration to you... but yeah all good otherwise

DEPLOYMENT.md
taylordowns2000: I _think_ 80, but I've actually been carrying around the `.prettierrc` for a wihle now and forgot to commit to main. Will add shortly. It shouldn't be an individual devs decision :-)

Dockerfile-dev
taylordowns2000: we haven't had arm64 suport anyway because of the rambo dependency but it looks like 14 hours ago that outstanding PR was just merged! i'll head back to my mac in a second and test this out. very exciting.

DEPLOYMENT.md
taylordowns2000: it was `proseWrap` @stuartc ! i think that likely one of your vsCodes wasn't wrapping markdown files cause there's an extra option needed for prettier. have added a very basic `.prettierrc` now:
```
{ "proseWrap": "always", "printWidth": 80 }
```

PR #216, opened 2022-07-25
Differentiate credentials - Add production tag

### LINKED ISSUE #192: Enable users to differentiate between staging and production credentials (2.4)
As a user that collaborates with others on workflows,
I would like to be able to differentiate between staging and production credentials
So that my colleagues and I can more easily identify the credentials we would like and mix them up less often. 

When a user creates a credential, they should have a toggle button that makes them choose whether to mark the credential as staging or production. The toggle button should be set to 'staging' by default. 

- [x] add checkbox toggle to the CredentialForm component
- [x] display "production" on the credential view and list with the `HeroIcon.Outline.exclamation`
- [x] add a boolean ? column to the credentials table for `production`

elias-ba:
Closes #192

### DIFF
--- a/lib/lightning/credentials/credential.ex
+++ b/lib/lightning/credentials/credential.ex
@@ -30,7 +30,7 @@ defmodule Lightning.Credentials.Credential do
   @doc false
   def changeset(credential, attrs) do
     credential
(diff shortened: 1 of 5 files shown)


stuartc (approved):
Good job!

PR #217, opened 2022-07-26
Credential Audit

### LINKED ISSUE #194: Generate an audit trail for credential configuration (2.6)
As a Lightning administrator, 
I would like credential creation and edits to be tracked 
So that I can audit these in the event a credential change breaks a workflow

- [ ] set up papertrail? on the credentials table to know when a credential has been changed and by whom
- [ ] add a tab in admin settings that displays a list of these changes

Note:
1. we should be able to see _that_ the body of a credential was changed and by whom, but not the credential body itself as this is encrypted at rest

stuartc:
**WIP**

- [x] Add Audit module that adds the base functionality to a given schema
- [x] Apply that to a new schema `Credentials.Audit`
- [x] Register a credential being created
- [x] Encrypt the `body` column in the Audit record as well
- [x] Register credentials being changed
- [x] Add audit section to admin area

Closes #194

### DIFF
--- a/lib/lightning/auditing.ex
+++ b/lib/lightning/auditing.ex
@@ -1,0 +1,13 @@
+defmodule Lightning.Auditing do
+  @moduledoc """
+  Context for working with Audit records.
+  """
+
+  import Ecto.Query
+  alias Lightning.Repo
+
+  def list_all(params \\ %{}) do
+    from(a in Lightning.Credentials.Audit, preload: [:actor])
+    |> Repo.paginate(params)
+  end
+end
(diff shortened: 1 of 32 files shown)


stuartc:
@taylordowns2000 could you give this a spin, it should all be working. I'm comfortable with the test coverage for now.

The Audit page could be prettier, but yeah can improve it as we have more...

taylordowns2000 (changes requested):
@stuartc , looks like this is working for all except the modification of `project_credentials` from the credentials screen - I didn't have enough time to build up context, but have added a failing test which I _think_ reproduces the bug as I see it on my local instance.

can you take a look? please start by trying to modify an existing credential, adding or removing a `project_credential` and then check out the failing test to confirm that it accurately captures that behaviour.

PR #221, opened 2022-08-02
Transfer Credential Ownership

elias-ba:
This PR ships code that implements the credential transferring feature. A credential can only be transferred to another user if that user is part of all the projects where the credential is being used. Otherwise a validation error message will pop up and block the process. For now transferring a credential is a matter of editing one and changing the `user_id` field. Later we may decide the best UI for this. I made sure to not display the transferring logic when creating a new credential.

### DIFF
--- a/lib/lightning_web/live/credential_live/form_component.ex
+++ b/lib/lightning_web/live/credential_live/form_component.ex
@@ -22,7 +22,14 @@ defmodule LightningWeb.CredentialLive.FormComponent do
        all_projects: all_projects,
        changeset: changeset,
        available_projects: filter_available_projects(changeset, all_projects),
-       selected_project: ""
+       selected_project: "",
+       users:
+         Enum.map(Lightning.Accounts.list_users(), fn user ->
+           [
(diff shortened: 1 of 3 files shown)


taylordowns2000 (changes requested):
thanks @elias-ba ! a few small behavior changes before the code review:
1. make the dropdown list display the first and last name of the other user, not their ID
2. change the validation error to `Invalid owner: {OWNER NAME} doesn't have access to {LIST_MISSING_PROJECTS}. Please grant them access or select another owner.`
3. grey out the "save" button when the form is invalid
4. implement the commented out test in line 115 of `credential_live_test`, but rename to `test "transfers a credential to a new owner"`

taylordowns2000 (changes requested):
hey @elias-ba , two things:
1. the save button still doesn't disable (see last review)
2. it looks like you lost the standard dropdown style on this last commit (see below)
image

elias-ba:
@taylordowns2000 this PR is now updated for integrating your pre-review change requests.

Issue #193, opened 2022-07-20
Transfer credential ownership (2.5)


amberrignell:
As an owner of credentials that are in use in several projects, 
I would like to transfer ownership to another user
So that they can access and edit credentials themselves if I leave the organisation 

- [ ] add a `transfer_ownership` function to the `credentials.ex` context which takes in a user and updates the `belongs_to` field to that user 
- [ ] add a 'transfer ownership' action to the credential_live next to show/edit/delete which redirects to transfer_credential_ownership.ex
- [ ] add transfer_credential_ownership.ex which displays a tooltip, a dropdown of users to select from (ids ? emails ?). When the user clicks 'transfer ownership' there should be a confirmation popup which says ("Are you sure? If you transfer ownership of this credential you will no longer be able to view or edit it yourself. "]

How do we handle share failures, for example if a user no longer exists (this technically shouldn't happen because the user is selecting from a dropdown of existing users)

amberrignell:
If a user is in one project, and wants to transfer credentials to someone in another project, what happens ? We stop them and the error message says 'this credential is shared with projects that the recipient cannot access - please update settings accordingly'. 
credentials are the only thing in an instance that we are even hiding from the 'admin' who has deployed that database. The user of the credential can pass around or shift to any other user in that system. An error will be thrown.

elias-ba:
@taylordowns2000 @stuartc @amberrignell are you guys sure the function `transfer_ownership` should go in the `credential.ex` module ? Why not in the `credentials.ex` module instead ?

Why do we need to have that function instead of just using the `update` function ?

taylordowns2000:
@elias-ba , great catch:

1. `credential.ex` was a typo, i've modified it in the issue to display `credentials.ex`
2. your new function `transfer_ownership` will _use_ `update` under the hood, but the logic to check if we _can_ transfer a cred to a new recipient based on their might also live in there. (or it might live in the changeset validation.)

elias-ba:
Very clear @taylordowns2000. Thanks

elias-ba:
@stuartc have you seen 

> As an owner of credentials that are in use in several projects, I would like to transfer ownership to another user So that they can access and edit credentials themselves if I leave the organisation
> 
> * [ ]  add a `transfer_ownership` function to the `credentials.ex` context which takes in a user and updates the `belongs_to` field to that user
[...]

Issue #191, opened 2022-07-20
Create credentials through a form (2.3)


amberrignell:
As a user that is not familiar with json, 
I would like to create credentials via a form
So that I know exactly what I need to input

- [ ] Create a bodyschemas constant that has the body schema for all of our adaptors OR find a way to dynamically get the expected credentials for each adaptor (I think depending on our timelines it would be ok to hardcode for now)
- [ ] Create a credential form component that takes in the body schema for an adaptor credential and generates a form with the required inputs.
- [ ] On submit, the form should pass in the inputs to create a credential for the user. 

Edge case: 
if the body schema is a function (ex Google Sheets), use that.

How can we redesign this ? Can we dynamically get the expected inputs for each language package ?

taylordowns2000:
The big question here is whether there is a way to allow users to _either_ create
1. a raw credential where they define keys and values via a simple interface. (this would be RE-USED for dataclips... in this sense, the only differences between credential body and dataclip body is that creds are encrypted and only visible for a single user)
2. a credential that is "linked" to a particular adaptor_version and displays all the necessary keys for that adaptor_version

This would save us having to maintain a list of credential types with bodySchemas.

If we go down the `#2` path, I'd want to consider how we extend existing adaptors with a `validate()` function which _both_ type-checks the provided `state.configuration` and attempts to login to a system with that config. This validate function (or the two functions that comprise it) would be used to define credential body schemas and to click a "test connection" button on a job step.

stuartc:
@taylordowns2000 why not both? These seem mutually exclusive in their feature contribution..

stuartc:
The 'holy grail' solution would be that the adaptors inform the app what schemas are available and the rules to apply, I'm not sure how we would go around that - not without building a fairly sophisticated runtime based form module. It would however completely decouple our FE from individual and custom deployments.

Re: 2. "adaptor or version bound credentials"

I like this, just based on the sense of comfort it may bring. I'm not sure how disconnected we can (reasonably) make our form system from our adaptors. Again keeping "data" like version and adaptor names in our codebase will require ongoing maintenance and redeploys for updates and changes for something that isn't directly related to the main functioning of Lightning.

taylordowns2000:
closed in #223

PR #223, opened 2022-08-05
Dynamic credential schemas using JSONSchema

stuartc:
This adds the first steps involved with dynamically changing how credentials for different adaptors appear.

For now there are two types, "raw" and "dhis2" - raw being the fallback to the existing interface. And DHIS2 being a form dynamically generated based on a (currently hard-coded) JSONSchema.

The schema is used to show the various inputs but also informs validations, currently it can:

- Check that something isn't blank
- Check that a field is present
- Check if something is URI or not
- Show an input as a password input (leaning on the `writeOnly` property in JSONSchema).

A full write up on how we use the schema and how to write schemas for other credential types will need to be done as we get further into this.

The implementation currently is a bit prickly tbh, I'm not really that happy with how the form component handles the two different possible paths (raw vs a json schema).
We will need to take another pass when we start extending it.

But all in all, this is definitely a start.


stuartc:
@taylordowns2000 ironed out the issues you observed, tests passing - please give it another look thanks! :+1:

taylordowns2000 (approved):

taylordowns2000:
closes #191 and closes #193

PR #230, opened 2022-08-10
various ux changes

taylordowns2000:
@stuartc , want to get a few UX changes in before a demo — the preloads _where_ to do the `Enum.map_join` are what I'd most appreciate your thoughts on, but if you're not completely disgusted by this implementation I'd say let's get them in because the UX is significantly enhanced by the visual changes they enable

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -249,7 +249,8 @@ defmodule Lightning.Invocation do
     from(r in Run,
       join: p in assoc(r, :project),
       where: p.id == ^project_id,
-      order_by: [desc: r.inserted_at, desc: r.started_at]
+      order_by: [desc: r.inserted_at, desc: r.started_at],
+      preload: :job
     )
     |> Repo.paginate(params)
   end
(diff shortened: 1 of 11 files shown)


lib/lightning/invocation.ex
```diff
@@ -270,11 +271,17 @@ defmodule Lightning.Invocation do
...
+    from(r in Run, where: r.event_id == ^event_id)
+    |> Repo.one!()
   end
 
   def get_run!(id), do: Repo.get!(Run, id)
 
+  @doc """
+  Fetches a run and preloads the job via the run's event.
+  """
+  def get_run_with_job!(id), do: get_run!(id) |> Repo.preload(:job)
```
stuartc, line 283: I think we should either split out the query from `get_run!` into in it's own function, and then use it in both places.

We want to avoid `Repo.preload` as much as possible as it results in n+1 queries. 

So ideally the `_with_job!` variant is `from(r in run_query, preload: [:job]) |> Repo.one!()` etc.

lib/lightning_web/live/credential_live/index.ex
```diff
@@ -43,5 +43,12 @@ defmodule LightningWeb.CredentialLive.Index do
 
   defp list_credentials(user_id) do
     Credentials.list_credentials_for_user(user_id)
+    |> Enum.map(fn c ->
+      project_names =
+        Map.get(c, :projects, [])
+        |> Enum.map_join(", ", fn p -> p.name end)
```
stuartc, line 49: While I'm not completely grossed out by that :sweat_smile:, I'm assuming all the projects are preloaded here.

Another way to do this would be via a query 

```
from([c, p] in credential_query,
  select_merge: %{
    project_names: fragment(
      "STRING_AGG(?, ', ') AS project_names",
      p.name
    )
  })
```

Something like that.. I don't think the frequency of this specific page warrants a lot of optimization (beyond ensuring projects are preloaded) for now.

stuartc (approved):
Looks good to me... As an aside, I don't know if you're also feeling the tension between our (lack) of queries and lots of functions that return data (vs. a query)... I'm not loving having lots and lots of functions just for reading the same data differently...

PR #254, opened 2022-08-11
Transfer Credential Ownership

elias-ba:
This PR ships code that implements the credential transferring feature. A credential can only be transferred to another user if that user is part of all the projects where the credential is being used. Otherwise a validation error message will pop up and block the process. For now transferring a credential is a matter of editing one and changing the user_id field. Later we may decide the best UI for this. I made sure to not display the transferring logic when creating a new credential.


elias-ba:
Closing to solve codecov

PR #255, opened 2022-08-11
Integrate Monaco code editor

josephjclark:
Integrates Monaco code editor into Lightning, with code assist and language services provided for adaptors that  export typings via d.ts.

## Next steps

* [x] Ensure proper two way sync in the job code with the server 
* [x] Load actual d.ts files
* [x] Experiment with settings some more
* [x] Fix the tooltip
* [x] Export package-fns in kit
* [x] Pass the adapter version properly
* [x]  ~~Hide imports?~~

## Implementation Notes

I've registered a new React component with Pheonix hook to load Monaco into the page. 

A hidden Pheonix form element is rendered inside the Editor component. React changes to the hidden input, which notifies and updates the Live View. Editor changes are only persisted when Save is pressed.

We use `fetchDTSListing` from `@openfn/compiler` to load the d.ts for the selected adaptor and version, then push those into monaco. Each d.ts is wrapped in a `declare module "<name>" { }` statement to ensure the typings match the import.  This seems to work (at least with the typings in common, which I've tested against).

Changing the adaptor and version will wipe the currently loaded typings. So you can load common@2.0.0-rc2 then downgrade to 1.7.1 and the type information will, correctly, be lost (it may take a moment).

## Hiding Imports

Apparently there's an API to hide imports in Monaco, but it doesn't seem to be available to my build.

In any case, IF we want to hide the imports (and the imports question is still open, as far as I'm concerned), we should only do so when we're ready to generate and insert a full import statement automagically. At the moment, the user has to write their own imports to get what they need - and one can't do that if the import statement is hidden!

So it's a bit of a catch-22. One to solve later, in any case.

## Further work

* #279
* #280


stuartc:
This is shaping up nicely! High five for using the delayed loading strategy - we'll probably have to use the LiveComponent events to communicate between the editor and the form component... but yeah this is great progress!

josephjclark:
Monaco is generally excellent but I am not a fan of this:

image

In Monaco, pressing ctrl + space to get code assist will present you a list of everything in the JS global space. Most of what's in there is not just useless, but it obscures the functions that job writers actually want.

I haven't found any good options for dealing with this yet, but I'd be really keen to curate the global scope. Or find a way to pin adaptor functions. I doubt Monaco will be very adaptable in this way though.

assets/package.json
```diff
@@ -9,16 +9,18 @@
...
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
     "react-markdown": "^8.0.3",
     "zustand": "^4.0.0-rc.2"
   },
   "devDependencies": {
     "@types/react": "^18.0.15",
-    "@types/react-dom": "^18.0.6"
+    "@types/react-dom": "^18.0.6",
+    "@openfn/compiler": "file:../../kit/packages/compiler"
```
josephjclark, line 24: Whoops - I didn't mean to check this in! I'll need to reverse it out and maybe update the compiler version number.

It's needed because I've had to expose extra exports in the compiler package.

josephjclark:
@stuartc I think this PR has gone about as far as I'm willing to take it for now - care to give me your thoughts?

taylordowns2000:
@josephjclark , if this is ready, can you click that "ready for review" button? (I marked this as a draft when it first got opened.)

Issue #260, opened 2022-08-12
Rename 'schema' to 'type' in create credential


amberrignell:
On the create credential page, the credential 'schema' should be renamed to 'type' so that it is less confusing to non-technical users. The 'Choose a credential schema' should say 'Choose a credential type'. 

![Group 1](https://user-images.githubusercontent.com/36554605/184351046-cb87ccb2-85c7-41a4-82f0-ec088279a05e.png)

taylordowns2000:
closed by #266

PR #266, opened 2022-08-12
change 'schema' to 'type'

avinashdhauni:


taylordowns2000:
this closes #260

taylordowns2000:
This is great, @avinashdhauni . Thanks so much.

avinashdhauni:
My pleasure. It was amazing contributing to the project. I am at a
beginning stage of programming and i didn't knew how to contribute to the
open source because the codes are so big and i wasnt able to understand.
But something like this was easy enough for me to complete. So I am
very grateful to you for this opportunity.

taylordowns2000:
Great to hear it, @avinashdhauni . What do you think of https://elixir-lang.org/ and https://www.phoenixframework.org/ ? Have you worked with these technologies before?

If you're up for it, I know that @amberrignell , the product manager for Lightning, would love to speak with you next week to hear how you discovered OpenFn, what it was like to get the project set up, and gather any feedback you might have on the application (or even on [OpenFn.org](https://www.openfn.org))

stuartc:
@avinashdhauni fyi I've changed the form a bit, since the column name is `schema` - when changing the html test for a label, it should be done in a block.

By changing the field name, the `for` attribute would also be changed to `for="credential-form_type"`, which breaks the browser behavior when clicking on a label and it selecting the associated input. Thanks for the PR!

See: https://github.com/OpenFn/Lightning/commit/acb7336d850c1ee03089bda5e509dce84adea5fa

PR #276, opened 2022-08-17
Rename users_tokens to user_tokens

stuartc:
Compounded table names should be singular-plural, so `users_tokens` has been renamed to `user_tokens` - all indexes and constraints have been renamed as well.


taylordowns2000 (approved):
thanks @stuartc !

PR #277, opened 2022-08-17
update create credential test

taylordowns2000:


taylordowns2000:
missed coverage on project_credentials adding/removing. will update

PR #284, opened 2022-08-22
docs: update database setup and asdf plugin

mtuchi:
## Description
Added instruction to troubleshoot database connection error when you run `mix ecto.create` or `
env $(cat .env | grep -v "#" | xargs ) iex -S mix phx.server`

#### The error looks like this 
```sh
[error] Postgrex.Protocol (#PID<0.637.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (${POSTGRES_HOST}:5432): non-existing domain - :nxdomain
[error] Postgrex.Protocol (#PID<0.636.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (${POSTGRES_HOST}:5432): non-existing domain - :nxdomain
```
For some reason if the `DATABASE_URL` contain environment variables the variables get interpreted as strings on not the actual value of the variable. Also added instruction for [asdf-plugins](https://github.com/asdf-vm/asdf-plugins)

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,13 @@ environments. Included in the repo is a `.tool-versions` file that is read by
 asdf in order to dynamically make the specified versions of Elixir and Erlang
 available.
 
+> P.S make sure you have added the following
+> [asdf plugin](https://github.com/asdf-vm/asdf-plugins) before you install
+
+- [Erlang](https://github.com/asdf-vm/asdf-erlang)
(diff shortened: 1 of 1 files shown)


taylordowns2000:
thanks @mtuchi , this is great. i've actually adjusted the base `.env.example` file according to your suggestions. check it out! https://github.com/OpenFn/Lightning/commit/14fe8f742043ade9f642770a82dccf637b5ee658

PR #292, opened 2022-08-25
Allow the use of 3rd-party OIDC authentication

stuartc:
This adds the ability for administrators to configure a 3rd party authentication provider.

image

There are a lot of moving parts here:

We split the concept of a 'handler' and the persisted configuration.
One of the reasons for this is that we rely on the `.well-known` file for a given provider. This file is retrieved every time a handler is initialized.

We cache the handler using the Cachex library (we plan to use this more down the line), this avoids querying the database too much (and in respect to the above, not querying the auth provider repeatedly).

Once a provider has been set up, the login page gets a 'via external provider' button.

In order to remove the button, the entire auth provider configuration needs to be removed. (See "Remove" button in screenshot).

**Notes**

- Has only been tested with Keycloak locally at present.
- Has a caching layer that is not cluster aware, meaning changes to the config can take upwards of 30 minutes to synchronize across a cluster, this can be added with a little effort down the line. 
- We could have/should actually stored the `.well-known` in the database, or used it to fill in the required fields (that are tedious at best to do by hand).


stuartc:
@taylordowns2000 next up is testing this with google and seeing where there might be cracks in this implementation.

stuartc:
@taylordowns2000 so we need to test this with Google before it's done done... 

But I have some thoughts about my choice of retrieving the `.well-known` file every time a handler (the thing that lives in memory) is initialised. It's difficult to test (I mean it _is_ tested), but just feels like it may be too brittle in the wild.

I think I should change how this works, to rather fetch the file when setting it up and saving it - or pulling out the details it needs into fields to be stored in the db... 

I'm not sure how much more time we should spend on this though..

stuartc:
@taylordowns2000 I've tested this against Google, worked after a tiny tweak that works with keycloak as well. :+1: 

Also ironed out some of those caching concerns I brought up - but as we decided to leave for now is some of the edge cases that can be resolved by re-saving the settings or a restart.

taylordowns2000 (approved):

PR #293, opened 2022-08-26
Fix monaco jumping cursor

josephjclark:
Monaco is currently broken in two hilarious ways!

1) Every keystroke is triggering the MutationObserver to fire, which causes the component to re-render like mad, which (for some reason but perhaps not surprisingly) is causing the cursor to skip around.

2) If you select multiple jobs in in the workflow editor, Monaco will update - but only if the adaptor happens to be different. This isn't very good.

This MR fixes this by tightening up the MutationObserver logic and explicitly tracking the selected job id.

Also, as a bonus, when changing adaptors, we now instantly clear out the type definitions in Monaco before we go and look for the new dts. This means intellisense will update immediately and drop the old adaptor types instead of waiting for the new stuff to come down.

### DIFF
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -72,6 +72,7 @@ export default function Editor({ source, adaptor, onChange }: EditorProps) {
   
   useEffect(() => {
     if (adaptor) {
+      setLib([]); // instantly clear intelligence
       loadDTS(adaptor).then(l => setLib(l));
     }
   }, [adaptor])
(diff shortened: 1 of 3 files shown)


assets/js/editor/Editor.tsx
```diff
@@ -72,6 +72,7 @@ export default function Editor({ source, adaptor, onChange }: EditorProps) {
   
   useEffect(() => {
     if (adaptor) {
+      setLib([]); // instantly clear intelligence
```
stuartc, line 75: Bonus points for this comment.

stuartc (approved):
Good catch! I like the extra options passed into the observer to curb excessive calls.

I'm still a little puzzled how the previous guard/if wasn't filtering all the callbacks. 

Anyway lgtm! :+1:

PR #294, opened 2022-08-26
Send Account Deletion Email

### LINKED ISSUE #291: Send email to user notifying them of account deletion
### User story

As a user, I'd like to be notified if my account has been scheduled for deletion cause it's a big deal and I'd like a papertrail in my inbox.

### Details

This comes out of #195 

### Implementation notes

- [x] Write an email template (taylor to do) , use "Hi #{user}, your account is going to be deleted." for now.
- [x] Review `Mailer.ex` and the code in `user_notifier.ex`
- [x] Add a send email call  to `Account.schedule_for_deletion()` that notifies the user when their deletion has been scheduled

### Release notes

- Sends emails to notify a user when their account is marked for deletion

### Tests

- [x] When a user has scheduled their account for deletion, assert that an email has been sent to the right address with the right text.

### QA

- [x] This has been QA’d by Amber, branch can be merged after review

elias-ba:
Ships code for sending account deletion email to the user. Closes #291

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,7 +5,12 @@
 # uncommented option that means it's either mandatory to set or it's being
 # overwritten in development to make your life easier.
 
+# Choose an admin email address and configure a mailer. If you don't specify
+# mailer details the local test adaptor will be used and mail previews can be
+# view at localhost:4000/dev/mailbox
 EMAIL_ADMIN='admin@openfn.org'
+# MAILGUN_API_KEY='some-key'
+# MAILGUN_DOMAIN='some-domain'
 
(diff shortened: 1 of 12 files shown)


taylordowns2000 (approved):
piping into a `case`! this might be the most elixiry thing @elias-ba and i approve 😎

the big thing i changed, and i think this is important, is how you're testing the business logic. in essence, you've unit tested the hex package with your `user_notifier_test.exs`. That's fine, but it doesn't test whether a user gets an email sent to them when their account is scheduled for deletion. check out the new lines i've added to the profile live test and the user live test. these more important, IMO. 

image

PR #303, opened 2022-08-29
node 18

taylordowns2000:
adds node 18


taylordowns2000:
@elias-ba , you'd mentioned you were interested in this. it turned out to be a change made during our standup, then an adjustment to test character spacing... so nothing too exciting, but still have a look FYI

PR #314, opened 2022-09-02
Create workflows - Back-end

elias-ba:
This PR ships code that allows to assign a workflow to a Job depending on the trigger type of that Job.

Note that when we delete a workflow all the jobs associated to it will be deleted (CASCADE).

### DIFF
--- a/lib/lightning/jobs/trigger.ex
+++ b/lib/lightning/jobs/trigger.ex
@@ -35,9 +35,8 @@ defmodule Lightning.Jobs.Trigger do
     belongs_to :job, Job
     belongs_to :upstream_job, Job
 
-    field :type, Ecto.Enum,
-      values: @trigger_types,
-      default: :webhook
+    field :type, Ecto.Enum, values: @trigger_types
+    # default: :webhook
 
     timestamps()
   end
@@ -91,6 +90,9 @@ defmodule Lightning.Jobs.Trigger do
         changeset
         |> validate_cron()
         |> put_change(:upstream_job_id, nil)
+
+      nil ->
+        changeset
     end
   end
 end
--- a/lib/lightning/jobs/trigger.ex
+++ b/lib/lightning/jobs/trigger.ex
@@ -35,8 +35,7 @@ defmodule Lightning.Jobs.Trigger do
     belongs_to :job, Job
     belongs_to :upstream_job, Job
 
-    field :type, Ecto.Enum, values: @trigger_types
-    # default: :webhook
+    field :type, Ecto.Enum, values: @trigger_types, default: :webhook
 
     timestamps()
   end
--- a/lib/lightning_web/live/job_live/form_component.ex
+++ b/lib/lightning_web/live/job_live/form_component.ex
@@ -7,7 +7,13 @@ defmodule LightningWeb.JobLive.FormComponent do
   import Phoenix.LiveView,
     only: [assign: 2, assign: 3]
 
-  alias Lightning.{Jobs, AdaptorRegistry, Projects}
+  alias Lightning.{
+    Jobs,
+    AdaptorRegistry,
+    Projects,
+    Workflows,
+    Workflows.Workflow
+  }
 
   defmacro __using__(_opts) do
     quote do
@@ -27,14 +33,33 @@ defmodule LightningWeb.JobLive.FormComponent do
       defdelegate validate(params, socket),
         to: LightningWeb.JobLive.FormComponent
 
+      def insert_workflow_id(params) do
+        params["job"]["trigger"]["type"]
+        |> case do
+          n when n in ["webhook", "cron"] ->
+            {:ok, %Workflow{id: workflow_id}} = Workflows.create_workflow()
+
+            job_attrs =
+              Map.get(params, "job")
+              |> Map.put("workflow_id", workflow_id)
+
+            %{"job" => job_attrs}
+
+          n when n in ["success", "failure"] ->
+            params
+        end
+      end
+
       @impl true
       def handle_event(event, params, socket) do
+        params_with_workflow = insert_workflow_id(params)
+
         case event do
           "validate" ->
-            {:noreply, validate(params, socket)}
+            {:noreply, validate(params_with_workflow, socket)}
 
           "save" ->
-            {:noreply, save(params, socket)}
+            {:noreply, save(params_with_workflow, socket)}
         end
       end
 
--- a/lib/lightning_web/live/job_live/form_component.ex
+++ b/lib/lightning_web/live/job_live/form_component.ex
@@ -7,11 +7,7 @@ defmodule LightningWeb.JobLive.FormComponent do
   import Phoenix.LiveView,
     only: [assign: 2, assign: 3]
 
-  alias Lightning.{
-    Jobs,
-    AdaptorRegistry,
-    Projects
-  }
+  alias Lightning.{Jobs, AdaptorRegistry, Projects}
 
   defmacro __using__(_opts) do
     quote do
--- a/lib/lightning_web/live/job_live/form_component.ex
+++ b/lib/lightning_web/live/job_live/form_component.ex
@@ -37,7 +37,8 @@ defmodule LightningWeb.JobLive.FormComponent do
         params["job"]["trigger"]["type"]
         |> case do
           n when n in ["webhook", "cron"] ->
-            {:ok, %Workflow{id: workflow_id}} = Workflows.create_workflow()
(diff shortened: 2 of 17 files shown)


lib/lightning/jobs/trigger.ex
```diff
@@ -35,9 +35,8 @@ defmodule Lightning.Jobs.Trigger do
     belongs_to :job, Job
     belongs_to :upstream_job, Job
 
-    field :type, Ecto.Enum,
-      values: @trigger_types,
-      default: :webhook
+    field :type, Ecto.Enum, values: @trigger_types
+    # default: :webhook
```
elias-ba, line 39: @taylordowns2000 @stuartc I did comment this out (didn't remove yet) because I want you to tell me what you think about it.

taylordowns2000, line 39: why remove the default? (maybe this is a stu question.)

lib/lightning_web/live/job_live/form_component.ex
```diff
@@ -7,7 +7,11 @@ defmodule LightningWeb.JobLive.FormComponent do
   import Phoenix.LiveView,
     only: [assign: 2, assign: 3]
 
-  alias Lightning.{Jobs, AdaptorRegistry, Projects}
+  alias Lightning.{
```
taylordowns2000, line 10: this makes the review harder --> doesn't seem related to the PR

lib/lightning_web/live/job_live/inspector_form_component.ex
```diff
@@ -31,8 +31,11 @@ defmodule LightningWeb.JobLive.InspectorFormComponent do
 
       :new ->
         case Jobs.create_job(
-               job_params
-               |> Map.put("project_id", socket.assigns.job.project_id)
+               Map.put(
```
taylordowns2000, line 34: same here --> is this part of the PR? if not please remove

priv/repo/migrations/20220829103340_add_workflows_table.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Lightning.Repo.Migrations.AddWorkflowsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:workflows, primary_key: false) do
+      add :id, :binary_id, primary_key: true
+      add :name, :string
```
taylordowns2000, line 7: i think we need a `project_id` field and then a unique constraint on `project_id + name`

test/lightning/jobs_test.exs
```diff
@@ -101,6 +104,146 @@ defmodule Lightning.JobsTest do
...
+        Jobs.create_job(%{
+          body: "some body",
+          enabled: true,
+          name: "some name",
+          adaptor: "@openfn/language-common",
+          trigger: %{type: "cron"},
+          project_id: project_id
+        })
+
+      {:ok, %Job{} = upstream_job_a} =
```
taylordowns2000, line 223: this is a "downstream" job, not upstream, right?

test/lightning/jobs_test.exs
```diff
@@ -101,6 +104,146 @@ defmodule Lightning.JobsTest do
...
+          enabled: true,
+          name: "some name",
+          adaptor: "@openfn/language-common",
+          trigger: %{type: "on_job_success", upstream_job_id: parent_job_1.id},
+          project_id: project_id
+        })
+
+      assert upstream_job_a.workflow_id == parent_job_1.workflow_id
+
+      {:ok, %Job{} = upstream_job_b} =
```
taylordowns2000, line 235: this is not a different job (b vs a), it's your same old job but you've updated it, right?

test/lightning/workflows_test.exs
```diff
@@ -0,0 +1,64 @@
...
+      workflow = workflow_fixture()
+      update_attrs = %{name: "some-updated-name"}
+
+      assert {:ok, %Workflow{} = workflow} =
+               Workflows.update_workflow(workflow, update_attrs)
+
+      assert workflow.name == "some-updated-name"
+    end
+
+    test "delete_workflow/1 deletes the workflow" do
```
taylordowns2000, line 50: this only tests a small fraction of the "delete_workflow" functionality. you've added a cascade, which is SUPER SERIOUS! make sure this test proves that if you delete a workflow you permanently delete every single job associated with it. (eish... this has huge audit trail implications.) update the test please, and then let's discuss with stu and amber

test/lightning/workflows_test.exs
```diff
@@ -0,0 +1,64 @@
...
+    test "create_workflow/1 with valid data creates a workflow" do
+      valid_attrs = %{name: "some-name"}
+
+      assert {:ok, %Workflow{} = workflow} =
+               Workflows.create_workflow(valid_attrs)
+
+      assert workflow.name == "some-name"
+    end
+
+    test "update_workflow/2 with valid data updates the workflow" do
```
taylordowns2000, line 40: once uniqueness is added, show that you can't name two workflows the same thing if they're in the same project

taylordowns2000 (changes requested):
looks great @elias-ba . big change is adding `project_id` (and the unique constraint) to your project model and the migration. in addition, please address each comment below and then set up an hour-long call with Amber to:

1. change the language and potentially adjust the logic in those 4 key tests: there are both some naming issues and some opportunities for logical clarity. is there an assertion for every single important business requirement?
2. discuss the cascade

lib/lightning/jobs/trigger.ex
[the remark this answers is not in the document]
elias-ba: @taylordowns2000 this is what I was mentioning [here](https://github.com/OpenFn/Lightning/pull/314#discussion_r961679662). Let's get @stuartc involved. I did show him that during one of our pairing sessions last week but I am not sure we did have a chance to discuss about.

The main reason why I decided to remove the default value is having would drop automatic changeset changes to be set when we create a trigger with type `webhook` (`webhook` being the default value there). That would be a problem to tests some scenarios like going from a `webhook` trigger to a `cron` one. I know that Ecto has some methods for forcing changesets changes and things like that but I didn't wanna take a solution that would add extra implementation time. At least now we can talk about this and decide what to do.

I am in favor of dropping the default because why should `webhook` be the default ? What is a default trigger type ? Are we trying to say that among all types `webhook` is the most natural one (for the user) ? These are my business reasoning when I am working on this.

What do you think guys ?

priv/repo/migrations/20220829103340_add_workflows_table.exs
elias-ba: Like a `belongs to` relationship between `workflows` and `projects` ?

test/lightning/jobs_test.exs
elias-ba: Yes it is. will change it.

test/lightning/jobs_test.exs
elias-ba: Yes

lib/lightning_web/live/job_live/inspector_form_component.ex
elias-ba: Yeah this was the first attempt of this work. After that things have changed / moved a lot and I forgot to restore the file to it's original format. But nothing too serious, will change it to a pipe.

lib/lightning_web/live/job_live/form_component.ex
elias-ba: Well just `mix format` I guess. but I will change that.

test/lightning/jobs_test.exs
elias-ba: Done

test/lightning/jobs_test.exs
elias-ba: Done

elias-ba: Done

elias-ba: Done

test/lightning/workflows_test.exs
elias-ba: Done

test/lightning/workflows_test.exs
elias-ba: Done

lib/lightning_web/live/job_live/form_component.ex
stuartc: @taylordowns2000 In this case, and will always (un/fortunately) lean on `mix format`, so almost impossible to avoid this. Our formatting checker might fail the PR checks if we fail to conform.

stuartc (approved):
Great work @elias-ba!

PR #315, opened 2022-09-06
Allow users to rename workflows

### LINKED ISSUE #239: Allow users to create and name workflows
### User story 
As a user, I would like to name my workflow
So that I can easily refer to it and understand which workflow a run is related to

### Details
We want to make the required database changes to have workflows as an entity. For now, they can just have an ID and a name (or label, see Stu's comments below).

Change the 'New job' button to 'New workflow'. When the user clicks 'New workflow', it should create a new workflow with a job that has default values. 

Screenshot 2022-08-26 at 19 06 45

Screenshot 2022-08-29 at 08 52 19

### Implementation notes
We will need to add a workflows table, that has a label (intentionally different to a name, leaving room for a url safe string if needed):
workflows: name, id.
jobs: workflow_id fk on jobs that cannot be null. 
All jobs will need to belong to a workflow, so a workflow_id fk on jobs.

Label above or near the trigger that has the workflow name. This means we wouldn't have to rearange the structure. 

Create a job and assign it to the workflow when user clicks the + icon. 

### Release notes 
Allow a user to create and name workflows

### Tests 

For Elias: 
There is a workflows table which has a name and id column. The jobs table has a workflow_id foreign key.
A job has to belong to a workflow (workflow_id cannot be null). 

test "create_job/1 with a cron or webhook trigger creates a new workflow and returns a job with THAT workflow_id" do
test "create_job/1 with an upstream job returns a job with the upstream job's workflow_id" do
test "update_job/2 from a cron to a webhook trigger does NOT create a new workflow" do 
test "update_job/2 from upstream_job A (in workflow 1) to upstream_job B (in workflow 2) changes the updated job's workflow_id to 2" do

For another issue: 
On the workflows page, there is a button that says 'new workflow'. When I click it, there is a modal on the page asking for a workflow name. When I hit 'create', I can see the workflow name on the canvas. 
Below the workflow name is a block with the job name set to 'Choose a trigger'. 

### QA
- [ ] This has been QA’d by Amber, branch can be merged after review

stuartc:
image

- Move WorkflowDiagram related actions from DashboardLive into
  WorkflowLive
- Move WorkflowDiagram component into liveview, so that we can subscribe
  to channels (i.e. updating of the diagram when someone changes
  something).
- Integrate `@openfn/workflow-diagram@0.0.8` and use the new Store
  interface for updating it.
- Remove `component_mounted` event from WorkflowDiagram hook, using a
  MutationObserver and a Base64 encoded JSON payload.
- Fixed an issue where the compiler component would try and load a
  'nothing adaptor', added a condition to check an adaptor is actually
  selected.
- Removed previous Workflow CTE queries, replaced by the introduction of
  the Workflow model, see (lib/lightning/workflows.ex:112).
  
Closes #239


stuartc:
Please ignore the codecov failure, I moved an html.heex file into a render function and now the coverage is slightly lower. I didn't ever notice that our templates are only considered for coverage if they are in the live view itself, I mean it makes sense - just never thought about it.

taylordowns2000 (approved):
very happy about DashboardLive to WorkflowLive :-)

PR #342, opened 2022-09-21
Allow HTTP adaptor credentials to be created via a form

### LINKED ISSUE #327: Allow HTTP adaptor credentials to be created via a form
### User story
As a user that's creating a credential for the http adaptor, I would like a form which tells me exactly which inputs are needed, so that I don't have to read through documentation. 

### Details

When creating a new credential, the schema option should have a Type option for HTTP.
When selecting the HTTP type, the form should change to show the correct fields for the HTTP credential.
Any required fields, and any validations specific to the HTTP credential type should be reflected in the form. (i.e. if `this_field` is required, failure to set it renders the form unsaveable and a validation error/message appears next to the input).

### Implementation notes

Currently we only have support for two types: `raw` and `DHIS2`, which is hardcoded here: `lib/lightning_web/live/credential_live/form_component.ex:29`.

We need to be able to (at first) be able to load schemas from disk, I suggest putting them in `priv/schemas` for now.

In a future issue we will move these schemas into adaptors, and fetch them on application startup - and cache them in memory (or on disk).

- [x] Create a schema for the HTTP adaptor, see the hardcoded example referenced above.
- [x] Move the current `fake_body_schema` into a file in `priv/schemas` called `dhis2.json`.
- [x] When the DHIS2 or HTTP schema type is selected in the dropdown, read the schema from disk and render the form.

elias-ba:
Later on we will read the JSON Schemas from the adaptor but for now they are stored in `./priv/schemas/{adaptor_name}.json`.

## Related issue

Fixes #327 


lib/lightning_web/live/credential_live/form_component.ex
```diff
@@ -169,20 +140,6 @@ defmodule LightningWeb.CredentialLive.FormComponent do
           schema_changeset: schema_changeset,
           changeset: changeset
         )
-
```
stuartc, line 172: The purpose of the tuple `{schema, schema_type}` was to avoid loading the schema from disk with every event.

I see you're no longer using the `schema`, so we need to either remove that from the case statement - or we change this around to only reload the schema when the schema type is changed.

If this feels like too much effort (like will require a whole boat load of logic) then we can leave it and wrap this in a schemaless changeset in the future as an optimization.

stuartc (changes requested):
Added one comment, let me know your position on that. Happy to merge on your take.

lib/lightning_web/live/credential_live/form_component.ex
elias-ba: Mhm @stuartc thanks for explaning the purpose of `{schema, schema_type}`. I did remove it because I couldn't see it's value but now I do understand and kind of wanna put it back. But what would you really suggest me to do there ? I know that we have multiple ways we can do this but what do you think would be the best ?

stuartc: So the idea is that we create a new 'schema' (`Lightning.Credentials.Schema`) when we have a schema type value that we need a schema for - but only if we don't already have a schema _or_ the schema we have is different to the last one.

Currently the `Schema` struct has no information about it's name etc. So we don't have a simple way to just compare what is loaded and what is requested.
I think the way I did it before only worked because only one of the schema types actually had a a schema.

I guess my option to just change the `{_schema, schema_type}` to `schema_type`, and load it every time was because I haven't taken enough time to think about a more robust solution - I feel like avoiding really simple "if before != current" kinda logic.

So I guess for now, lets remove the leftovers as to not confuse someone else and we come back to it. Again, only because I don't have a solid solution right now; and we _will_ be coming back to this when we add more schemas...

elias-ba: @stuartc this is now done. Codecov is still complaining tho but I don't see the reason. So tell me if this is ready to merge or if I still need to address something. Thanks

PR #349, opened 2022-09-27
Scheduled deletion should not appear when creating a new user

### LINKED ISSUE #344: Scheduled deletion should not appear when creating a new user
When a superuser creates a new user, the form includes a 'scheduled for deletion' input. This should not be here, since no one would ever create a user and schedule them for deletion at the same time. 

1. Go to admin settings -> users -> new user
2. See that the form has a 'scheduled for deletion' input (see screenshot) and a disabled? input

happen.

The form should only have the first name, last name, email and password inputs.

The following input should not be there: 'scheduled for deletion' and 'disabled?'.

Screenshot 2022-09-23 at 11 28 42

To be clear, these changes should be made to the form only when a new user is being created. 
When a superuser edits a user, the form should stay as it currently is.

elias-ba:
`scheduled_deletion` should not appear neither in :new nor in :edit

## Related issue

Fixes #344 

### DIFF
--- a/lib/lightning_web/live/user_live/form_component.html.heex
+++ b/lib/lightning_web/live/user_live/form_component.html.heex
@@ -81,55 +81,35 @@
         ) %>
       </div>
     </div>
-
     <div class="hidden sm:block" aria-hidden="true">
       <div class="py-5"></div>
     </div>
-
-    <div class="grid grid-cols-6 gap-6">
-      <div class="col-span-3">
-        <%= label(f, :scheduled_deletion,
-          class: "block text-sm font-medium text-secondary-700"
-        ) %>
-        <%= text_input(f, :scheduled_deletion,
-          class: "block w-full rounded-md",
-          phx_debounce: "blur"
-        ) %>
-        <%= error_tag(f, :scheduled_deletion,
-          class:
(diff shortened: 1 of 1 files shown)


taylordowns2000 (changes requested):
it looks like you're hiding `disabled` from new, but keeping it on edit.
and then it looks like you're removing `scheduled_deletion` entirely.

please make sure that `scheduled_deletion` appears in `:edit`. chat to @amberrignell with Qs. otherwise, the pattern used to conditionally hide `disabled` (assuming that was something Amber requested) looks good.

elias-ba:
> it looks like you're hiding `disabled` from new, but keeping it on edit. and then it looks like you're removing `scheduled_deletion` entirely.
> 
> please make sure that `scheduled_deletion` appears in `:edit`. chat to @amberrignell with Qs. otherwise, the pattern used to conditionally hide `disabled` (assuming that was something Amber requested) looks good.

@taylordowns2000 your requests have been addressed. Please review and merge to main if all good. cc @amberrignell FYI we have agreed that we'll keep scheduled_deletion input in user edit for when we wanna cancel deletion. But hide it in new user.

PR #359, opened 2022-10-12
352 reverse relationship between triggers and jobs

### LINKED ISSUE #352: Reverse relationship between Triggers and Jobs
In the near future we will be needing to have a Trigger assigned to either a Job or a Workflow; this is currently not possible because a Trigger _belongs to_ a Job and cannot be created without being associated with one.

In order to facilitate these future requirements, we want to reverse the relationship.
This means that a Trigger will also need to belong to a Workflow (currently this is derived via it's related Job).

Instead of the `triggers` tables having a `job_id`, the `jobs` table will have a `trigger_id` column.

Rough steps:

1. Create a migration that changes the `jobs` and `triggers` tables. (ensure you specify `binary_id` or `UUID` column types)
2. Change both Job and Trigger schemas to: `belongs_to Trigger` and `has_many Job` respectively vs. `has_one Trigger` and `belongs_to Job`.
3. Make any changes to the Job UI/form and/or changesets in order to ensure a trigger gets both updated or created in the appropriate situations. 

Some gut-check assumptions:
1. A trigger should _not_ have a `job_id` column — multiple jobs can use the same trigger.
3. A trigger _should_ have a `workflow_id` column — it belongs to a workflow.
4. A job should have a `trigger_id` column — later we will allow a single job to have multiple triggers, but for now we’d like each job to have a single trigger_id.
5. A workflow should have 

Currently triggers are created _via_ a Job changeset, [see: `job.ex#76`](https://github.com/OpenFn/Lightning/blob/main/lib/lightning/jobs/job.ex#L76)
Where the form passes in params for `trigger`. This may not require any changes - but do keep in
mind that the Trigger record is created at the same time (in this case just before) the Job.

image

- - -

**Resources**

https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3
https://hexdocs.pm/ecto/Ecto.Schema.html#belongs_to/3
https://hexdocs.pm/ecto/Ecto.Changeset.html#cast_assoc/3

wanecode:
## Related issue

Fixes #352 reverse relationship between triggers and jobs


stuartc:
@wanecode I've oriented myself with the issue you have been facing; this is a tough one - still thinking of the best solution while not making our existing jobs->trigger interface (which has been really nice) difficult to work with. Will see what comes of it tomorrow am.

PR #361, opened 2022-10-12
Configure Cron Expression Thru Form

### LINKED ISSUE #233: Configure a cron expression through a form
If Cron is selected, there should be a radio list containing:

- Every hour
- Every day
- Every week
- Every month 
- Custom

On selecting one of the options an additional component should be displayed allowing a required day/time to be selected:

- Every hour - select minute
- Every day - select time of day
- Every week - select day of week and time
- Every month - select day of month and time
- Custom - a monospace font styled text input that accepts a cron expression.

All initial times, day of week and day of months can default to 00:00, Monday and the 1st respectively.

Internally we want to save these selections as cron expressions:

i.e. Every day at 04:05 should be persisted as: `5 4 * * *`
Every week on Monday at 00:00 should be: `0 0 * * MON`.

elias-ba:
## Related issue

Fixes #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 - Add seed data for demo site
 - create adaptor credentials through a form
 - Redirect users to projects list page when they click on Admin Settings menu
+- Add radio buttons for selecting options to configure cron expression thru form
 
 ## [0.2.0] - 2022-09-12
 
--- a/lib/lightning/helpers.ex
+++ b/lib/lightning/helpers.ex
@@ -32,4 +32,99 @@ defmodule Lightning.Helpers do
       {:ok, body_map} -> body_map
     end
   end
+
+  def cron_values_to_expression(
+        %{
+          "weekday" => weekday,
+          "hours" => hours,
+          "minutes" => minutes
+        } = trigger_params
+      ) do
+    Map.put(
+      trigger_params,
+      "cron_expression",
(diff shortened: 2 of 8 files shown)


elias-ba:
@stuartc please can you move forward w/ this ?

stuartc:
@elias-ba I've taken a good look at the work you've done, I've made some notes - we will schedule some time to make a plan to get this wrapped up in a way that is more resilient.

- I've got some naming proposals like:
  - Change `periodicity` to `frequency`
  - `hours` to `hour` and others from plural to singular.
- We should be using the existing label component instead of using the
  `prompt` attribute for selects.
- Remove unused `on_cron_type_change`
- Remove unused `parse_cron_expression`
- We don't need to attach the form to inputs that don't map to a schema/changeset.
  (This may not entirely be true for reasons stated below)
- We need validations, you can save a broken cronjob which causes the job
  to be inaccessible afterwards.
- We do need a changeset, but have to find a good way to integrate it.  
  This is something I've also struggled to find a pattern that fits naturally
  in Phoenix.
- We should use the existing Crontab to parse and manipulate cron strings

elias-ba:
Thank you so much @stuartc for your feedback. This is great. I am happy to schedule sometime tomorrow or Monday to talk about all this w/ you.

elias-ba:
Closing this PR to open new one according to all the new directions we are taking.

PR #374, opened 2022-10-20
355 associate runs with attempts

### LINKED ISSUE #355: Associate Runs with Attempts
In order to support retries but still be able to reference a previously failed context (e.g. same dataclip, started from a webhook, that failed half-way through it's workflow). By grouping Runs in attempts, we can show users related events like retries.

We need two new tables: `attempts` and `attempt_runs`

Create a table and model for `Lightning.Attempt`, which allows for Runs to be assigned (and created via its changeset).

A Run has many attempts through `attempt_runs`, this relationship should be added to the Run schema.
An Attempt must have a Reason associated with it.

> Note we do not need a model for `attempt_runs` as for now it is purely a join table. This means that it doesn't have an `id` either.

Add a compound primary key to `attempt_runs`, this will enforce uniqueness and serve as an index.

> It may not be necessary to add the `workflow_id` reference to `attempts` at this point, in order to keep things simpler.

wanecode:
## Related issue 

Fixes #355 associate runs with attemps

### DIFF
--- a/lib/lightning/attempt_service.ex
+++ b/lib/lightning/attempt_service.ex
@@ -1,0 +1,56 @@
+defmodule Lightning.AttemptService do
+  @moduledoc """
+  The Attempts context.
+  """
+
+  import Ecto.Query, warn: false
+  alias Lightning.Repo
+  alias Lightning.Attempt
+  alias Lightning.Invocation.{Event, Run}
+
+  @doc """
+  Creates a reason.
+
+  ## Examples
+
+      iex> create_attempt(%{field: value})
+      {:ok, %Attempt{}}
+
+      iex> create_attempt(%{field: bad_value})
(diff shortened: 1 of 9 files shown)


wanecode:
Hey @stuartc.

wanecode:
Hey @stuartc. 
We have timestamps on the attempt_runs table.
Problem is:
Depending on how we setup the join table migration , Ecto will insert automatically timestamps or raise a not null violation.
see https://github.com/elixir-ecto/ecto/issues/1260.
Basically we need a AttemptRun model if we want to use timestamps function.
So i skipped it for now.

stuartc:
@wanecode thats perfect, I don't think we need timestamps on that table - and since we're avoiding making it a schema (or an entity) that seems like a realistic compromise.

PR #387, opened 2022-10-26
Improve job inspector to be a panel

### LINKED ISSUE #380: Make job inspector a panel, not a pop-up
The current job inspector appears in a popup on the right side of the screen. (See first image.)

Make it appear in a panel, not a pop-up. (See second image.) The panel should behave like the inspector on chrome.

It should be draggable and the contents _inside_ the panel (i.e., the form) should take up 100% of the panel. It should _not_ extend above the top menu, but rather sit inside the parent component which contains the top menu for user login, etc.

Note that this issue does _not_ involve styling for the job form, or making any changes to the job form. It's focused on the div that contains the job form.

While CSS and JS will be involved, your implementation should focus on functionality, not cosmetics.

## Implementation notes:
1. From @stuartc : "The changes required are to stretch the div to the bottom of the page (make is absolute positioning), and to resize it when the window/viewport is resized."
2. More from @stuartc: "Should the content of the panel require scrolling (i.e. viewport is small) then the scrolling must happen inside the panel, and should not affect the elements outside of it.. i.e. when scrolling past the end of the panel it should not scroll the diagram." - https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
3. Possible helpful example for draggable: https://stackoverflow.com/questions/68446571/how-to-do-a-draggable-expandlble-div-react-tailwind
4. A thought from Stu on why we think making this div appear "on top" of the canvas rather than "side by side" with the canvas: "having the panel be on top in this specific case gives us the opportunity to refocus the diagram without it reacting to a container size change without us saying so... if however we still want to resize the diagram we will already have the places to control that with our callbacks for viewport resizes for the panel."

mtuchi:
Style job inspector to behave like a panel instead of a pop-up 

## Related issue

Fixes #380 

### DIFF
--- a/lib/lightning_web/live/job_live/inspector_form_component.ex
+++ b/lib/lightning_web/live/job_live/inspector_form_component.ex
@@ -74,6 +74,12 @@ defmodule LightningWeb.JobLive.InspectorFormComponent do
     end
   end
 
+  def editor_height(form, id) do
+    height = if form.data.id && id, do: 33.7, else: 38
+
+    "rounded-md border border-secondary-300 shadow-sm overscroll-y-auto h-[calc(100vh_-_#{height}rem)] bg-vs-dark"
+  end
+
   @impl true
   def render(assigns) do
     ~H"""
@@ -196,7 +202,7 @@ defmodule LightningWeb.JobLive.InspectorFormComponent do
               phx-hook="Editor"
               phx-update="ignore"
               id="editor-component"
-              class="rounded-md border border-secondary-300 shadow-sm overscroll-y-auto h-[calc(100vh_-_33.7rem)] bg-vs-dark"
+              class={editor_height(f, @id)}
               data-adaptor={Phoenix.HTML.Form.input_value(f, :adaptor)}
               data-hidden-input={Phoenix.HTML.Form.input_id(f, :body)}
               data-job-id={@id}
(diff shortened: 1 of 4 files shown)


lib/lightning_web/live/job_live/inspector_form_component.ex
```diff
@@ -74,6 +74,12 @@ defmodule LightningWeb.JobLive.InspectorFormComponent do
     end
   end
 
+  def editor_height(form, id) do
+    height = if form.data.id && id, do: 33.7, else: 38
+
+    "rounded-md border border-secondary-300 shadow-sm overscroll-y-auto h-[calc(100vh_-_#{height}rem)] bg-vs-dark"
+  end
+
```
stuartc, line 82: Can you explain the logic here? What are the circumstances the height is different?

lib/lightning_web/live/job_live/inspector_form_component.ex
```diff
@@ -188,7 +194,7 @@ defmodule LightningWeb.JobLive.InspectorFormComponent do
               phx-hook="Editor"
               phx-update="ignore"
               id="editor-component"
-              class="rounded-md border border-secondary-300 shadow-sm h-96 bg-vs-dark"
+              class={editor_height(f, @id)}
```
stuartc, line 197: We will need to refactor this so we don't call functions, as per my notes yesterday.

stuartc (changes requested):
Is that it? :smile: if this is the entire solution then fantastic work.

Let's just make the editor height change more clear, and we can change the class around from using a function to being put into assigns.

All of these main components will be split out in the near future so we will also get the opportunity to have the logic inside another component.

stuartc (approved):

Issue #347, opened 2022-09-26
Run history queries (2/2) - runs history expanded view


amberrignell:
### User story
Given a project_id and source dataclip, 
Return an ordered list of runs with their status, associated job name and run time. 

### Details
This query will be used to populate the runs history page when a user clicks into a workflow run
![Screenshot 2022-09-26 at 08 46 02](https://user-images.githubusercontent.com/36554605/192202191-55ed8a72-5614-4522-8afa-786376a125b3.png)

(note: this screenshot is missing the run time next to each job run, but we will be adding it)
note: as long as the architecture supports the 'greyed out/jobs that haven't been run', this can be out of scope for v0 Lightning

### Implementation notes

This may involve extending the existing query to return some nested data, _or_ the ability to request this nested data on demand.

PR #393, opened 2022-11-01
346 runs history queries 12 runs history collapsed view

### LINKED ISSUE #346: Runs history queries (1/2) - runs history collapsed view
### User story
As a user monitoring my workflows, I would like to determine the final status of a WorkOrder run, so that I know if every step has completed successfully. 

### Details
Given a project ID and a WorkOrder, return the final status of a workflow run. 

### Implementation notes
See this document for notes: https://docs.google.com/spreadsheets/d/1YvvskT8pJckTVbc18JW0qo38pNAjlgLp2OM7ZRx2QCk/edit#gid=0

### Tests
Given a workflow with Job A, B and C. 
When run A,B and C are successful, I expect to see a wokflow run status of 'Success'. 
When run A and B are successful but C fails, I expect to see a wokflow run status of 'Failure'. 
When run A times out, I expect to see a wokflow run status of 'Timeout'. 
When run B crashed, I expect to see a wokflow run status of 'Crashed'. 

@taylordowns2000 should this implementation be returning status codes that then get changed in the UI or the actual words 'Success', 'Failure' etc ?

wanecode:
## Related issue

#347 Run history queries 2/2

Fixes #346 Runs history queries 1/2


lib/lightning_web/live/run_live/components.ex
```diff
@@ -100,21 +30,24 @@ defmodule LightningWeb.RunLive.Components do
...
         class="my-auto whitespace-nowrap p-6 font-medium text-gray-900 dark:text-white"
       >
-        workFlowName
+        <%= @workflow.name %>
       </th>
+      <td class="my-auto p-6"><%= @reason.dataclip_id %></td>
       <td class="my-auto p-6">12i78iy</td>
-      <td class="my-auto p-6">UTC 24:20:60</td>
+      <td class="my-auto p-6">
+        <%= @last_attempt.last_run.finished_at |> Calendar.strftime("%c") %>
```
stuartc, line 43: I managed to get the app to throw an exception with runs that haven't finished. I've fixed the issue in the WorkOrder component, but not here.

We need a test for this.

Also the name of this function component is `work_order`, should it not be attempt or something else? I'm confused by the `Components.WorkOrder` and this?

Is this file being used anymore? What about `AttemptComponent`? Please remove this file is we're not using it.

stuartc (changes requested):
This is great overall @wanecode, really great. Could you address some of my comments.

I have done a tiny bit of changes here: 9802577541dab6cc681e5fcf1316a0a55fbcdd63

- Added `pending` status for Runs that are in progress
- Removed use of Pipeline to make fixtures for WorkOrderLive tests
- Remove codecov skip comments for RunLive WorkOrder components

lib/lightning_web/live/run_live/components.ex
wanecode: Only pill function are needed in this file. Removed work_order function

lib/lightning_web/live/run_live/attempt_component.ex
```diff
@@ -0,0 +1,64 @@
...
+              <%= case @last_run.exit_code do %>
+                <% val when val > 0-> %>
+                  <span class="my-auto ml-2 whitespace-nowrap rounded-full bg-red-200 py-2 px-4 text-center align-baseline text-xs font-medium leading-none text-red-800">
+                    Failure
+                  </span>
+                <% val when val == 0 -> %>
+                  <span class="my-auto ml-2 whitespace-nowrap rounded-full bg-green-200 py-2 px-4 text-center align-baseline text-xs font-medium leading-none text-green-800">
+                    Success
+                  </span>
+                <% _ -> %>
```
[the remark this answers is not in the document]
wanecode: I think we should have it here too to keep things in sync

PR #402, opened 2022-11-04
Cron expression builder

### LINKED ISSUE #233: Configure a cron expression through a form

elias-ba:
## Related issue

Fixes #233 

### DIFF
--- a/test/lightning_web/live/workflow_live_test.exs
+++ b/test/lightning_web/live/workflow_live_test.exs
@@ -161,11 +161,11 @@ defmodule LightningWeb.WorkflowLiveTest do
 
     test "get_cron_data/1" do
       assert LightningWeb.JobLive.CronSetupComponent.get_cron_data("5 0 * 8 *")
-             |> Map.get(:frequency) == :custom
+             |> Map.get(:frequency) == "custom"
 
       assert LightningWeb.JobLive.CronSetupComponent.get_cron_data("5 0 8 * *") ==
                %{
-                 :frequency => :monthly,
+                 :frequency => "monthly",
                  :minute => "05",
                  :hour => "00",
                  :monthday => "08"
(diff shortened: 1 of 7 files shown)


test/lightning_web/live/workflow_live_test.exs
```diff
@@ -153,6 +154,160 @@ defmodule LightningWeb.WorkflowLiveTest do
...
+
+      view
+      |> render_patch(
+        Routes.project_workflow_path(conn, :edit_job, project.id, job.id)
+      )
+
+      assert view
+             |> has_element?("#frequency option[selected][value=monthly]"),
+             "Should have the option that was previously selected"
+    end
```
stuartc, line 308: @elias-ba I added a test here to show where the feature doesn't work reliably, whenever I change the value, save the job and return the value is always 'hourly'.

stuartc (changes requested):
@elias-ba please address the test assertion I added.

test/lightning_web/live/workflow_live_test.exs
elias-ba: @stuartc I think this is now fixed with the code I pushed lately. Sorry for this interference

stuartc (commented):
Great! Merged!

Issue #285, opened 2022-08-23
Filter runs by date


amberrignell:
### User story
As a user that's looking for a workflow run that happened on a specific date, I would like to filter my workflow runs by date, so that I can more easily find it.

### Details
The user should be given a date time selector to pick a start date and an end date.

The user needs a way to clear the date filters

### Release notes 
- Filter runs by run date

### Tests 
On the runs history page, there is a date picker with the label ‘Before’ and another with the label ‘After’. 
Given a workflow run a on the page that was run on the 23/08/2022 and another workflow b that was run on the 29/08/2022. 
When the user sets the ‘after’ date to 25/08/2022, workflow run b is no longer on the page. 
When the user sets the ‘before’ date to 28/08/2022, workflow run b AND workflow run a are not on the page. 
Given a user has selected only a before date, show all runs that happened before that date. 

The user can clear the date for both 'before' and 'after'.

Issue #243, opened 2022-08-10
Filter runs by workflow name


amberrignell:
### User story 
As a user that is focussed on a specific workflow, I would like to be able to filter runs by workflow name, so that I don't have to view runs that are not relevant to me 

### Context
Give the user a dropdown of workflows in the project where they can select a workflow to filter by. When the user selects a workflow, they should only see workorders / runs related to that workflow. 

### Release notes
Filter runs by workflow name

### User acceptance criteria
There's a title 'Filter by workflow' on the search bar
When I click the input, I see a dropdown of all the workfflows in the project
There is no workflow from another project in that dropdown
When I select a workflow, I only see runs for that workflow
Work orders are still ordered by date in the runs history (latest first)

If the user has already selected/unselected some options in the 'filter by status', these do not change when the user selects a workflow (the old search parameters still apply)

PR #424, opened 2022-11-14
Filtering runs #241 #243 #285

### LINKED ISSUE #241: Filter run history by status
### User story
As a user that is searching for a specific run or just monitoring the health of my integrations, I would like to filter my workflow runs by status (success, failure, crash, timeout)

### Context 
We'll need to add a panel for the search console (where we'll add more filters and search functionality later). See screenshot below:

The status should be checkboxes that can be selected/unselected. These should all be selected by default. 

Screenshot 2022-08-22 at 23 28 15

### Implementation notes
This is an OR search (ie if I select 'success' and 'failure' I want a run to see either or

### Release notes 
- Filter runs history by status (success, failure, crash, timeout).

### Tests 
When a user navigates to the runs history page, there is a section in the search panel that says 'Filter by run status' and which contains four checkboxes that are all selected. 
There is a run on the page with a final status of 'failure'.
When I unselect 'failure', that run disappears/is hidden from the runs history. 
If I select it again, it reappears. 
Same for crash success timeout. 

### QA
[ ] This has been QA’d by Amber, branch can be merged after review

wanecode:
## Related issue 

Fixes #241, #243, #285


amberrignell:
@wanecode if you can add the labels 'After' and 'Before' to the date pickers, and update the changelog then you're good to go 🎊

stuartc (approved):

PR #435, opened 2022-11-16
Docs panel

josephjclark:
This PR replaces the existing compiler/adaptor docs component with a new documentation panel.

This PR is basically just glue code - the heavy implementation logic is off in the `adaptor-docs` package in `kit`. Most improvements will take place over there.

* [x] Render docs below the editor
* [x] Insert text snippets into Monaco

Before this can be tested, we need to publish new `adaptor-docs` and `describe-package` versions over in `kit`.

## Future Work

This PR is just supposed to be a starting point: there's a whole bunch of work we know we need to do soon, including:

* Enable columns layout if there's enough horizontal space
* Size the side panel sensibly (maybe even user-driven)
* Enable the docs panel to be hidden
* Magically auto-import adaptor functions into Monaco, so that code-assist and tooltips work

## Text Insertion

There's a lot to say about this!

First off, we don't insert, we append. This is because insertion is really fiddly:
* If the editor isn't focused, it's not clear where the snippet will be inserted
* Unless the cursor is in precisely the right place, the snippet will probably break the code and almost definitely require fiddly refactoring

Instead, we're doing an append. We add the snippet to the bottom of the code, seperated by a new line. We also scroll to it, move the cursor to it, and give the editor focus.

The implementation uses DOM events now to talk between the Editor and Docs components. This breaks all the patterns in both React and Live View, so it's not something I expect to live for very long :sweat_smile:  Ultimately this needs to be a single React hierarchy which both components live inside - that will give us a much slicker integration.


assets/package.json
```diff
@@ -10,7 +10,8 @@
   "license": "ISC",
   "dependencies": {
     "@monaco-editor/react": "^4.4.5",
-    "@openfn/describe-package": "^0.0.8",
+    "@openfn/adaptor-docs": "file:../../kit/dist/openfn-adaptor-docs-0.0.1.tgz",
```
josephjclark, line 13: Obviously these dependencies are wrong but I can't update them properly until the actual packages are released. For now, use the export utility in `kit` to test locally.

josephjclark:
Ok @stuartc this one is also good to go!

stuartc (changes requested):
Dang, we have CSS issues with our approach. We might have them with the older component too?

What happens is that the component injects it's styles programmatically; which just messes with the existing tailwind config.

image
image

I don't think we can merge this until we have a strategy for the CSS. I'm not super clued up on the different approaches available for this kind of problem.

stuartc:
I think we’re going to have to produce a separate css file for these components to allow the consumer to decide when to import it.

And we should investigate tailwinds purge functionality to avoid duplication of styles.

josephjclark:
Oooh so that explains why my CSS was screwy. Dang.

Let me look into this.

My immediate reaction is that producing a seperate CSS for the components won't really help because when we import that into the page, it'll still duplicate styles and interfere with tailwind.

I do wonder why this is causing issue and `workflow-diagram` isn't....

stuartc:
@josephjclark https://github.com/OpenFn/kit/pull/85#issuecomment-1319637048 - managed to merge this in with the changes I made in https://github.com/OpenFn/kit/pull/87

PR #438, opened 2022-11-21
378 dataclip selector widget

### LINKED ISSUE #378: Dataclip Selector Widget
### User story
As a user that is updating a job, I would like to see an example input, so that I can esaily copy/paste keys into my job expression.

As a user that is testing a job, I would like to select different inputs, so that I can run my job with these. 

### Details
Change the 'input' widget into a dropdown which displays the id for the three latest matching dataclips. 
When a dataclip is selected, show the dataclips body (for now use a monospace text div).
The most recent dataclip should be selected by default. 

### Implementation notes
Extending the very basic "input" widget created in #238, modify that component so that given at least a `project_id` it can show a list of dataclips in a dropdown.

Show the three latest dataclips to have been used by that job

When a dataclip is selected, show the dataclips body (for now use a monospace text div).

It should take 100% of the height and width of its surrounding div: #381

### Release notes
- Show latest matching inputs for a job

### User acceptance criteria
- Given a job with no runs, when I go to the input panel I cannot see any input. There is an input dropdown but it has no options to select. 
- Given a job with several runs, when I go to the input panel I can see the latest input. When I select the dropdown, I can see the ID for 2 other inputs. When I select a different input ID, I can see it displayed. 
- If I select an input, then exit the job inspector and return to it, that same input is still selected and displayed.

wanecode:
## Related issue

Fixes #378 Dataclip Selector Widget

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to
 - Filter runs by workflow
 - Filter runs by date
 - Show page for Runs
+- Show latest matching inputs for a job
 
 ## [0.2.0] - 2022-09-12
 
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -37,6 +37,18 @@ defmodule Lightning.Invocation do
     list_dataclips_query(%Project{id: project_id}) |> Repo.all()
   end
 
+  def list_dataclips_for_job(%Lightning.Jobs.Job{id: job_id}) do
+    from(r in Run,
+      join: d in assoc(r, :input_dataclip),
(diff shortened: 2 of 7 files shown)


amberrignell:
@wanecode this looks great ! One small request: can you remove the 'blank/null' option from the dropdown if there are existing inputs to select from ? 
Screenshot 2022-11-21 at 11 32 40

amberrignell:
Perfect ! Last request - if it's not too much work could you format the input ? (so it looks the same as when you 'Beautify' in postman)

stuartc (approved):

PR #439, opened 2022-11-21
abbreviate standard adaptors

taylordowns2000:

Less technical users find it confusing to see so much duplicated "@openfn/language-" text on the dropdown. If accepted, this would only display that full name for _non-standard_ adaptors — i.e., those hosted by other npm orgs with other naming conventions.

## Related issue

Amber feedback from user interviews


taylordowns2000:
@stuartc , I'll happily add some tests for that function... can you let me know _where_ they should go?
1. given a standard name, does it produce `{x, y}`
2. given a non-standard name, does it produce `x`

stuartc:
@taylordowns2000 I think a new describe or test block at the bottom of `job_live_test.exs` is fine - since it's not used anywhere else.

PR #441, opened 2022-11-22
Runs history bug

### LINKED ISSUE #436: Order work orders by last run finished at date
### User story

As a user looking at my runs history page, I would like to see work orders sorted by date of the last run.finished_at descending, so that I can find the latest run more easily. **The bug is that we show "Last Run Finished At" in the headers and don't sort by anything.**

### Details

Sort the work orders displayed on the runs history page by last run date descending, with nulls coming first so that work orders with in-progress attempts appear first.

### Release notes
Order runs by date

### User acceptance criteria
- Given run A with latest run 11/11/22 and run B with latest run 15/11/22, run B appears ABOVE run A in the runs history.

taylordowns2000:
closes #436

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -393,7 +393,7 @@ defmodule Lightning.Invocation do
       where: ^filter_run_started_after_where(date_after),
       where: ^filter_run_started_before_where(date_before),
       distinct: true,
-      order_by: [desc: r.finished_at],
+      order_by: [desc_nulls_first: r.finished_at],
       preload: [
         reason:
           ^from(r in Lightning.InvocationReason,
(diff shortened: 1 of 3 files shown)


lib/lightning/invocation.ex
```diff
@@ -404,13 +405,23 @@ defmodule Lightning.Invocation do
...
   end
 
   def list_work_orders_for_project(%Project{} = project, filter, params) do
     list_work_orders_for_project_query(project, filter)
     |> Repo.paginate(params)
+    |> find_uniq_wo()
+  end
+
+  def find_uniq_wo(page) do
+    %{page | entries: Enum.uniq_by(page.entries, fn wo -> wo.id end)}
```
stuartc, line 424: @taylordowns2000 is it possible to uniq these via the query and not in memory? I'm cautious that a uniq after the query would result in less than the page limit specifies - assuming it's possible to have more than one unique result in the first place..

PR #455, opened 2022-11-25
Delete job from inspector

### LINKED ISSUE #416: Delete a job from the job inspector
### User story
As a user that has added a job I don't want to my workflow, I would like to be able to delete it from the job inspector, so that I can remove it. 

As a user that is trying to delete a job with downstream jobs, I would like to see a warning that explains why I am unable to delete it, so that I do not think this is simply a bug with the app. 

### Details
There's already the functionality to delete a job, it's just not on the job inspector yet. Once we add it we can get rid of the jobs/list page altogether. 

Users are currently unable to delete a job that has downstream jobs. That's the behaviour we want, but we should be giving feedback to explain why to the user. 

### Release notes
Delete a job from the job inspector
Display error message when user tries to delete a job with downstream jobs

### User acceptance criteria
There is no longer a jobs/list page or the button to navigate there. 
On the job inspector, there is a 'delete' button next to save and cancel. 
Given the last job in a workflow, when I click 'delete', the job gets deleted and I am taken back to the canvas (the job inspector panel is no longer open). 
Given a job with downstream jobs, when I click 'delete', I am given a pop up message that says 'Unable to delete this job because it has downstream jobs'.

elias-ba:
## Related issue

Fixes #416 


amberrignell:
Hey @elias-ba this looks good. If the user is deleting a job with downstream jobs, could you show them the "Unable to delete this job" error immediately (rather than showing the 'This action is irreversable ..." message first (this gives them the impression that it is possible, which is a it misleading).

amberrignell:
Looks good to me - thanks Elias

lib/lightning/jobs.ex
```diff
@@ -194,7 +194,12 @@ defmodule Lightning.Jobs do
 
   """
   def delete_job(%Job{} = job) do
-    Repo.delete(job)
+    job
+    |> Ecto.Changeset.change()
+    |> Ecto.Changeset.foreign_key_constraint(:trigger_id,
+      message: "This job is associated with downstream jobs"
+    )
+    |> Repo.delete()
```
stuartc, line 202: Lovely :+1:

lib/lightning_web/live/job_live/job_builder.ex
```diff
@@ -45,6 +46,18 @@ defmodule LightningWeb.JobLive.JobBuilder do
...
+  end
+
+  defp delete_job_title(job_id) do
+    if is_deletable(job_id) do
+      "Delete this job"
+    else
+      "Impossible to delete upstream jobs. Please delete all associated downstream jobs first."
+    end
+  end
+
```
stuartc, line 60: I have a concern here about performance, see: https://hexdocs.pm/phoenix_live_view/assigns-eex.html#change-tracking

I think that a `@is_deletable` assign should be added to `update/3`, without that we run the risk of the database being called with every render (i.e. `handle_event/3`).

lib/lightning_web/live/job_live/job_builder.ex
```diff
@@ -208,12 +221,55 @@ defmodule LightningWeb.JobLive.JobBuilder do
...
+              text="Delete"
+              phx-click="delete"
+              phx-target={@myself}
+              phx-value-id={@job_id}
+              disabled={!is_deletable(@job_id)}
+              data={[
+                confirm:
+                  "This action is irreversible, are you sure you want to continue?"
+              ]}
+              title={delete_job_title(@job_id)}
```
stuartc, line 236: Instead of calling a function here, use an expression like `if @is_deleteable, do: "x", else: "y"`

test/lightning/jobs_test.exs
```diff
@@ -284,6 +284,26 @@ defmodule Lightning.JobsTest do
...
+          adaptor: "@openfn/language-common",
+          trigger: %{type: "on_job_success", upstream_job_id: job.id},
+          workflow_id: job.workflow_id
+        })
+
+      {:error, changeset} = Jobs.delete_job(job)
+
+      assert {:trigger_id,
+              {"This job is associated with downstream jobs",
+               [constraint: :foreign, constraint_name: "jobs_trigger_id_fkey"]}} in changeset.errors
```
stuartc, line 305: By the way there is a helper for these kinds of matches.

Example from `AccountsTest`:

```
assert %{
  password: ["can't be blank"],
  email: ["can't be blank"]
} = errors_on(changeset)
```

test/lightning_web/live/job_live_test.exs
```diff
@@ -68,4 +70,60 @@ defmodule LightningWeb.JobLiveTest do
...
+      job: job
+    } do
+      {:ok, %Job{} = _} =
+        Jobs.create_job(%{
+          body: "some body",
+          enabled: true,
+          name: "some name",
+          adaptor: "@openfn/language-common",
+          trigger: %{type: "on_job_success", upstream_job_id: job.id},
+          workflow_id: job.workflow_id
```
stuartc, line 112: What about the functions from `JobFixtures`? Can you try replace the call to `create_job` with `workflow_job_fixture` from the fixtures?

stuartc (changes requested):
@elias-ba This is great! A few small changes but otherwise this is good to go.

stuartc (approved):

PR #460, opened 2022-11-28
align run duration with oban timeout and grace period

taylordowns2000:
This brings _some_ of the shutdown grace from OpenFn.org over to Lightning. We still need a dynamic queue service which unsubscribes nodes across larger clusters, but if we run with a single copy of the app during alpha and beta (which will be nice to get a sense of the load that a single node can tolerate) this should prevent us from exiting during running jobs when we restart nodes/perform upgrades.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,14 @@ ORIGINS=//localhost:*
 # SENTRY_DSN=https://some-url.ingest.sentry.io/some-id
 
 # ==============================================================================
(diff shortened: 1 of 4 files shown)


stuartc (commented):
@taylordowns2000 could you add some details on these ENV vars to `DEPLOYMENT.md` please.

stuartc (approved):

PR #477, opened 2022-12-01
Scrub Secrets From Run Results

### LINKED ISSUE #454: Scrub secrets from run_results
Even if it's replaced with a more sophisticated system later, we need to prevent secrets from being written/shown in the clear while storing or viewing run_results.

This concept implementation does that, albeit in a pretty simple/clunky way and _only_ for `state.configuration`. Secrets stored in other parts of the run_result are not scrubbed because they may be used by subsequent jobs in a workflow. We need to decide how to handle this _later_.

We should create tests which:

- [ ] Show that secrets are not stored in `run_results` when left in `state.configuration`.

```ex
  @doc """
  Scrubs values from all keys in configuration, will be replaced by extensions
  to scrubber.ex, which is currently only used for logs.
  """
  @spec scrub_result(body :: map()) :: map()
  def scrub_result(%{} = body) do
    case Map.get(body, :configuration) do
      nil ->
        body

      map ->
        safe_body = body

        Map.keys(map)
        |> Enum.each(fn k -> Map.put(safe_body, k, "***") end)

        safe_body
    end
  end
```

elias-ba:
@stuartc FYI we finally went with your suggestion of remove the `:configuration` key out of the run result. It's much simpler and more secure 😄 

## Related issue

Fixes #454 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 - Added a Delete job button in Inspector
 - Filter workflow runs by text/value in run logs or input body
+- Drop "configuration" key from Run output dataclips after completion
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


stuartc (commented):
@elias-ba great! thanks, could you update the CHANGELOG and I'll merge this!

PR #488, opened 2022-12-02
Multi-line truncate job label

### LINKED ISSUE #447: Long job names are ugly and confusing on the canvas
When a job name has too many characters, it overflows past the job node.

We could:

1. Set a character limit to prevent this from happening
2. Truncate the job name and show an elipsis, then display the full name on hover
3. Something else?

Create a long name for a job then view it from the canvas.

mtuchi:
## Any notes for the reviewer?
Style job label text to be truncated in multi-line, Using [Multi-line truncation with @tailwindcss/line-clamp plugin](https://tailwindcss.com/blog/multi-line-truncation-with-tailwindcss-line-clamp) 

https://www.loom.com/share/af61a76b30c0417c9969f0455e5cccdc
## Related issue

Fixes #447 

### DIFF
--- a/assets/js/workflow-diagram/src/nodes/JobNode.tsx
+++ b/assets/js/workflow-diagram/src/nodes/JobNode.tsx
@@ -41,7 +41,7 @@ const JobNode = ({
         'bg-white',
         'cursor-pointer',
         'h-full',
-        'p-2',
+        'p-1',
         'rounded-md',
         'shadow-sm',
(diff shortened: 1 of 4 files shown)


taylordowns2000 (changes requested):
is there a way to keep the center alignment @mtuchi ?

see how the elipsis doesn't seem to be part of the centering calculation?
image

PR #491, opened 2022-12-08
452 rename a workflow

### LINKED ISSUE #452: Rename a workflow
### User stories
- [ ] As a user that has just created an untitled workflow, I would like to be able to rename it from the page, so that I can give it a name and more easily find it from the workflows index page

### Details

When the user selects a workflow, show the workflow name in the breadcrumbs above

### Implementation notes

- Change the name of a workflow via blur event.

### Release notes
- Rename a workflow through the page heading

### Tests
I can see the workflow name at the top of the page. 
When I double click on the workflow name, I am able to rename it.

wanecode:
## Related issue

Fixes #452 Rename a workflow

### DIFF
--- a/lib/lightning_web/live/components/layout.ex
+++ b/lib/lightning_web/live/components/layout.ex
@@ -24,8 +24,7 @@ defmodule LightningWeb.Components.Layout do
     <div class="flex-none bg-white shadow-sm z-20">
       <div class="max-w-7xl mx-auto h-20 sm:px-6 lg:px-8 flex items-center">
         <h1 class="text-3xl font-bold text-secondary-900 flex items-center">
-          <%= @title %>
-          <%= if assigns[:details], do: render_slot(@details) %>
+          <%= if assigns[:title], do: render_slot(@title) %>
         </h1>
         <div class="grow"></div>
         <%= if assigns[:inner_block], do: render_slot(@inner_block) %>
--- a/lib/lightning_web/live/components/layout.ex
+++ b/lib/lightning_web/live/components/layout.ex
@@ -23,8 +23,9 @@ defmodule LightningWeb.Components.Layout do
     ~H"""
     <div class="flex-none bg-white shadow-sm z-20">
       <div class="max-w-7xl mx-auto h-20 sm:px-6 lg:px-8 flex items-center">
-        <h1 class="text-3xl font-bold text-secondary-900">
+        <h1 class="text-3xl font-bold text-secondary-900 flex items-center">
           <%= @title %>
+          <%= if assigns[:details], do: render_slot(@details) %>
         </h1>
         <div class="grow"></div>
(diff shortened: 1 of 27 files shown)


amberrignell:
Looking good ! Thanks @wanecode

lib/lightning_web/live/workflow_live.ex
```diff
@@ -15,6 +18,31 @@ defmodule LightningWeb.WorkflowLive do
     <Layout.page_content>
       <:header>
         <Layout.header title={@page_title} socket={@socket}>
+          <:details>
+            <%= case @live_action do %>
+              <% :index -> %>
+                <div>/ Workflows</div>
+              <% :new_job -> %>
+                <div>/ new workflow</div>
+              <% _ -> %>
```
stuartc, line 27: This introduces some strange UX, if you:

- Create a new workflow
- Set the name
- Click "Add your first Job"

This the workflow name changes from the one you entered to "new workflow", if you close the job panel it turns back.

What is the reason for changing the title here?

lib/lightning_web/live/workflow_live.ex
```diff
@@ -7,6 +7,9 @@ defmodule LightningWeb.WorkflowLive do
   alias Lightning.Workflows
   import LightningWeb.WorkflowLive.Components
 
+  def editable do
+  end
```
stuartc, line 11: Please remove this if it's not being used.

lib/lightning_web/live/components/layout.ex
```diff
@@ -23,8 +23,9 @@ defmodule LightningWeb.Components.Layout do
     ~H"""
     <div class="flex-none bg-white shadow-sm z-20">
       <div class="max-w-7xl mx-auto h-20 sm:px-6 lg:px-8 flex items-center">
-        <h1 class="text-3xl font-bold text-secondary-900">
+        <h1 class="text-3xl font-bold text-secondary-900 flex items-center">
           <%= @title %>
+          <%= if assigns[:details], do: render_slot(@details) %>
```
stuartc, line 28: I think we should avoid adding to the component interface. (`:details` slot vs `title` attr).

I prefer the named slot approach you just added, could you change the slot name to `:title` and then change all the templates to use `<:title>...</:title>` please.

stuartc (changes requested):
This looks good dude, a made a few requests for change.

As an aside, I'm not 100% crazy about the `WorkflowNameEditor` component doing a patch. But I think that doing it with `send_update` etc is probably more work than we want to do now. But lets keep in our minds that we shouldn't be patching back to the same URL just to trigger a refresh of the dom.

lib/lightning_web/live/components/layout.ex
wanecode: Agreed. Had some hesitations of changing all the page titles.

PR #494, opened 2022-12-09
Work-order-history-doesn't-work-properly-with-pagination

### LINKED ISSUE #443: Work order history doesn't work properly with pagination
By determining uniques _after_ pagination, we may not get the right number per page. We should re-write this query so that we can distinct _before_ calling `Repo.paginate`

wanecode:
## Related issue

Fixes #443 Work order history doesn't work properly with pagination

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -10,6 +10,8 @@ import Config
 config :lightning,
   ecto_repos: [Lightning.Repo]
 
+config :lightning, Lightning.Repo, types: Lightning.PostgrexTypes
+
 # Configures the endpoint
 config :lightning, LightningWeb.Endpoint,
   url: [host: "localhost"],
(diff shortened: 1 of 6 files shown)


amberrignell:
Just one small comment - I think you forgot to update the number that's displayed

In the screenshot below, it says "11 to 20 of 16 results" when it should say "11 to 16"
Screenshot 2022-12-14 at 16 44 41

config/config.exs
```diff
@@ -10,6 +10,8 @@ import Config
 config :lightning,
   ecto_repos: [Lightning.Repo]
 
+config :lightning, Lightning.Repo, types: Lightning.PostgrexTypes
```
stuartc, line 13: @wanecode do we still need this since you used `nullif`?

stuartc (approved):
Other than my comment on adding PostgrexTypes this is good to me.

Please let me know if we still need it, and if not please remove it from `config.exs`.

config/config.exs
wanecode: @stuartc. Yes . the max clause needs infinity. Or we can hard a "max date"

amberrignell:
Looks good, thanks Mamadou :)

PR #495, opened 2022-12-09
Handle time out jobs

### LINKED ISSUE #474: Runs over 65s never get updated
When a run takes more than 65s, it remains showing a status of "pending" in the UI.

1. Create a workflow with a job that has a setTimeout of more that 65s
```
fn(state => {
  return new Promise((resolve, reject) => {
    setTimeout(() => {
      console.log('wait, and then resolve');
      resolve(state);
    }, 65 * 1000);
  });
});
```
2. Click on run
3. Go to runs history and see that it stays as "pending" and never updates

happen.
If a job times out, it should throw an error and display as "Timeout" status in the runs history.

elias-ba:

This PR ships code that allows to mark timeout runs as `Failure`. It also introduces an Oban Manager to Lightning for a better Oban exception / errors handling.

This PR doesn't solve the root problem stated in the issue #474. It's more a workaround that captures any killed run that Engine returns as a Failing run. This is strong solution but needs more reflection and maybe a better engineering. This is the reason why we have decided to open [another issue](https://github.com/OpenFn/Lightning/issues/508) to think more in-depth about how Engine handles timeout jobs and other (system) failures that can happen with Rambo. The outcome of solving that issue will also produce a result on how to run the downstream jobs associated to any failing job.

cc: @taylordowns2000 @amberrignell 

## Related issue

Fixes #474 

### DIFF
--- a/lib/lightning/oban_manager.ex
+++ b/lib/lightning/oban_manager.ex
@@ -1,33 +1,32 @@
 defmodule ObanManager do
   require Logger
 
-  alias OpenFn.{
-    Repo,
-    Run,
-    RunService,
-    FlowService,
-    ObanQuery
-  }
+  # def handle_event([:oban, :circuit, :open], _measure, meta, _pid),
+  #   do: Logger.info("Circuit open #{inspect(meta, pretty: true)}")
 
-  def handle_event([:oban, :circuit, :open], _measure, meta, _pid),
-    do: Logger.info("Circuit open #{inspect(meta, pretty: true)}")
+  # def handle_event([:oban, :circuit, :trip], _measure, meta, _pid) do
+  #   Logger.error("Circuit tripped with #{inspect(meta, pretty: true)}")
 
-  def handle_event([:oban, :circuit, :trip], _measure, meta, _pid) do
-    Logger.error("Circuit tripped with #{inspect(meta, pretty: true)}")
+  #   context =
+  #     Map.take(meta, [:name])
+  #     |> scrub_context
 
-    context =
-      Map.take(meta, [:name])
-      |> scrub_context
-
-    Sentry.capture_exception(meta.error,
-      stacktrace: meta.stacktrace,
-      message: meta.message,
-      extra: context,
-      tags: %{type: "oban"}
-    )
-  end
+  #   Sentry.capture_exception(meta.error,
+  #     stacktrace: meta.stacktrace,
+  #     message: meta.message,
+  #     extra: context,
+  #     tags: %{type: "oban"}
+  #   )
+  # end
 
   def handle_event([:oban, :job, :exception] = first, measure, meta, _pid) do
+
+    # Lightning.Invocation.update_run(run, %{
+    #   finished_at: DateTime.utc_now(),
+    #   exit_code: result.exit_code,
+    #   log: scrubbed_log
+    # })
+
     IO.inspect(first, label: "first")
     IO.inspect(measure, label: "measure")
     IO.inspect(meta, label: "meta")
@@ -54,6 +53,7 @@ defmodule ObanManager do
     # error = meta.error
 
     # dead? = meta.attempt >= meta.max_attempts
+
     # timeout? = Map.get(error, :reason) == :timeout
 
     # if timeout? do
@@ -88,17 +88,17 @@ defmodule ObanManager do
     #     ])
 
     # handle_crashed_run(
-    #     get_in(args, ["run", "id"]),
-    #     get_in(args, ["state"]),
-    #     error,
-    #     timeout?
-    #   )
+    #   get_in(args, ["run", "id"]),
+    #   get_in(args, ["state"]),
+    #   error,
+    #   timeout?
+    # )
   end
 
   @doc """
   Attempt to create a record of the crash, updating an existing run in the db.
   """
-  def handle_crashed_run(run_id, state, error, timeout?) do
+  # def handle_crashed_run(run_id, state, error, timeout?) do
     # run = Repo.get(Run, run_id)
 
     # result =
@@ -153,12 +153,12 @@ defmodule ObanManager do
     # |> RunService.ensure_result_handled(run)
     # |> FlowService.ensure_flow_controlled(state)
     # |> clean_timeouts_for_run()
-  end
+  # end
 
   @doc """
   Attempt to create a record of the crash, updating an existing run in the db.
   """
-  def handle_orphaned_run(run_id, state) do
+  # def handle_orphaned_run(run_id, state) do
     # run = Repo.get(Run, run_id)
 
     # %Engine.Result{
@@ -180,7 +180,7 @@ defmodule ObanManager do
     # }
     # |> RunService.ensure_result_handled(run)
     # |> FlowService.ensure_flow_controlled(state)
-  end
+  # end
 
   @doc """
   Given a run which we are certain has been handled properly (i.e., exit_code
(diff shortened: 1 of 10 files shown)


lib/lightning/oban_manager.ex
```diff
@@ -0,0 +1,203 @@
...
+    #       timeout?
+    #     ])
+
+    # handle_crashed_run(
+    #     get_in(args, ["run", "id"]),
+    #     get_in(args, ["state"]),
+    #     error,
+    #     timeout?
+    #   )
+  end
```
stuartc, line 88: @elias-ba I'd like there to be a unit test for this function at least. What we discussed was that we weren't going to test that the function is _called_ (we can assume Oban is tested) - what do **do** want to test is that when it's called that it marks the job as finished.

lib/lightning/oban_manager.ex
```diff
@@ -0,0 +1,203 @@
...
+
+  #   ObanQuery.timeouts_for_run(id)
+  #   |> Repo.delete_all()
+  # end
+
+  defp scrub_context(context) do
+    {_private_state, safe_context} = pop_in(context, [:args, "state"])
+    safe_context
+  end
+end
```
stuartc, line 203: I have no idea what this is for in the context of Lightning, we don't pass state around in our job arguments. Unless theres something else we use it for please remove it.

lib/lightning/pipeline.ex
```diff
@@ -28,6 +28,11 @@ defmodule Lightning.Pipeline do
     |> process()
   end
 
+  # @impl Oban.Worker
+  # def timeout(_job),
+  #   do:
+  #     Application.get_env(:lightning, :max_run_duration)
+
```
stuartc, line 35: Why is this commented out? Where else is the timeout set?

stuartc (changes requested):

lib/lightning/pipeline.ex
elias-ba: @stuartc I introduced this to be able generate timeout exceptions and handle them in the `handle_event([:oban, :job, :exception], measure, meta, _pid)` in the Oban manager but then I could test that the exception is fired but I completely couldn't capture it in the Oban manager. I left it commented out to get your attention and pair on it during our 1-1. This is very related to your comment [here](https://github.com/OpenFn/Lightning/pull/495#discussion_r1045711919).

amberrignell:
Screenshot 2022-12-12 at 14 08 41
-> could you make sure it says "Timeout" instead of failure next to the run itself too ?

Screenshot 2022-12-12 at 14 08 22
-> could we have an exit code ?

@elias-ba looking much better ! Can you look at these two screenshots above and let me know if you need to discuss ?

lib/lightning/oban_manager.ex
elias-ba: Done

lib/lightning/pipeline.ex
elias-ba: All good now

lib/lightning/oban_manager.ex
```diff
@@ -0,0 +1,72 @@
...
+    Sentry.capture_exception(meta.error,
+      stacktrace: meta.stacktrace,
+      message: meta.message,
+      extra: context,
+      tags: %{type: "oban"}
+    )
+  end
+
+  def handle_event([:oban, :job, :exception], measure, meta, _pid) do
+    Repo.get!(AttemptRun, meta.args["attempt_run_id"])
```
stuartc, line 31: We have more than one module that uses `Oban.Worker`. So if `Lightning.Accounts` for example crashes this function will also fail (since `"attempt_run_id"` won't be in `meta.args`.

Please add a condition to check the error is from `Pipeline` before updating the Run.

See: https://github.com/sorentwo/oban#reporting-errors

From what I can see here:

```
meta.job
|> Map.take([:id, :args, :meta, :queue, :worker])  # <== I'm guessing worker will be Lightning.Pipeline
```

_NOTE_: I see the Oban README says `meta.job` and we use just `meta`. Can you please double check which one is correct.

stuartc: We do still want to do the rest of the Sentry and logging stuff no matter the worker.

lib/lightning_web/live/run_live/components.ex
```diff
@@ -225,8 +238,8 @@ defmodule LightningWeb.RunLive.Components do
     """
   end
 
-  attr :line, :string, required: true
-  attr :num, :integer, required: true
+  attr(:line, :string, required: true)
+  attr(:num, :integer, required: true)
```
stuartc, line 242: How are you formatting these files? There are rules set up in `.formatter.exs` to treat things like `attr` and in the case of our schemas to not wrap `field` in parentheses.

Can you check your VSCode settings and perhaps try and add this (if you don't already have it):

```
  "[elixir]": {
    "editor.wordBasedSuggestions": true,
    "editor.trimAutoWhitespace": true,
    "editor.defaultFormatter": "JakeBecker.elixir-ls"    <=== this
  },
```

test/lightning/crash_test.exs
```diff
@@ -0,0 +1,116 @@
...
+          :run,
+          Run.changeset(%Run{}, %{
+            project_id: job.workflow.project_id,
+            job_id: job.id,
+            input_dataclip_id: dataclip.id
+          })
+        )
+        |> Repo.insert()
+
+      assert attempt_run.run.finished_at |> Kernel.is_nil()
```
stuartc, line 98: Do you have to write this like that?

What about `assert attempt_run.run.finished_at |> is_nil()` or `refute attempt_run.run.finished_at` since `nil` is falsy.

test/lightning/crash_test.exs
```diff
@@ -0,0 +1,116 @@
...
+          args: %{"attempt_run_id" => attempt_run.id},
+          error: %CaseClauseError{term: :killed},
+          stacktrace: []
+        },
+        nil
+      )
+
+      run = Repo.get!(Run, attempt_run.run.id)
+
+      assert run.finished_at |> Kernel.is_nil() |> Kernel.not()
```
stuartc, line 113: Again, please use `refute` instead of `assert` when testing for a negative condition.

With `refute` you don't need the not.

A tip here, reaching into `Kernel` is generally a smell, since a large part of the language itself provides patterns and syntax that (internally use `Kernel`) make calling it directly unnecessary.

stuartc (changes requested):

lib/lightning/oban_manager.ex
elias-ba: Fixed

elias-ba: For `meta.job` vs `meta` I have double checked and the correct map to use is `meta.job`.

test/lightning/crash_test.exs
elias-ba: This is much better. Thanks for the suggestion

test/lightning/crash_test.exs
elias-ba: Thanks a lot man for the suggestions. This is much better and I have updated the PR

lib/lightning_web/live/run_live/components.ex
elias-ba: Done. Thanks

lib/lightning/oban_manager.ex
elias-ba: I say the correct map to use is `meta.job` because `:queue` doesn't exist in `meta` but is inside `meta.job`

amberrignell:
Looks good to me ! Thanks Elias. 
@taylordowns2000 for context when you're reviewing, we are marking these as failure for the time being because the error is not specific enough for us to be sure when it's a timeout or a different error. Elias has created an issue to improve this down the line.

PR #499, opened 2022-12-12
Fix Credential creation on demo.openfn.org

### LINKED ISSUE #445: New credential not working on demo
When I try to create a new credential on demo and select DHIS2 or HTTP, the credential window closes and no credential is saved. 

Weirdly this works fine when running main locally ...

1. Go to demo
2. Click on a job and click new credential
3. Select http or dhis2
4. See the form closes

happen.
Display the DHIS2 or HTTP form and allow the user to save their credential. 

https://images.zenhubusercontent.com/454419290/ec15385e-4686-4ab9-a603-1e412e4d9bac/screen_recording_2022_11_22_at_11_27_02.mov

stuartc:

The issue was because the credential schemas were not included in the image.

I also made two other changes:

- Sentry wasn't correctly set up to catch exceptions outside of `Plug`, meaning liveview exceptions were never logged.
- Silence 404s on Sentry, no point gathering errors from bots poking around.

## Related issue

Fixes #445 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,6 +107,7 @@ ENV ADAPTORS_PATH /app/priv/openfn
 # Only copy the final release and the adaptor directory from the build stage
 COPY --from=builder --chown=lightning:root /app/_build/${MIX_ENV}/rel/lightning ./
 COPY --from=builder --chown=lightning:root /app/priv/openfn ./priv/openfn
+COPY --from=builder --chown=lightning:root /app/priv/schemas ./priv/schemas
 
 USER lightning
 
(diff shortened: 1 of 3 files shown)


stuartc:
@taylordowns2000 could you take a peek here, was actually a really simple fix.

There is some work that would be desirable in the future, that being that because we're not serving vanilla Plug/controller pages, and our exceptions are essentially logged via a Logger (`Logger.add_backend(Sentry.LoggerBackend)`) we have zero context on the error.

So I think we need to play around with places to add context, like the user and maybe the URL they are on etc.

taylordowns2000 (approved):

PR #500, opened 2022-12-12
Move Engine Handler et. al into Lightning.Runtime

### LINKED ISSUE #442: Get new runtime into Lightning
Decouple it from engine so that it doesn't affect platform too.

stuartc:

Before using this, make sure to run: `mix lightning.install_runtime`.

The older `mix openfn.runtime.install` task is no longer available.

## Related issue

Fixes #442 

## Follow up considerations

- We no longer have tests for checking the version of the `cli` and the adaptor from the logs, I've added an issue here: https://github.com/OpenFn/kit/issues/130 
- `cli` and our new runtime logs using prefixed tags like this: `[CLI]`, `[R/T]` etc, we should consider filtering these and tagging individual log lines in order to control what we show users.
- We're not using the autoinstall functionality of `cli` just yet, we need a plan for this - where we still preload a bunch of adaptors for deployment; but store them in a way that is compatible with `cli`.


stuartc:
@taylordowns2000 @wanecode I have marked you as reviewers. I'm going to merge anyway for now - but please give the code a rough look over - there is _a lot_ of new code, mostly from Engine - but it's important to me that we all get a rough sense of the structure.

PR #504, opened 2022-12-13
Editor improvements

### LINKED ISSUE #279: Monaco: hide unhelpful language constructs when suggesting
## The problem

When the user hits ctrl + space in Monaco, they are presented with a bunch of suggestions.

image

Most of these are unhelpful - I doubt our users want to create AudioBuffers any time soon.

This is a bit of a problem because really when a user hits ctrl + space, what we want to suggest is:
* The API exposed by the current adaptor
* Maybe some code snippets (like `for` or `if`)
* Some key "globals" from our environment, like the state object.

### 1. Create our own lib.d.ts

We can pass an option called [noLib](https://www.typescriptlang.org/tsconfig/#noLib) to the typescript compiler, which prevents it from loading the default type definitions in [lib.d.ts](https://github.com/microsoft/TypeScript/blob/main/lib/lib.d.ts).

Unfortunately this removes even very basic language definitions like arrays, and typescript then can't compile anything, so you lose all type tracking.

We could provide our own lib.d.ts file with a custom environment. But who knows what we can really afford to cut out - I expect we'll still end up with an environment full of (and I say this with the greatest respect) rubbish.

The trick here is that we don't really want to remove this stuff from the environment, we just don't want to suggest it. Which leads to option 2....

### 2. Write our own suggester

Monaco does give us an API to write our own suggestion function via [registerCompletionItemProvider](https://microsoft.github.io/monaco-editor/api/modules/monaco.languages.html#registerCompletionItemProvider)

Documentation is minimal and I suspect this will be a difficult thing to do. Does the provider have any idea what sort of context we're in? How much of our own analysis do we have to write?

There is a simple example in the [monaco playground](https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-completion-provider-example) but it doesn't really address my questions around complexity.

## Footnote

I have spun this issue out of #255 as it's probably quite a difficult one and beyond the scope of the original work.

josephjclark:
This PR contains two improvements to the code editor:

1) Tooltips and code assist should now be available without adding an import statement
2) Code complete suggestions are now much smarter (no environment junk or keywords are made available)

The  diff is tiny but the benefit substantial, so I wanted to make this available to merge ASAP.

## Issues

Closes #279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to
 
 ## [Unreleased]
 
+- Better code-assist and intelliense in the Job Editor
+
 ### Added
 
 - Added a Delete job button in Inspector
@@ -24,8 +26,8 @@ and this project adheres to
 - Updated @openfn/describe-package to 0.0.10
 - Create an follow a manual Run from the Job Inspector
 - View all workflows in a project on the workflows index page
-- Move @openfn/workflow-diagram into the application, the NPM module
-  is now deprecated.
+- Move @openfn/workflow-diagram into the application, the NPM module is now
+  deprecated.
 
 ### Fixed
 
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -28,6 +28,13 @@ const options: MonacoProps['options'] = {
 
   codeLens: false,
   wordBasedSuggestions: false,
+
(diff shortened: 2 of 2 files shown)


stuartc (approved):
This is great, took me a few seconds to work out how this worked. Very nice. Thanks!

PR #505, opened 2022-12-13
adaptor-docs: import component into this repo

josephjclark:
This PR just imports the adaptor-docs component out of kit and into this repo.

There should be no diff in the core component here and in kit.

**Update**

This now includes fixes from #506.

### DIFF
--- a/assets/js/adaptor-docs/Docs.tsx
+++ b/assets/js/adaptor-docs/Docs.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback } from 'react';
-import Docs from '@openfn/adaptor-docs';
+import DocsPanel from '@openfn/adaptor-docs';
 
 type DocsProps = {
   adaptor: string; // name of the adaptor to load. aka specfier.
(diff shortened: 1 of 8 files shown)


josephjclark:
@amberrignell This has been code-approved, but maybe you can run a bit of basic QA before merging?

* Caption should no longer appear in examples
* You should be able to add multiple snippets
* Markdown should render properly (ie you'll see lists and stuff)
* If for any reason there's an error (and there will be if you select the `latest` version of an adaptor) you'll get an error message, rather than infinite loading

josephjclark:
There are two blocks on this branch and I don't think I can resolve them until the new year :(

1) The latest release of `describe-package` does not seem to include the latest fixes, so the caption-example fix is AWOL
2) Since moving to `jsdelivr` we can no longer fetch files from `@latest`, which is actually really harmful to the jobs panel (it kills both the editor and the docs panel). 

The first issue should be fixed by re-releasing and maybe version bumping `describe-package`. The second is resolved by fixing #509 (which I think is really easy). Alternatively we can roll back the `jsdelivr` fix in `describe-package`.

Either way I need to wait for @stuartc's input on this.

Issue #503, opened 2022-12-13
adaptor-docs: <caption> tag included in examples


josephjclark:
When rending examples in the `adaptor-docs` component, sometimes there's a caption in the example

![image](https://user-images.githubusercontent.com/7052509/207362183-909958d0-1f7b-466b-aa36-40018e98883e.png)

This just needs to be rendered more smartly

(Raised this here and not in kit as the component will be moving - in any case I intend to fix it soon)

Issue #502, opened 2022-12-13
adaptor-docs: render markdown properly


josephjclark:
Some JSDoc documentation in the adaptors contains markdown syntax in the description. Ie, `dhis2.create`.

We should be mindful to render markdown test as HTML.

PR #506, opened 2022-12-13
adaptor-docs fixes

josephjclark:
Several fixes to adaptor docs.

Note that this is branched off #505 and so is targeted at that (for now)

* [x] #503 Render examples with captions
* [x] Fix  some strange behaviour with add (adding multple items should work better now)
* [x]  #502 Render markdown
* [x] Handle errors a bit better (especially with @latest)

## Error handling

if you set an adaptor to `latest` and go to the editor, the docs panel will fail to load. Well, actually, it'll just say "loading" forever.

That issue will be fixed separately. But for now you'll at least see an error message should something go wrong.

image

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -11,12 +11,13 @@
   "license": "ISC",
   "dependencies": {
     "@monaco-editor/react": "^4.4.5",
-    "@openfn/describe-package": "^0.0.10",
+    "@openfn/describe-package": "file:../../kit/dist/openfn-describe-package-0.0.11.tgz",
     "@tailwindcss/container-queries": "^0.1.0",
     "@tailwindcss/forms": "^0.5.3",
     "classcat": "^5.0.4",
     "cronstrue": "^2.14.0",
     "elkjs": "^0.8.2",
+    "marked": "^4.2.4",
     "monaco-editor": "^0.34.0",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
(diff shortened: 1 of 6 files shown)


josephjclark:
I've added `marked` (an extra 100kb of js to the pile) so that we can parse markdown on the fly. This only added a few minutes and makes a small improvement to the presentation of docs.

This is reasonably safe since the documentation comes out of code we control. Even so, I think we should be sanitising and parsing in adaptor-docs really. Maybe I should go off and do that instead? Or raise a ticket to do it soon.

assets/package.json
```diff
@@ -11,12 +11,13 @@
   "license": "ISC",
   "dependencies": {
     "@monaco-editor/react": "^4.4.5",
-    "@openfn/describe-package": "^0.0.10",
+    "@openfn/describe-package": "file:../../kit/dist/openfn-describe-package-0.0.11.tgz",
```
josephjclark, line 14: This is temporary and will be reverted, but I kinda had to check it in...

stuartc (approved):
All good here, changes are nice and clean - thanks!

stuartc:
@amberrignell when you're done reviewing this, can you delete the branch please. 
unless @josephjclark, would it be easier to review in the other branch?

josephjclark:
@stuartc Yeah, let's delete this and ask @amberrignell  to do some QA on the other branch and merge when she's happy. I'll sort that out.

PR #516, opened 2022-12-16
Create new project for user on registration

### LINKED ISSUE #430: Create new project for user on registration
### User story
As a user that is registering for Lightning, I would like to be added to a new project as an admin when I register, so that I can start building workflows

Demo workflow?

### Implementation notes
To avoid having lots of projects with the same name, generate a random name for the project the user gets added to (same as platform).

elias-ba:

This PR ships code that allows to create sample project for a new registered user. The feature is enabled thru the `INIT_PROJECT_FOR_NEW_USER` environment variable. If that variable is set to `true` then whenever a user is creating an account, the system will create a project named `openhie_project` and configure a sample workflow in it (the one in demo.ex). cc @amberrignell 

## Related issue

Fixes #430 

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -58,6 +58,11 @@ config :lightning,
        System.get_env("QUEUE_RESULT_RETENTION_PERIOD", "60")
        |> String.to_integer()
 
+config :lightning,
+       :init_project_for_new_user,
+       System.get_env("INIT_PROJECT_FOR_NEW_USER", "false")
+       |> String.to_atom()
+
 # If you've booted up with a SENTRY_DSN environment variable, use Sentry!
 config :sentry,
   filter: Lightning.SentryEventFilter,
--- a/.env.example
+++ b/.env.example
@@ -120,3 +120,6 @@ LIGHTNING_LISTEN_ADDRESS=0.0.0.0
 #DOCKER_POSTGRES_MEMORY=0
 #DOCKER_WEB_CPUS=0
 #DOCKER_WEB_MEMORY=0
+
+# Give this variable the value of true if you want the system to create a sample project for a new registered user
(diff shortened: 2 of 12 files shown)


taylordowns2000 (changes requested):
Hey @elias-ba , this isn't ready:

1. Please read the "implementation notes" section of the issue.
2. The user registration should not use "demo" — it's not a demo. The demo module could use a project setup helper.

I'm going to implement these changes now, but wanted you to see them when you're back from holiday. Also, @amberrignell , the initial project is not specified in the issue. What initial project should be created?

elias-ba:
> Hey @elias-ba , this isn't ready:
> 
> 1. Please read the "implementation notes" section of the issue.
> 2. The user registration should not use "demo" — it's not a demo. The demo module could use a project setup helper.
> 
> I'm going to implement these changes now, but wanted you to see them when you're back from holiday. Also, @amberrignell , the initial project is not specified in the issue. What initial project should be created?

@taylordowns2000 I am interested in knowing what you have added / changed. For context @amberrignell and I did exchange a lot about this issue and why is it needed. We did agree on a way of implementing it before I start. This also the reason why I did reuse some part of what we did in "Demo" module.

config/runtime.exs
```diff
@@ -58,6 +58,11 @@ config :lightning,
        System.get_env("QUEUE_RESULT_RETENTION_PERIOD", "60")
        |> String.to_integer()
 
+config :lightning,
+       :init_project_for_new_user,
+       System.get_env("INIT_PROJECT_FOR_NEW_USER", "false")
+       |> String.to_atom()
```
stuartc, line 64: I did not know that turning "false" into an atom made a boolean :ok_hand:

PR #520, opened 2022-12-19
482 post input to webhook from the input tab

wanecode:
## Related issue

Fixes #


taylordowns2000 (changes requested):
@wanecode , this doesn't work:

1. When the trigger type is cron (image 1)
2. When the trigger type is webhook but there isn't any prior dataclip (image 2)

Please fix it for both cases.

image

image

taylordowns2000:
qa by taylor

PR #523, opened 2022-12-21
509 job panel should always pass an explicit version specifier

### LINKED ISSUE #509: Job Panel should always pass an explicit version specifier
### User story

The adaptor-docs panel won't populate its contents properly if the adaptor version is set to `latest`

### Details

Changes in how `describe-package` works over in kit mean we need to always pass an explicit adaptor version to the docs builder. This is probably correct and a good thing.

But it does mean docgen no longer supports `@latest` as a valid version specifier.

Lightning does have all the information it needs to know which the latest version of an adaptor is. Either we should drop the word `latest` entirely, or when `latest` is selected, it passes the actual latest version through. That's probably worth some discussion.

### Implementation notes

In `job_builder_component.ex` the adaptor is passed as a `name@version` specifier to the `adaptor-docs`  component. 

Lightning needs to ensure that `@latest` is never sent, but an actual version number is passed instead.

### Release notes
.

### User acceptance criteria

API documentation for an adaptor is shown when the latest adaptor version is selected.

wanecode:
## Related issue

Fixes #509 Job Panel should always pass an explicit version specifier

### DIFF
--- a/lib/lightning/adaptor_registry.ex
+++ b/lib/lightning/adaptor_registry.ex
@@ -172,6 +172,19 @@ defmodule Lightning.AdaptorRegistry do
     {:reply, versions, state}
   end
 
+  @impl GenServer
+  def handle_call({:latest_for, module_name}, _from, state) do
+    latest =
+      state
+      |> Enum.find(fn %{name: name} -> name == module_name end)
+      |> case do
+        nil -> nil
+        %{latest: latest} -> latest
+      end
+
+    {:reply, latest, state}
+  end
+
   @doc """
   Get the current in-process list of adaptors.
   This call will wait behind the `:continue` message when the process starts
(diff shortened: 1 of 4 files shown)


taylordowns2000 (changes requested):
@wanecode i might be misunderstanding the code, but i don't think this actually gets to the desired behavior. first, a few quick things:

1. change what currently says `latest (vM.m.p)` to `vX.Y.Z (auto-upgrade)`
2. ensure that you can select _either_ `v4.2.3` or `v4.2.3 (auto-upgrade)` for http. currently, there's no way to choose to lock onto the highest version.
3. in your test, it looks like you consider v1.6 the latest in a list that includes v2+. why?

image

---

then, when that's done... can you confirm that:

1. by selecting http `v4.2.3 (auto-upgrade)` you'll get the latest version each time you boot up the app — in other words, if `v4.2.4` is detected, that version will be loaded.
2. this whole thing works produces _exactly_ the same behavior that it did before your changes with respect to how versions are chosen for running a job set to "latest", except that (a) the words appear as indicated above and (b) when you click on the editor panel for the job, it _knows_ which version the current "latest" version is so that it can pull in the correct documentation.

wanecode:
@taylordowns2000 Changes have been made so that either v4.2.3 or v4.2.3 (auto-upgrade) can be selected. 
When auto-upgrade is selected the adaptor stored in database is "....@latest" and the correct version is fetched when needed and passed to the Editor and Docs components. So i confirmed points 1 and 2. cc @amberrignell @stuartc

taylordowns2000 (approved):

PR #525, opened 2022-12-23
273 allow a user to delete a credential

### LINKED ISSUE #273: Allow a user to delete a credential
### User story
As a user that is no longer using a credential, I would like to delete it, so that it doesn't clutter up my credentials page.

### Details
When I delete a credential, it should not delete the audit trail for this credential. 
If I try to delete a credential that is being used in a workflow, it should prevent me from doing so and tell me why. 

--> at the moment we block credential deletion if it has access to a project, but this doesn't mean it's being used ... how difficult would it be to check if a credential is being used within the user's workflows ?

### Implementation notes
1. See planned tests marked with `@tag :skip`
2. See delete button placement, after "Edit" in `lib/lightning_web/live/credential_live/index.html.heex` 👇 

```html.eex
|
<span>
  <%= link("Delete",
    to: "#",
    phx_click: "delete",
    phx_value_id: credential.id,
    data: [confirm: "Are you sure?"]
  ) %>
</span>
```

### Release notes 

Allow any user to delete a credential that they own. 

### Tests 
- [ ] User cannot delete a credential that is being used in a workflow
- [x] User can delete a credential that is not being used
- [x] User gets confirmation message when a credential is deleted
- [x] Credential does not get deleted from the audit trail

### QA
[ ] This has been QA’d by Amber, branch can be merged after review

wanecode:

- Credential is considered to be undeletable if a job refering to it (via project_credential) has been ran.
- To allow "deleted" record in audit trail, we removed referential integrity (credential foreign key: row_id is now a simple field)
- needs ecto.reset or ecto.migrate

## Related issue

Fixes #273 allow a user to delete a credential

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to
 - Added plausible analytics
 - Allow user to click on Webhook Trigger Node to copy webhook URL on workflow
   diagram
+- Allow any user to delete a credential that they own
 
 ### Changed
 
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -208,7 +208,41 @@ defmodule Lightning.Credentials do
 
   """
   def delete_credential(%Credential{} = credential) do
-    Repo.delete(credential)
+    is_used = job_uses_credential?(credential)
+
+    if is_used do
+      {:error,
+       %Credential{}
+       |> Ecto.Changeset.change()
+       |> Ecto.Changeset.add_error(
+         :job_using_credential,
+         "Can't delete. This credential is being used by at least one job"
+       )}
+    else
+      # Repo.delete(credential)
+      Multi.new()
+      |> Multi.delete(:credential, credential)
+      |> Multi.insert(:audit, fn _ ->
+        Audit.event("deleted", nil, credential.user_id, %{
+          before: %{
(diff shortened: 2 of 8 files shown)


taylordowns2000 (changes requested):
@wanecode , what's the reason for blocking the deletion of a credential if it's ever been used?

@stuartc , can we take this opportunity to drop the `credentials_audit` table in favor of the `audit` table which has a target `table` and `row_id` column?

I don't like having a separate table for each one, and I'm not sure what it gains us if we specifically can't maintain referential integrity because we want a log of changes to include deletions. Feels like we're getting closer and closer to the model we used in papertrail

wanecode:
> @wanecode , what's the reason for blocking the deletion of a credential if it's ever been used?
> 
> @stuartc , can we take this opportunity to drop the `credentials_audit` table in favor of the `audit` table which has a target `table` and `row_id` column?
> 
> I don't like having a separate table for each one, and I'm not sure what it gains us if we specifically can't maintain referential integrity because we want a log of changes to include deletions. Feels like we're getting closer and closer to the model we used in papertrail

@taylordowns2000 , the deletion is blocked if a job (that has at least one run) refer to the credential. For now, there is no easy way to relate a credential to a run (other than the dataclip body containing the credential body)

PR #527, opened 2023-01-02
Show input dataclip from run-history only

### LINKED ISSUE #472: Add input tab to a run page
### User story
As someone debugging a run, I'd like to view the run log, output, _and_ the input state that was used to start the run.

### Details

1. Add another tab (like OUTPUT and LOG below) which allows the user to see the input dataclip which started this run.
2. Make sure it only shows when viewing a run from the runs history, NOT on the output tab of the inspector.

**It should show from here:**

**But not from here:**

mtuchi:
## Any notes for the reviewer?
I have created a `show_input_dataclip` attribute and set it to `true` when `run_viewer` is being used from a run history page 

## Related issue

Fixes #472


mtuchi:
@taylordowns2000 I have removed the `inspect.io`

taylordowns2000 (approved):

PR #548, opened 2023-01-05
Fix tooltips in Monaco

josephjclark:
Hovering over text at the top of the Monaco editor in the Job Editor panel can result in the tooltip being clipped.

image

This PR fixes it so that the whole tooltip is always visible.

image

The problem is caused by zindex and overflow controls in the parenting dom hierarchy. Fixing the parent page will result in overflow hell and I'm not up for that.

Instead, we can tell Monaco to render its tooltips and menus outside of its own dom.

The Editor component will, on mount, attach a div container to `document.body` and render its overlays in there. It will be a good citizen and remove this extra div when unmounted.

While I was in the area, I caught a couple of issues in the snippet code, so I've fixed those. I'll call them out in comments.

### DIFF
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react';
+import React, { useState, useCallback, useEffect, useRef } from 'react';
 import Monaco from "@monaco-editor/react";
 import type { EditorProps as MonacoProps } from  "@monaco-editor/react/lib/types";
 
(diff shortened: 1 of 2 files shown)


assets/js/editor/Editor.tsx
```diff
@@ -99,23 +100,39 @@ export default function Editor({ source, adaptor, onChange }: EditorProps) {
...
+      fixedOverflowWidgets: true
+    })
 
-    document.addEventListener('insert-snippet', handleInsertSnippet)
-    
     return () => {
-      document.removeEventListener(handleInsertSnippet);
+      overflowNode.parentNode?.removeChild(overflowNode);
+      if (listeners.current?.insertSnippet) {
+        document.removeEventListener('insert-snippet', listeners.current.insertSnippet);
```
josephjclark, line 132: The code to remove the insert-snippet was in completely the wrong place. Strange this, it should have never worked. I have a feeling there's been a funny merge conflit.

Anyway, if you wanted to test this, you can add some logging to the listener and mount/unmount the editor a few times (by showing/dismissing the panel) and you'll see that it only ever triggers one insert-snippet event.

PR #552, opened 2023-01-06
Cache credential schemas from npm #471

### LINKED ISSUE #471: cache credential schemas from npm
### User story
As a user that would like to build a workflow, I would like to be able to create my credentials through a form, so that I know what is required and I don't have to look at the adaptor documentation.

### Details
- This list of schemas can _not_ include any schemas that rely on Oauth. They won't work. (Dynamics, GoogleSheets, others?)

### Implementation notes
Write a mix task to pull down a predefined list of credential schemas from unpackage / js deliver and show those credentials schema options in the credential form

### Release notes
- Create any credential through a form except for OAuth

### User acceptance criteria
- Given a user has just selected "new credential", when they select the type dropdown, there is the same number of credential options as adaptors. 
- When a user selects a credential for an Oauth, they can see an alert that says 'This credential requires OAuth authentication, we do not currently support it.'

wanecode:
## Related issue

Fixes #471 cache credential schemas from npm


lib/mix/tasks/install_schemas.ex
```diff
@@ -0,0 +1,126 @@
...
+    )
+    |> case do
+      {:error, _} ->
+        raise "Unable to access #{package_name}"
+
+      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
+        write_schema(package_name, body)
+
+      {:ok, %HTTPoison.Response{status_code: status_code}} ->
+        raise "Unable to fetch #{package_name} configuration schema. status=#{status_code}"
```
stuartc, line 83: I think we need to adjust this behaviour here, when a language pack doesn't have a configuration schema the whole stream stops - resulting in an inconsistent set of results. For example `language-divoc` doesn't have a schema yet.

For now I think logging a warning instead of throwing is about as strict as we can be.

test/lightning/credentials/schema_test.exs
```diff
@@ -89,17 +97,20 @@ defmodule Lightning.Credentials.SchemaTest do
       assert Ecto.Changeset.get_field(changeset, :password),
              "Should be able to find existing keys via atoms"
 
-      errors = errors_on(changeset)
+      errors = errors_on(changeset) |> IO.inspect()
```
stuartc, line 100: Please remove the inspect.

test/lightning/credentials/schema_test.exs
```diff
@@ -89,17 +97,20 @@ defmodule Lightning.Credentials.SchemaTest do
...
       schema = Schema.new(schema_map, %{"username" => "initial_username"})
 
       changeset =
         Schema.changeset(schema, %{
           "password" => "pass",
-          "hostUrl" => "http://localhost"
+          "hostUrl" => "http://localhost",
+          "number" => 100
         })
+        |> IO.inspect()
```
stuartc, line 113: And here.

lib/lightning/credentials/schema.ex
```diff
@@ -66,6 +66,9 @@ defmodule Lightning.Credentials.Schema do
 
       %{actual: "null", expected: ["string"]} ->
         add_error(changeset, field, "can't be blank")
+
+      _ ->
+        add_error(changeset, field, "can't be blank")
```
stuartc, line 71: I think we can make this a tiny bit clearer by changing the one above to:

```
%{actual: "null", expected: expected} when is_list(expected) ->
  add_error(changeset, field, "can't be blank")
```

I mean the amount of variance we get back from the JSON schema library is significant,  but for now I think the example I suggested covers both scenarios with "I wasn't expecting `null`"...

test/lightning/install_schemas_test.exs
```diff
@@ -0,0 +1,174 @@
...
+      File
+      |> expect(:rm_rf, fn _ -> nil end)
+      |> expect(:mkdir_p, fn _ -> nil end)
+      |> expect(:open!, fn "priv/schemas/asana.json", [:write] -> nil end)
+      |> expect(:open!, fn "priv/schemas/primero.json", [:write] -> nil end)
+      |> expect(:close, 2, fn _ -> nil end)
+
+      IO
+      |> expect(:binwrite, fn _, ~s({"name": "language-asana"}) -> nil end)
+      |> expect(:binwrite, fn _, ~s({"name": "language-primero"}) -> nil end)
```
stuartc, line 68: I appreciate the effort in trying to stub both the HTTP and filesystem calls, but I think it's a little overkill.

Next time, or in the future I'd propose having the location of where to persist them be stubbed to `/tmp/priv/schemas` - and while stubbing the http calls, test that the files were actually written. It's probably easier imo.

stuartc (changes requested):
Nice work dude, a few change requests - namely changing the exception in the mix task and the leftover `IO.inspects`.

test/lightning/install_schemas_test.exs
wanecode: Totally agree. Here we're just ensuring that the expected content will be written.

test/lightning/install_schemas_test.exs
```diff
@@ -116,9 +119,13 @@ defmodule Lightning.InstallSchemasTest do
         {:ok, %HTTPoison.Response{status_code: 400}}
       end)
 
-      assert_raise RuntimeError, fn ->
-        InstallSchemas.persist_schema("@openfn/language-asana")
-      end
+      {_result, log} =
+        with_log(fn ->
```
stuartc, line 123: Awesome! Haven't seen this before! :tada:

lib/lightning/credentials/schema.ex
```diff
@@ -67,7 +67,7 @@ defmodule Lightning.Credentials.Schema do
       %{actual: "null", expected: ["string"]} ->
         add_error(changeset, field, "can't be blank")
 
-      _ ->
+      %{actual: "null", expected: expected} when is_list(expected) ->
         add_error(changeset, field, "can't be blank")
```
stuartc, line 71: Ah I meant to replace the one above. Don't worry, I'll change it.

stuartc (approved):

PR #564, opened 2023-01-09
Display rerun button

### LINKED ISSUE #538: Retry button not showing
On the runs history page, the retry button does not show up even though it is clickable. 

1. Go to runs history
2. Move the mouse to the right of a run and click
3. See run is retried, BUT user cannot see the button. 

happen.

elias-ba:
## Related issue

Fixes #538 

### DIFF
--- a/lib/lightning_web/live/run_live/components.ex
+++ b/lib/lightning_web/live/run_live/components.ex
@@ -113,7 +113,7 @@ defmodule LightningWeb.RunLive.Components do
             </span>
           </.link>
           <span
-            class="pl-2 opacity-0 group-hover:opacity-100 hover:underline hover:underline-offset-2 cursor-pointer"
+            class="pl-2 group-hover:opacity-100 hover:underline hover:underline-offset-2 cursor-pointer"
             phx-click="rerun"
             phx-value-attempt_id={@attempt.id}
             phx-value-run_id={@run.id}


taylordowns2000 (commented):
@stuartc , is there something _else_ going on here? i'm unconvinced by the idea that you would have added `opacity-0` and that it would make the link completely invisible when you first added this feature. there must have been something else that later made this not visible. thoughts?

(cc @elias-ba - possible that your fix is right, but something tells me it's missing the bigger picture. i could be wrong.)

elias-ba:
> @stuartc , is there something _else_ going on here? i'm unconvinced by the idea that you would have added `opacity-0` and that it would make the link completely invisible when you first added this feature. there must have been something else that later made this not visible. thoughts?
> 
> (cc @elias-ba - possible that your fix is right, but something tells me it's missing the bigger picture. i could be wrong.)

@taylordowns2000, I have discussed briefly w/ @stuartc and @amberrignell this morning about this. The `opacity-0` was not added mistakenly. The intention behind was to make the `rerun` button appear once the mouse enter that `div` (area). But for some reasons it was not working well. Stu seems to be happy about the current fix, because we're more focusing on the features themselves than their presentation.

I see 2 options:
1 - We can decide to style that button now
2 - Leave it like this for now and later make it's stylization part of a bigger UI/Presentation issue/feature.

PR #566, opened 2023-01-09
Remove Jobs list page

### LINKED ISSUE #510: Remove jobs list page
### User story
As a new user of Lightning, I would NOT like to see the jobs list page, so that I do not get confused about what it is for and I only see my jobs in the context of their workflow

elias-ba:
## Related issue

Fixes #510 


amberrignell:
Nice one 😄

stuartc (approved):

PR #580, opened 2023-01-09
Add default adaptor and job body

### LINKED ISSUE #563: Make default job adaptor `common@latest`
Right now it defaults to asana, but no version is selected. Note that we want to go with this default setting **_instead_** of having no adaptor selected at all for a new job. (I know this was discussed earlier. Apologies for the back-and-forth.)

image

elias-ba:

1. Adaptor list defaults to `common@latest`
2. Job body defaults to `// Get started by adding operations from your adaptor here`

## Related issue

Fixes #563 


amberrignell (approved):

stuartc (approved):
All good here @elias-ba.

Just a thought here, since the job body by default isn't blank - we're allowing a user to save a job that is just a comment.

i.e. there is no warning that the job is invalid when creating a new one - I mean it's technically not invalid.

@taylordowns2000 your thoughts here? What happens now is that you won't get that little warning icon on the Body tab when creating a new job. Ideally in the future, we can detect _how_ valid a job is by analysing the ast - but that is work for the future.

PR #582, opened 2023-01-09
532 output message does not update if job is run and no output to show

### LINKED ISSUE #532: output message does not update if job is run and no output to show

When 
- case 1: a job which does not return anything is run, 
- case 2: a job which fails is run,
the following message remains on the output tab "Nothing here yet.The resulting dataclip will appear here when the run finishes successfully.". 

1. Create a job which does not return any output
Case 1: nothing returned
```
fn(state=> {
    console.log(state)
})
```

Case 2: failing job 
```
fn(state=> {
    throw "error"
})
```

2. Run it 
3. See output tab displays the following: "Nothing here yet.The resulting dataclip will appear here when the run finishes successfully."

**Expected behavior**

Case 1 (nothing returned) should display: "There is no output for this run. Check your job expression to ensure that the final operation returns something."
Case 2 (job fails) should display: "This run failed; there is no output. See the logs for more information."

wanecode:
## Related issue

Fixes #532 output message does not update if job is run and no output to show

### DIFF
--- a/lib/lightning_web/live/run_live/components.ex
+++ b/lib/lightning_web/live/run_live/components.ex
@@ -155,7 +155,7 @@ defmodule LightningWeb.RunLive.Components do
     </.toggle_bar>
     <%= if @show_input_dataclip do %>
       <div id="input_section" style="display: none;" class="@container">
-        <.dataclip_view dataclip={@run.input_dataclip} run={@run} />
+        <.dataclip_view dataclip={@run.input_dataclip} />
       </div>
     <% end %>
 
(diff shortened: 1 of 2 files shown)


amberrignell:
Looks good to me, thanks @wanecode 🥇

lib/lightning_web/live/run_live/components.ex
```diff
@@ -155,7 +155,7 @@ defmodule LightningWeb.RunLive.Components do
     </.toggle_bar>
     <%= if @show_input_dataclip do %>
       <div id="input_section" style="display: none;" class="@container">
-        <.dataclip_view dataclip={@run.input_dataclip} />
+        <.dataclip_view dataclip={@run.input_dataclip} run={@run} />
       </div>
```
stuartc, line 159: @wanecode I think instead of passing the run into the dataclip view, we should have the markup and logic here instead. I'd like to avoid adding coupling between dataclips and runs here - ideally the `dataclip_view` component has no idea about a run.

We could take the existing code further as well and have the `dataclip_view` component expect a dataclip, but I think we should at least try and avoid increasing the number of dependencies.

stuartc (changes requested):
@wanecode I have an opinion about the `dataclip_view` changes here, would you take a look and let me know if you disagree?

stuartc (approved):

PR #591, opened 2023-01-11
Refocus diagram on layout change

### LINKED ISSUE #483: Refocus the canvas when inspector opens up
### User story

When clicking on a node in the workflow-diagram, the view should re-center after the job panel opens.

At the moment, clicking on a node often results in this:

image

We should automatically re-center the view so it looks like this:

image

stuartc:

This change observes changes to the containing element and triggers a 'fitView' - which is the same behaviour when seeing the diagram for the first time.

This means it will respond to browser window size changes and when the Job panel appears.

It should be noted that if a user moves and/or zooms around, the fitView function will essentially reset the view.

We can discuss later about how we might want to improve this, such as remembering the last position the view had, and return to that. There is a lot we can do around this behaviour like zooming and focusing and zooming back out etc, but it requires some careful logic.

Unless we want that (or something different now), let's chat.

## Related issue

Fixes #483 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to
   diagram
 - Allow any user to delete a credential that they own
 - Create any credential through a form except for OAuth
+- Refit all diagram nodes on browser and container resize
 
 ### Changed
 
--- a/assets/js/workflow-diagram/component.tsx
+++ b/assets/js/workflow-diagram/component.tsx
@@ -14,6 +14,7 @@ export function mount(el: Element | DocumentFragment) {
   function update({ onNodeClick, onPaneClick, onJobAddClick }: UpdateParams) {
     return componentRoot.render(
       <WorkflowDiagram
+        ref={el}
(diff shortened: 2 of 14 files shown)


taylordowns2000 (changes requested):
looks awesome! some thangs:

1. refocus should happen when clicking the "new" button. (you said maybe this is because of a different layout for new vs. edit.
2. when clicking "new", the previously highlighted node should lose focus/highlight.
3. when closing the job panel via "cancel", the window should refocus also.
4. when closing the job panel via "cancel", the previously highlighed job should lose focus/highlight.

stuartc:
@taylordowns2000 so maintaining _which_ one must be selected was pretty tough and I was going to do it in a separate PR, but I think I've got it... 

I also managed to make the trigger 'unselectable' which makes the code to maintain which one is selected much easier. 

Anyway please have another look. :pray:

taylordowns2000 (approved):
perfect! thank you

PR #600, opened 2023-01-16
Project settings > project details page

### LINKED ISSUE #579: Add project settings and project settings > project pages
User story
As a project administrator, I want a place to view all of my project settings. 
As a project administrator, I want to be able to change a project name and description. 

Details
Add a project settings tab to the navbar. 
Add a 'project' section ('credentials' and 'collaboration' will be added in subsequent issues) where an admin user can change the name and description for a project. 

The 'Danger zone' and deleting a project is out of scope for this issue. 

Implementation notes
url should be projects/:uuid/settings
you will need to add a description to the projects table 

Release notes

User acceptance criteria
- [x] there is a project setting tab on the navbar
- [x] given a user that does not have an admin role, they can view project settings > project but are unable to perform any actions (rename project or change description)
- [x] an admin is able to change the project name
- [x] an admin is able to change a project description

elias-ba:

This PR ships code to add a project setting page with all the constraints needed and the different tabs.

Much of this work was done in pair programming with @mtuchi 🥇 

## Related issue

Fixes #579 

### DIFF
--- a/lib/lightning/projects/policy.ex
+++ b/lib/lightning/projects/policy.ex
@@ -8,6 +8,12 @@ defmodule Lightning.Projects.Policy do
   alias Lightning.Projects
   alias Lightning.Projects.Project
 
+  # Users with admin level access to a project ccan access project settings with read/write rights
+  def authorize(action, %User{} = user, %Project{} = project)
+      when action in [:access_project_settings] do
+    Projects.get_project_user_role(user, project) == :admin
+  end
+
   # Project members can read a project
   def authorize(action, user, %Project{id: _id} = project)
       when action in [:read, :index] do
--- a/lib/lightning/projects/policy.ex
+++ b/lib/lightning/projects/policy.ex
@@ -8,9 +8,8 @@ defmodule Lightning.Projects.Policy do
   alias Lightning.Projects
   alias Lightning.Projects.Project
 
-  # Users with admin level access to a project ccan access project settings with read/write rights
-  def authorize(action, %User{} = user, %Project{} = project)
-      when action in [:edit] do
+  # Users with admin level access to a project can access project settings with read/write rights
+  def authorize(:edit, %User{} = user, %Project{} = project) do
     Projects.get_project_user_role(user, project) == :admin
   end
 
--- a/lib/lightning/projects/policy.ex
+++ b/lib/lightning/projects/policy.ex
@@ -10,7 +10,7 @@ defmodule Lightning.Projects.Policy do
 
   # Users with admin level access to a project ccan access project settings with read/write rights
   def authorize(action, %User{} = user, %Project{} = project)
-      when action in [:access_project_settings] do
+      when action in [:edit] do
     Projects.get_project_user_role(user, project) == :admin
   end
 
--- a/lib/lightning_web/live/project_live/settings.ex
+++ b/lib/lightning_web/live/project_live/settings.ex
@@ -29,7 +29,7 @@ defmodule LightningWeb.ProjectLive.Settings do
   end
 
   defp apply_action(socket, :index, _params) do
-    socket |> assign(:page_title, "Projects settings")
+    socket |> assign(:page_title, "Project settings")
   end
 
   @impl true
--- a/lib/lightning_web/live/project_live/settings.ex
+++ b/lib/lightning_web/live/project_live/settings.ex
@@ -48,9 +48,10 @@ defmodule LightningWeb.ProjectLive.Settings do
 
   defp save_project(socket, project_params) do
     case Projects.update_project(socket.assigns.project, project_params) do
-      {:ok, _project} ->
+      {:ok, project} ->
         {:noreply,
          socket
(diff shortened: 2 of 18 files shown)


amberrignell (changes requested):
@mtuchi @elias-ba looking good ! 2 small comments
- when I change a project name, the project name displayed in the navbar does not update (unless I reload the page)
- could you change the name in the navbar from 'project settings' to 'Settings' so that it looks nicer ? (you can leave the title on the page itself as Project settings)

amberrignell:
Looks good to me, thank you for addressing my comments :)

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -0,0 +1,61 @@
...
+    changeset =
+      socket.assigns.project
+      |> Projects.change_project(project_params)
+      |> Map.put(:action, :validate)
+
+    {:noreply, assign(socket, :changeset, changeset)}
+  end
+
+  def handle_event("save", %{"project" => project_params} = _params, socket) do
+    save_project(socket, project_params)
```
stuartc, line 46: I'm a little concerned that we are relying solely on disabled fields to protect changes to the project. I didn't manage to easily break this, but I think with a well crafted LiveSocket event in Javascript on could easily save params that aren't in the form.

I don't think this branch is the place to address this, but we should have a discussion about approaches that we can take to lock this stuff down

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -0,0 +1,112 @@
...
+      <div class="pr-4 pt-0 sticky top-0 self-start rounded-md">
+        <div class="flex-none">
+          <LightningWeb.Components.Common.tab_bar
+            orientation="vertical"
+            id={@project.id}
+            default_hash="settings"
+          >
+            <LightningWeb.Components.Common.tab_item
+              orientation="vertical"
+              hash="settings"
```
stuartc, line 18: I think the default hash should be `project` no? and the project tab should have a `project` hash? Right?

I mean I guess I see why it's called `settings` since it's the project settings..
But imo the uri looks weird with `.../settings#settings`.

Not super strong opinions on this.. your thoughts?

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -0,0 +1,112 @@
...
+              phx-submit="save"
+            >
+              <div class="hidden sm:block" aria-hidden="true">
+                <div class="py-2"></div>
+              </div>
+              <div class="grid grid-cols gap-6">
+                <div class="md:col-span-2">
+                  <LightningWeb.Components.Form.text_field
+                    form={f}
+                    disabled={!@is_user_admin}
```
stuartc, line 64: Again for context, this should be in the form of `!@can_edit_project` or similar... basically my rough rule for these variables is to start with the word `can_...` and use that to find the verb and subject..

stuartc (changes requested):
This is really great dude, super stoked with the reuse of the tab components :1st_place_medal: .

Can you address my comments about using a policy module for determining if a user can edit the project settings please.

lib/lightning_web/live/project_live/settings.html.heex
elias-ba: I agree. This is done, thanks.

lib/lightning_web/live/project_live/settings.html.heex
elias-ba: Done

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  on_mount({LightningWeb.Hooks, :project_scope})
+
+  @impl true
+  def mount(_params, _session, socket) do
+    {:ok,
+     socket
+     |> assign(
+       active_menu_item: :settings,
+       is_user_admin:
```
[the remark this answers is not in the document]
elias-ba: Done

lib/lightning/projects/policy.ex
```diff
@@ -8,6 +8,12 @@ defmodule Lightning.Projects.Policy do
   alias Lightning.Projects
   alias Lightning.Projects.Project
 
+  # Users with admin level access to a project ccan access project settings with read/write rights
+  def authorize(action, %User{} = user, %Project{} = project)
+      when action in [:access_project_settings] do
```
stuartc: No need for a guard clause here, can do this:

```
def authorize(:access_project_settings, %User{} = user, %Project{} = project)
```

We only use guard clauses (in these places) when more than one action has the same authorize logic, like a case statement but for function matches.

Also why is it called `access_project_settings` when a normal user can still see the settings - should it not be `change_project_settings` or something like that?

elias-ba: I turned it to `:edit` and removed the guard clause. See my last commit. Thanks

stuartc (approved):

PR #601, opened 2023-01-17
libcluster

taylordowns2000:
**Note, a bug in erlang `25.1` necessitates `25.2` for the local gossip strategy to work. See changes to .tool-versions**

Enables for erlang node clustering. Note that a different strategy is used for local clusters (gossip) than for kubernetes.

To test locally, start one node called `node-a`:
```sh
env $(cat .env | grep -v "#" | xargs ) iex --sname node-a -S mix phx.server
```

Then start another called `node-b`:
```sh
env $(cat .env | grep -v "#" | xargs ) LIGHTNING_LISTEN_ADDRESS=4001 PORT=4001 iex --sname node-b -S mix phx.server
```

You'll notice `[info] [libcluster:local] connected to :"node-b@your-machine"` and an annoying false-alarm about `Name conflict terminating {Swoosh.Adapters.Local.Storage.Memory` - but I've checked and this isn't something that (a) does anything bad and (b) will happen on production.

You can view all the nodes in your cluster (including the node you're iex-ing from) with:
```sh
[Node.self] ++ Node.list
```


stuartc (commented):
Cool, this looks good @taylordowns2000 - I don't know if now is the time to add some notes to the README/DEPLOYMENT docs or not; but I think some details around `K8S_HEADLESS_SERVICE` is necessary.

Also in our deployment docs, we say that the default address binding is `0.0.0.0`, but I'm seeing the loopback address in `runtime.exs` - to be fair this looks like an inconsistency that existed before this...

stuartc:
@taylordowns2000 happy to merge this once after your take on my comment above. :+1:

PR #613, opened 2023-01-20
Roles and permissions

### LINKED ISSUE #594: User roles and permissions
Review current authorization implementation to follow best practice and avoid hard-coding roles. 
https://owasp-top-10-proactive-controls-2018.readthedocs.io/en/latest/c7-enforce-access-controls.html#don-t-hardcode-roles

Currently, access controls happen on user roles in separate policy files - this pattern makes it difficult to see every single action a user with a given role can perform. 

```
  # Super users can do anything
  def authorize(_action, %User{role: :superuser}, _params), do: true

  # Regular users can't access user management page
  def authorize(action, %User{role: :user} = _user, _params)
      when action in ~w[index show]a,
      do: false

  # Default blacklist
  def authorize(_action, _user, _params), do: false
end
```

Expected:
- [x] There is one place in the app where I can see all of the permissions granted for each role (or at least a very easy way of auditing it)
- [x] All functions / possible actions in Lightning have an authorize statement around them

**Define permissions for viewer, editor and admin of a project_user:** 
Note: below list is just indication / example, we need to think about this carefully
- viewer: [view_workflow, view_project_members, view_credentials_list, view_runs] <- run a job ?
- editor: [view_workflow, edit_workflow, view_project_members, view_credentials_list, view_runs, run_job]
- admin: [view_workflow, edit_workflow, delete_workflow, add_member, remove_member, run_job, view_credentials_list, view_runs]

See roles and permissions doc for context: https://docs.google.com/spreadsheets/d/1mlWjlvVdhupI3Ph0msf9nE2K7Uoe-qklOj81cG0N19c/edit?usp=drive_open&ouid=117811280250615103178 

Acceptance criteria: 
- [x] Given a project_user with role viewer, when they open the job editor, the save button is disabled.

elias-ba:
## Related issue

Fixes #594 


lib/lightning_web/live/job_live/job_builder.ex
```diff
@@ -226,7 +228,7 @@ defmodule LightningWeb.JobLive.JobBuilder do
             to: @return_to
           ) %>
           <Form.submit_button
-            disabled={!@changeset.valid?}
+            disabled={!@changeset.valid? or !@can_edit}
```
stuartc, line 231: Hmm, I find this double not with an or pretty confusing - it does appear to do the right thing..

Would it be clearer as `!(@can_edit && @changeset.valid?)`?
Like staggering the importance, "they can edit something" and then "is it valid"; and then inverting whatever the outcome of that is for `disabled`..

lib/lightning/policies/permissions.ex
```diff
@@ -0,0 +1,11 @@
...
+  @moduledoc """
+  The Bodyguard permissions module.
+  """
+  def can(policy, action, user, params \\ []) do
+    case Bodyguard.permit(policy, action, user, params) do
+      :ok -> true
+      {:error, :unauthorized} -> false
+    end
+  end
+end
```
stuartc, line 11: I really like where you're going with this.

lib/lightning_web/live/components/form.ex
```diff
@@ -305,11 +305,12 @@ defmodule LightningWeb.Components.Form do
...
       )
+      |> assign_new(:disabled, fn -> false end)
 
     ~H"""
     <div class="flex items-start">
       <div class="flex items-center h-5">
-        <%= checkbox(@form, @id, class: @checkbox_classes) %>
+        <%= checkbox(@form, @id, disabled: @disabled, class: @checkbox_classes) %>
       </div>
       <div class="ml-3 text-sm">
```
stuartc, line 315: Instead of making a new assigns to fill a keyword list (which is what `checkbox/3` takes at the end).

Liveview has some conventions around this:

```
 attr :opts, :global   # <== This filters out all the standard HTML attrs (like disabled)

 def label_field(assigns) do
 ....

    assigns =
      assigns
      |> update(:opts, fn opts ->
        assigns_to_attributes(opts)
        |> Keyword.put_new(:class, label_classes)
      end)

     ~H"""
    <%= label(@form, @id, @title, @opts) %>
     """
   end
```

These are some changes I have for `label_field` in my current branch, I wanted to be able to override the class, but it will work with `disabled` because it's a standard HTML attribute.

And `assigns_to_attributes/2` turns assigns into a keyword list allowing us to easily drop the resulting valid in `label/4` or in your case `checkbox/3`.

Check out: https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#assigns_to_attributes/2
and https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#attr/3

The `attr/3` function was introduced not so long ago so I've only just started adding it to our components as I go.

stuartc (commented):
This is looking good @elias-ba, good work so far - I added two comments that may improve some small things..

config/config.exs
```diff
@@ -33,7 +33,7 @@ config :joken, default_signer: "secret"
 # Configures the mechanism for erlang node clustering
 config :libcluster,
   topologies: [
-    local: [strategy: Cluster.Strategy.Gossip]
+    # local: [strategy: Cluster.Strategy.Gossip]
```
stuartc, line 36: @taylordowns2000 why is this commented out?

lib/lightning/application.ex
```diff
@@ -9,7 +9,7 @@ defmodule Lightning.Application do
   @impl true
   def start(_type, _args) do
     # mnesia startup
-    :mnesia.stop()
+    # :mnesia.stop()
```
stuartc, line 12: @taylordowns2000 This is the recommended pattern for mnesia, while not compulsory in our current application shape - in the event of other applications that are started before or after this one that also use mnesia, you must always stop mnesia before creating a schema. I think we should leave this here.

lib/lightning/projects/project_user.ex
```diff
@@ -26,11 +26,11 @@ defmodule Lightning.Projects.ProjectUser do
...
-    belongs_to :user, User
-    belongs_to :project, Project
-    field :delete, :boolean, virtual: true
-    field :failure_alert, :boolean, default: true
-    field :role, RolesEnum, default: :editor
+    belongs_to(:user, User)
+    belongs_to(:project, Project)
+    field(:delete, :boolean, virtual: true)
+    field(:failure_alert, :boolean, default: true)
+    field(:role, RolesEnum, default: :editor)
```
stuartc: @taylordowns2000 please fix your formatter.

stuartc (commented):
@elias-ba can you please: 

1. Rebase this branch (and squash the merge commit in here if necessary)
2. Make the recommended changes that I commented on (from @taylordowns2000's changes)

config/config.exs
stuartc: @elias-ba can you change this back for this PR.

lib/lightning/application.ex
stuartc: @elias-ba please change this back.

lib/lightning/projects/project_user.ex
stuartc: @elias-ba undo these formatting changes.

config/config.exs
elias-ba: Done

lib/lightning/application.ex
elias-ba: Done

lib/lightning/projects/project_user.ex
elias-ba: Done

Issue #536, opened 2023-01-04
Handle OAuth credentials


amberrignell:
### User story
As a user writing a job that requires OAuth credentials, I would like the authentication flow to be handled for me, so that I don't have to manually fetch and set a token to run my job.

### Implementation notes
closes #623

### Release notes
- [ ] Support OAuth credentials. 

### User acceptance criteria

For a given OAuth credential, 
- [ ] the user is asked for the correct inputs (client secret, client id etc) instead of an access token. 
- [ ] when an access token expires, a refresh token is fetched. 
- [ ] given a credential which is NOT Oauth, the Oauth flow is not called. 
- [ ] when a user edits their credentials, they can see the current values. (see #623)

PR #620, opened 2023-02-01
OAuth Credentials

stuartc:
# OAuth Credentials

Starting with support for only Google Sheets.

We are adding:

- [x] The ability to select a '3rd' type of credential: `googlesheets`
- [x] The new credential type has it's own LiveComponent
- [x] Provide a 'system wide' Google OAuth client (we already use uberauth/oauth2)
- [x] Add a Google Authorize button which gets it's URL from the client (above)
- [x] Work out the best way to communicate back to the component from a regular controller
- [x] Add a route and controller action for handling OAuth2 'code' strategy callbacks.
- [x] Add an encrypted state string to the callback url.
- [x] In the callback controller forward the decode state to the component.
- [x] Add handling for code key exchange inside the component
- [x] Add the refresh token, and access token to the body of the credential.
- [x] Fetch the users information and show their image on the form.
- [x] Test with a real Google client id and secret
- [x] Ensure existing OAuth credentials render correctly in the form
- [x] Render userinfo when editing

## Remaining

- [x] Uncover and list the touch points where things can go wrong
  - [x] Handle non-200 when retrieving userinfo
  - [x] Handle non-200 when refreshing token
  - [x] Handle non-200 when getting token
  - [x] **Should _never_ happen but** handle code for token exchange when a `refresh_token` isn't provided  
- [x] UI changes/tweaks
- [x] Unit testing 
- [x] Mark as 'needing attention' when token has expired or userinfo fails
- [x] **Two buttons** on credential (@taylordowns2000) are:
  - [x] "test" (e.g. `userInfo` endpoint for Google - it will `renew` if necessary),
  - [x] "re-authorize" (only if _last_ `renew` attempt failed with a specific error.)

## Testing

In order to test this, the follow new ENV variables are required:

- `GOOGLE_CLIENT_ID`
- `GOOGLE_CLIENT_SECRET`

Or else when trying to create new Google Sheet credential, a notice is visible saying the client isn't configured.

## Related issue

Re #536 

-> needs 3 new issues created before closing


stuartc:
**Happy path**

image

**One of the sad paths**
image


cc/ @taylordowns2000

taylordowns2000 (approved):
looks good and works well from my side. if you're happy with the UX/UI @stuartc then we should:
**1. address your unchecked checkboxes and then merge this.**
2. new issue for running a job with this credential type
3. new issue for DB changes: `:oauth_client` type of credential (two credential types: `:secrets`—or something "standard"—and `:oauth_client`) which can have `:secrets` as children
4. new issue for switching from using ENV to using `:oauth_client` credentials
that sound right?

stuartc:
@taylordowns2000 

> 1. address your unchecked checkboxes and then merge this.

The renew/error feedback stuff has been addressed now.
There aren't any specific renew or refresh buttons but rather specific links on each kind of error:

image

image

image


> 2. new issue for running a job with this credential type

Over here https://github.com/OpenFn/Lightning/issues/646

> 3. new issue for DB changes: :oauth_client type of credential (two credential types: :secrets—or something "standard"—and :oauth_client) which can have :secrets as children
> 4. new issue for switching from using ENV to using :oauth_client credentials
that sound right?

We should plan these out later.

PR #622, opened 2023-02-08
574 import projectyaml

### LINKED ISSUE #574: Import project.yaml
Write a function which, given a project.yaml file with the format defined in [projects_as_code](https://github.com/OpenFn/projects-as-code), creates a project with the given workflows and jobs. 

Out of scope: http or cli interface

The function should take a map, create the resources and return the models that it makes. If you call the function on it's own, no users get created. (when we add the cli, the user that is calling the function will be added as the project owner). 

### User acceptance criteria: 
- [ ] Given a data structure with a workflow A and workflow B, when I call the function xxx with it, I can see that project on the admin settings>projects page
- [ ] When I add myself as a member of that project, I can see 2 workflows on my workflows index page
- [ ] The function should return the models that it makes. 
- [ ] If the project already exists, throw an error. 

(notes for other issues)
1. Fix export project in Lightning 
2. Write function that creates a project based on a data structure (this issue)
3. Deploy through the CLI - add a command to the CLI which calls a lightning endpoint and deploys a new project from a project.yaml file

wanecode:

Write a function which, given a project.yaml file with the format defined in [projects_as_code](https://github.com/OpenFn/projects-as-code), creates a project with the given workflows and jobs.

Out of scope: http or cli interface

The function should take a map, create the resources and return the models that it makes. If you call the function on it's own, no users get created. (when we add the cli, the user that is calling the function will be added as the project owner).

## Fisrt attempt: use ecto association features 

Relying on existing ecto association mechanism let us benefits fro the changesets validation pipeline
Problem: we need to generate a credential ID before  adding it to the project (ie: before the project gets saved).

## Second attempt (current): use multi

At first, try to create credential one by one (Mutli.insert) using chansgeset for validation (problem: this leads to having credentials keyed on the resulting map)

Second, use insert_all to keep the resulting map closer to the model we need (proper list of credentials, proper list of workflows). But then, we'll need to "bulk" validate before insert_all

Insert_all implies not using the schema (have to manually add timestamps)
Finally getting back to the keyed credentials in the multi space %{key1: credential1, key2: credential2}

## Tests

- Tests happy path: one big map with all required keys creates all the models 
- Test missing workflows key -> empty project
- Test missing project name -> classic error in changeset (way of confirming every known entity in the map will benefit of its own validation mechanism)
- Test credentials (if job has a credential key and and there is no corresponding credential in input,an error is added on job changeset )

 

## Related issue

Fixes #574 Import project.yaml

### DIFF
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -247,7 +247,9 @@ defmodule Lightning.Projects do
       {:ok, string}
 
   """
-  def import_project(data) do
-    nil
+  def import_project(attrs) do
+    attrs
+    |> Project.import_changeset()
+    |> Repo.insert()
   end
 end
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -6,7 +6,7 @@ defmodule Lightning.Projects do
   import Ecto.Query, warn: false
   alias Lightning.Repo
 
-  alias Lightning.Projects.{Project, ProjectCredential, ProjectUser}
+  alias Lightning.Projects.{Project, ProjectCredential}
   alias Lightning.Accounts.User
   alias Lightning.ExportUtils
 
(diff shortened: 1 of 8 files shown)


lib/lightning/projects.ex
```diff
@@ -236,4 +236,150 @@ defmodule Lightning.Projects do
...
+      {:ok, string}
+
+  """
+  def import_project(project_data, user) do
+    ImportUtils.import_multi_for_project(project_data, user)
+    |> Repo.transaction()
+  end
+end
+
+defmodule ImportUtils do
```
stuartc, line 256: @wanecode I think this module should be under `Projects`, right now it's at the root namespace.

So it should be either `Lightning.Projects.ImportUtils`, but I think it might be better named as `Lightning.Projects.Importer`.

And with that, can it be moved to it's own file under `lightning/projects/importer.ex`?

lib/lightning/projects.ex
```diff
@@ -236,4 +236,150 @@ defmodule Lightning.Projects do
...
+
+    %Workflow{}
+    |> cast(
+      workflow,
+      [:name, :project_id, :id]
+    )
+    |> cast_assoc(:jobs,
+      with:
+        {ImportUtils, :import_job_changeset, [workflow_id, import_transaction]}
+    )
```
stuartc, line 315: :+1:

stuartc (changes requested):
@wanecode this is great! Could you move that importer module into it's own file and this should be good to go.

lib/lightning/projects.ex
```diff
@@ -232,6 +230,7 @@ defmodule Lightning.Projects do
       {:ok, string}
 
   """
+  @spec export_project(:yaml, any) :: {:ok, binary}
```
stuartc, line 233: @wanecode I believe the convention is to have the `@spec` between the doc tag and the `def`.

test/lightning/importer_test.exs
```diff
@@ -36,15 +36,15 @@ defmodule Lightning.ImportProjectsTest do
                 adaptor: "language-fhir",
                 enabled: true,
                 credential: "xyz",
-                body: "> fn(state => state)"
+                body: "fn(state => state)"
```
stuartc, line 39: @wanecode I changed these to match what the body actually looks like, it appears you added the prepending `>` from the YAML examples - which is a signal in YAML to create a single string from a multiline indented to the left margin of the `>`.

stuartc (approved):

PR #630, opened 2023-02-09
Ensure terms are accepted before allowing user registration

### LINKED ISSUE #531: Prevent user from registering if they haven't selected checkbox
### User story
As OpenFn, I would like to prevent users from registering if they have not check the lightning beta checkbox, so that I ensure they have read the notice. 

### Details
Disable the checkbox by default. If the user hasn't selected it, prevent them from registering.

elias-ba:
## Implementation plan
- [x] Change User registration to use schemaless changeset
- [x] Track the changes on the terms and conditions checkbox
- [x] Add validation error on the changeset when the terms and conditions are not accepted (variable is not true), this error should block the registration process
- [x] Add UI changes to place the error message right above the register button
- [x] Write a phoenix controller tests to test the scenario above

PS: The user registration page is not using liveview; it uses the regular phoenix controllers. Currently all validations requires page reloading (checked with @amberrignell and @stuartc about this plan and it seemed the right thing to do)

Given that we're now using schemaless changesets for user registration, most of the tests are failing. I need to address those

**QUESTION: Should superusers also accept terms and conditions before registering for an account ?**

## Related issue

Fixes #531 

### DIFF
--- a/lib/lightning/accounts.ex
+++ b/lib/lightning/accounts.ex
@@ -108,27 +108,6 @@ defmodule Lightning.Accounts do
   end
 
   @doc """
-  Registers a superuser.
-
-  ## Examples
-      iex> register_superuser(%{field: value})
-      {:ok, %User{}}
-
-      iex> register_superuser(%{field: bad_value})
-      {:error, %Ecto.Changeset{}}
-  """
-
-  def register_superuser(attrs) do
-    %User{}
-    |> User.superuser_registration_changeset(attrs)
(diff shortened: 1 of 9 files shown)


amberrignell:
Looks good to me elias, nice work

stuartc (approved):
@elias-ba good stuff here! I made two tiny changes here: https://github.com/OpenFn/Lightning/pull/630/commits/a639e18dd85bca142f269d1a5474b650a1138efb

PR #631, opened 2023-02-09
Prove that uninvited members cannot view things in projects

taylordowns2000:
@stuartc , @elias-ba , this is what i'd like to be able to show with `thief` - we use this same concept in the v1


elias-ba (commented):
@taylordowns2000 this seems good and important to me. Let's see what @stuartc thinks about it. But definitely a yes for me.

Issue #585, opened 2023-01-09
Project digest


amberrignell:
### User story
As a project user, I would like to receive a digest via email, so that I can get an update on how the workflows within my project are doing AND see the project activity. 

### Details
In the Digest column in project settings > collaboration (created by #588), allow a project user to select 'never', 'daily', 'weekly', 'monthly'. 

Write a function that emails a user the following digest at the frequency selected: 

```
Subject: ${frequency} digest for project ${project name} 

Body: 

Workflow A:
- 12 workorders correctly processed this ${month/day/week} 
-> sum of work orders that have a status of success, where the **last run** is in the past ${month/day/week} 
- 4 of these work orders initially failed,  but were rerun and eventually processed correctly
- 2 work orders failed and still need addressing
```

### Implementation notes
Keep in mind we will be adding a feature which allows administrators to prevent any logs from being sent in emails. 

Daily: send at 10am every day
Weekly: send on Monday at 10am
Monthly: send on the 1st of the month at 10am

Add a digest column to the project_users table. The default should be set to 'weekly'. 
Add a Digest column to the collaboration page. 

### Release notes
- Configure project to receive daily/monthly/weekly digest

### User acceptance criteria
- [ ] The default is set to 'never'
- [ ] Given a project user for project 'Sync OpenHIE' with digest set to 'weekly', I should receive an email on Monday at 10am with the subject: 'Weekly digest for project 'Sync OpenHIE'.

amberrignell:
Updated phrasing in the email after request for clarification from Elias: 
"4 of these work orders initially failed,  but were rerun and eventually processed correctly"

amberrignell:
When I change the project digect and alerts values in project settings, I don't get a flash confirmation message and the values aren't updated. 

https://images.zenhubusercontent.com/454419290/8ccec269-7305-4d71-8254-4378e17f5cd3/screen_recording_2023_03_08_at_09_54_01.mov

amberrignell:
closed. by #682

PR #648, opened 2023-02-16
Allow users to edit project digest and alerts settings

### LINKED ISSUE #638: Allow users to edit project digest and alerts settings
### User story
As a project user, I would like to edit the frequency of my project digest and alerts, so that I can be notified on failures and get a summary of my project activity at my desired frequency.

### Details
Context: issues [583](https://app.zenhub.com/workspaces/product-engineering-6305f22ada44914905485ab1/issues/gh/openfn/lightning/585) and [581](https://app.zenhub.com/workspaces/product-engineering-6305f22ada44914905485ab1/issues/gh/openfn/lightning/581) added the functionality to send users project digests and alerts on run failures at different frequencies. 

See rough sketch below for guidance:

### Implementation notes
N/A - @stuartc please read through but since this is mostly UI I don't expect many notes, just an estimate.

### Release notes
- A user can enable/disable realtime alerts for a project.
- A user can edit the frequency at which they receive a project digest. (Possible values are never, daily, weekly and monthly)

### User acceptance criteria
- [ ] In project settings > collaboration there is both a "digest" and a "realtime alerts" column 
- [ ] A viewer, editor, admin or owner can edit THEIR OWN digest and alerts frequency: in the row next to their name, digest and realtime alerts are dropdowns that display the current value set for digest and realtime alerts.
- [ ] When an editor clicks the dropdown for digest, they can select from 'never', 'daily', 'weekly' and 'monthly'.
- [ ] When an editor clicks the dropdown for realtime alerts, they can select from "enabled" or "disabled"
- [ ] A viewer, editor, admin, owner can ONLY edit their _own_ digest and realtime alerts settings. The digest and realtime alerts values displayed for other users in the 'collaboration' section are read-only. 

~@taylordowns2000 this is where we need to start thinking about roles ... should a viewer be able to receive notifications and project digests ? And if not then we should set the default values accordingly. Will update the release notes accordingly after we've discussed.~ @amberrignell , everyone on a project should be able to edit their own notifications (seems like a good start) but we should implement these permissions via the permissions module anyway — and later we may add the ability for an admin/owner to edit the notifications settings of _other people_ who belong to their project.

elias-ba:
## Implementation plan

The idea of using forms / ecto.changesets in this context was not the best one in terms of performance. Instead I am using the phoenix bindings and listening the click events on these components (selectbox for digest values and checkbox for failure alerts). For every click I send I get the id of the current project user and the value of the component. With these values I update the project user and reassign it to the liveview socket. The logic of all this will be controlled by the roles and permissions.

## Related issue

Fixes #638 and #585 

### DIFF
--- a/lib/lightning_web/live/project_live/settings.ex
+++ b/lib/lightning_web/live/project_live/settings.ex
@@ -4,6 +4,10 @@ defmodule LightningWeb.ProjectLive.Settings do
   """
   use LightningWeb, :live_view
 
+  alias Lightning.Policies.Permissions
+  alias Lightning.Accounts.User
+  alias Lightning.Policies.ProjectUsers
+  alias Lightning.Projects.ProjectUser
   alias Lightning.{Projects, Credentials}
 
   on_mount({LightningWeb.Hooks, :project_scope})
@@ -33,8 +37,17 @@ defmodule LightningWeb.ProjectLive.Settings do
        can_edit_project: can_edit_project,
        credentials: credentials,
        project_users: project_users,
-       changeset: Projects.change_project(socket.assigns.project)
+       current_user: socket.assigns.current_user,
+       project_changeset: Projects.change_project(socket.assigns.project)
      )}
+  end
+
+  def can_edit_project_user(
+        %User{} = current_user,
+        %ProjectUser{} = project_user
+      ) do
+    ProjectUsers
+    |> Permissions.can(:edit_project_user, current_user, project_user)
   end
 
   @impl true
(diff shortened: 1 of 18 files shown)


lib/lightning/workflows.ex
```diff
@@ -190,4 +190,118 @@ defmodule Lightning.Workflows do
...
     end)
   end
+
+  defp filter_digest(digest) do
+    case digest do
+      :daily ->
+        dynamic([r], fragment("age(now(), ?) <= '1 days'", r.finished_at))
+
+      :weekly ->
+        dynamic([r], fragment("age(now(), ?) <= '7 days'", r.finished_at))
```
stuartc, line 200: I think we should avoid using the word digest here, we have an opportunity to have a more generic function like:

`finished_before_query` or similar.

```
def finished_before_query(period) do
  period = [case period do
    :daily -> 1
    ...
  end |> to_string(), "days"]
  |> Enum.join(" ")

  dynamic([r], fragment("age(now(), ?) <= ?", r.finished_at, period))
end
```

Note that I'm passing in the period as well, this will allow Postgres to get more of an opportunity to reuse a query plan (it will cache one query instead of 3).

Also it looks like this query returns _all_ runs before a given date. Do we not need it to be between two dates?  
And I see you're using `age(now() ..` that makes me a little nervous because:

1. After the TZ issues we saw recently with Mtuchi's setup
2. Postgres will need to cast all of the `finished_at` values into ages in order to compare them - removing the chance of any index being used.
  I think we should create the 'correct' date using Timex or similar and pass that in as a `where: r.finished_at >= ^from_date`.

lib/lightning/workflows.ex
```diff
@@ -190,4 +190,118 @@ defmodule Lightning.Workflows do
...
+              where: w.id == ^workflow.id,
+              where: parent_as(:attempts).work_order_id == wo.id,
+              select: count(wo.id)
+            )
+          ),
+      select: count(attempt.id)
+    )
+  end
+
+  defp failed_workorders_query(workflow, digest) do
```
stuartc, line 272: On the face of it these queries look good. We do need to:

1. Move them into `WorkOrder` because they have nothing to do with workflows besides a single where clause.
2. Refactor the common parts into queries.

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -53,13 +70,148 @@ defmodule LightningWeb.ProjectLive.Settings do
...
+        socket
+      ) do
+    project_user = Projects.get_project_user!(project_user_id)
+
+    changeset =
+      {%{failure_alert: project_user.failure_alert}, %{failure_alert: :boolean}}
+      |> Ecto.Changeset.cast(%{failure_alert: value}, [:failure_alert])
+
+    case Ecto.Changeset.get_change(changeset, :failure_alert) do
+      x when not is_nil(x) ->
```
stuartc, line 95: Lets avoid using `x` and other unrelated letters for variables. Rather use `failure_alert`.

Also instead of this pattern, maybe the below reads easier:

```
case val do
  nil -> {:noreply, socket}
  val -> ...
end
```

test/lightning/projects_test.exs
```diff
@@ -153,4 +224,76 @@ defmodule Lightning.ProjectsTest do
...
+
+      {:ok, %{project_users: weekly_project_users}} =
+        Projects.perform(%Oban.Job{args: %{"type" => "weekly_project_digest"}})
+
+      {:ok, %{project_users: monthly_project_users}} =
+        Projects.perform(%Oban.Job{args: %{"type" => "monthly_project_digest"}})
+
+      assert daily_project_users |> length() == 3
+      assert weekly_project_users |> length() == 3
+      assert monthly_project_users |> length() == 3
```
stuartc, line 296: Perhaps mix these counts up, this doesn't prove that you're _not_ just getting all the project users for a project. Have at least one of the project users have no digest settings.

test/lightning/workflows_test.exs
```diff
@@ -207,4 +209,108 @@ defmodule Lightning.WorkflowsTest do
...
+
+      assert Workflows.get_digest_data(
+               workflow |> Repo.preload(:work_orders),
+               :daily
+             ) == %{
+               failed_workorders: 2,
+               rerun_workorders: 0,
+               successful_workorders: 3,
+               workflow_name: workflow.name
+             }
```
stuartc, line 305: All of these scenarios need to have data to _prove_ they aren't including other runs... Add some data you expect to not be found.

test/lightning_web/live/project_live_test.exs
```diff
@@ -4,6 +4,7 @@ defmodule LightningWeb.ProjectLiveTest do
   import Phoenix.LiveViewTest
   import Lightning.ProjectsFixtures
   import Lightning.AccountsFixtures
+  import SweetXml
```
stuartc, line 7: I think we should be using Floki instead of `SweetXml`... Check the code base for other uses of `Floki`.

PR #661, opened 2023-02-23
add `tooltip` component

### LINKED ISSUE #650: Add tooltips to the setup tab
### User story
As a new user that's not familiar with OpenFn concepts, I would like there to be tooltips next to the different input labels on the setup page, so that I can get a brief explanation/description of what adaptors, credentials etc are used for. 

### Details
Add a tooltip component that can be reused across the app. 
Add the following tooltips to the job inspector: 
- trigger: When your job will run. Select webhook to trigger is from an external system or cron to trigger it at a recurring point in time. 
- adaptor: Which system to connect to. This will update the adaptor documentation in the editor with system-specific operations to select from.
- credential: How to connect. The credentials you need for authentication in the selected system.  

### Implementation notes

We should start with using the `title` attribute on all elements that require it.

If we want to later add some styling to the tooltips we can use CSS like this _as an example_:

```css
[title]:hover::after {
  content: attr(title) ;
  position: absolute ;
  top: 1.1em ;
  left: 1em ;
  min-width: 200px ;
  border: 1px #808080 solid ;
  padding: 8px ;
  color: black ;
  background-color: #cfc ;
  z-index: 1 ;
}
```

### Release notes
- Added tooltips to trigger, credential and adaptor inputs on the setup tab

### User acceptance criteria
- [ ] There is a tooltip (?)  next to the trigger, adaptor and credential input labels
- [ ] When I hover over the ?, I can see the description

mtuchi:
## Implementation checklist
- [x] create a tooltip component 
- [x] write a test for the component

## Related issue

Fixes #650 


mtuchi:
🚧 
[Screencast from 24-02-2023 06:28:35 ALASIRI.webm](https://user-images.githubusercontent.com/6592749/221230780-0a65aa19-b212-4845-abb1-567218b194e0.webm)

stuartc (approved):
@mtuchi thanks for this, I made a bunch of small changes:

- Use `aria-label` instead of `data-label` in an attempt to respect accessibility needs.
- Moved the CSS into Tailwind classes
- Use Heroicons instead of inline svg
- Remove unused :inner_block from tooltip
- Dry up test with helper

https://github.com/OpenFn/Lightning/pull/661/commits/15e01857a789a6117b1c781634839738e20c2986

PR #675, opened 2023-03-03
Consolidate Job Editor into a single panel

josephjclark:
This PR consolidates the job editor in to a single React component - merging the Editor, Adaptor docs and Metadata thingy all into one.

Since it's all React now, I've also made a nice (?) dynamic layout controller.

image

I've done this on a separate branch in case the consolidation or layout goes awry  - and besides this is really orthogonal to the main magic stuff. The goal of this branch isn't really to develop the magic components any further, but to consolidate the layout and architecture.

The Pheonix stuff might want a look at.

This should (shortly) be in a state to merge into the main magic feature branch where this stuff will continue to be worked on. But we have an opportunity too for a mini-review.

Big things to do:

* [ ]  Persist panel config (horiz/vert and expand/collapse)
* [ ] Also remember which panel is selected when changing jobs
* [ ] Drive panel selection from the URL(maybe? the hash is already in use)


assets/css/app.css
```diff
@@ -106,6 +106,10 @@
   animation: 0.2s ease-out 0s normal forwards 1 fade-out-keys;
 }
 
+.flex-col-reverse {
```
josephjclark, line 109: idk why I had to add this, because it's documented in tailwind, but it didn't seem to apply. Could be a version thing?

assets/js/editor/Editor.tsx
```diff
@@ -189,8 +190,16 @@ export default function Editor({ source, adaptor, onChange, metadata }: EditorPr
       // ensure the editor has focus
       editor.focus();
     };
+    
+    // This is a temporary (and poor) solution to make the editor respond to resizing
+    listeners.current.updateLayout = (e: Event) => {
```
josephjclark, line 195: I will need to sit and figure this out at some stage, but not necessarily in this PR (or even for the first release, it works better than I thought it would)

PR #694, opened 2023-03-14
Detect when encryption keys are invalid

### LINKED ISSUE #684: New credential not saving
When creating a new credential on `v0.4.4` it's not persisting. All migrations have been run. Is this an ENV issue?

https://www.loom.com/share/bb4c3444fe274255802b12ff20b2bcb1

1. Go to the canvas
2. Click on a job
3. Click new credential
4. Click save
5. Modal closes but nothing happens

stuartc:
Fixes #684

Encryption keys (for AES GCM) must be exactly 256-bits in size. When Lightning.Vault starts up it now checks for:

- That the key can be correctly decoded
- A key must be exactly 256-bits in size

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to
 
 ### Changed
 
+- Improved errors when decoding encryption keys for use with Cloak.
+  [#684](https://github.com/OpenFn/Lightning/issues/684)
+
 ### Fixed
 
 ## [0.4.4] - 2023-03-10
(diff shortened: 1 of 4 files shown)


taylordowns2000 (approved):
@amberrignell , this looks good to me from my phone. Want to spin this up locally and make sure the app boots properly then merge?

amberrignell:
Tested the following and all looks good to me :) 
- when I add a random string for the encrypton key 'gersyudgfve', the app does not boot up
- when I add a base64 encoded string that is not 256-bits, the app does not boot up
- when the string is encoded base64 AND exactly 256 bits, the app boots up
- I can generate a key with the suggested command `mix lightning.gen_encryption_key`

PR #697, opened 2023-03-15
Load test for webhook endpoint performance

### LINKED ISSUE #634: load test for webhook endpoint performance
### Details
Write a test which sends 10,000? a configurable number of requests to a webhook trigger and measure performance. 

### User acceptance criteria
What should we run these in ? 
Given this is the first one, what else needs to be put into place ?

elias-ba:
## Plan

K6 is pretty new to me and most of this issue is about using some key metrics and thresholds to evaluate the performance of Lightning under heavy loads of webhook API calls. For this here are the steps I am intending to take:
1. Keep reading about k6. Currently I am exploiting these resources: [Load testing restful APIs with k6](https://k6.io/blog/load-testing-restful-apis-with-k6/), [Load testing](https://k6.io/docs/test-types/load-testing/)
2. I will practice what I am reading in Lightning by creating a local webhook workflow and using a k6 script to target the workflow. I have already layed out the script [here](https://github.com/OpenFn/Lightning/blob/7a53b8e15fc9f4ecfbb37f21e7114551cd72d39f/test/webhook_load_test.js).
3. Then I will need to have some key metrics and hyperparameters to measure using k6. My understanding is k6 has what they call thresholds and durations. These allows to measure the amount of time the system takes to respond under some number of calls and also to set some threshold that we need to target. This is where I am not super clear in my understanding and probably I will need @amberrignell @taylordowns2000 and @stuartc to help make it clearer.
4. I need to find a way to make my script generate some data file (CSV) of the load testing results

## Related issue

Fixes #634 


stuartc (commented):
I made a few changes here and there, namely pointing users to use the demo data instead of making their own workflow.

I also added `k6` to our `.tool-versions` so a simple `asdf install` will download it for them.

PR #699, opened 2023-03-15
Job Panel resize

### LINKED ISSUE #681: Resize right-hand panel
The tabbed panel on the right should be resizable - as in, the user can hoizontally expand or shrink the panel to show more or less of the workflow diagram.

image

This is particularly important for job editors, as the job editing panel needs a lot more horizontal space than the other tabs.

A couple of considerations:
* This is probably a pure javascript solution
* The last size should persist to the browser (as a propotion with min/max, rather than a pixel value)

josephjclark:
This PR implements a resizer, allowing you to change the relative sizes of the workflow diagram and the job editor.

It would close #681

This inserts a draggable div and, on when it is dragged, resizes the parent node to fit the drag position. 

A mask is inserted under the cursor to help prevent nasty interactions with the rest of the page as you drrag.

There is a min and max width set on the panel size - the max is a bit arbitrary.

Some notes:

* If you drag outside the page, page text gets select weirdly. I am unsure how to prevent this
* The size is only persisted in client javascript, so while the page is loading you may see the panel jump
* The workflow dialog looks awful when narrow but I think that's just how it is.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -29,6 +29,9 @@ import JobEditor from './job-editor';
 import WorkflowDiagram from './workflow-diagram';
 import TabSelector from './tab-selector';
 
+let dragEndListener;
+let dragListener;
+
 let Hooks = {
   WorkflowDiagram,
   TabSelector,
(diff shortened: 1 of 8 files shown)


josephjclark:
@stuartc I've udpated the resizer to be a proper pheonix component. Feels much better now! Do you want a quick look before we merge this down?

stuartc (approved):

PR #710, opened 2023-03-17
Metadata CLI Service

### LINKED ISSUE #621: Magic functions: call CLI to generate metadata
Lightning needs to push magic function metadata into the Editor component.

There are two parts to this:

* Call out to the CLI to generate metadata
* Provide an endpoint so that the client app can fetch data down

# CLI

The CLI provides a command (currently unreleased, so that's a problem) to generate metadata for a credential (well, a configuration object).

The CLI will write metadata to disk (inside its repo dir, as a json file) and "return" to stdout the path it saved to.

Use it like this:

```
openfn metadata -s path/to/config -a adaptor-name=path/to/adaptor
```
You can also pass state straight to stdin:

```
openfn metadata -S "{ configuration: {} }" -a adaptor-name
```

This is when it would be useful to split stdout and stderr, or have a no log mode which only returns the final output. But `openfn | tail n 1` will return the path.

The CLI will cache any metadata it generates, so if the first call with a given credential is slow, the second should be significantly faster. I expect Lightning to add its own cache management at some stage but this gives us a leg-up while getting started.

Here's a sample run:

image

There's an open PR this in kit https://github.com/OpenFn/kit/pull/125. There's actually a bit of work to do - right now it only works if you run the adaptors out of the monorepo (since no adaptors with `metadata` functions are published). There's a to-do list in a comment at the bottom of the PR.

# Endpoint

I think for now the easiest approach is to add an endpoint to Lightning. The Editor component can call this for the metadata json. 

Maybe later we'll swap this out for a lazy-loaded attribute provided by lightning, if that's more Elixir idiomatic. That's not really for me to decide.

Probably the endpoint is like this:

POST /metadata/:credential-id?adaptor=dhis2

Or something. Maybe /metadata/credential/adaptor, or metadata?credential=&adaptor=. Probably that last one because it's not a very path-like query is it? credential and adaptor are sort of equal hierarchically.

Anyway, whatever - Lightning exposes an api like that. It returns the metadata result from the CLI, verbatim.

Ultimately we need a set of error codes for:
* 400 Invalid arguments
* 403 bad credential
* 501  metadata function not available for this adaptor/version
* 500 internal error (catch-all)
* ? No data returned by function (why would this happen? We should get at least an empty stub)

stuartc:
Building out a module for calling out the to OpenFn CLI getting metadata.

- [x] Limit concurrency
- [x] Call the CLI with the correct arguments.
- [ ] Provide useful errors for later use in the front end when the CLI or the CLI Worker throws errors.
- [x] Add a `handle_event` handler to the JobBuilder that can respond with metadata when requested. (to sync with @josephjclark)
- [ ] Provide a recommended path _or_ solution that caters for caching.

## Related issue

Fixes #621

### DIFF
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -27,7 +27,7 @@ const spinner = (
       cy="12"
       r="10"
       stroke="currentColor"
-      stroke-width="4"
+      strokeWidth="4"
     ></circle>
     <path
       className="opacity-75"
--- a/assets/js/editor/index.tsx
+++ b/assets/js/editor/index.tsx
@@ -9,7 +9,12 @@ interface EditorEntrypoint {
   changeEvent: string;
   field?: HTMLTextAreaElement | null;
   handleContentChange(content: string): void;
-  pushEventTo(target: HTMLElement, event: string, payload: {}): void;
+  pushEventTo(
+    target: HTMLElement,
+    event: string,
+    payload: {},
+    callback?: (reply, ref) => void
+  ): void;
   mounted(): void;
   observer: MutationObserver | null;
   render(): void;
@@ -38,6 +43,10 @@ export default {
       this.setupObserver();
       this.render();
     });
+
+    this.pushEventTo(this.el, 'request_metadata', {}, (reply, ref) => {
+      console.log({ reply, ref });
+    });
   },
   handleContentChange(content: string) {
     this.pushEventTo(this.el, this.changeEvent, { source: content });
@@ -50,7 +59,7 @@ export default {
           adaptor={adaptor}
           source={source}
           onChange={src => this.handleContentChange(src)}
-          disabled={disabled == "true"}
+          disabled={disabled == 'true'}
         />
       );
     }
--- a/assets/js/editor/index.tsx
+++ b/assets/js/editor/index.tsx
@@ -2,20 +2,26 @@ import React from 'react';
 import { createRoot } from 'react-dom/client';
 import type Editor from './Editor';
 
-interface EditorEntrypoint {
-  componentRoot: ReturnType<typeof createRoot> | null;
+interface ViewHook {
   destroyed(): void;
   el: HTMLElement;
-  changeEvent: string;
-  field?: HTMLTextAreaElement | null;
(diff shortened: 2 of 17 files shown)


stuartc:
> @stuartc I've been thinking about the `handle_event` pattern and it's just so... un-reacty. If we want the components to be standalone react then they shouldn't support `handle_event` to fetch information.
> 
> Which means `handle_event` needs to be done in the mount point and needs to pass through the context name as a top level prop on the JobEditor component. Ok, maybe we could use React Context, but it's still happening at the top of the app.
> 
> So isn't it easier just to pass the context name as an attribute to the JobEditor component? The same way we pass through the adaptor specifier?
> 
> Just not convinced we're on the right track here.

I'm kinda following you, but can't be sure. I'm not proposing we add a function to the Editor component that knows how to talk to Phoenix.

What I'm proposing is we pass in a function that when called will do whatever (in this case use the Phoenix socket to fetch the metadata). So the function will need to be built with the context of the hook, like it needs the right instance of the socket etc.

So the call to the editor would be something like:

```
<EditorComponent
  adaptor={adaptor}
  source={source}
  onChange={src => this.handleContentChange(src)}
  metadataFetcher={this.fetchMetadata}
  disabled={disabled == "true"}
/>
```

Is this the exact concern you had? Or does this make you feel :performing_arts:?

josephjclark:
@stuartc I think we're talking cross purposes a bit.

There are two bits of information we need:

1) metadata
2) credential name

Metadata does indeed need to be an async function which calls up to Pheonix, and handle event is fine for this.

Credential name should be passed  down via attribute data, just like the adaptor specifier.

I am probably going to just remove the credential name thing - it's really not important :) I wanna focus our energy on the actual metadata.

stuartc:
@josephjclark here is an example of the fetch function for getting metadata:

https://github.com/OpenFn/Lightning/pull/710/commits/827d4aa752ffe9c45113653a1300bcb6e2bdde1c#diff-c0db3977c9a3708d4a625153e1bc4935e536dc1b1bd3dcb3de990e243150922cR53-R70

How does that feel?

If that works, whats the next step? You rebase on top of mine, merge?

PR #711, opened 2023-03-17
Stress load test

### LINKED ISSUE #654: Stress load test
### User story
As a Lightning developer/contributor, I would like an easy way to test how Lightning behaves when it has a heavy workflow (many jobs to execute at the same time), so that I can trust it and be aware of its performance.

### Details

Create a test harness that allows us to programmatically create a set of jobs and push a configurable number of dataclips into the system.

We want to be able to profile variances of both message count and message size (many dataclips of different average sizes).

Things we need to measure at increasing levels of load:

1. Time to persist a dataclip
2. Time from dataclip created to start of job (per workflow).
3. Time to invoke a `fn(x) => x;` job (`run.started_at` to `run.finished_at`)

In order to get metrics out of the test run, we should strongly consider using the
telemetry library as these event/metrics can later be used for monitoring.

**Additional Metrics**

- memory usage
- cpu usage?
- atom count? (if this climbs linearly or significantly it could indicate a problem)

**Important observations**

- What happens when it fails?
- Does the rapid failure of other components (ecto/postgres) cause jobs to fail prematurely?
- How much does it take to make it break?

### Implementation notes

I think a good place to start is to write an exs script that can call a function in the app that is expected to produce a telemetry event. Then see how we can get telemetry events out (be it by filtering a log or writing to a file or another application)

### Release notes

- Stress load test: executing many jobs at the same time. 

### User acceptance criteria

- [ ] The test is NOT in our unit tests
- [ ] The test produces structured (CSV or JSON) data that we can easily chart and keep the results.

elias-ba:
## Plan

This plan will evolve a lot. But after many discussions, for now, this are the actions that I am prioritizing.

- [x] Layout a .exs script for creating a arbitrary number of jobs and dataclips
- [x] Figure out how to add telemetry to that function
- [x] View telemetry event
- [x] Get the data out

This should give us a solid foundation for our load testing lib that we can adjust accordingly

### To finish this issue

- [x] Move the load test to benchmarks folder
- [x] Delete metrics.ex
- [x] Write a line to the CSV with raw data (event_name, timestamp, duration)

### Future steps (in another issue)

- Move telemetry spans into the application code (ie polish patterns about span and attach_events so that we don’t end up with them littered everywhere in the app)
- Having a map of the functions we want to measure (pass-through functions)
- Decide what we want to measure
- Find a way to trace a span -> potentially look into [Open Telemetry](https://opentelemetry.io/) for this

## Related issue

Fixes #654 


stuartc (approved):
@elias-ba great job! I made a few tweaks (nothing major) but this is good to go for the next issue.

@amberrignell could you give this a look over. I added a tiny instruction on how to run it at the top of the `load_tests.exs` file.

There are two things to add to the list of things we need to do for the _next_ load test issue and that is:

1. Have this script (and any other elixir based scripts), use an Ecto sandbox or have some kind of cleanup strategy - that way changes get wiped after the test is run and we don't smash our local dev databases with 1000s of jobs (when we run that many etc)
2. Update the BENCHMARKING.md document to have a nice pre-amble about what is there - this is probably _as_ we finish the next issue.

amberrignell:
@stuartc have created a new issue for next steps and moved your comments into there: https://github.com/OpenFn/Lightning/issues/745

amberrignell (approved):

PR #748, opened 2023-03-29
release v0.4.7

amberrignell:
## Notes for the reviewer
Making another release so we can have the fix for failure alerts on app.openfn.org


amberrignell:
@taylordowns2000 this looks good, happy for you to merge. Only fix I wasn't able to double-check on demo is the user registration because that's only enabled on app.openfn.org but I QA'd it locally before it was merged to main.

PR #777, opened 2023-04-05
add ability to save using ctrl + s

### LINKED ISSUE #568: Better save for jobs panel
1. When creating a **_new_** job, the "Save" button should say "Create" rather than save because the record is not stored until the button is pressed. This will help with a current UX frustration where users cannot tell if they are creating a new job or editing an existing one.
2. As anyone, I'd like to be able to save my changes to a job without closing the jobs panel. (Having the panel close is particularly frustrating when I'm first building a job, as certain features are blocked until the job is saved.)
3. As a developer-user, I'd like `command-s` or `control-s` to save my job expression and _not_ close the panel.
4. The "cancel" button should instead say "close" ~and if I attempt to close or click away with unsaved changes I should get a warning indicating that i have unsaved changes~.

mtuchi:
## Notes for the reviewer

Change cancel to close and save the job without closing the job panel

## Related issue

Fixes #568 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to
   [#147](https://github.com/OpenFn/Lightning/issues/147)
 - Add `last-used at` to API tokens
   [#722](https://github.com/OpenFn/Lightning/issues/722)
+- Improved "save" for job builder; users can now press `Ctrl + S` or `⌘ + S` to
+  save new or updated jobs job panel will _not_ close. (Click elsewhere in the
+  canvas or click the "Close" button to close.)
+  [#568](https://github.com/OpenFn/Lightning/issues/568)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


taylordowns2000 (changes requested):
nice work @mtuchi . a few things:
1. it seems like issue items 1 and 4 haven't been dealt with. please mark this as draft until they're ready. (if we merged now, that issue would get closed by the `Fixes #568` tag in the PR)
2. for issue item 2, is it standard to `|> push_patch(...)` on a socket if no navigation is to take place? maybe @elias-ba can help here. at first glance, it feels strange to push this navigation event from the current URL _to_ the current URL. ideally, we'd display the flash message but not actually do any navigation.
3. i can't quite tell if the comments left in app.js are intentional. i notice that you're proposing to `preventDefault` across the _whole_ application, rather than just the job edit page. and then you've hooked up a new action for only the job edit page. is this intentional? are you putting a placeholder in for future developers to add more "control-s" behaviors elsewhere? let me know what you're thinking in app.js or make adjustments as necessary.

thanks!

PR #784, opened 2023-04-10
Fix issue when selecting from Postgres with a `::timestamptz` fragment

### LINKED ISSUE #782: refactor invocation.ex last run finished at
Line 443-ish is causing time zone problems and looks unnecessarily complex 

```ex
fragment(
              "nullif(max(coalesce(?, 'infinity')::timestamptz), 'infinity')",
              r.finished_at
            )
```

Use regular ecto `max` function, and maybe an `any` with an `or` to put a nil finished at at the top.

Ever since moving from UTC to BST (UTC+1) my tests fail with this:
```
  2) test runs list_work_orders_for_project/3 returns paginated workorders with ordering (Lightning.InvocationTest)
     test/lightning/invocation_test.exs:477
     Assertion with == failed
     code:  assert expected_order == page_one_result
     left:  [
              %{
                id: "86594e6d-8bcf-464b-b3ad-016c5f20dc52",
                last_finished_at: ~U[2023-04-09 09:04:01.505027Z]
              },
              %{
                id: "7281653f-d41a-449f-8e18-143727b3fb14",
                last_finished_at: ~U[2023-04-09 09:03:51.505027Z]
              },
              %{
                id: "e499d15d-1f29-4b19-a313-47cb81e27a69",
                last_finished_at: ~U[2023-04-09 09:03:41.505027Z]
              }
            ]
     right: [
              %{
                id: "86594e6d-8bcf-464b-b3ad-016c5f20dc52",
                last_finished_at: ~U[2023-04-09 08:04:01.505027Z]
              },
              %{
                id: "7281653f-d41a-449f-8e18-143727b3fb14",
                last_finished_at: ~U[2023-04-09 08:03:51.505027Z]
              },
              %{
                id: "e499d15d-1f29-4b19-a313-47cb81e27a69",
                last_finished_at: ~U[2023-04-09 08:03:41.505027Z]
              }
            ]
```

taylordowns2000:
## Notes for the reviewer

This is ugly, all around. I don't like the implementation because it relies heavily on fragments, and since the UK changed from UTC to UTC+1 the results returned from the DB have been off by one hour, cause the [tests here](https://github.com/OpenFn/Lightning/blob/main/test/lightning/invocation_test.exs) to fail.

There are three commits on this branch. The first removes an unnecessary fragment. The second uses a `distinct` and `order_by: [desc_nulls_first: r.finished_at]` to obviate the need for the first problematic fragment: `"nullif(max(coalesce(?, 'infinity')::timestamptz), 'infinity')",`

The third commit is the ugly one. Everything seems to work if we let Postgres return naive timestamps and then convert those to UTC after the query is done. I don't like this, but wonder if until we can refactor this query to avoid using that fragment we've got to do it: https://elixirforum.com/t/why-use-utc-datetime-over-naive-datetime-for-ecto/32532

## Related issue

Fixes #782 


stuartc (commented):
Thanks for taking a deeper look at this, it's been bugging me for a while. I get a bit turned around as to the correct way to fix this; your changes that remove the fragments look good. 

Just FYI there was _some_ reason there was a `nullif` function, something about unfinished runs not coming out on top. Lets make sure we haven't broken that.

As for the last commit (mapping over the `last_finished_at` and casting it from naive), I agree - really not ideal.

I quickly tried changing the timestamp schemas to `timestamps(type: :utc_datetime_usec)` from the `:naive_datetime_usec`. The tests passed after tweaking a test helper. While the underlying type on pg is `timestamp without time zone` I have a hunch this might be a replacement for the map casting.

> Removing like 21 of `model_helpers.ex` after changing all the timestamp types fixed the tests.

Are you able to create a test (even if it's a throwaway) to recreate this? I run pg inside a container with the timezone set as UTC so I can't recreate these kinds of issues without re-initing my db container.

taylordowns2000:
@stuartc , i'm nervous changing the models without changing the DB columns... or at least without knowing exactly how/why we'd want to do that. (this post was confusing, but seemed to say that naive is _good_ for performance. can you read/explain this?)

and one more q: why do a bunch of fields currently use naive while others use UTC?

Runs has `add :finished_at, :utc_datetime_usec` in the migration, and `field :finished_at, :utc_datetime_usec` in the model, but most (like `users` - use naive: `field :confirmed_at, :naive_datetime`)

I could be missing something, but it feels like these should all be the same.

stuartc:
@taylordowns2000 no you're right to be cautious;

If you take a look at the resulting schemas in psql you can see in the case of users:

```
 updated_at         | timestamp(0) without time zone | 
```

and in the case of runs:

```
 started_at         | timestamp without time zone | 
 finished_at        | timestamp without time zone | 
 inserted_at        | timestamp without time zone | 
 updated_at         | timestamp without time zone | 
```

`timestamp(0)` is second resolution and `timestamp` from what I understand is the max resolution (or usec).

Both of them are created without a time zone in postgres. So the real difference here comes from the `field` type in the schema.

From what I understand is that Ecto, for the differences between `utc_datetime` and `datetime` is that with `utc_datetime` will save a regular datetime (not-naive) as UTC if it's something like +3 to start with. As for a naive datetime it does nothing since there is no timezone info and assumes it's utc essentially.

If we are intentionally not using the timezone stuff in postgres (which ecto does by default), then it's up to us to provide the correct (i.e. _must_ be UTC) datetime values to Ecto.

We either make doubly sure that everything going into models are UTC or naive but are assumed to be correctly UTC. I believe one way is to use `:utc_datetime` which would do any coercion for us, but then again we should be feeding in the correct values in the first place.

I do see we have quite a bit of variance between our model schemas and their datetimes and we _should_ pick one (or two in the case of `*_usec`).

As for the overall issue here, I need to recreate this issue locally to see where the rouge timestamps are coming from.

taylordowns2000:
Hey @stuartc , all of the above works like a charm _except_ that we're running into issues with that `fragment` usage for one of our queries. It makes sense, and I'm happy to leave the `get_and_update` as that query will get refactored soon anyway. Thanks for all the help here.

PR #788, opened 2023-04-12
Fix project digest queries

### LINKED ISSUE #768: Incorrect project digest queries
**Describe the bug** 
The project digest email doesn't look right, it seems the queries are incorrect. On app.openfn.org, I got the following email: 

Primero demo:
- 36 workorders correctly processed this week
- 15 failed work orders that were rerun and then processed correctly
- 36 work orders that failed/still need addressing 

But when I went to the history page and filtered for the past week, I could only see 9 work orders. 

**Implementation notes**
1. Fix the queries. (See https://github.com/OpenFn/Lightning/blob/main/test/lightning/workorders_test.exs#L13) 
2. Update the email to the following, adding links to the filtered view on the history page and including the date range used for the queries
3. Update the email subject to it passes in the correct project_digest value (currently says 'Weekly project digest' when the digest is set to daily)

```
Hi xx, 

Here's a daily/weekly/monthly digest for project_name project activity between date_range.  

workflow_name:
- xx workorders correctly processed this week
- xx failed work orders that were rerun and then processed correctly
- xx work orders that failed/still need addressing
Click here to view this in the history page.

workflow_name:
- xx workorders correctly processed this week
- xx failed work orders that were rerun and then processed correctly
- xx work orders that failed/still need addressing
Click here to view this in the history page.
```

elias-ba:
## Related issue

Fixes #768 


amberrignell (changes requested):
Looking good Elias. One thing: for the work orders that failed and were then retried, we want the count of work order, rather than the count of attempts that failed. 

For example, given: 

wo1:
attempt 3: success
attempt 2: success
attempt 1: failure

wo2:
attempt 3: success
attempt 2: failure
attempt 1: failure

I expect to see:
2 wo that failed then were rerun and succeeded (rather than 3)

lib/lightning/accounts/user_notifier.ex
```diff
@@ -155,26 +154,86 @@ defmodule Lightning.Accounts.UserNotifier do
...
+        "filters[search_term]" => "",
+        "filters[success]" => true,
+        "filters[timeout]" => true,
+        "filters[wo_date_after]" => "",
+        "ilters[wo_date_before]" => "",
+        "filters[date_after]" => start_date,
+        "filters[date_before]" => end_date,
+        "filters[workflow_id]" => workflow.id
+      })
+    )
```
stuartc, line 177: We should be using verified routes here, we need to add `use LightningWeb, :html` to add support for the `~p` sigil. From there we can use it to also build the query string:

```
params = %{page: 1, direction: "asc"}
~p"/posts?#{params}"
```

See: https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html

lib/lightning/digest_email_worker.ex
```diff
@@ -43,17 +50,116 @@ defmodule Lightning.DigestEmailWorker do
...
+        search_fields: [:body, :log],
+        search_term: "",
+        workflow_id: workflow.id,
+        date_after: start_date,
+        date_before: end_date,
+        wo_date_after: "",
+        wo_date_before: ""
+      ],
+      %{}
+    )
```
stuartc, line 129: This interface feels a little off to me namely:

1. Things like `search_term` being an empty string, is that a default or an explicit search for `""` (likely not the case). Could these empty values be considered defaults and if they can, would we be able to omit them?
2. The 3 arguments (the 2nd being the filter, and the 3rd being pagination params), I think the filter should be changed to a map, and the params be defaulted to an empty map (`\\ %{}`).

Perhaps it's a good moment to create a struct for the search params (if needed), something that can hold defaults and coercion to keywords or uri params etc.

Something like:

```elixir
defmodule WorkorderSearch do
  defstruct [:status, :search_fields, ...]

  def new(attrs) do
    // possibly merge in defaults
    struct!(__MODULE__, attrs)
  end

  def to_params(search) do
    Map.from_struct(search) // possibly strip empty/null values?
    |> stringify_keys()
  end
end
```

This kind of approach would be used to replace part of `build_digest_url` and ensure the building up of search params stays in one place

stuartc (commented):
Looking good so far, take a look at my comment about the struct for working with search params. I think it's a good idea to unify the creation of the URI params for the mailer, and the handling of LiveView search params using a single common struct.

Also what happened to the `Workorders` module? I see it's been deleted here? What changes invalided its use?

taylordowns2000 (changes requested):
@elias-ba , I'm happy with this once you've addressed Stu's last review: https://github.com/OpenFn/Lightning/pull/788#pullrequestreview-1391351674

Thanks!

lib/lightning/digest_email_worker.ex
elias-ba: Done

CHANGELOG.md
```diff
@@ -33,6 +36,11 @@ and this project adheres to
 
 ### Fixed
 
+=======
+
+- Incorrect project digest queries
+  [#768](https://github.com/OpenFn/Lightning/issues/768)]
+
```
stuartc, line 43: Please remove the duplicate CHANGELOG line here.

lib/lightning/accounts/user_notifier.ex
```diff
@@ -155,26 +156,83 @@ defmodule Lightning.Accounts.UserNotifier do
...
+        "search_term" => true,
+        "success" => true,
+        "timeout" => true,
+        "wo_date_after" => "",
+        "wo_date_before" => "",
+        "date_after" => start_date |> DateTime.to_string(),
+        "date_before" => end_date |> DateTime.to_string(),
+        "workflow_id" => workflow.id
+      }
+    }
```
stuartc, line 176: Why aren't we using the SearchParams struct here? 

The point of that module is to handle the defaults and add a layer of type checking to ensure that changes to the SearchParams (and/or the form) stay consistent across the app.

test/lightning/accounts/user_notifier_test.exs
```diff
@@ -76,53 +77,249 @@ defmodule Lightning.Accounts.UserNotifierTest do
...
-    test "deliver_project_digest/4" do
-      user = %User{email: "real@email.com"}
+    test "build_digest_url/3" do
+      workflow = Lightning.WorkflowsFixtures.workflow_fixture(name: "Workflow A")
+      start_date = Timex.now()
+      end_date = Timex.now() |> Timex.shift(days: 2)
+
+      query =
+        UserNotifier.build_digest_url(workflow, start_date, end_date)
+        |> URI.decode_query()
```
stuartc, line 87: Please add a test to verify the path that is returned here is correct.

Use `URI.parse/1` and match on the `:path` key to check that the url has the correct path.

Oh and just thinking out loud here, you might want to use the `:rfc3986` option when decoding the query. Just try it. The default decoding pattern for decoding queries is `:www_form` which is slightly different to URI query params.

See [`decode_query/3`](https://hexdocs.pm/elixir/1.14.4/URI.html#decode_query/3)

stuartc (changes requested):
Super close dude, please address my last few points.

lib/lightning/accounts/user_notifier.ex
elias-ba: Done

test/lightning/accounts/user_notifier_test.exs
elias-ba: Done. Thanks for the recommendation

elias-ba:
> Super close dude, please address my last few points.

@stuartc all your change requests have been addressed. Please merge if this ready for you. Thanks

amberrignell (approved):
Nice one Elias! @stuartc  for context when you're reading, we added the 'pending' to the search query for failed,crashed and timeout work orders. We've identified a bug where workorders are marked as timeout in the UI but considered 'pending' (see recording in slack).
 I'll create an issue for this.

stuartc (approved):
:+1: thanks! Looks good to me, please check your diffs before push - I see a few places where the formatter has added parens where it doesn't need to.

I also changed the url builder function to use [`url/1`](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html#url/1) from VerifiedRoutes - which basically is the same thing just using Phoenixs functions.

PR #798, opened 2023-04-19
Expand scope of `delete project` function

### LINKED ISSUE #759: Expand scope and tests for "delete_project" function
- Either via appropriate use of cascades _or_ via well-planned associated deletes, we should ensure that the `Projects.delete_project` function removes all associated project activity.
- Note that it does _not_ remove `users` or `credentials`, but _does_ remove `project_users` and `project_credentials`.

## Testing

- Ensure that an attempted project deletion tries to delete a project with every different type of associated resource, including:

- [ ] workorders, attempts, run
- [ ] audit trail events
- [ ] project users
- [ ] project credentials
- [ ] jobs and triggers

mtuchi:
## Implimentation plan
- [x] Delete all association activity 
- [x] Delete the project
- [x] Update `Delete project` tests

## Related issue

Fixes #759 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to
 
 ### Changed
 
+- The `delete-project` function now delete all associated activities
+  [#759](https://github.com/OpenFn/Lightning/issues/759)
+
 ### Fixed
 
 ## [0.5.1] - 2023-04-12
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -161,13 +161,13 @@ defmodule Lightning.Projects do
 
       project_workorders_query(project) |> Repo.delete_all()
 
-      project_runs_invocation_reason(project) |> Repo.delete_all()
+      project_users_invocation_reasons(project) |> Repo.delete_all()
 
       project_runs_query(project) |> Repo.delete_all()
 
       project_jobs_query(project) |> Repo.delete_all()
 
-      project_trigger_invocation_reason(project) |> Repo.delete_all()
+      project_workflows_invocation_reason(project) |> Repo.delete_all()
 
       project_triggers_query(project) |> Repo.delete_all()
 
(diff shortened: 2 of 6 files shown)


taylordowns2000 (changes requested):
hey @mtuchi , thanks for this. in the tests, it doesn't look like you're asserting the existence of anything before you run the delete. how are you certain that all of those things exist? and do you know that you're deleting the resources for project A rather than project B?

please update your tests so that they _proove_ that this new function does what it says on the tin :-)

mtuchi:
@taylordowns2000 I have updated the test for project deletion, Not the project activities will be asserted before deletion

taylordowns2000 (changes requested):
1. please remove the comments from `lib/lightning/projects.ex` - they don't any provide additional information to the developer
2. please remove the second arguments you're providing to assert _or_ make it more valuable than the code and/or the error message provided by the failed assertion. for example: "There should be only one workorder" actually makes it harder to understand what's happening in line 214 because it repeats the code above it without providing the developer with any context or _reason_ that there should only be one workorder (matching a certain project).
3. let's talk about `full_project_fixture` tomorrow. it doesn't look like this creates a full project. rather, it looks like it might actually create two projects.

mtuchi:
@taylordowns2000 
1. I have removed the comments in `lib/lightning/projects.ex`
2. Let's chat about this in the AM tomorrow
3. Let's chat in the AM

mtuchi:
### Outstanding changing 
- [x] Delete runs for project
- [x] Test deleted runs in a project
- [ ] Verify that every related model that we want to delete is deleted
- [ ] Add a docstring to delete the project function
- [ ] Create functions for querying project-related data (invocation reason, runs, jobs, workflows, project users, project credentials, dataclip)
- [x] Remove new project added in project_fixture and use `%{ project: p1 } = full_project_fixture()
%{ project: p2 } = full_project_fixture()`

stuartc (approved):
Looks good to me @mtuchi 

While I don't think we should change this here, but all the deleting functions definitely need to be put in their own service. There are a lot of alias to modules outside of the project context that end up defeating the point of contexts (i.e. isolating behaviour).

@taylordowns2000 you can merge this when you're happy with it.

Issue #799, opened 2023-04-19
Better delete defaults for GDPR


taylordowns2000:
N.B.: All of the following assumes that the super user allows physical deletes. If a single tenant instance run by the MoH cannot allow deletes for any projects or users, this doesn't apply and both users and projects get marked as deleted/disabled, but never physically removed unless the super does so manually.

So, assuming real deletes happen, here's how they should happen:

- If a project owner requests for a project to be deleted, all related activity should be purged.
- If a user requests for their account to be deleted and there is still activity related to their account,
  - the `jobs` using their `credentials` must be set to `credential_id = null`,
  - the `project_credentials` linked to their `credentials` should be deleted
  - the `user` and `credential` records themselves cannot be purged until no runs remain which are linked
    - to their user via `invocation_reason`
    - or to their credentials via `run.credential_id`

That is to say that they will remain in the system until their project related audit trail has expired or is removed by the project owner. This will be explained to them over email... if you run jobs in a project that you do not own, you are agreeing to not be forgotten until that audit trail has expired. (Which may be never if you're on a single tenant MoH server.)

PR #801, opened 2023-04-20
Link runs directly to credentials

### LINKED ISSUE #800: Add credential id to run
To complete the audit trail we need to track credential id on runs, not just jobs.

taylordowns2000:
## Notes for the reviewer

@stuartc , this is a funky one with implications here on #800 and also on #799. I think that stamping the `credential_id` onto the run (for now with the on_start call, later with some even emitted from the RTM) is actually a really reasonable part of our usability/history (for audit) generation.

This gets around the frustration our v1 users have when they don't know which credential was used for a particular run if the job's credential has been changed recently. Note that it doesn't help you instantly see when a credential itself has been edited, but that can be done via the audit trail and this makes that journey to discovery easier.

image

see #799 for why users _cant_ be deleted if their credentials have been used in runs that still exist on the system.

## Related issue

Fixes #800 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to
 
 - Add `workflow_edges` table in preparation for new workflow editor
   implementation [#794](https://github.com/OpenFn/Lightning/issues/794)
+- Stamped `credential_id` on run directly for easier auditing of the history
+  interface. Admins can now see which credential was used to run a run.
+  [#800](https://github.com/OpenFn/Lightning/issues/800)
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


stuartc:
@taylordowns2000 is this the solution we really want? I mean yes on the face, even with the caveats it's fine but given the caveats and how we actually envision this working in the future; do we want to change our models now knowing that we need to plan out something more comprehensive like a snapshotted workflow?

taylordowns2000:
merging but noting that this may present a (useful) doorstop for a bigger design decision later... how do we guarantee the audit trail at the level of the plan.

Issue #793, opened 2023-04-18
Nodes and Edges Refactor


stuartc:
This epic outlines the changes required to introduce the concept of edges and
nodes into Lightning. Re: #778

Currently our Triggers belong to Jobs and Workflows, this makes it difficult to
create a trigger as the workflow and the job must at least have an ID; and
ideally they should be persisted first.

We have some 'fancy' `cast_assoc/3` functions that allow us to create a trigger
via a job that copies the jobs `workflow_id` to the trigger.

By introducing the concept of edges and nodes we can create a trigger without
associating it to a job; and vice-versa for a job.

Another use case is having conditions between node (jobs or triggers; but just
jobs for now), allowing for users to conditionally control the flow of their
workflows based on data.

Below represents a Job that has a condition on it, if the condition is true then
the subsequent job will be executed.

```mermaid
---
title: WorkflowEdge with Condition
---
flowchart LR
    A[Job A]-->|state.data == true|B[Job B]
```

```mermaid
---
title: WorkflowEdge with Runtime Condition
---
flowchart LR
    A[Job A]-->|on_success or on_failure|B[Job B]
```

## ER Diagram

Given the ER diagram below, we can see that we have a `WorkflowEdge` model that
holds the relationship between a each node in the Workflow.

```mermaid
erDiagram
    Workflow {
        uuid id PK
    }
    WorkflowEdge }o--|| Workflow : belongs_to
    WorkflowEdge o|--o| Job : source
    WorkflowEdge o|--o| Job : target
    WorkflowEdge o|--o| Trigger : triggerable
    WorkflowEdge {
        uuid id PK
        uuid workflow_id FK
        uuid source_trigger_id FK
        uuid source_job_id FK
        string precondition
        uuid target_job_id FK
    }
    Trigger }o--|| Workflow : belongs_to
    Trigger {
        uuid id PK
        uuid workflow_id FK
    }
    Job }o--|| Workflow : belongs_to
    Job {
        uuid id PK
        uuid workflow_id FK
    }
```

The `WorkflowEdge` model has a `precondition` field that will hold an expression
that will be evaluated at runtime to determine if the edge should be followed.

> _\*TDB_: In order to represent "on success" and "on failure" conditions we
> will need to represent these as `true` or `state.error` respectively. I have
> concerns about leaning on the `state` for this since it's user modifiable, and
> also get overridden in the case of a failure.

**Triggers**

Are no longer associated to a Job, but instead to a Workflow. This means that
[...]

PR #824, opened 2023-05-05
New Workflow Diagram

josephjclark:
This PR updates react-flow and adds a new diagram, using `d3-hierarchy` for layout and the new nodes/edges model for the workflow itself.

Most of the effort is in a standalone dev server, which has a little demo app, pictured:

Screenshot from 2023-05-05 17-41-52

## Architecture

The `WorkflowDiagram` component is responsible for rendering a chart with react-flow. A zustand store is the contract between the chart and the app.

The `WorkflowDiagram` controls the state of the chart. It works out a layout with `d3-hierarchy` and passes the positions into the chart.

Changes can be pushed into the component form outside, in which case existing layout and selection state will be preserved, and new nodes will animate into place.

If the diagram wants to make a change (ie adding a node), it should call a callback requesting the wrapping application to add the changes to the store.

The idea is that the component itself doesn't use the store - it just takes a model of the workflow, converts it to a react-flow structure, and publishes targeted events when it wants to instigate changes. The wrapping app/component then interfaces that with the store.

I think this is easier and cleaner - but I am still a bit concerned about the state management side of things and wonder if the store needs pushing further down.

I hit a really big problem with selection. We can't prevent react-flow selecting a node on-click. This causes a conflcit when adding a new node: clicking the plus icon counts as a click on the node, which changes selection. There's a messy hack right now to ignore the selection change event to allow us to select the new placeholder, but it makes me grumpy.

## Placeholder nodes

I'm playing with creating placeholder nodes in the diagram.  These are special nodes which are designed to go with the new form. There are limits on how many placeholders can be created (the idea is to stop users spamming loads of empty new nodes).

If the form is closed without changes, I am tempted to remove the placeholder.

Ideally placeholders wouldn't be published to the server. As soon as they are edited (ie given a name or adaptor) they  are converted to real nodes.

## Demo

The best way to play right now is from the demo app.

Start the server: `cd assets && npm start`

Go to `localhost:3000`

The control bar in the top right pushes state into the component - representing navigation or server side changes. Add Job will either add from node a (TODO: use a random node) or from the seleced node. New nodes will animate in, but a whole new chart should just appear.

The form right now is a read only representation of a node.

The change events displays patches from the store.

Adding a node from within the diagram will create a placeholder node. Placeholders can't have new children. TODO restrict a parent node to only have one child placeholder. TODO As soon as the placeholder is given a name, adaptor or expression, convert it into a real node.

## Integration

I've started to integrate with Lightning properly but there's not actually a lot I can do yet. Hopefully as the backend gets developed the chart will respond appropriately without too much work.

Access the new workflow by replacing `/w/` with `w/w-new` in the workflow page url.

The sidepanel is using Stu's example and doesn't do much. Sample data is sent to the page for each workflow so the diagrams aren't very interesting.

I don't think I'm handling the store-diagram relationship very well in the actual chart and need to spend a bit more time on this.

## TODO

* [ ] Finish off placeholder nodes
* [ ] Stare at the store some more
* [ ] I really don't like the use of refs for posittion and selection, so I'd like to stare hard at that for a while too
* [ ] I'm not sure I've done node sizing very well and, at the very least, I'd like to set the size constants centrally and re-use them where needed.
* [ ] Properly control selection of edges
* [ ] Draw error edges to one side (but how do we we know?)
* [ ] ... Probably more


josephjclark:
There's some strange stuff in the diff, I hope that's just because I'm out of date. I'll review tomorrow and get it straight

josephjclark:
@stuartc more work will be needed on this but this is probably a good time to review and merge the progress so far :)

assets/js/workflow-editor/component.tsx
```diff
@@ -1,102 +1,82 @@
...
+      // TODO listen to change events from the diagram and upadte the store accordingly
+      <WorkflowDiagram
+        ref={el}
+        workflow={identifyPlaceholders(model)}
+        onSelectionChange={handleSelectionChange}
+        requestChange={handleRequestChange}/>
+    );
+  }
+
+  const unsubscribe = workflowStore.subscribe(render)
```
stuartc, line 79: Whats the reasoning for subscribing here? Does `useStore` not cut it?

stuartc (approved):
Sho, this is a lot; this is a great start!

We must sync up for a chat about the store and how the updating works now that we have a lot more surface area to consider.

I'll add some TODOs as we go to the epic #793 for cleanup, I encourage you to do the same - assuming they don't quite belong in the story that currently in progess.

PR #842, opened 2023-05-16
Better placeholders

josephjclark:
Experimental PR for better placeholder nodes.
image

The new placeholder has commit and cancel options and steals focus, instantly demanding a name. Escape and Enter will commit or remove the placeholder.

This makes it quicker, easier, and more satisfying to sketch out new jobs in a workflow. It also helps makes placeholders a bit more transient, encouraging a name to be set early (so you're less likely to create a chart with many placeholders).

This video uses the keyboard to create an cancel a couple of placeholders:

https://github.com/OpenFn/Lightning/assets/7052509/f517c34f-f7f1-4e66-b500-8e799c2d76e2

## Issues

There are some UX-y issues which I've managed to mitigate:

* Only one placeholder can be created at a time, which prevents all sorts of cleanup issues.
* If the side panel is open, adding a new node does not open the side panel. But if the panel is already visible, adding a new node will update the panel's contents. I do not want to hide the panel, even though it would be a bit neater, because the chart will jump all over the place.
* The placeholder node and side panel act differently. As soon as the side panel is edited, the placeholder status is replaced. But with the placeholder's own text input, the node isn't created until enter is pressed. The UX in the diagram is WAY nicer. Controlling the Lightning form is harder and out of my control.
* Later, in Lightning, we may consider a special Placeholder side panel form, which only lets you edit the name of the node, and doesn't save until you hit enter.
* For large workflows, the auto-zoom doesn't work very well, and the text input in the node gets very small. But this is not likely to be a problem in practice.
* When a node is first committed, we set a empty job expression on it. This isn't ideal, but it prevents the node reverting back to a placeholder state when its name is removed (which can easily happen and looks very bad). 

IMPORTANT NOTE: To get this last bit working I've removed the default job string from `lib/lighting/jobs/job.ex`. If this is unacceptable we're gonna need another solution.

There is one bug on this branch I can't really fix:

Pressing enter while the new form has focus will trigger an error, as the form posts to a URL that doesn't exist. This might crop up as you're adding nodes.

## Next steps

I've raised #843 for more keyboard control, but that's not something to worry about right now.

### DIFF
--- a/lib/lightning/jobs/job.ex
+++ b/lib/lightning/jobs/job.ex
@@ -41,8 +41,7 @@ defmodule Lightning.Jobs.Job do
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
   schema "jobs" do
-    field :body, :string,
-      default: "// Get started by adding operations from your adaptor here"
+    field :body, :string
 
     field :enabled, :boolean, default: true
     field :name, :string
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -8,7 +8,7 @@ import createCompletionProvider from './magic-completion';
 // static imports for core lib
 import dts_es5 from './lib/es5.min.dts';
 
-const DEFAULT_TEXT = '// Get started by adding operations from the API reference\n';
+export const DEFAULT_TEXT = '// Get started by adding operations from the API reference\n';
 
 type EditorProps = {
   source?: string;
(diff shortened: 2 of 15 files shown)


lib/lightning/jobs/job.ex
```diff
@@ -41,8 +41,7 @@ defmodule Lightning.Jobs.Job do
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
   schema "jobs" do
-    field :body, :string,
-      default: "// Get started by adding operations from your adaptor here"
```
josephjclark, line 45: @stuartc just wanna flag this diff before I merge down - I've removed the default `body` setting for a job. It makes it easier for the client to distinguish placeholder nodes from real ones.

Does this seem alright?

PR #845, opened 2023-05-17
[epic] Nodes and edges release

josephjclark:
Release PR for the new nodes-and-edges style workflow

* [x] New edge-based model
* [x] New workflow diagram
* [x] New job editor forms
* [x] Old workflow diagram removed

Amber's quick notes: 
**Create new workflow**
- [ ] first thing on the page should be either the trigger form open, or button "create a trigger to start your workflow"
- [x] When I change a trigger from web hook to cron, I expect the cron form to show up (and the copy web hook URL to disappear)
- [x] Trigger node should display the webhook URL or cron expression, not "New Trigger" 
- [ ] Tooltip should display fully (currently cut off by form size)
- [ ] Should be able to save a new workflow (generate a random name for the workflow)

**Job setup**
- [x] When a user selected an adaptor, the default version that is selected should the latest stable version (ex: 4.0.1 instead of latest >=4.0.1)
- [ ] When I create a new job, the job form should open
- [x] Should be able to rename a job
- [x] can be shared via url

**Job editor**
- [x] should have input, editor and output
- [x] should display the job code
- [x] should load the adaptor docs
- [x] can be shared via url
- [ ] should display the adaptor same way as the adaptor dropdown ("salesforce" rather than "language-salesforce" etc)

**Run a job**
- [x] input tab displays 3 latest inputs from runs for that job
- [x] allows user to run the job with a new input
- [x] when the user clicks run, displays the output on the output tab (and the run appears in the run history too)

**Delete job**
- [x] delete job button should be on the job setup form
- [x] when a job is deleted, its associated edge should be deleted
- [x] cannot delete a job that has downstream jobs


stuartc:
Ah dang, I think I know what the issue is. I’ll take a look shortly. Should be pretty simple.On 17 May 2023, at 17:02, josephjclark ***@***.***> wrote:﻿
@stuartc We seem to have an issue on this branch where hitting return in the form tries to post to an unregistered route:

open /w-new/
Select a job
Click in the job name field
Hit return
Boom 💥

taylordowns2000 (dismissed):

PR #851, opened 2023-05-29
Store log lines as rows in a table

### LINKED ISSUE #514: Store log lines as rows in a table
### Details

Currently logs are persisted in a single column as an array of strings.

By moving logs into a separate table we:

- Gain the ability to persist logs more easily as they arrive.
- Filter and tag individual lines by prefixes or by `STDOUT/STDERR`.
- Have individual timestamps for each line.
- Make accessing Runs more efficient by not having to remember to _not_ load the logs every time.

elias-ba:
## Implementation plan

- [x] Create the `run_logs` table
- [x] Make sure associations between the `runs` table and the `run_logs` table is established
- [ ] Update `runs` creation to now write the logs in the `run_logs` table. Scan the code and change things like `create / update runs`.
- [ ] Update Workorders search queries to search on the `run_logs` table when log filter is checked.
- [ ] Adjust tests 

## Related issue

Fixes #514 


lib/lightning/pipeline/runner.ex
```diff
@@ -47,21 +48,34 @@ defmodule Lightning.Pipeline.Runner do
...
       dataclip_result = Runner.create_dataclip_from_result(result, run)
 
       Lightning.FailureAlerter.alert_on_failure(run)
 
       dataclip_result
     end
+
+    defp save_run_logs(logs, run) do
+      Enum.each(logs, fn log ->
+        if log != "" do
```
stuartc, line 74: Interesting... I'm not sure we want to strip out log lines that are empty?

I mean it seems silly to persist them, so maybe the shape of the lines (i.e. whether we include blank ones or not) lives outside of here..

test/lightning/pipeline/state_assembler_test.exs
```diff
@@ -74,7 +74,7 @@ defmodule Lightning.Pipeline.StateAssemblerTest do
 
       assert StateAssembler.assemble(run) |> Jason.decode!() == %{
                "configuration" => %{"other" => "credential"},
-               "error" => ["I've failed, log log"],
+               "error" => "I've failed, log log",
```
stuartc, line 77: The error value in job state is expected to be an array.

lib/lightning/pipeline/runner.ex
elias-ba: Done

priv/repo/migrations/20230531122638_rename_run_logs_table.exs
```diff
@@ -0,0 +1,11 @@
...
+  use Ecto.Migration
+
+  def change do
+    rename table(:run_logs), to: table(:log_lines)
+
+    # Rename the index
+    execute "DROP INDEX IF EXISTS run_logs_run_id_index"
+    create index(:log_lines, [:run_id])
+  end
+end
```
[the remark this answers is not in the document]
elias-ba: Fixed

test/lightning/pipeline/state_assembler_test.exs
elias-ba: Reverted

lib/lightning_web/live/run_live/run_viewer_live.ex
```diff
@@ -41,5 +42,6 @@ defmodule LightningWeb.RunLive.RunViewerLive do
       preload: :output_dataclip
     )
     |> Lightning.Repo.one()
+    |> Repo.preload(:log_lines)
```
stuartc, line 45: FYI I'm undecided whether we should preload logs or not, I see situations where it might be better to load them only when the logs tab is rendered. I've seen issues already with the way we are rendering them (i.e. browser crash) - so just a heads up that we need to revisit how and when we render log lines in LiveView.

stuartc (approved):
Great job @elias-ba :heart:

PR #853, opened 2023-06-01
Form for trigger nodes in the new side panel

### LINKED ISSUE #810: New side panel/form for trigger
- Create a Phoenix form that will be rendered in the canvas side-panel to create/edit triggers
- Be careful to implement authorization based on the new patterns
- these are webhook or cron triggers (flow/fail are edges now)

- make a changeset for the trigger
- pull out the trigger parts from the job form and put it in its own trigger form 
- when you click on the trigger node, open the panel 
- users should be able to save a trigger independently

elias-ba:
## Implementation plan
- [x] When a Trigger node is clicked, make sure the side panel renders the trigger form with the cron builder
- [x] Interacting with the trigger form should call the validate function with the trigger type.
- [x] When the trigger type is cron, the validate function should receive the trigger type and the cron expression coming from the cron setup component
- [x] Make sure validate is validating the trigger form changeset
- [x] Add liveview tests for the trigger form - most of these tests had already been written, we will probably need to copy paste and adjust

## Related issue

Fixes #810 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/job_live/cron_setup_component.ex
+++ b/lib/lightning_web/live/job_live/cron_setup_component.ex
@@ -105,12 +105,7 @@ defmodule LightningWeb.JobLive.CronSetupComponent do
 
   @impl true
   def update(
-        %{
-          id: id,
-          form: form,
-          on_change: on_change,
-          disabled: disabled
-        },
+        %{id: id, form: form, on_change: on_change, disabled: disabled},
         socket
       ) do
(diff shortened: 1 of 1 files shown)


stuartc (approved):
Great stuff @elias-ba

stuartc:
We're not going to add anything to the CHANGELOG for now, will take care of it before the release branch hits main.

PR #854, opened 2023-06-01
Check and renew OAuth credentials when running a job

mtuchi:
## Steps

- [x] ~~Write a function that can tell the difference between a token that is about to expire and one that is not~~
- [x] Write a test for `Google.refresh_token()`
- [x] Write `Credentials.maybe_refresh_token/` which takes a credential and checks if its token is about to expire and renew it and update the credential, otherwise return the same credential.
- [x] Write a test for `Credentials.maybe_refresh_token/1`. That test should capture the scenario when a credential is not about to expire (return the same credential) and the one when a credential is about to expire (renew it, update its body, and return it)
- [x] Make sure when running a job that requires an OAuth credential, the access_token is refreshed. That requires calling `Credentials.maybe_refresh_token/1` in the runner module.

## Notes to reviewer

**@stuartc @zacck for some reasons I can't figure out, Dialyzer is complaining about missing @spec on the `Google.refresh_token/2` and `Credentials.maybe_refresh_token/1` functions. I have made sure to write / edit the @spec of all those two function but still can't pass Dialyzer's warnings about this.**

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to
 
 ### Added
 
+- Check and renew OAuth credentials when running a job
+  [#646](https://github.com/OpenFn/Lightning/issues/646)
+
 ### Changed
 
 ### Fixed
--- a/config/config.exs
+++ b/config/config.exs
@@ -112,7 +112,7 @@ config :lightning, Lightning.FailureAlerter,
 # Disables / Hides the credential transfer feature for beta (in LightningWeb.CredentialLive.Edit)
 config :lightning, LightningWeb,
   allow_credential_transfer: false,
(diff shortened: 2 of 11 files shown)


amberrignell:
Closing this PR for now as it won't be worked on for a while. Left a comment on the issue so that whoever picks it up can keep going on the same branch with the existing work that has already been done

taylordowns2000:
sounds good. coincidentally, @mtuchi and i were just discussing this as it relates to the new `msgraph` adaptor (which relies on oauth)

taylordowns2000 (changes requested):
td note to self

taylordowns2000 (approved):

PR #861, opened 2023-06-08
Changes from deploy CLI integration

stuartc:
**Notable changes**

Changing unauthorized API responses from 404s to 401s.

This is necessary because we need to be able to tell the difference
between problems with the users API key (not their project or their role
isn't correct) and _actually_ new projects so that the state merging
works correctly.

## Related issue

Re: https://github.com/OpenFn/kit/issues/245

## Review checklist


stuartc:
We need to change authorisation failures to 403s, but keep the 401 for invalid tokens etc.

taylordowns2000 (approved):
This is great, @stuartc . I've made some changes that I'd like to discuss tomorrow AM briefly but it looks (and works, with the new `openfnx deploy` like a charm!)

PR #865, opened 2023-06-09
Bulk rerun work orders from start

### LINKED ISSUE #659: Bulk rerun work orders (from start)
### User story
As a user that had a workflow which was running successfully, who then realised the logic isn't what I wanted, I would like to be able to bulk reprocess workflows, so that each workorder can be rerun with my updated logic.

### Details
Allow users to select work orders from the history page
Allow users to reprocess the work orders they have selected

1. Selecting work orders
Using the [following component](https://tailwindui.com/components/application-ui/lists/tables#component-390db22a862687fe88b313692989ffaf), add a checkbox column to the history page. 

- There should be a checkbox in the first column (to the left of 'Workflow'). 
- Every work order should have a checkbox, unselected by default
- When the user selects the top-level checkbox in the column heading, all work orders on the page get selected
- A user can select several individual work orders using the checkboxes within the work order rows
image

If a user has selected some work orders, then changes the filters, their selection is lost. 

2. Reprocessing work orders
- When a user selects either the top-level checkbox OR work order level checkboxes, a 'rerun from start' button should appear. 
- When the user clicks 'Rerun from start', the user should see a pop up modal which confirms which work orders have been selected (how many work orders will be reprocessed, and the search filters used to select these work orders)

Use [the following](https://tailwindui.com/components/application-ui/overlays/modals#component-f39bfe26a95719b1611d27f969be9a24) modal component

**A.Top-level select**
Message: `You've selected the number_of_workorders_on_page work orders from page age_number of results for workorders matching the following query search_params. There are a total of number_of_workorders_matching_query work orders matching this query.`

_Example: You've selected the 12 work orders from page 1 of results for workorders matching the following query:  received between x and y date, for workflow A, with a status of 'Failed'. There are a total of 32 work orders matching this query._

Buttons: Cancel, Rerun all work orders on this page from start, Rerun all work orders matching the query. 

 **B.Individual work orders selected**
Message: ` You've selected number_of_selected_work_orders work orders to rerun from the start (first job). This will create a new attempt for these within the same work order`
_Example: You've selected 12 work orders to reprocess from the start (first job). This will create a new attempt for these within the same work order._
Buttons: Cancel, Rerun from start

- When the user clicks 'Rerun from start', the users gets an alert "xx work orders will be bulk rerun" (so that they have feedback that their action was applied), and the work orders all get reprocessed in the same order in which they were received

### Implementation notes

Please read through the [platform-app implementation](https://github.com/OpenFn/platform-app/pulls?q=is%3Apr+is%3Aclosed+bulk) first
Write a function that can take a list of work order IDs  and create a new attempt for each one, in the order in which they have been passed to the function. 
Write a function that can take a query, and create a new attempt for each work order returned by the query, in the order in which they are returned.

The function rerun_run_from_start parameters would be: 
source: (list of IDs, or parameters) 
user: (used to add to the invocation reason)

### Release notes
- Bulk rerun work orders from start

### User acceptance criteria
- [ ] I can 'select all' work orders on the history page
- [ ] Work orders get reprocessed in the order in which they were received
- [ ] when I click 'Rerun from start', I see a pop-up which confirms how many work orders will be reprocessed, and the search filters used to select these work orders (ex: Reprocess 12 work orders which were received between x and y date, for workflow A, with a status of 'Failed')
- [ ] If a user changes the filters, they lose their selected items. 
- [ ] If I click the top level select, all workorders are selected. If I unselect one, the top-level select is also unselected. If I click unselect from the top level, all of them are unselected.

midigofrank:
### UI 
- [x] Add an unselected checkbox to the first column (to the left of 'Workflow') to every row in the workflows history table
- [x] Display rerun from start when any of the checkboxes is checked
- [x] Check all the checkboxes in the table when the top-level heading checkbox is selected
- [x] Uncheck the checkboxes when the user changes the filter
- [x] Add confirmation modal
- [x] Add relevant tests
### LOGIC
- [x] Add function to rerun work orders selected by the user. ??

## Related issue

Fixes #659 

## Review checklist

- [x] Amber has **QA'd** this feature

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -75,23 +75,16 @@ Hooks.Copy = {
   },
 };
 
-Hooks.CheckboxState ={
+// Sets the checkbox to indeterminate state if the element has the
+// `indeterminate` class
+Hooks.CheckboxIndeterminate = {
   mounted() {
-    if (this.el.classList.contains('indeterminate')) {
-      this.el.indeterminate = true;
-    } else {
-      this.el.indeterminate = false;
-    }
+    this.el.indeterminate = this.el.classList.contains('indeterminate');
   },
   updated() {
-    if (this.el.classList.contains('indeterminate')) {
-      this.el.indeterminate = true;
-    } else {
-      this.el.indeterminate = false;
-    }
-  }, 
-  
-}
+    this.el.indeterminate = this.el.classList.contains('indeterminate');
+  },
+};
 
 // @ts-ignore
 let csrfToken = document
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -75,6 +75,24 @@ Hooks.Copy = {
   },
 };
 
+Hooks.CheckboxState ={
+  mounted() {
+    if (this.el.classList.contains('indeterminate')) {
+      this.el.indeterminate = true;
+    } else {
+      this.el.indeterminate = false;
+    }
+  },
+  updated() {
+    if (this.el.classList.contains('indeterminate')) {
+      this.el.indeterminate = true;
+    } else {
+      this.el.indeterminate = false;
+    }
+  }, 
+  
+}
+
 // @ts-ignore
 let csrfToken = document
   .querySelector("meta[name='csrf-token']")
--- a/lib/lightning/attempt_service.ex
+++ b/lib/lightning/attempt_service.ex
@@ -111,6 +111,12 @@ defmodule Lightning.AttemptService do
     end)
   end
 
+  @doc """
+  Creates new Attempts for each pair of corresponding AttemptRun and
+  InvocationReason.
+  """
+  @spec retry_many([AttemptRun.t()], [Lightning.InvocationReason.t()]) ::
+          Ecto.Multi.t()
   def retry_many(
         [%AttemptRun{} | _other_runs] = attempt_runs,
         [%InvocationReason{} | _other_reasons] = reasons
@@ -255,6 +261,11 @@ defmodule Lightning.AttemptService do
     |> Repo.one()
   end
 
+  @doc """
+  Returns a list of AttemptRun structs that should be rerun for the given list
+  of work order ids.
+  """
+  @spec list_for_rerun_from_start([Ecto.UUID.t()]) :: [AttemptRun.t()]
   def list_for_rerun_from_start(order_ids) when is_list(order_ids) do
     attempt_run_numbers_query =
       from(ar in AttemptRun,
--- a/lib/lightning/attempt_service.ex
+++ b/lib/lightning/attempt_service.ex
@@ -133,7 +133,7 @@ defmodule Lightning.AttemptService do
         Enum.map(attempt_runs, fn %{attempt: %{work_order: work_order}, run: run} ->
           %{
             work_order_id: work_order.id,
-            reason_id: reasons_map[run.id],
+            reason_id: Map.fetch!(reasons_map, run.id),
             inserted_at: now,
             updated_at: now
           }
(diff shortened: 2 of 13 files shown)


taylordowns2000 (changes requested):
looks great @midigofrank . i've made some minor language tweaks but have one code tailwind thing i'd like to request:

- [ ] when stretching the cancel button, align items center (image 1 and 2). cancel should be in the middle.

image
image

after this is done, we're good from a UX perspective and simply awaiting @zacck and/or @stuartc 's thoughts on the code

stuartc:
@taylordowns2000 I noticed when retrying a Workorder from start, it starts 'from the first job', as in it copies the first job and goes from there:

image

Did you experience this? If I hit rerun on the first job it reruns _that_ job.

taylordowns2000:
hrmmm @stuartc this works properly for me. 
image


**EDIT.... I SEE YOUR EDIT!**
maybe the demo runs are still not being built properly. @amberrignell , can you check this out?

stuartc:
> **EDIT.... I SEE YOUR EDIT!** maybe the demo runs are still not being built properly. @amberrignell , can you check this out?

Yeah this is because of the demo data issue... I just reset my db and looked at the resulting rows in the db.

taylordowns2000:
awesome. ok, but i think `list_for_rerun_from_start` is the biggest gnarliest thing to look at here —> for a bunch of workorders how do you find which `attempt_runs` to send for bulk retry. does the logic/query structure look good to you?

lib/lightning/attempt_service.ex
```diff
@@ -157,6 +255,46 @@ defmodule Lightning.AttemptService do
...
+    attempt_run_numbers_query =
+      from(ar in AttemptRun,
+        join: att in assoc(ar, :attempt),
+        join: r in assoc(ar, :run),
+        where: att.work_order_id in ^order_ids,
+        select: %{
+          id: ar.id,
+          row_num:
+            row_number()
+            |> over(partition_by: att.work_order_id, order_by: r.started_at)
```
stuartc, line 268: What is the purpose of windowing these rows, below we're only pulling out the first one - wouldn't it make more sense to pick the first row from a query ordered by started_at?

I also want to suggest that we try order by `r.started_at OR r.inserted_at`, because although rare it's not impossible for a run to have not started (and won't ever be - due to configuration issues).

stuartc (changes requested):
This is fantastic work :+1:, really well put together!

I made two comments if you wouldn't mind addressing them.

zacck (commented):
Otherwise mostly looks good to me, albeit that might be because

assets/js/app.js
```diff
@@ -75,6 +75,24 @@ Hooks.Copy = {
...
+  },
+  updated() {
+    if (this.el.classList.contains('indeterminate')) {
+      this.el.indeterminate = true;
+    } else {
+      this.el.indeterminate = false;
+    }
+  }, 
+  
+}
```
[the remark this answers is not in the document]
stuartc: Ah yeah! Derp! :sweat_smile: totally forgot the indeterminate was a prop on the element and not just a class. Thanks!

lib/lightning/attempt_service.ex
stuartc: Yeah thanks @midigofrank, it got my attention we were selecting `num_workorders * runs_for_all_attempts`, sorting and numbering them just to drop all but one (for each workorder).

I would have thought dropping the ones that aren't the 'first' before joining would be more efficient. But eh, not a hill I'm willing to die on! :sweat_smile:

After playing around, the way I was expecting it to be is likely much more work than the negligible (if not zero) performance gain. 

Thanks!

stuartc (approved):

PR #889, opened 2023-06-20
SQL script for converting triggers to edges

### LINKED ISSUE #875: Write/run sql script to convert triggers
To keep projects built on app.openfn.org, we must execute a set of small SQL scripts before running the `drop_x` migrations from the `nodes-and-eges-release` branch.

This issue is to create a SQL script that does the following:

fetch a list of triggers from the database
check what each trigger is doing, and make a new edge for it based on the criteria below: 
- [ ] for each trigger where type = `cron`, create an edge from that trigger to its associated job with condition `'always'`
- [ ] for each trigger where type = `on_job_success`, create an edge from the upstream job to the downstream job with condition `'on_job_success'`
- [ ] for each trigger where type = `on_job_failure`, create an edge from the upstream job to the downstream job with condition `'on_job_failure'`

elias-ba:
## Implementation plan

- [x] for each trigger where type = `cron`, create an edge from that trigger to its associated job with condition `'always'`
- [x] for each trigger where type = `on_job_success`, create an edge from the upstream job to the downstream job with condition `'on_job_success'`
- [x] for each trigger where type = `on_job_failure`, create an edge from the upstream job to the downstream job with condition `'on_job_failure'`

## Related issue

Fixes #875 

## Review checklist


lib/lightning_web/live/credential_live/edit.ex
zacck, line 1: Hey @elias-ba  why this change?

zacck (changes requested):

lib/lightning_web/live/credential_live/edit.ex
elias-ba: I am suprised by this being in this branch. But it's still a change we need in the app. That code is no longer needed.

taylordowns2000 (approved):

PR #906, opened 2023-06-26
Bulk rerun workorders from a specific job

### LINKED ISSUE #833: Bulk rerun workorders (from a specific job)
### User story

As a user that has a multi-job workflow where a data is

1. fetched from system A
2. transformed to structure B
3. and _then_ loaded to system C

when I realize that the _second job_ is not working properly, I would like to bulk reprocess work orders _from that specific job_, so that I don't need to fetch all the data again in step 1.

Building on #865 , rather than initiating a new attempt from the first `attempt_run` in the last attempt for each selected workorder, this feature will allow the user to initiate a new attempt from from the the "job X" attempt_run (if it exists) in the last attempt for each selected workorder.

### Details

When a user has selected work orders that all belong to the SAME WORKFLOW, add a "Rerun from..." button next to the rerun button (rename to rerun instead of rerun from start).

Screenshot 2023-06-21 at 15 11 56

The options in the list are all of the jobs that are currently in the workflow. 

### Implementation notes

1. The "rerun from job X" interface should only appear when workorders for a SINGLE workflow are selected.
2. If "all" workflows matching the query are selected, the query must include a single workflow condition.
3. If you rerun 10 workorders "from job 7" but the final attempt for each of those workorders does not contain an attempt_run for "job 7", then only those containing an attempt_run for job 7 will be rerun.
4. This might be done by filtering out runs where `job_id != 7` [here](https://github.com/OpenFn/Lightning/pull/865/files#diff-3ec2daa312c9b31faa10f1083f49db9def6a22db73999a3dcfb9430ab8295f64R276-R277) - the important thing is that the filtering happens AFTER we have already selected the most recent attempt for each workorder.

Filtering sequence:
1. for these 4 selected workorders (4 rows)
2. find last attempt for each (4 rows)
3. find the run where `job_id = 7` in the last attempt (3 rows if one attempt does not include a job 7 attempt run)
4. create 3 new attempts from those 3 `attempt_runs`

### User acceptance criteria

- [x] The "rerun from..." button appears when I select multiple workorders that belong to the same workflow
- [x] The "rerun from..." button disappears (but the "rerun" button remains) when I select multiple workorders that belong to _different_ workflows.
- [x] When I "rerun from job X" and the final attempt for a selected workorder does not include job X, no new attempt will be made for that workorder

midigofrank:
### UI 
- [x] Display rerun from... button when the selected workorders are only for a single workflow
- [x] Add confirmation modal
- [x] Order the steps accordingly in the modal
- [x] Add relevant tests
### LOGIC
- [x] Add function to filter attempt runs based on given job_id

## Related issue

Fixes #833 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 - Add public API token to the demo site setup script
 - Check and renew OAuth credentials when running a job
   [#646](https://github.com/OpenFn/Lightning/issues/646)
+- Add feature to bulk rerun work orders from a specific step in their workflow;
+  e.g., "rerun these 50 work orders, starting each at step 4."
+  [#906](https://github.com/OpenFn/Lightning/pull/906)
 
 ### Fixed
 
--- a/lib/lightning/attempt_service.ex
+++ b/lib/lightning/attempt_service.ex
@@ -203,6 +203,10 @@ defmodule Lightning.AttemptService do
     )
   end
 
+  def retry_many([], []) do
+    Multi.new()
+  end
+
   defp attempt_runs_attrs(attempt_id, runs, timestamp) do
     Enum.map(runs, fn run ->
       %{
@@ -307,6 +311,72 @@ defmodule Lightning.AttemptService do
       )
 
     Repo.all(first_attempt_runs_query)
+  end
+
+  @doc """
+  Returns a list of AttemptRun structs that should be rerun for the given list
+  of workorder ids that are associated to the given Job
+  """
+  @spec list_for_rerun_from_job(
+          workorders :: [Ecto.UUID.t(), ...],
+          job_id :: Ecto.UUID.t()
+        ) :: [
+          AttemptRun.t(),
+          ...
+        ]
+  def list_for_rerun_from_job(order_ids, job_id)
+      when is_list(order_ids) do
+    last_attempts_query =
+      from(att in Lightning.Attempt,
+        where: att.work_order_id in ^order_ids,
+        group_by: att.work_order_id,
+        select: %{
+          work_order_id: att.work_order_id,
+          last_inserted_at: max(att.inserted_at)
+        }
+      )
+
+    last_attempt_runs_query =
+      from(ar in AttemptRun,
+        join: att in assoc(ar, :attempt),
+        join: last in subquery(last_attempts_query),
+        on:
+          last.work_order_id == att.work_order_id and
+            att.inserted_at == last.last_inserted_at,
+        join: r in assoc(ar, :run),
+        on: r.job_id == ^job_id,
+        group_by: ar.attempt_id,
+        select: %{
+          attempt_id: ar.attempt_id,
+          last_inserted_at: max(ar.inserted_at)
+        }
+      )
+
+    attempt_runs_query =
+      from(ar in AttemptRun,
+        join: last in subquery(last_attempt_runs_query),
+        on:
+          ar.attempt_id == last.attempt_id and
+            ar.inserted_at == last.last_inserted_at,
+        order_by: ar.inserted_at,
+        preload: [
+          :attempt,
+          run:
+            ^from(r in Run,
+              select: [
+                :id,
+                :job_id,
+                :started_at,
+                :finished_at,
+                :exit_code,
+                :input_dataclip_id,
+                :output_dataclip_id
+              ]
+            )
+        ]
+      )
+
+    Repo.all(attempt_runs_query)
   end
 
   @doc """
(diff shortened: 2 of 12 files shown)


taylordowns2000:
hey @midigofrank , it's not quite possible to do an ordered list (as there are multiple _branches_ in workflows. please go alphabetical for now).

taylordowns2000 (changes requested):
Looking good. A few things:

1. Possible regression on #659  - When I select three workorders from 2 different workflows, the modal tells me that i've only selected two of workorders. I'm pretty sure this bug is introduced by the use of `selected_workflow_count(...)` which is now being used for to set the value of `selected_count`. You still need that `@selected_count` attribute to be the count of how many work orders have been selected. (Don't change it!) To deliver this new feature, you've also got to introduce something along the lines of `@distinct_workflows_in_selection` which will count the number of different workflows.

image

2. You've gone off-spec with the "Oops, chosen step hasn't been run in the latest attempt of any of the selected workorders" concept and I really like the concept behind it, but if you're adding to the UX we need to make it really work. there are three big issues with how it's implemented now, which would not be problematic if the modal succesfull sent the workorder to the backend, closed, and notified the user that `0 workorders have succesfully been enqueued.`. The problems are: (a) the modal stays open after that message is displayed but you don't let me pick another step and try again. instead the user must dismiss the notification, close the modal, and then re-open the modal to get the button to work again. (b) after closing the modal, the screen focus is lost and you must click TWICE to get anything to work again on that page. did a process die? (c) dismissing the notification also dismisses the modal. 

i like where you're heading with this, but please address those new UX problems!

3. the re-run from doesn't appear to work. in the image below, i've selected a workorder that failed on the "transform data to FHIR standard" step and opted to re-run it from that step. no new attempt is made and i get the warning message from point 2.
image

in this next image, i'd expect 1 of the 3 selected workorders to result in a new attempt since the final attempt of that third workorder has the step i've chosen.

image

ping me on slack if you want to chat through any of this.

taylordowns2000:
Ok. I’ve sent you a note over there. Meet at 930 EAT?

amberrignell (changes requested):
Hey @midigofrank - looking good but there's a small of bug: 

When I select rerun from... for one workflow then another, sometimes the job options do not get updated. To reproduce, create a workflow 1 and a workflow 2 with several runs each, then click "rerun from" for one after the other. You will see that sometimes the job options show to rerun workflow 1 are the jobs from workflow 2.
(I think this is related to liveview not updating)

Please also fix the second modal following [Taylor's fix](https://github.com/OpenFn/Lightning/commit/9b74275e29319a7f75aa9dd640fa93fbf8362093) for the first

amberrignell:
Oh and sometimes it seems I need to click the "rerun from.." button twice for it to work? Do you have a fix for that?

taylordowns2000:
hey @midigofrank , i found and fixed the issue that @amberrignell identified, but I still want your eyes and opinions here: https://github.com/OpenFn/Lightning/pull/906/commits/e67773d288233d04c87a65efd2de46b91e8c64b6

I think the issue is a double-whammy:

1. the root cause is that this `selected_work_orders` assign is not a list of work orders. It's way more complicated than that. It's a Map, with 0..N keys. Each key has a list value with [0..N] items. The items inside those lists are workorders. Does it need to be this complicated? Could we not have one assign for `selected_workorders` and another for `selected_workflows`? Is there something driving this complexity?
2. Assuming that yes, we need that list of workorders to be grouped into a map and keyed by workflow_id, then the issue is in how you use that `Map.update`. I don't know much about Map.update, so I've proposed a fairly hacky solution... but it's not enough to remove the value from the list of workorders under each key. If there are no workorders left under that key, you also need to remove the _key itself_. Otherwise if you select a workorder from WF1, then unselect it and select a workorder from WF2, your `workflow_id={@selected_work_orders |> Map.keys() |> hd()}` code (which also seems a little icky to me) won't work properly. It will randomly the `hd()` and show the user a list of jobs for a workflow that has no selected items.

Regardless of whether or not you want to change the naming/design to reduce the complexity, would you please add a test that locks in this new behavior and shows that the old "select-then-unselect-workorders-from-different-workflows" bug is gone?

taylordowns2000 (changes requested):
hey @midigofrank  looks great!

@amberrignell pointed out that the order of these retries getting sent to the processor is reversed. i don’ know if this was on the spec or not (apologies if not) but i’ve pushed an extension to your tests which now fails: https://github.com/OpenFn/Lightning/pull/906/commits/02348b83175a8af88c7a407f92b8a40b69d52fd1

i’ve also included in the commit a sample fix, though I don’t think it’s working the way we want it to.

can you try to ensure that the order we send these retries is: retry the oldest associated work_order FIRST and the newest associated work_order LAST?

feel free to overwrite all of my code in that commit to guarantee that the order we send these for reprocessing is `[asc: work_order.inserted_at]`

PR #932, opened 2023-07-05
fix: change timestamp fields to contain usec

midigofrank:
## Notes for the reviewer

The database tables fields were defined without  `usec`. By default, the `timestamps()` function in `Ecto.Migration` generates fields of type `:naive_datetime`. 

Below is a link to the docs: https://hexdocs.pm/ecto_sql/Ecto.Migration.html#timestamps/1


taylordowns2000 (approved):
Thanks Frank. @stuartc , if you're happy with this DB change it will take care of the issue we detected with work order ordering.

PR #936, opened 2023-07-11
Workflow Diagram: Fixes and refactors

josephjclark:
This PR introduces a bunch of fixes, refactors and improvements to the workflow diagram. The cosmetic stuff will be based on top of this.

Coming back into this after a while was pretty  difficult. Complexity and various changes since I last looked haven't helped (I don't know why the store is down in the component now, but I don't intend to move it back)

* [x] Remove Context - we just don't need it (yet)
* [x] Simplify the `workflow`, `_workflow` and `model` stuff
* [x] Move some code out of Workflow Diagram (it's massive, there's way too much stuff in it)
* [x] Fix (placeholder) node removal  - this looked to be totally broken
* [x] Don't run layout on every single update
* [x] Fix edge selection
* [x] Don't blow up the chart when adding a placeholder
* [x] Accept selection on mount

I have mostly fixed the edge selection thing. There seems to be a `react-flow` bug if you: select a node, select a link, then click the background. This will cause the app to deselect the link, but note that react-flow will still style the link as selected, and will continue to do so as you click other links.

I think this is minor enough to leave open as an issue, maybe I can take time to make a nice repro and report it on react-flow. I don't think I want to spend more time on it here.

### DIFF
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -23,7 +23,7 @@ import toWorkflow from './util/to-workflow';
 import { useStore } from 'zustand';
 import { shallow } from 'zustand/shallow';
 import { WorkflowContext } from '../workflow-editor/component';
-import { WorkflowProps } from '../workflow-editor/store';
+
 import { FIT_DURATION, FIT_PADDING } from './constants';
 import type { Flow, Positions } from './types';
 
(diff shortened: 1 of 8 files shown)


assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -12,135 +11,88 @@ import ReactFlow, {
...
     useEffect(() => {
       const { positions, selectedId } = chartCache.current;
       const newModel = fromWorkflow(workflow, positions, selectedId);
 
-      console.debug('UPDATING WORKFLOW', newModel, selectedId);
+      //console.log('UPDATING WORKFLOW', newModel, selectedId);
       if (flow && newModel.nodes.length) {
-        layout(newModel, setModel, flow, 200).then(positions => {
-          // trigger selection on new nodes once they've been passed back through to us
-          if (chartCache.current.deferSelection) {
```
josephjclark, line 132: Very happy to see the back of all this!!

taylordowns2000 (changes requested):
@josephjclark , this seems almost perfect, but i'm still noticing that when i:
1. refresh my page
2. select an edge
3. change the edge condition
4. then hit `ctrl-s`
the save doesn't work. if i touch other parts of the diagram (trigger or job) it seems to work. are you getting that same issue?

josephjclark:
> @josephjclark , this seems almost perfect, but i'm still noticing that when i:
> 
> 1. refresh my page
> 2. select an edge
> 3. change the edge condition
> 4. then hit `ctrl-s`
>    the save doesn't work. if i touch other parts of the diagram (trigger or job) it seems to work. are you getting that same issue?

I have not noticed this - I'll take a look!

I've also just remembered  that I need to ensure that selection is remembered properly on load and refresh - I'll get that sorted in the morning.

josephjclark:
@taylordowns2000 I am noticing a bunch of issues around save and I don't think they're directly related to changes on this branch.

Most edits seem to be OK, occasionally I don't get the flash message. I think that if the editor widget is closed/hidden, the save fails.

On `nodes-and-edges-release` If the editor is closed, the save fails every time. I can't really check that on edges on that branch because edge selection is so flaky!

josephjclark:
I would like @stuartc to review this but as he's going to be dark for a few days and the changes are important, I'm gonna suggest we merge as soon as it passes QA.

stuartc:
@josephjclark if you think it's ready, I would be happy to merge it... I'm doing 'delete a node' right now and the layout/render logic may or may not need tweaking so best I be on the latest and greatest.

Issue #910, opened 2023-06-28
Trigger webhook URL and cronExpression no longer passed to Trigger.txs


taylordowns2000:
The trigger object passed to `getTriggerLabels` no longer contains `webhookUrl` and `cronExpression` so it always looks like the trigger on the canvas is a new and unsaved trigger. (no human readable label.)

![image.png](https://images.zenhubusercontent.com/566ac5e0638e718c995f5356/3c65545a-63a5-4662-858a-dc1bdb45cca0)

![image.png](https://images.zenhubusercontent.com/566ac5e0638e718c995f5356/f5a6459f-11b0-41d5-a541-e61be75a9c27)

josephjclark:
This seems to be working properly in #942

PR #942, opened 2023-07-14
Workflow Diagram: Update styles

### LINKED ISSUE #898: Update workflow canvas
Update the workflow canvas to match the new designs. A few things:

- [ ] Square lines, not curved
- [ ] Boxes that take icons based on their adaptors. If no icon is loaded for that adaptor it should instead display _________
- [ ] In the middle of each edge we'd like a clickable edge label that can either be an icon or text.
- [ ] 

cc @stuartc , i think there were some implementation notes you wanted to add here.

josephjclark:
This PR updates the canvas style of the workflow diagam, as outlined in #898

image

The way nodes are styled, with arrows pointing to the icon and space for long labels, the whole chart is technically offset in a weird way. Connectors go to the left edge of the node, rather than to the center. In practice this seems to be working out nicely. I think the layout animation is a bit jerkier (a different easing function may help).

I've hard-coded a couple of adaptor icons just to get a feel for where they'll go and how they'll look. Obviously we need to work out the actual strategy for loading icons.

I do expect the look, feel and behaviour of the chart to involve  - this is just a big shove in the right direction.

TODO:

* [x] Use blue lines to indicate selection
* [ ] Use nicer icons for edge labels
* [ ] Use different fills for selected node / edge label types
* [x] Various code tidyups
* [ ] Work out the strategy for icon loading (load direct from the adaptor package using jsdelivr as a CDN?)
* [x] Try and make the layout animation a bit smoother
* [x] Don't blow up if trigger is the only node
* [x] Allow triggers to add new nodes
* [x] #910

There's a bunch of stuff I have not done yet, as the functionality is not yet supported:
* Disabling edges
* Showing status indicators
* Colouring nodes by last-run status

## Related issue

Closes #898 

## Review checklist


josephjclark:
This is almost ready: the biggest thing now is the adaptor icons, which I think we should solve in a different PR

stuartc (approved):
@josephjclark this is great!

One thing, I need to make an issue for it - that is it's possible for the URL to change from say clicking close on a panel, we need to propagate the new `?s=abc` to the component.

I have however tweaked the initialSelection to use search params instead of the hash. :+1:

stuartc:
@josephjclark issue outlining the desired behaviour for tracking selection from the URL from pushState here: https://github.com/OpenFn/Lightning/issues/944

PR #947, opened 2023-07-21
Support Backup Codes for Multi-Factor Authentication

### LINKED ISSUE #937: Support Backup Codes for Multi-Factor Authentication
# User story
As a user, I want to have the option to use backup codes for multi-factor authentication (MFA) so that I can securely access my account even when my primary authentication factors are unavailable.

# User acceptance criteria
- Users should be able to generate and access backup codes during the initial setup process.
- Users should be able to enter backup codes as a valid authentication factor during the login process.
- Backup codes should be unique for each user and should not be reused.
- Backup codes should be one-time use and become invalid once used for authentication.
- Backup codes should be encrypted and stored securely to prevent unauthorized access.

midigofrank:
### Implementation plan
- [x] create `users_backup_codes` table
- [x] Generate backup codes when users add TOTP
- [x] Maybe generate new backup codes when users edit TOTP (add new device)?
- [x] Add page to display backup codes
- [x] Require reauthentication to view backup codes
- [x] Implement option to use backup codes on the 2FA page
- [x] Fix UI
- [x] Allow regenerating of backup codes
- [ ] Maybe encrypt the backup codes?
- [x] Add tests

## Related issue

Fixes #937 

## Review checklist


taylordowns2000 (approved):
@stuartc , the UX is great. i've made some small UI changes, cleaned up the table names, and implemented encryption as that wasn't done yet.

everything works as expected on the front end, but i'm hitting a problem with `backup_codes: build_list(10, :backup_code)` since moving from `:string` to `:binary` in the DB. do you have any idea why this might be?

@midigofrank , would love your input on this as well. and please take a look at my last few commits to see the changes i've made. want to make sure we're on the same page!

lib/lightning_web/router.ex
```diff
@@ -77,15 +77,35 @@ defmodule LightningWeb.Router do
...
 
     get "/profile/confirm_email/:token",
         UserConfirmationController,
         :confirm_email
 
     get "/download/yaml", DownloadsController, :download_project_yaml
 
+    live_session :auth, on_mount: LightningWeb.InitAssigns do
+      live "/auth/confirm_access", ReAuthenticateLive.New, :new
+    end
```
stuartc, line 91: Why is this in it's own `live_session`? Is it because we need to ensure we're in a browser redirect to apply a new cookie?

PR #967, opened 2023-07-25
Workflow Diagram: Fix deselect

### LINKED ISSUE #944: Propagate url pushState/changes to Workflow Diagram selection
Currently if you select a node in the diagram, it correctly selects the node and updates the URL. However if the panel is closed via the cross (`pushState`) the url is updated and the panel closes however the diagram shows the node as selected.

In order to reselect the node, it has to be unselected by clicking on another node or the canvas before triggering a URL change again.

josephjclark:
This PR fixes selection in the workflow diagram where selection change is not driven by the diagram itself.

Selection is now driven entirely by the URL. When Pheonix changes the URL to update the selection, the Diagram component listens to the change and pushes the new selection into react  flow.

This allows the app generally to control the selection in the diagram and makes the URL the owner of selection, which makes sense to me.

Right now this is only used in practice by the X icon at the top of the forms.

Here's the rough flow when the user clicks on the diagram:
* React-flow will internally update selection state (boo, I can't stop this)
* Our component gets the click event and calls the `onSelectionChange` callback passed in as a prop (ie it passes the selection outside the component)
* The mount point in `workflow-editor/index.ts` gets the selection change event and updates the URL appropriately (if it wants to)
* LiveView publishes a `phx:navigate` event, which the workflow-editor receives, extracts the selection from, and re-renders the component
* The diagram component receives the new selection as a prop, updates some internal state, and re-renders the nodes for stying.

This additional re-render at the end is actually important for styling edges when clicking on an edge label (if you comment out the navigate event, clicking an edge label will select the label but not the whole edge).

Adding a new placeholder node seems to maintain selection and update the form correctly.

## LiveView Version Bump

Unfortunately to get this to work I've have had to version bump LiveView.

To get the notification that the URL has changed, I've had to use the `phx:nagivate` event introduced in liveView 0.19.5.

Nothing else seemed to work worked: not `addEventListener('popstate')` or `page-loading-start`. I figure this is something to do with how the URL is updated.

It's not ideal but it seems OK. I don't like being on the cutting edge for something as seemingly basic as this. If there's another way to track the URL change I'm all for it!

## Related issue

Fixes #944

## Review checklist


assets/js/workflow-editor/index.ts
```diff
@@ -89,9 +93,6 @@ export default {
     // between the current state and the server state and send those diffs
     // to the server.
   },
-  unselectNode() {
-    this.liveSocket.pushHistoryPatch(this.el.dataset.baseUrl!, 'push', this.el);
```
josephjclark, line 93: This, so far as  I can tell, is unused (it was copied from the old diagram)

stuartc (approved):

PR #978, opened 2023-07-26
require edge condition, set default to on_success, fix #977

### LINKED ISSUE #977: When a new job is created, the default edge will never be run
Note how the edges between a newly created job (bottom-most) or between a downstream job and its parent when the edge hasn't been specifically modified and saved (right-side) is an `∞` symbol. This is invalid and will not let the downstream job run

Until we support the new edge conditions, this needs to go back to being an `on_success` edge.

taylordowns2000:
This fixes #977

## Review checklist

### DIFF
--- a/test/lightning/workflows/edge_test.exs
+++ b/test/lightning/workflows/edge_test.exs
@@ -15,9 +15,7 @@ defmodule Lightning.Workflows.EdgeTest do
     end
 
     test "edges must have a condition" do
-      changeset =
-        Edge.changeset(%Edge{}, %{workflow_id: Ecto.UUID.generate()})
-        |> IO.inspect()
+      changeset = Edge.changeset(%Edge{}, %{workflow_id: Ecto.UUID.generate()})
 
       refute changeset.valid?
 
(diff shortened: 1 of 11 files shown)


test/lightning/workflows/edge_test.exs
```diff
@@ -5,10 +5,27 @@ defmodule Lightning.Workflows.EdgeTest do
...
+          condition: :on_job_success
+        })
+
       assert changeset.valid?
     end
 
+    test "edges must have a condition" do
+      changeset =
+        Edge.changeset(%Edge{}, %{workflow_id: Ecto.UUID.generate()})
+        |> IO.inspect()
```
zacck, line 20: Do we need this inspect?

zacck (commented):
Looks good to me apart from the Inspect

test/lightning/workflows/edge_test.exs
taylordowns2000: ack. nope. forgot to remove it, thank you!

stuartc (approved):
Looks good from a cursory reading through

PR #982, opened 2023-07-27
Sentry releases

taylordowns2000:
This adds more information to our Sentry config to help debugging

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -242,7 +242,6 @@ if config_env() == :test do
     pool_size: :erlang.system_info(:schedulers_online) + 4
 end
 
-# If you've booted up with a SENTRY_DSN environment variable, use Sentry!
 release =
   case Application.get_env(:lightning, :image_info) do
     [image_tag: image_tag, branch: _branch, commit: commit] ->
@@ -260,6 +259,7 @@ config :sentry,
   tags: %{
     host: Application.get_env(:lightning, LightningWeb.Endpoint)[:url][:host]
   },
+  # If you've booted up with a SENTRY_DSN environment variable, use Sentry!
   included_environments:
     if(System.get_env("SENTRY_DSN"), do: [config_env()], else: []),
   release: release,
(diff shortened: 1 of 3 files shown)


config/runtime.exs
```diff
@@ -248,3 +241,27 @@ if config_env() == :test do
   config :lightning, Lightning.Repo,
     pool_size: :erlang.system_info(:schedulers_online) + 4
 end
+
+# If you've booted up with a SENTRY_DSN environment variable, use Sentry!
```
stuartc, line 245: Could you move this comment down to the `if` below `included_environments:`.

lib/lightning_web/live/components/common.ex
```diff
@@ -4,36 +4,47 @@ defmodule LightningWeb.Components.Common do
...
-
-        # If running in docker and tag doesn't match :vsn, display :vsn and warning.
-        image != nil and image != vsn ->
-          {commit,
-           "Warning: detected image tag that does not match application version #{vsn}; image tag '#{image}' built from #{commit}",
-           :warn}
-
-        true ->
-          {vsn, "Lightning #{vsn}", :no_docker}
-      end
```
stuartc, line 36: What was the rationale behind moving this into a different function, function components are functions as well and since nothing else is calling `version_tuple` directly besides this component I don't see it as a good reason to abstract it.

I say that because the `Common` module can/is imported in its entirety into LiveViews, so by adding functions we add more reasons for the compiler to have to recompile more modules if anything is touched. My recommendation is either one of these:

1. Move the logic back
2. Make `version_tuple` a `defp`

stuartc (changes requested):
Tiny nit-pick, other than that all good.. lemme know your thoughts/changes and I'll merge it.

lib/lightning_web/live/components/common.ex
taylordowns2000: derp. i moved it out so i could use it elsewhere in the app. then realized i needed something different and forgot to move it back. thanks for catching this!

PR #992, opened 2023-07-27
Show workorders with runs in last 30 days by default

taylordowns2000:
## Why

To help with https://github.com/OpenFn/Lightning/issues/968 , this borrows a trick from platform-app. By default, show the last 30 days of activity. If the user then wants to pick a date further in the past, let them but they'll expect it to be a bit slower.

## Notes to reviewer, open questions for standup?

1. Why is there a `project_id` parameter showing up at the back of the url-bound filters when you navigate to the history page?
2. Why are we asserting redirect every single time in the tests? feels like we should assert the the redirect does what we expect once, and not repeat that test over and over (i've deleted most of them... the assertion is an optional third argument.)
3. Note that we want to return pending stuff even if we have a before date: https://github.com/OpenFn/Lightning/pull/992/files#diff-b400713bb29056407bd66ecb6e7a8663bde8d8adbf8a5a8f68f6bf3ffb8366eaR353
4. Should [this](https://github.com/OpenFn/Lightning/pull/992/files#diff-0d2174d557f95f0bf8356930daa8a2e91380ae2dc36400d9f39ef8f56cc9b8beR103-R107) really be an error? Is that how a redirect works in LiveView?

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to
 
 ### Changed
 
+- Unless otherwise specified, only show workorders with activity in last 14 days
+  [#968](https://github.com/OpenFn/Lightning/issues/968)
+
 ### Fixed
 
 ## [v0.7.0-pre4] - 2023-07-27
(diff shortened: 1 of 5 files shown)


taylordowns2000:
merging, reviewed in paring with @stuartc

PR #993, opened 2023-07-27
feat: admins can require MFA for their project users

### LINKED ISSUE #892: Add "require MFA" option to the project settings page
Related to #364

### LINKED ISSUE #893: When a user attempts to access a project and does NOT have MFA but belongs to a project with an MFA requirement, redirect to user profile and ask them to enable MFA.
Part of #364

midigofrank:
## Notes for the reviewer

- [x] Add `requires_mfa` field to `projects` table
- [x] Add security tab to project settings page
- [x] Use `can_edit_project` permission to determine who can toggle MFA
- [x] add relevant tests

## Related issue

Fixes #892 
Fixes #893 

## Review checklist

### DIFF
--- a/test/lightning_web/live/workflow_live/index_test.exs
+++ b/test/lightning_web/live/workflow_live/index_test.exs
@@ -25,20 +25,15 @@ defmodule LightningWeb.WorkflowLive.IndexTest do
       user = user_with_mfa_fixture()
       conn = log_in_user(conn, user)
 
-      {:ok, project} =
-        Lightning.Projects.create_project(%{
-          name: "project-1",
+      project =
+        insert(:project,
           requires_mfa: true,
-          project_users: [%{user_id: user.id, role: :admin}]
-        })
+          project_users: [%{user: user, role: :admin}]
+        )
 
       create_workflow(%{project: project})
 
-      {:ok, view, _html} =
-        live(
-          conn,
-          ~p"/projects/#{project.id}/w"
-        )
+      {:ok, view, _html} = live(conn, ~p"/projects/#{project}/w")
 
       assert element(view, "#workflows-#{project.id}", "No workflows yet")
 
@@ -47,10 +42,7 @@ defmodule LightningWeb.WorkflowLive.IndexTest do
         {conn, _user} = setup_project_user(conn, project, role)
 
         assert {:error, {:redirect, %{to: "/mfa_required"}}} =
-                 live(
-                   conn,
-                   ~p"/projects/#{project.id}/w"
-                 )
+                 live(conn, ~p"/projects/#{project}/w")
       end)
     end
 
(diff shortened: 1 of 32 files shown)


taylordowns2000 (changes requested):
thanks @midigofrank .
1. please add a test showing that editors and viewers cannot flip the toggle
2. please either continue work on this branch (and we'll wait to merge until the requirement in enforced) or remove the line from the changelog.

re: 2, either one is fine, but we can't say "Project owners can require MFA for their users" in the change-log until that's actually true! my vote would be to continue your work here, let's hold off on the merge until the enforcement is done.

taylordowns2000:
@midigofrank , i'm marking this as draft until it's ready with the enforcement

taylordowns2000:
Admin please.

taylordowns2000 (approved):
@stuartc , i've approved this PR from @midigofrank - it looks great!

please note that I've also _added_ a new bit... the `:user_scope`.

I'm not sure if this is the best way to do it, but I noticed your use of a `:project_scope` hook that could pull in available projects when mounting certain live views. I hope i've done it right. basically, even when outside the project scope, we now need a list of available projects so that the user can access project X, even if project Y requires MFA.

with your approval this is good to merge.

stuartc:
@taylordowns2000 I'm not entirely sure about the change you made with the projects list, if I go to my users profile (which does have access to a project), I see:

image

So it looks like the `projects` assign isn't being passed down to the menu component...

However!

I don't think there should be a project list in the user areas... since almost everything is different... for example the menu items are different as well..

I think we should provide a way to "leave" the user area, like we have with super users - provide visual ques and/or colour changes to show that the user isn't in the root of the application...

test/lightning_web/live/workflow_live/index_test.exs
```diff
@@ -17,6 +18,42 @@ defmodule LightningWeb.WorkflowLive.IndexTest do
              |> element("#workflows-#{project.id}", "No workflows yet")
     end
 
+    test "only users with MFA enabled can access workflows for a project with MFA requirement",
+         %{
+           conn: conn
+         } do
+      user = user_with_mfa_fixture()
```
stuartc, line 25: Could you move the new `user_with_mfa_fixture` into the Factories module and rewrite it to conform with ExMachinas' API.

test/lightning_web/live/workflow_live/index_test.exs
```diff
@@ -17,6 +18,34 @@ defmodule LightningWeb.WorkflowLive.IndexTest do
...
+      conn = log_in_user(conn, user)
+
+      project =
+        insert(:project,
+          requires_mfa: true,
+          project_users: [%{user: user, role: :admin}]
+        )
```
stuartc, line 32: I've changed one or two of the previous calls to `Projects.create_project(...)`, as with the request about the mfa fixture, please when writing new code could you try and not use things for the fixtures module and rather use/extend the `Factories` module.

stuartc (changes requested):
Hey @midigofrank, lovely work, thank you!

I don't think this was really announced anywhere you would have seen, but we're trying to move off of our fixtures modules - and use ExMachina.

So while that would be _lot_ of work just changing everything - we do want to try and get and new tests written without the use of the fixtures. I changed one or two of them on your changes for an example - there should be a lot more around the project.

I'd like to ask you change at least the new tets you have introduced to use the Factories instead.

And please move the MFA user fixture into the factories module.

Other than that, great work!

stuartc:
@midigofrank excellent point, I don't want to get rid of 'scenario' style helpers... Perhaps there is a place to port these to use the factories, but live in a 'Scenarios' module - or something similar; don't want to hijack a generic name like that for just models/data.

So for now, I definitely don't have an issue with using those more complex fixtures. Happy to let the use cases settle and make a call when we have time or it's clearer.

stuartc (approved):

PR #995, opened 2023-07-28
Ensure credentials are reloaded after a new credential was made

### LINKED ISSUE #951: Create new credential fails to populate job edit form
When adding a credential via the credential modal on the job editor, that credential no longer gets passed into the form.

When I create this new credential, I expect that on save* the credential that i've just built will be selected in my current job editor form.

*save does work, btw—the cred is added to the DB but is not passed to the form and is unavailable for selection until the workflow canvas is refreshed

stuartc:
## Notes for the reviewer

In an effort to reduce reloading credentials over and over again, the prior fix broke new credentials from being selected.

## Related issue

Fixes #951 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to
 
 ### Fixed
 
+- Fix issue creating a new credential from the Job editor where the new
+  credential was not being set on the job.
+  [#951](https://github.com/OpenFn/Lightning/issues/951)
+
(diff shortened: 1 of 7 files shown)


taylordowns2000 (approved):
i've pushed a change to this branch to fix the linting stuff. everything else is great.

https://github.com/OpenFn/Lightning/commit/d82634f0c993d718aa1859633cb523d2bcbd7662

PR #997, opened 2023-07-28
Elixir 1.15

taylordowns2000:
This _should_ fix that annoying Elixir-LS issue. (I've tried to trigger it a few times and so far so good.) Also, everything appears to work well on Elixir 1.15 and Erlang 26.

**Update**

1. https://github.com/jeremyjh/dialyxir/pull/512 - when this is merged we'll get cleaner dialyzer output
2. there are a handful of dialyzer errors - should we be worried about these?

## Review checklist


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       elixir_version:
         description: Elixir version
         type: string
-        default: "1.14.1"
+        default: "1.15.4-otp-25"
```
stuartc: Changing this to `1.15.4` since it looks like it's based on OTP26 already: https://github.com/erlef/docker-elixir/blob/5ac15c1d2bbdbf051f198e471c93f604f71d87eb/1.15/Dockerfile#L1

stuartc (approved):
Thanks for this @taylordowns2000, re: the dialyzer warnings - they are fine for now (i.e. not functional difference) but we will have to address them. I gave it a quick go around, and it looks like `%Task{ref: reference()}` is marked as `opaque` now, which means we're not supposed to directly reference it.

The two modules that have this issue will need to be refactored a bit, will have to schedule that in.

PR #1006, opened 2023-08-01
Workflow Diagram: Update placeholders

### LINKED ISSUE #1001: Cannot save new job if editor panel is used to create name
https://www.loom.com/share/410646fee7404be4b35e9c8a7c61c49a

josephjclark:
This PR makes a few changes to workflow placeholders.

Placeholders are now a client-side only affair. The store/server should not see a placeholder until it's committed - at which point its not a placeholder any more.

The WorkflowDiagram component now has its own private state of placeholders, which it adds to the Flow Model but does not push back to the store. To help all this and tidy the code, I've created a `usePlaceholders` hook which I think is a bit nicer.

When a placeholder is committed, we add it to the store, much as before, where the server will see it and write to the database.

I've made an executive decision and removed the starting placeholder from the new Workflow Page. I was starting to push a prop to tell the component to initialise with a placeholder but I realised it didn't feel like very good UX anyway? We may need to spin another ticket out for this.

TODO:

* [ ] Right now the client is defaulting a lot of stuff - should the server do more?
* [ ] The server should show some kind of form when `?placeholder=true` (we can rename the flag easily from `workflow-editor/index.ts:onSelectionChange`)
* [ ]  Are we happy with new workflow behaviour?

## Related issue

Fixes #1001

## Review checklist


taylordowns2000 (approved):
@stuartc , @josephjclark , this is perfect. 

good to go from me. and @stuartc , this PR _already_ removes the panel when you click new. I don't know if you spoke with Joe about this, but it's just what I wanted (and we discussed!)

the behavior is spot on. please merge if you're OK with the code.

assets/js/workflow-editor/index.ts
```diff
@@ -94,19 +124,58 @@ export default {
...
+      for (const i of everything) {
+        if (id === i.id) {
+          return i;
+        }
+      }
+    }
+  },
   onSelectionChange(id?: string) {
-    const currentUrl = new URL(window.location.href);
+    const currentUrl = new URL(this.baseUrl!);
```
stuartc, line 159: @josephjclark note this change, I'm keeping a copy of the url from `handle_params` in the liveview, and passing it to the component... it seems that between liveviews dom diffing and the browsers history patch updates, the component _can_ mount and fire a patch _before_ (and therefore based on an old url) `window.location` has been updated. 

Resulting in a url like `.../w?s=...` so you never actually see the new workflow template.

PR #1013, opened 2023-08-04
workflow-diagram: quick fix to flaky icons

josephjclark:
This is not a good fix, but I don't have the bandwidth today to go deeper.

This is a bandaid to the problem of icons on new workflow nodes occasionally vanished.

The issue occurs because committed placeholder jobs are not given an adaptor until they are fed back into the app from the server. That is, the client creates a job like `{ id, name, body }`, which the server accepts as a patch, converts into a job like `{ id, name, body, adaptor }`, and sends back to the store.

Somehow those updated jobs from the server don't get fed into the app properly - they miss a render, basically, and so don't get assigned an adaptor.

This may be related to saving? It likely means that something is broken in the sync, which would affect concurrency, and may affect future server-side defaulting.

Anyway: as a quick fix I've set the default adaptor in the placeholder node, so that the client always has an adaptor and so always shows an icon.

It's a band-aid. We can accept or wait until I have time to look into a better fix next week.

### DIFF
--- a/assets/js/workflow-diagram/usePlaceholders.ts
+++ b/assets/js/workflow-diagram/usePlaceholders.ts
@@ -28,6 +28,7 @@ export const create = (parentNode: Flow.Node) => {
     },
     data: {
       body: DEFAULT_TEXT,
+      adaptor: '@openfn/language-common@latest',
     },
   });
 


josephjclark:
I will raise an issue for the deeper fix either way.

taylordowns2000 (approved):
~this is great. thank you @josephjclark . unless @stuartc disagrees i'd imagine that the deeper fix is low priority, but it's good to track in a new issue nonetheless~

**edit:** sorry guys, changes requested... realize that this doesn't actually get the fix in for me.

taylordowns2000 (changes requested):
ah sorry, i'm realizing that this doesn't quite solve the problem on my end:

image

can we track this in an issue and prioritize it for after any super urgent security stuff?

josephjclark:
Sorry @taylordowns2000 , I `git stash`'d half the solution away - this should work now

stuartc:
@josephjclark this smells like what I was talking about the other day, a symptom of the ridged decoupling of state… the server only assigns a default adaptor to an incoming patch set _if_ the only key it provides is an id key… basically the same issue as the placeholders but on the server… the proper solution would be to add the default adaptor to the hook via attrs and use that in the patchset

PR #1017, opened 2023-08-08
Fix cutting-off tooltips

### LINKED ISSUE #972: Tooltip for credential select in Job Edit form is cut off
The hover tooltip is cut off by the job edit box

image

elias-ba:
## Notes for the reviewer

In this PR, I've proposed a solution to enhance the way we render tooltips by leveraging the Tippy.js library in conjunction with hooks.

One of the primary motivations behind choosing [Tippy.js](https://github.com/atomiks/tippyjs) is its extensive customizability, allowing us to control nearly every facet of the tooltip programmatically.

Given that the issue I'm addressing is categorized as a 1-point issue, I've intentionally avoided delving deep into its engineering intricacies (Ex: I have written a custom CSS to handle the style of the tooltip - I know that there's a better way to do this using Tippy's props). Instead, I've focused on an efficient solution that meets our needs while ensuring future adaptability due to Tippy's flexibility.

## Related issue

Fixes #972 

## Review checklist

### DIFF
--- a/assets/js/hooks/index.ts
+++ b/assets/js/hooks/index.ts
@@ -1,4 +1,4 @@
-import tippy from 'tippy.js';
+import tippy, { Instance as TippyInstance } from 'tippy.js';
 import { PhoenixHook } from './PhoenixHook';
 
 export const Flash = {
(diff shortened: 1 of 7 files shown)


assets/js/hooks/index.ts
```diff
@@ -16,6 +17,15 @@ export const Flash = {
   },
 } as PhoenixHook<{ timer: ReturnType<typeof setTimeout> }>;
 
+export const Tooltip = {
+  mounted() {
+    let content = this.el.ariaLabel;
+    tippy(this.el, {
+      content: `${content}`,
```
stuartc, line 24: Why is the `content` variable inside a string interpolation? Is this to ensure it's not undefined or null?

If that's the case, can you instead _not_ call `tippy` if the `ariaLabel` is falsey, and call `console.debug` with a message saying that the Tooltip hook was mounted but the element doesn't have a label.

Also I don't see an `destroyed` function, are we sure that `tippy` doesn't have/need an destroy/cleanup functions?

stuartc (changes requested):
Looks good at first glance, left a comment about the hook... I'll fire this up in my browser and check it out after the changes.

assets/js/hooks/index.ts
stuartc: @elias-ba I have made the changes here: https://github.com/OpenFn/Lightning/pull/1017/commits/f1675e007fbc8bf291b1c1e4e7c40f839b4af06b

stuartc (approved):

PR #1023, opened 2023-08-14
Save job before manual run

### LINKED ISSUE #959: Change Run button to be Save + Run
**Problem:**
The Run button in the Job Editor, currently only runs the job in it's last saved state - so it's easy for a user to edit the job body and click Run only to find that their changes **aren't** reflecting.

You _can_ work with thing by clicking Save at the bottom right and/or using the Ctrl/Cmd+S keyboard shortcut; this isn't particularly obvious though.

Saving is currently bound to the entire workflow, so in order to save a job the workflow must be valid.

**Solution:**
Change the _Run_ button to save the job and then run the job.
Update the button wording to reflect the fact it saves as "Save & Run".

**Notes:**
- When the workflow changeset is invalid, the button must be disabled.
- When the manual run chageset is invalid, the button is also disabled. (Existing behaviour)

elias-ba:
## Related issue

Fixes #959 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/job_live/manual_run_component.ex
+++ b/lib/lightning_web/live/job_live/manual_run_component.ex
@@ -119,14 +119,19 @@ defmodule LightningWeb.JobLive.ManualRunComponent do
 
   @impl true
   def update(
-        %{save_status: save_status, manual_workorder: manual_workorder},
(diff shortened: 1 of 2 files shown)


elias-ba:
Closing in favor of #1024.

PR #1024, opened 2023-08-15
Keybindings to save and run a job

### LINKED ISSUE #960: Ctrl/Cmd + Enter keybindings to save and run a job
Add Ctrl/Cmd + Enter as a keybinding to call the event dispatched by the "Save + Run" button on the Job editor component.

The binding _must_ only be registered when the component is mounted.

See: [`assets/js/hooks/index.ts`](https://github.com/OpenFn/Lightning/blob/main/assets/js/hooks/index.ts) and `SubmitViaCtrlS` for an example of a hook that achieve this.

When the command is invoked and saving isn't possible, a flash message should be shown explaining as such.

Depends on #959 since we don't want to let people quickly run a job until saving before running is implemented.

--------

- [x] Copy paste existing Ctrl/Cmd+s function/hook
- [x] Bind it to the ManualRun form as a `phx-hook` attr

elias-ba:
## Related issue

Fixes #960 and #959

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to
   [937](https://github.com/OpenFn/Lightning/issues/937)
 - Log a warning in the console when the Editor/docs component is given latest
   [#958](https://github.com/OpenFn/Lightning/issues/958)
+- Ctrl/Cmd + Enter keybindings to save and run a job
+  [#960](https://github.com/OpenFn/Lightning/issues/960)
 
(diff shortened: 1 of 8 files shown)


taylordowns2000:
@stuartc , we've broken this apart from #1047 so it's ready for your review - i don't know how long the code in this PR will live if/when we shift to a more durable approach to this component, but if you're OK with the implementation we can merge. (and handle the #1047 bug separately)

PR #1031, opened 2023-08-16
Improve workflow name edition

### LINKED ISSUE #961: Improve feedback when a Workflow name is invalid
It's not obvious that the Workflow name is a required field.

This requires some UX/UI input and design to get this right.

elias-ba:
## Related issue

Fixes #961 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/components.ex
+++ b/lib/lightning_web/live/workflow_live/components.ex
@@ -387,7 +387,7 @@ defmodule LightningWeb.WorkflowLive.Components do
           ) %>
         <% else %>
           <%= wokflow_name_from_form(@form) %>
-          <button phx-click="edit" type="button">
+          <button phx-click="edit" type="button" id="edit_workflow_name">
             <Icon.pencil class="h-5 w-5 mx-2 inline-block" />
           </button>
         <% end %>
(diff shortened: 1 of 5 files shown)


lib/lightning_web/live/workflow_live/components.ex
```diff
@@ -372,26 +373,41 @@ defmodule LightningWeb.WorkflowLive.Components do
...
-        <div
-          class="absolute inset-x-0 bottom-0
-                 peer-hover:border-t peer-hover:border-gray-300
-                 peer-focus:border-t-2 peer-focus:border-indigo-600
-                 peer-invalid:border-t-2 peer-invalid:border-rose-400"
-          aria-hidden="true"
-        >
-        </div>
+      <div class="relative flex items-center space-x-4">
+        <%= if @editing do %>
```
stuartc, line 377: We shouldn't be storing the editing state on the server. We can use the `focus` pseudo selector for that.

If you look at the diff, the previous approach was to use the `peer` facility of Tailwind to control the visibility of the border at the bottom.

```
<div
  class="absolute inset-x-0 bottom-0
         peer-hover:border-t peer-hover:border-gray-300
         peer-focus:border-t-2 peer-focus:border-indigo-600
         peer-invalid:border-t-2 peer-invalid:border-rose-400"
   aria-hidden="true"
>...
```

Where the text input had the `peer` class on it... so the elements are next to each other but hovering and focus states on the text input would control the line at the bottom.

I suggest using this approach with the pencil icon, i.e. render it all the time, but have it's visibility controlled by the `peer-focus:...` pseudo selectors.

And then using `phx-click` use the `JS` helpers on LiveView to set the focussed element to the text input.

stuartc (changes requested):

elias-ba:
Closing in favor of #1040

PR #1044, opened 2023-08-23
Warn before closing a form with unsaved changes

### LINKED ISSUE #780: Warn before closing a form with unsaved changes
Starting with the job builder because it's the most sensitive, we need an elegant and generic way to detect when a form has unsaved changes (maybe through the validate hook?) then then display a warning before the user closes.

This is particularly important on the job builder since clicking anywhere _else_ on the canvas will close the form.

elias-ba:
## Related issue

Fixes #780 

## Review checklist


taylordowns2000:
Closing as the approach will be quite different when we have a full canvas + inspector React application

PR #1056, opened 2023-08-25
hack, temporarily fixes #1054 until next gen yaml exporter is created

### LINKED ISSUE #1054: Ensure same order for generated YAML
To prevent large diffs from showing up unintentionally, we should ensure the same order each time we generate YAML and send back a project.json payload via the API.

This [commit looks great](https://github.com/OpenFn/demo-openhie/commit/6a8b8137ae2b192ca9bdc554a6ce1483b33e6833): i've made two small changes... adding a project description and changing a single job name.

This [commit is misleading](https://github.com/OpenFn/demo-openhie/commit/a320a2cfc53fbf304cabf70a01b9e38ee3b76c10): all I've done is change the name of one job, and since the order gets flipped (maybe it's sorting by last updated?) it looks like I've made a big change.

taylordowns2000:
## Notes for the reviewer

This is a temporary fix. The order of the jobs coming back from the DB was only flaking this test (and creating random diffs on the github sync) because we're ordering on `inserted_at`. Normally, this wouldn't present a problem but for the demo projects and the test project, all jobs have exactly the same `inserted_at` timestamp because our precision is `:seconds`.

Rather than dialing up the precision, or sorting on something that might change when a name changes (alpha-sort on job name?) I'd like to politely "make this go away" by setting the inserted_at for our demo projects and in the tests.

Note that this does _not_ fix the random `inserted_at` sort when a user deploys to a Lightning instance from scratch, but in the OpenFn world that's an edge case right now and I imagine that before that happens very much we'll have a new, better way of sorting YAML.

If you @stuartc , are super against this, then I think the backup would be to say to users: "Every time you pull a project from Lightning your project.yaml and your projectState.json will have their jobs alphabetically sorted, regardless of what your YAML looked like before. If you deploy a YAML to Lightning that isn't alphabetically sorted, you'll get back a sorted projectState.json, but if you then pull down changes from Lightning to your repo you'll notice a phantom diff, where we sort your YAML for you."

## Related issue

Fixes #1054

## Review checklist

### DIFF
--- a/lib/lightning/export_utils.ex
+++ b/lib/lightning/export_utils.ex
@@ -175,7 +175,9 @@ defmodule Lightning.ExportUtils do
 
   def build_yaml_tree(workflows, project) do
     workflows_map =
-      Enum.reduce(workflows, %{}, fn workflow, acc ->
+      workflows
+      |> Enum.sort_by(& &1.inserted_at, NaiveDateTime)
+      |> Enum.reduce(%{}, fn workflow, acc ->
         ytree = build_workflow_yaml_tree(workflow)
         Map.put(acc, hyphenate(workflow.name), ytree)
       end)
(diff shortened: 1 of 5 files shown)


taylordowns2000:
@stuartc , thoughts on this?

stuartc (approved):
I think I’m okay with this, or at least don’t have the context to argue strongly either way. If it’s getting in the way more than the cost of debating the correct way, I’m happy with merging this.

PR #1075, opened 2023-09-04
Ensure Unique Job Names Within the Same Workflow

### LINKED ISSUE #1053: Make job names unique per workflow
In response to https://github.com/OpenFn/kit/pull/379#discussion_r1305482115

- [x] We should check that all job names are unique _per workflow_.
- [x] We should check that no special characters are allowed in job names.
- [ ] ~We should consider sending the spaces-to-hyphens slug in the project.json via the provisioning API - maybe send slug in payload?~ (not just yet)

elias-ba:
## Notes for the reviewer

This PR introduces a database constraint and accompanying changeset validations to ensure that job names within a given workflow are unique.

- [x] Added a unique index on the jobs table combining name and workflow_id fields.
- [x] Modified the Job changeset function to incorporate the unique_constraint validation.
- [x] Introduced a unit test to ensure the validity of the new constraint and its proper functioning.

With these changes in place, users will be prevented from inadvertently creating jobs with the same name in a given workflow,

## Related issue

Fixes #1053 

## Review checklist

### DIFF
--- a/test/lightning/invocation_test.exs
+++ b/test/lightning/invocation_test.exs
@@ -6,10 +6,7 @@ defmodule Lightning.InvocationTest do
   alias Lightning.Invocation
   alias Lightning.Invocation.Run
   alias Lightning.Repo
-  import Lightning.InvocationFixtures
-  import Lightning.ProjectsFixtures
-  import Lightning.JobsFixtures
-  import Lightning.WorkflowsFixtures
+  import Lightning.Factories
 
   describe "dataclips" do
     alias Lightning.Invocation.Dataclip
@@ -17,19 +14,21 @@ defmodule Lightning.InvocationTest do
     @invalid_attrs %{body: nil, type: nil}
 
     test "list_dataclips/0 returns all dataclips" do
-      dataclip = dataclip_fixture()
-      assert Invocation.list_dataclips() == [dataclip]
+      dataclip = insert(:dataclip)
+
+      assert Invocation.list_dataclips()
+             |> Enum.map(fn dataclip -> dataclip.id end) == [dataclip.id]
     end
 
     test "list_dataclips/1 returns dataclips for project, desc by inserted_at" do
-      project = project_fixture()
+      project = insert(:project)
 
       old_dataclip =
-        dataclip_fixture(project_id: project.id)
+        insert(:dataclip, project: project)
         |> shift_inserted_at!(days: -2)
 
       new_dataclip =
-        dataclip_fixture(project_id: project.id)
+        insert(:dataclip, project: project)
         |> shift_inserted_at!(days: -1)
 
       assert Invocation.list_dataclips(project)
@@ -38,8 +37,10 @@ defmodule Lightning.InvocationTest do
     end
 
     test "get_dataclip!/1 returns the dataclip with given id" do
-      dataclip = dataclip_fixture()
-      assert Invocation.get_dataclip!(dataclip.id) == dataclip
+      dataclip = insert(:dataclip)
+
+      assert Invocation.get_dataclip!(dataclip.id) |> Repo.preload(:project) ==
+               dataclip
 
       assert_raise Ecto.NoResultsError, fn ->
         Invocation.get_dataclip!(Ecto.UUID.generate())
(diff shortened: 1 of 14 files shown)


stuartc (changes requested):
Cool man, looks fine but I think instead of using the fixtures it would be better to update the job factory to automatically generate a unique name:

```diff
diff --git a/test/support/factories.ex b/test/support/factories.ex
index 4363d3338..b9b2e671d 100644
--- a/test/support/factories.ex
+++ b/test/support/factories.ex
@@ -27,6 +27,7 @@ defmodule Lightning.Factories do
   def job_factory do
     %Lightning.Jobs.Job{
       id: fn -> Ecto.UUID.generate() end,
+      name: sequence(:job_name, &"job-#{&1}"),
       workflow: build(:workflow),
       body: "console.log('hello!');"
     }
```

And then instead of using `job_fixture` use `insert(:job, ...)`.

Remember that the factories (ExMachina) doesn't care about ids 99% of the time, so the `insert(..., workflow_id: workflow_1.id)` would need to be changed to `insert(..., workflow: workflow_1)`.

elias-ba:
> Cool man, looks fine but I think instead of using the fixtures it would be better to update the job factory to automatically generate a unique name:
> 
> ```diff
> diff --git a/test/support/factories.ex b/test/support/factories.ex
> index 4363d3338..b9b2e671d 100644
> --- a/test/support/factories.ex
> +++ b/test/support/factories.ex
> @@ -27,6 +27,7 @@ defmodule Lightning.Factories do
>    def job_factory do
>      %Lightning.Jobs.Job{
>        id: fn -> Ecto.UUID.generate() end,
> +      name: sequence(:job_name, &"job-#{&1}"),
>        workflow: build(:workflow),
>        body: "console.log('hello!');"
>      }
> ```
> 
> And then instead of using `job_fixture` use `insert(:job, ...)`.
> 
> Remember that the factories (ExMachina) doesn't care about ids 99% of the time, so the `insert(..., workflow_id: workflow_1.id)` would need to be changed to `insert(..., workflow: workflow_1)`.

@stuartc this is now done. Can you check again and let me know. Thanks for your feedback

test/lightning/invocation_test.exs
```diff
@@ -226,40 +233,38 @@ defmodule Lightning.InvocationTest do
...
+        insert(:reason, type: :webhook, dataclip: dataclip, trigger: trigger)
 
-      wo_one = work_order_fixture(workflow_id: workflow.id)
-      wo_four = work_order_fixture(workflow_id: workflow.id)
-      wo_two = work_order_fixture(workflow_id: workflow.id)
-      wo_three = work_order_fixture(workflow_id: workflow.id)
+      wo_one = insert(:workorder, reason: reason, workflow: workflow)
+      wo_four = insert(:workorder, reason: reason, workflow: workflow)
+      wo_two = insert(:workorder, reason: reason, workflow: workflow)
+      wo_three = insert(:workorder, reason: reason, workflow: workflow)
```
stuartc, line 267: A little tip here, ExMachina has a `insert_list/3` function, this could be written like this:

```elixir
[wo_one, wo_four, wo_two, wo_three] = 
  insert_list(4, :workorder, reason: reason, workflow: workflow)
```

test/lightning/invocation_test.exs
```diff
@@ -368,16 +373,16 @@ defmodule Lightning.InvocationTest do
...
 
     test "list_work_orders_for_project/1 returns runs ordered by desc finished_at" do
-      fixture = workflow_job_fixture(workflow_name: "chw-help")
-      job_one = fixture.job
-      trigger = fixture.trigger
+      project = insert(:project)
+      workflow = insert(:workflow, name: "chw-help", project: project)
+      job_one = insert(:job, workflow: workflow)
+      trigger = insert(:trigger, workflow: workflow)
+      dataclip = insert(:dataclip)
```
stuartc, line 380: Even though these technically satisfy the test (the functions don't care about edges).

There are helpers for creating workflows:

```elixir
job = build(:job)
trigger = build(:trigger)
workflow =
  build(:workflow, project: project)
  |> with_job(job)
  |> with_trigger(trigger)
  |> with_edge({trigger, job})
  |> insert()
```

The job and trigger models (in this example) are not persisted (the workflow insert persists them), and would need to be reloaded to access their assocs. However they _do_ have their `:id` preset, so you may not need to reload them after the workflow insert if a function only relies on the their id.

test/lightning/invocation_test.exs
```diff
@@ -515,16 +520,15 @@ defmodule Lightning.InvocationTest do
...
+      project = insert(:project)
       now = Timex.now()
 
-      fixture =
-        workflow_job_fixture(workflow_name: "chw-help", project_id: project.id)
-
-      job1 = fixture.job
+      workflow = insert(:workflow, name: "chw-help", project: project)
+      job1 = insert(:job, workflow: workflow)
+      trigger = insert(:trigger, workflow: workflow)
```
stuartc, line 528: As I mentioned above, you can use the `with_job|trigger` helpers to associate the job/trigger with a workflow and have the `insert()` persist the jobs/triggers when the workflow is created.

stuartc:
@elias-ba this is looking amazing, thanks for the hard work changing over the fixtures to factories.. I left a comment or two.

I'll take another look once you've address the case-sensitive issue that @taylordowns2000 brought up.

elias-ba:
> @elias-ba this is looking amazing, thanks for the hard work changing over the fixtures to factories.. I left a comment or two.
> 
> I'll take another look once you've address the case-sensitive issue that @taylordowns2000 brought up.

@stuartc I believe this is now descent enough to be merged. I have scanned all the tests and added your recommendations. I am sure there will still be some tiny things here and there that need more love; but as this was not the primary intention of the issue, I am happy we merge now and I will keep doing this type of improvements on our tests anytime I have the opportunity to do so. Thanks a lot for your guiding.

PR #1082, opened 2023-09-05
Updating erlang image tag in Dockerfile-dev

piische-tph:
The current tag is not working

## Related issue

Fixes #

## Review checklist

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -12,9 +12,9 @@
 #   - https://pkgs.org/ - resource for finding needed packages
 #   - Ex: hexpm/elixir:1.13.2-erlang-24.2.1-debian-bullseye-20210902-slim
 #
-ARG ELIXIR_VERSION=1.15.4
+ARG ELIXIR_VERSION=1.15.5
 ARG OTP_VERSION=26.0.2
-ARG DEBIAN_VERSION=bullseye-20221004-slim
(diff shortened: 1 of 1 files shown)


stuartc:
@piische-tph thanks for this, you're right - the bullseye image isn't available anymore.

This dev dockerfile has fallen out of date, the main `Dockerfile` uses a bookworm image.

Instead of changing the Elixir version as well, could you try the bookworm image used in `Dockerfile`? See: f87434c1d27ef1defa24f1e888dcfcc76c1d35ba

I'd like to know your success using this image (once the image/versions are correct), it's been a while since I used it so if more needs to be done to it - we can plan some further work on it.

piische:
Hey @stuartc,
Thanks, I tried with the  bookworm image and this also worked to get the docker running.

I was able to run all commands and the compose up also worked, I have to wait for our server admin to open a port for the postgres before I have everything settled and can test it, but the setup so far worked perfectly fine.

Best regards,
Patrick

stuartc:
@piische-tph if you could update the `Dockerfile-dev` to use the bookworm image and drop the change to the Elixir version (since the production image uses 1.15.4 it's best to keep the two aligned) I'd be happy to merge that change.

stuartc:
Thanks @piische-tph!

PR #1086, opened 2023-09-08
 LocalRuntime process that manages Runtime Server child process

### LINKED ISSUE #990: LocalRuntime process that manages a long running Runtime Server child process
In order to ease development and default installations of Lightning, the Lightning application should be able to upkeep a long running (external) process.

## Technical Notes

- Create a Genserver that locates and spawns a Runtime server.
- The process passes in a `lightning-url` 
- If the process exits, restart it - backing off exponentially (possibly exiting eventually, taking down the application?)
- All logs emitted from the process should be visible in the Elixir processes logs.
- Think about how this thing handles SIGTERM or SIGQUIT (when you press control-c to stop elixir app from IEX it should shut this thing down)

midigofrank:
## Notes for the reviewer

- [x] Start GenServer with config
- [x] Kill the runtime process when the app exists
- [x] Handle logs
- [x] Configure version and executable path
- [x] Add tests for SIGTERM
- [x] ~~Figure out how to capture logs from the GenServer~~. Settled on connecting to the `Port` and asserting on the received message
- [x] The terminate callback doesn't seem to work when I remove the `IO.inspect/2` 
- [x] Eventually kill the application when the `RuntimeManager` is restarted multiple time consecutively.
- [ ] Sometimes Zombie processes are left behind during tests. This happens very rarely
- [ ] Restart the process exponentially

## Related issue

Fixes #990 

## Review checklist


stuartc (commented):
This is great! I'm pretty happy about where it is right now... I did occasionally spot the broken pipe issue you were talking about and we can see in CI... But yeah it's not common.

I'd like you to try and tweak the tests to not use `:trap_exit`, and instead use monitors - not that I'm sure it will fix the broken pipe. We may be dealing with the reason things like `Rambo` exist.

stuartc:
> Hey @stuartc , by default the main supervisor exits if the child consistently fails. There's an option to ensure that this happens via: `:auto_shutdown` option. We can set it to `:any_significant` but I'm a bit skeptical on changing this to main `Supervisor`. https://hexdocs.pm/elixir/Supervisor.html#module-automatic-shutdown

I'm ok with the RuntimeManager process taking down the whole application if it fails consistently. When working in dev mode, I think it's appropriate for a user to sort out missing dependencies etc instead of wondering why jobs aren't being processed. In production the workers will be running in different containers.

PR #1088, opened 2023-09-08
Fix: Failed runs not displaying in inspector

### LINKED ISSUE #1084: Finished run never appears in inspector when it fails
https://www.loom.com/share/dc7e70148d284ddca465bfb86b1c05f2

elias-ba:
## Notes for the reviewer

**Problem**
When a run completes, we attempt to build its output dataclip from the generated state file in the run result. We have identified an issue where if a run exits with a code of 1, the state file contains an empty string. This leads to a scenario where the Jason.Decode function attempts to parse an empty string, resulting in an error. Consequently, the run is not updated, causing the frontend to be unaware of the new status of the run. As a result, it remains in the 'pending' status.

**Proposed Solution**:
For the time being, this PR introduces a solution where the run is updated with an empty body if the content of the state file is unparsable. This ensures that runs no longer remain stuck in a pending state due to unparsable state files.

**Note**:
While the proposed fix addresses the issue at hand, there might be alternative solutions worth exploring. This fix is open to discussion, and I'm keen on gathering feedback and other potential strategies for addressing this problem.

## Related issue

Fixes #1084 

## Review checklist


taylordowns2000 (approved):
thanks @elias-ba ! from a product perspective this now behaves exactly as i'd like it to. i see 3 failing tests, and don't want to wade into the technical issues here, but assuming the tests simply need updating i'd like to merge this to `main` first thing Monday morning so we can go back to using the job editor panel asap

@stuartc , given that this whole thing will be dropped/re-built once we're operating at an attempt level (and listening for the _first_ run in an attempt?) and/or moving to a better UX/DX for this panel, i'd say don't stress too much about the implementation. if it works, let's ship it. it will be replaced in a few weeks.

lib/lightning/pipeline/runner.ex
```diff
@@ -170,46 +172,92 @@ defmodule Lightning.Pipeline.Runner do
...
+    end
+  end
 
-        res
+  defp decode_file_content({:ok, data}) do
+    case Jason.decode(data) do
+      {:ok, body} -> {:ok, body}
+      error -> error
     end
   end
```
stuartc, line 220: I don't understand the point of these case statements

lib/lightning/pipeline/runner.ex
```diff
@@ -170,46 +172,92 @@ defmodule Lightning.Pipeline.Runner do
...
-          body: scrub_result(body)
-        }
-      })
-    else
-      res = {:error, %Jason.DecodeError{position: pos}} ->
-        Logger.info(
-          "Got JSON decoding error when trying to parse: #{result.final_state_path}:#{pos}"
-        )
-
-        res
```
stuartc, line 202: I don't see the point of changing this. What is the real difference between these changes and the existing implementation?

lib/lightning/pipeline/runner.ex
```diff
@@ -170,46 +172,92 @@ defmodule Lightning.Pipeline.Runner do
...
+      "Got JSON decoding error when trying to parse: #{run.final_state_path}:#{pos}"
+    )
+
+    run
+    |> Repo.preload(:output_dataclip)
+    |> Invocation.update_run(%{
+      output_dataclip: nil
+    })
+
+    {:error, error}
```
stuartc, line 250: Is this the actual fix here? If so then rather put these calls inside the `else` block for the `with` in the original implementation no?

stuartc (changes requested):

elias-ba:
@stuartc I reverted the refactor I made to the function. But here are the reasons why I made it tho:

- The initial implementation combined multiple tasks into one function: reading a file, decoding its content, logging errors, and updating a run with the relevant dataclip. I wanted to refactor these tasks down into separate, smaller functions to make code more readable.

- I also thought that handling errors errors using pattern matching in function heads would be easier for debugging. For example if we need to modify how a specific error is handled or how data is processed, we can go directly to the relevant function.

stuartc (approved):

PR #1111, opened 2023-09-14
Improve User Creation Form

### LINKED ISSUE #731: Creating a new user without a password fails and there is no user feedback
Loom: https://www.loom.com/share/49a94552044b4f4d88946833d0a390a1?sid=0563d7ea-a985-43a1-a65a-ade86106c4c7
Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=173-1640&mode=design

- [x] Disable "Save" button until all fields are filled and valid
- [x] Display validation **only** on the field that has been blurred (not all of them)
- [x] Update styling for the validation text to be red
- [x] Update the validation text to be "<field name> can't be blank"

elias-ba:
## Notes for the reviewer

This PR proposes code for improving how we render error messages in our forms and prepares for using the NewInput module as replacement to our old Form module.

## Related issue

Fixes #731 

## Review checklist


lib/lightning_web/live/components/form.ex
```diff
@@ -275,35 +276,35 @@ defmodule LightningWeb.Components.Form do
...
+  #     |> update(:opts, &assigns_to_attributes/1)
 
-    ~H"""
-    <%= error_tag(@form, @field, @opts) %>
-    """
-  end
+  #   ~H"""
+  #   <%= Phoenix.HTML.Form.error_tag(@form, @field, @opts) %>
+  #   """
+  # end
```
stuartc, line 307: If `error/1` is no longer used in this module (great!) then please remove it.

lib/lightning/accounts/user.ex
```diff
@@ -143,14 +156,22 @@ defmodule Lightning.Accounts.User do
...
+      message: "Oh no! The maximum length for passwords is 72 characters"
+    )
     |> maybe_hash_password(opts)
   end
 
   defp validate_name(changeset) do
     changeset
-    |> validate_required([:first_name, :last_name])
+    |> validate_required(:first_name, message: "First name can't be blank")
+    |> validate_required(:last_name, message: "Last name can't be blank")
```
stuartc, line 174: I'm not crazy about changing all the validations to include the field name in the message. This is best addressed in the view layer/components - where the field name is prefixed to the message.

I say this because the data structure underneath ends up becoming `[{:first_name, [message: "First name ..."]}]` and if we ever _don't_ want the prefix like for API error messages etc, we can't easily take them out.

If we want all our error to show the field name in them, (@NickOpenFn when rendering errors do we want all of the message to include the field name?, I personally think it's redundant because it's right next to the invalid input - but I defer to you); then we should change the `error/1` component to be used with something like this:

```heex
<.error :for={msg <- @errors}><%= @label %> <%= msg %></.error>
```

stuartc (commented):
@elias-ba well done :clap: this was a massive set of changes... It's looking great, excited for the rest of them to match the new stuff!

I have some comments on prefixing the fields name to all the error messages. So not rejecting right now - but I think we can solve it a little differently.

elias-ba:
> @elias-ba well done 👏 this was a massive set of changes... It's looking great, excited for the rest of them to match the new stuff!
> 
> I have some comments on prefixing the fields name to all the error messages. So not rejecting right now - but I think we can solve it a little differently.

Hey @stuartc and @NickOpenFn this is looking now much better. I have introduced usage of gettext to translate Ecto errors to more humanized errors. Let me know if you need anymore change.

PR #1119, opened 2023-09-16
for #271, set up audit_events

### LINKED ISSUE #44: Create generic module which creates audit events
- [https://luizdamim.com/blog/tracking-changes-with-context-using-phoenix-and-ecto/](https://luizdamim.com/blog/tracking-changes-with-context-using-phoenix-and-ecto/)

### LINKED ISSUE #271: ATNA audit trail refactor/expand functionality
- [x] remove cascades
- [x] drop FK constraint on `row`
- [x] drop FK constraint on `actor`
- [x] ensure there's an index on the `actor`
- [x] rename `row` relationship to `item`
- [x] ensure there's an index on the `item`
- [x] add `item_type` column
- [x] change `metadata` to `changes` jsonb
- [x] keep `{ before: ..., after: ...}` in `changes`
- [x] change `credentials_audit` to `audit_events` -> each separate audit MODEL is its own module, with its own structs, handling, etc., but they use the same DB schema: `audit_events`

taylordowns2000:
**First, remember that audit trails are weird.** There is no longer a foreign key on `actor_id` (because actors can get deleted and we sometimes want audit trails to live on) or `item_id` (because it's polymorphic).

This fixes #271 and fixes #44 so we can start wiring up many more audit events, but there is a big open question in my mind:

**How do we want to store `item_type` in the `audit_events` table?** Right now, I'm storing it as the whole string version of a model in Lightning (e.g., "Elixir.Lightning.Credentials.Credential") so that we could (in the future) make it possible for auditors to access the items in question using some automatic `Repo.get` interface. Imagine something weird like this:

```ex
audit.item_type
|> String.to_existing_atom()
|> Repo.get(audit.item_id)
```

We don't do that right now, so we almost might as well simply write "Credential" in the DB for `item_type` but I wanted to consider the benefits and drawbacks of storing the whole official model name.

**Note that there is also a pretty weird `try_to_get_actors` (almost a "maybe preload") without a proper ecto relationship** thing going on [here](https://github.com/OpenFn/Lightning/pull/1119/files#diff-09d9a2ead0f010271c4bf0ba7ddd893d56882ef968f7c7008013cb6713a7ecc3R14) that I'm not proud of, but I still think it's useful to display actor info on the audit trail screen.

In the end, no matter which way we go with the two funky bits above, it's probably better to get this _in_ so that we can add the other audit events as we build out the administrative features and come back to these enhancements later.

image

## Review checklist

### DIFF
--- a/lib/lightning/auditing.ex
+++ b/lib/lightning/auditing.ex
@@ -7,10 +7,18 @@ defmodule Lightning.Auditing do
   alias Lightning.Repo
 
   def list_all(params \\ %{}) do
-    from(a in Lightning.Credentials.Audit,
-      # preload: [:actor_id],
-      order_by: [desc: a.inserted_at]
-    )
-    |> Repo.paginate(params)
+      from(a in Lightning.Credentials.Audit,
+        order_by: [desc: a.inserted_at]
+      )
+      |> Repo.paginate(params)
+      |> try_to_get_actors()
+  end
+
+  defp try_to_get_actors(page) do
+    page
+    |> Map.put(:entries, Enum.map(page.entries, fn entry ->
+      entry
+      |> Map.put(:actor, Repo.get(Lightning.Accounts.User, entry.actor_id))
+    end))
   end
 end
--- a/lib/lightning/auditing.ex
+++ b/lib/lightning/auditing.ex
@@ -8,20 +8,11 @@ defmodule Lightning.Auditing do
 
   def list_all(params \\ %{}) do
     from(a in Lightning.Credentials.Audit,
+      left_join: u in Lightning.Accounts.User,
+      on: [id: a.actor_id],
+      select_merge: %{actor: u},
       order_by: [desc: a.inserted_at]
     )
     |> Repo.paginate(params)
-    |> try_to_get_actors()
-  end
-
-  defp try_to_get_actors(page) do
-    page
-    |> Map.put(
-      :entries,
-      Enum.map(page.entries, fn entry ->
-        entry
-        |> Map.put(:actor, Repo.get(Lightning.Accounts.User, entry.actor_id))
-      end)
-    )
   end
 end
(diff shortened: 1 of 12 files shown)


lib/lightning/auditing.ex
```diff
@@ -8,9 +8,20 @@ defmodule Lightning.Auditing do
...
+
+  defp try_to_get_actors(page) do
+    page
+    |> Map.put(
+      :entries,
+      Enum.map(page.entries, fn entry ->
+        entry
+        |> Map.put(:actor, Repo.get(Lightning.Accounts.User, entry.actor_id))
+      end)
+    )
```
stuartc, line 25: I don't think we need to remove the preload, we can do a `left_join` to the users table and then preload on that join instead.

Also we could just not have a constraint/fk on the `actor_id`, making it our responsibility to reference users properly.

test/lightning/auditing_test.exs
```diff
@@ -10,7 +10,7 @@ defmodule Lightning.AuditingTest do
 
       %{entries: [entry]} = Auditing.list_all()
 
-      assert entry.row_id == credential_id
+      assert entry.item_id == credential_id
```
stuartc, line 13: :ok_hand:

stuartc (commented):
Re: using the models module name. I kind of agree with your sentiment here.

While it is nice to be able to query/create these things using model names. There is a foot gun waiting in assuming these models won't change, audit events are _not_ the original models - and blindly applying a 6 month old map over an Ecto model will never end up working in the long run.

We need to think about how we want to use these audit events down the line, if it's just for reading then we don't have to think very hard - if we ever want to restore from them then we have to do some planning around how we want to approach it - these persisted maps will get stale and we need to have a strategy for applying (or knowing we can't apply) these stale maps over the current model.

taylordowns2000:
todo:
1. change to "credentials" - use name, not module name
2. address preload

PR #1147, opened 2023-09-22
Crash and wrong error message when setting up version control

### LINKED ISSUE #1112: Crash and wrong error message when setting up version control
https://www.loom.com/share/7add73420ff448ebb4a7d65af3869d66?sid=199cf13e-be48-4f88-b647-0c457e14ea21

1. When trying to set up Github version control from an "admin" project_user the Github page doesn't redirect them back to the right page.
2. When going back to the workflow page, no repo or branch is available.
3. The app displays a "bad github cert" message, but the cert is not bad because we know it works properly for the superuser.
4. No error message is sent to Sentry.
5. No error is visible in the logs.
6. It appears that a project_repo_connection has been built

Github version control works perfectly for the `superuser` on app.openfn.org

here is the link the loading stops on: `http://app.openfn.org/setup_vcs?installation_id=41428355&setup_action=update`

midigofrank:
## Notes for the reviewer

- [x] Introduces `GithubError` exception for easier tracking
- [x] Adds `unique_constraint` to `project_repo_connections` for each `project`
- [x] Raise when we don't find a `pending` `connection` while trying to update `installation id`
- [x] add entire body of error to the sentry capture so that we are never hiding the errors from future developers!
- [x] Send a text response back to the user whenever they update the app settings on github, e.g. add more `repo` and they get redirected to the app. `Github installation updated successfully; you may close this page or navigate to any OpenFn project which uses this installation: <INSTALLATION_ID>` @NickOpenFn we may need to improve this later
- [x] check to see that we don’t capture an unknown error anywhere. (it’s fine if we do, though Taylor points out that it may be risky! do you REALLY know for sure that we will never get another type of error?)
- [x] Add tests

## Current flow for Github Installation
github-installation

## Related issue

Fixes #1112 

## Review checklist

### DIFF
--- a/lib/lightning/version_control/github_client.ex
+++ b/lib/lightning/version_control/github_client.ex
@@ -45,45 +45,49 @@ defmodule Lightning.VersionControl.GithubClient do
       Application.get_env(:lightning, :github_app)
       |> Map.new()
 
-    with {:ok, auth_token, _} <- GithubToken.build(cert, app_id),
-         client <-
+    with {:ok, auth_token, _} <- GithubToken.build(cert, app_id) do
+      client =
+        Tesla.client([
+          {Tesla.Middleware.Headers,
+           [
+             {"Authorization", "Bearer #{auth_token}"}
+           ]}
+        ])
+
+      case post(
+             client,
+             "/app/installations/#{installation_id}/access_tokens",
+             ""
+           ) do
+        {:ok, %{status: 201} = installation_token_resp} ->
+          installation_token = installation_token_resp.body["token"]
+
+          {:ok,
            Tesla.client([
              {Tesla.Middleware.Headers,
               [
-                {"Authorization", "Bearer #{auth_token}"}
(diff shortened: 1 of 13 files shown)


lib/lightning/version_control/github_client.ex
```diff
@@ -126,11 +67,22 @@ defmodule Lightning.VersionControl.GithubClient do
...
+            body
+          )
+
+        Sentry.capture_exception(error)
+        Logger.error(inspect(resp, label: "Unexpected Github Response: "))
+        {:error, error}
+
+      resp ->
+        Logger.error(inspect(resp, label: "Unexpected Github Response: "))
+        error = GithubError.invalid_pem("Github Cert is misconfigured")
```
taylordowns2000, line 83: this is not necessarily an invalid pem. you should instead send "Error" with no misleading description.

taylordowns2000 (approved):

PR #1162, opened 2023-09-28
Merge jobs context into workflows context

elias-ba:
## Related issue

Fixes #

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -86,7 +86,7 @@ config :lightning,
       Application.get_env(:lightning, :schemas_path) || "./priv"
 
 base_oban_cron = [
-  {"* * * * *", Lightning.Jobs.Scheduler},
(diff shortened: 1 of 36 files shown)


stuartc (changes requested):
Ah @elias-ba we must have misunderstood each other - I was under the impression that the Job modules (as in `Jobs.x`) would be moved into `Workflows`, so `Jobs.Job` would become `Workflows.Job`.

elias-ba:
> Ah @elias-ba we must have misunderstood each other - I was under the impression that the Job modules (as in `Jobs.x`) would be moved into `Workflows`, so `Jobs.Job` would become `Workflows.Job`.

Thanks @stuartc for clarifying this. I have made the changes you requested.

stuartc (approved):

PR #1168, opened 2023-10-03
Extend LiveDashboard

rorymckinley:
* Enable Db Information
* Enable Os Information

## Notes for the reviewer

This PR makes it possible for the LiveDashboard to display OS- and DB-related data.

I am not sure how useful the OS-level data would be in a containerised environment but I think it would be useful
in situations where the infrastructure is either a metal box or a VM.

image

The DB information will, I think, be more generally useful when it becomes time to answer the question: 'Why does the database hate me?'.

image

## Related issue

Fixes #

## Review checklist


rorymckinley:
It appears that the code coverage failure is due to enabling `os_mon`. Will dig a bit and see if it can be resolved.

image

rorymckinley:
Turns out my previous diagnosis was incorrect - original failure was caused by a flickering test. When I fixed that, I exposed another flickering test and, in fixing that, another flickering test surfaced. So, I am making this PR draft until I can fix the last test.

test/lightning_web/live/project_live_test.exs
```diff
@@ -508,7 +508,14 @@ defmodule LightningWeb.ProjectLiveTest do
           ~p"/projects/#{project.id}/settings#vcs"
         )
 
-      Process.sleep(15)
+      Enum.take_while(0..30, fn _index ->
```
rorymckinley, line 511: I would have preferred something that could have looked at some underlying state to determine when to break out of the loop but (within the constraints of my limited Elixir) - it appears that an unsupervised Task runs and then passes the state to the Live view - so it does not look like there is something other than the rendered view that the test can reach into.

rorymckinley:
The case for the last flickering test () is less [clear cut](https://github.com/rorymckinley/Lightning/blob/main/test/lightning_web/live/work_order_live_test.exs#L422). I can't find an obvious element that could result in the test flickering - there does not appear to be an obvious asynchronous component.

So, going to leave this for now.

taylordowns2000 (approved):

PR #1175, opened 2023-10-04
Initial implementation of workorder metrics

### LINKED ISSUE #1171: Move key telemetry spans into app
Right now, our load_testing script (see BENCHMARKING.md) is adding spans around key functions in an `.exs` file. This ticket is to experiment with adding those spans into the app itself so we can monitor them first in Live Dashboard and then later in grafana/something like it.

The proposed spans are:
1. rolling average, time to store new HTTP request bodies in the DB when sent to a webhook trigger URL
2. rolling average, time to start executing a new workorder for a webhook triggered attempt
3. rolling average, how long is it taking for an attempt to be claimed off the queue? (this helps us understand if worker apps are being auto-scaled efficiently, or if they are being overworked)
4. rolling average, how long is it taking for users to access the history page? (historically, this is the heaviest read from the DB)

rorymckinley:
## Notes for the reviewer

I have not added any tests for this - I am unsure of the ROI. In addition, while there is a [package](https://hex.pm/packages/telemetry_test) that can enable testing of Telemetry, it is very new and I am uncertain if it should be used. However, the introduction of the telemetry code results in me being dinged on code coverage.

## Related issue

Fixes #1171

## Review checklist

### DIFF
--- a/lib/lightning_web/controllers/webhooks_controller.ex
+++ b/lib/lightning_web/controllers/webhooks_controller.ex
@@ -6,33 +6,44 @@ defmodule LightningWeb.WebhooksController do
   # this gets hit when someone asks to run a workflow by API
   @spec create(Plug.Conn.t(), %{path: binary()}) :: Plug.Conn.t()
   def create(conn, %{"path" => path}) do
-    path
-    |> Enum.join("/")
-    |> Workflows.get_edge_by_webhook()
-    |> case do
-      nil ->
-        put_status(conn, :not_found)
-        |> json(%{})
+    source_trigger_id = path |> List.last()
 
-      %Workflows.Edge{target_job: %Jobs.Job{enabled: false}} ->
-        put_status(conn, :forbidden)
-        |> json(%{
-          message:
-            "Unable to process request, trigger is disabled. Enable it on OpenFn to allow requests to this endpoint."
-        })
+    :telemetry.span(
+      [:lightning, :workorder, :webhook],
+      %{source_trigger_id: source_trigger_id},
+      fn ->
+        result =
+          path
+          |> Enum.join("/")
+          |> Workflows.get_edge_by_webhook()
+          |> case do
+            nil ->
+              put_status(conn, :not_found)
+              |> json(%{})
 
-      edge ->
-        {:ok, %{work_order: work_order, attempt_run: attempt_run}} =
-          WorkOrderService.create_webhook_workorder(edge, conn.body_params)
+            %Workflows.Edge{target_job: %Jobs.Job{enabled: false}} ->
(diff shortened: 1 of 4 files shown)


lib/lightning_web/controllers/webhooks_controller.ex
```diff
@@ -6,33 +6,44 @@ defmodule LightningWeb.WebhooksController do
...
-          attempt_id: attempt_run.attempt_id
-        }
-
-        conn
-        |> json(resp)
-    end
+    source_trigger_id = path |> List.last()
+
+    :telemetry.span(
+      [:lightning, :workorder, :webhook],
```
rorymckinley, line 12: It feels like it may be useful to also pull the workflow_id for this request as that may be more useful? The problem is that, if I am reading the code correctly, I can only get that from the edge (if I reuse the existing code). However, that would mean that I would move  a potentially expensive operation outside the span, where it would not be measured.

lib/lightning_web/telemetry.ex
```diff
@@ -63,7 +63,19 @@ defmodule LightningWeb.Telemetry do
...
-      summary("vm.total_run_queue_lengths.io")
+      summary("vm.total_run_queue_lengths.io"),
+
+      # Business metrics
+      summary("lightning.workorder.webhook.stop.duration",
+        tags: [:source_trigger_id],
+        unit: {:native, :millisecond},
+        description: "Time taken to trigger a new workorder"
+      ),
+      distribution("lightning.workorder.webhook.stop.duration",
```
rorymckinley, line 74: @taylordowns2000 `distribution` gives a breakdown of the distribution of the request durations, e.g. (x requests took n seconds, y requests took x + 1 seconds). I thought this might be useful.

rorymckinley:
@taylordowns2000 @stuartc Not ready for showtime yet - just wanted to see if the code is on track for "rolling average, time to store new HTTP request bodies in the DB when sent to a webhook trigger URL".

Also, questions/concerns:

1. I see you use a particular branch naming scheme - does it follow the pattern "#{issue_id}-#{clever_branch_name}"?
2. I did a very unscientific test of performance before and after applying the metric (using Postman on my desktop). Performance did not seem to be impacted by the change, but I am wondering if this should be feature-flagged, at least initially? That way, if you do see negative effects out in the wild it is easy to disable it?
3. What is the policy re: adding new hex packages? I wanted to add a test for the telemetry and I see that there is this [package](https://hex.pm/packages/telemetry_test) that is designed for testing telemetry but it is relatively new and does not appear to have a lot of exposure - so at the very least, it may be wise for me to audit the code before I use it.

taylordowns2000 (approved):

PR #1184, opened 2023-10-09
Pre-Release 0.10.0

stuartc:
## Notes for the reviewer

**Major Changes** - adding the new worker architecture

## Related issue

Fixes #

## Review checklist


test/lightning_web/live/work_order_live_test.exs
```diff
@@ -46,35 +46,25 @@ defmodule LightningWeb.RunWorkOrderTest do
...
 
       dataclip = insert(:dataclip)
 
-      reason =
-        insert(:reason,
-          type: :webhook,
-          trigger: trigger,
-          dataclip: dataclip
-        )
-
```
stuartc, line 61: Reasons are now deprecated and 'conceptually' exist on the WorkOrder or Attempt directly.

In the case of WorkOrder, there are now `trigger` and `dataclip` fields.

We must hunt down all mentions of reason, probably best starting with this file.

test/support/factories.ex
```diff
@@ -212,11 +247,63 @@ defmodule Lightning.Factories do
...
+      end)
+    end)
+    |> with_edge({trigger, jobs |> Enum.at(0)})
+    |> with_edge({jobs |> Enum.at(0), jobs |> Enum.at(1)})
+    |> with_edge({jobs |> Enum.at(1), jobs |> Enum.at(2)})
+    |> with_edge({jobs |> Enum.at(2), jobs |> Enum.at(3)})
+    |> with_edge({jobs |> Enum.at(0), jobs |> Enum.at(4)})
+    |> with_edge({jobs |> Enum.at(4), jobs |> Enum.at(5)})
+    |> with_edge({jobs |> Enum.at(5), jobs |> Enum.at(6)})
+  end
```
stuartc, line 305: Here is the replacement for `JobFixtures.workflow_scenario/1`, it's return value is a little different - in that it's the Workflow (and not a map), you're expected to destructure/match jobs/triggers etc from it.

`workflow_scenario/1` has been marked as deprecated, there will be warnings from the compiler. We need to replace all calls to workflow_scenario with the new factory.

Issue #1159, opened 2023-09-27
Webhook Auth: Display associated Workflows with trigger


NickOpenFn:
Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=365-7222&mode=design&t=UBx8DlulgRxWP9rd-4

On the credentials table and delete modal, we have the option that users can see workflows with associated credentials (very handy in terms of management). 

When this modal displays, the workflow names should be hyperlinked and opened in a new tab.

**Routing:** 
- If the modal is opened from the app, the close button will close the modal
- If the modal is opened from another modal, the close button will reopen the previous modal with correct state

NickOpenFn:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Thunderbolt--Lightning-6305f22ada44914905485ab1/issues/OpenFn/Lightning/1159?planning-poker) with Zenhub @elias-ba @midigofrank

Issue #1158, opened 2023-09-27
Webhook Auth: Deletion of a Webhook credential


NickOpenFn:
Deletion of a Webhook credential can be quite destructive, and requires a confirmation step.

Deletion tech items:
- Soft delete possible
- Removal of credential from workflows
- 
- Impacted areas?
- Logs?

NickOpenFn:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Thunderbolt--Lightning-6305f22ada44914905485ab1/issues/OpenFn/Lightning/1158?planning-poker) with Zenhub @elias-ba @midigofrank

Issue #1157, opened 2023-09-27
Webhook Auth: Add icon on the canvas for Triggers that have authentication enabled


NickOpenFn:
If the trigger has an authentication method available, display an icon on the top left of the trigger.
Icon from hero icons

Figma (`Webhook: Trigger icon`): https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=365-7331&mode=design&t=UBx8DlulgRxWP9rd-4

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBczNnIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--be2dcb9e8504b6350cc96dd447c55d6b9c36663f/image.png)

Issue #1153, opened 2023-09-26
Webhook Auth: Creation of webhook auth_method overview table in canvas


NickOpenFn:
- Creation of the modal
- Gathering of the data for the table 
- Ability to select and apply webhook_auth_methods to the Webhook on save for selected items (update of Webhook authentication method)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbEhYIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--06f35e6e4ffa401ea6e2ff716bdc041328d3657a/image.png)

NickOpenFn:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Thunderbolt--Lightning-6305f22ada44914905485ab1/issues/OpenFn/Lightning/1153?planning-poker) with Zenhub @elias-ba @midigofrank

PR #1198, opened 2023-10-12
Webhook Authentication Methods

### LINKED ISSUE #1152: Webhook Auth: Webhook Authentication Database Credential's Table
**Create a table webhook_credentials table in the DB**
- Migration required
- Ecto model that represents the table

Context webhook_credentials.ex
- Include all CRUD operations for the Webhook
- Test changeset and operations

Liveview:
- Edit Webhook panel in canvas to include the Webhook authentication section ( #1148 )

elias-ba:
## Notes for the reviewer

**Only merge this after @jyeshe work on #860 is merged into the `release-0.10.0` branch**

## Related issue

Fixes #1152 #1148 #1150 #1149 #1153 #1157 #1159 #1158 

## Review checklist


lib/lightning/jobs/trigger.ex
```diff
@@ -54,7 +60,8 @@ defmodule Lightning.Jobs.Trigger do
         :custom_path,
         :type,
         :workflow_id,
-        :cron_expression
+        :cron_expression,
+        :has_auth_method
```
stuartc, line 64: Do we need `:has_auth_method` here? Since it's a virtual field, I don't see it being used anywhere other than in select queries..

lib/lightning/jobs/trigger.ex
```diff
@@ -109,4 +116,20 @@ defmodule Lightning.Jobs.Trigger do
...
+      preload: [webhook_auth_methods: wam],
+      select: %{
+        t
+        | has_auth_method:
+            fragment(
+              "CASE WHEN ? IS NULL THEN ? ELSE ? END",
+              wam.id,
+              false,
+              true
+            )
```
stuartc, line 132: I'm not 100% sold on having `has_auth_method` as a virtual field, and also on using a case statement here (I might be wrong here but does it not return duplicates).

Anyway, my main issue with this is, if we're always preloading the auth methods, we can determine if there are any auth methods _afterwards_ with `Enum.any?/1`.

lib/lightning/workflows.ex
```diff
@@ -208,6 +208,67 @@ defmodule Lightning.Workflows do
...
+    from(e in Edge,
+      join: j in Job,
+      on: j.id == e.target_job_id,
+      left_join: t in Trigger,
+      on: e.source_trigger_id == t.id,
+      where: t.id == ^trigger_id,
+      preload: [:source_trigger, :target_job]
+    )
+    |> Repo.one()
+  end
```
stuartc, line 270: If we rebase against the RTM release branch, we may not need this anymore.

stuartc (commented):
@elias-ba @midigofrank this is such great work! :tada: 

I have left a few comments, and I still need to dig a little deeper on the new LV Hooks you added.

Could we try and rebase this onto the release branch, the webhook controller (and the tests for it) hopefully will be the only things that will need to be changed a fair amount.

PR #1203, opened 2023-10-17
1192 track webhook result

### LINKED ISSUE #1192: Add additional data to telemetry span in Webhook controller, to include HTTP response code in metadata
If the metadata includes the HTTP return code then we can separate the metrics for 200s from those for 404s.

rorymckinley:
## Notes for the reviewer

I have added telemetry metadata that indicates the response of the call and then filtered the telemetry metrics to only show for the cases where the call state was `:ok`. My reasoning is that the execution time between an `:ok` versus a `:not_found`  may differ greatly and this could lead to skewed results.

## Related issue

Fixes #1192 

## Review checklist


taylordowns2000 (approved):

taylordowns2000:
Really good call on successful webhook url only.

rorymckinley:
@taylordowns2000 I finally managed to appease the CodeCoverage gods :) - ready to be merged in.

taylordowns2000 (approved):

PR #1214, opened 2023-10-18
Prevent git push without credentials

jyeshe:
## Notes for the reviewer

This prevents "state.json" or similar content to be pushed even temporarily to the public git repo.

Additionally it enables git commit and push hooks for code formatting and analysis.

## Related issue

Keep going and help automating a bit GDPR (and equivalent) compliance.

## Review checklist


taylordowns2000:
Oooh actually, @jyeshe , this should probably go in the https://github.com/OpenFn/project template

jyeshe:
> @jyeshe , this is super cool, but i'd suggest it doesn't belong in Lightning. Would it make more sense in OpenFn/kit or OpenFn/adaptors? (My rational is this... if a developer is writing job state and expressions on their local filesystem, it's because they're using the CLI, not Lightning. What do you think? In other words, I love the concept but think it's in the wrong repo!)

Hey Taylor, thanks for clarifying. For this check I have looked at this: https://github.com/OpenFn/Lightning/blob/2b93dce6d7f057601a928658f28fd7600f001e91/lib/lightning_web/controllers/api/provisioning_controller.ex#L38
However it would happen only if someone accidentally dump it in the future to any file for debugging. No doubt this wouldn't be merged (: I will adapt this PR to the scope of dev tools and apply the check to the project template. Thanks

NickOpenFn:
Blocked until @stuartc is back

stuartc:
@jyeshe @taylordowns2000 can I get an explanation of what these changes do, I mean I see the git hooks - but no changes to the README or really what the justification is. 

I can figure it out but this does change things for anyone contributing to the project - for these kinds of things I feel it's fair to have a proposal as why it's a good change.

Aside:

```
   plt_add_apps: [:mix, :ex_unit],
```

Curious why `:ex_unit` needed to be addedd to the plt?

jyeshe:
> @jyeshe @taylordowns2000 can I get an explanation of what these changes do, I mean I see the git hooks - but no changes to the README or really what the justification is.
> 
> I can figure it out but this does change things for anyone contributing to the project - for these kinds of things I feel it's fair to have a proposal as why it's a good change.
> 
> Aside:
> 
> ```
>    plt_add_apps: [:mix, :ex_unit],
> ```
> 
> Curious why `:ex_unit` needed to be addedd to the plt?

Basically it avoids code to be committed without proper formatting. From time to time we need to push again because it was not formatted (making the diff more precise as well). On a `git push` it runs the `mix credo` which is also super fast on the project instead of waiting for the CI.

The CI only passed with the `:ex_unit`.

jyeshe:
Have not been used for a long time. Moving to the umbrella app.

PR #1231, opened 2023-10-23
Enable adaptor icons to load gracefully

### LINKED ISSUE #1140: Improve Icon loading on Canvas so the icons load in gracefully
**Overview:**
1. The icons on the canvas look great, however they are loading in and rendering very abruptly.
2. If there is any delay on the icons loading the text displays - which creates a strange experience of it loading "twice"

Possible Solutions:
- Add in a animation & wording to display on the nodes (this is good regardless of issue 2 above)
- Preload & cache the images to be ready to display when user logs in on the workflow overview page
- If preloading is not possible, then perhaps we improve the perceived display of these with animations. We delay the text by showing and then image from showing by .250ms, and then _hopefully_ the images will be ready and display before the text. 
- Another potential solution would be only to render the canvas once the images are ready.

josephjclark:
## Notes for the reviewer

This PR fixes the load of adaptor icons in the Workflow Diagram so that data is shared between all components, and the icons load "gracefuly".

The fix is to use a hook which shares a single promise between all consuming components. The first call to the hook creates a promise to fetch and parse the icon manifest. Subsequent calls will just add a `then()` to this promise chain. After the data has been downloaded once, the promise will resolve instantly.

If the icon manifest changes during a user session, the icons will not update. A refresh of the page will download the latest icons, though.

Note that in the event of an error (ie the icons failing to load), the component will still fallback nicely to textural representation. An error will be logged to the console.

## QA Notes

If your internet connection is fast enough you won't see the issue - but if you look at the Diagram on `main` and check the network log, you'll notice that `adaptor_icons.json` is downloaded once per job node. On slow connections or large charts, this causes icons to pop in one at a time.

Against this branch, you should only see `adaptor_icons.json` be downloaded once.

## Related issue

Fixes #1140

## Review checklist


josephjclark:
@midigofrank yes, of course, I'll sort it out now

josephjclark:
@midigofrank I've just rebased this.

I can't test locally because my migration is failing. I don't really have time to sort it out today - can you give it a quick test before merging?

PR #1237, opened 2023-10-27
Janitor

taylordowns2000:
## Notes for the reviewer

1. Decision needed around `claimed_at` or `started_at`
2. is this the right direction/place to put things

## Review checklist

### DIFF
--- a/lib/lightning/janitor.ex
+++ b/lib/lightning/janitor.ex
@@ -15,28 +15,28 @@ defmodule Lightning.Janitor do
 
   require Logger
 
-  import Ecto.Query
   alias Lightning.{Repo, Attempt}
+  alias Lightning.Attempts
 
   @doc """
-  Takes a type to either deletes completed Oban jobs, preserving
-  discarded/cancelled for inspection or marks orphaned Oban jobs as cancelled,
-  preserving them for inspection.
+  The perform function takes an `%Oban.Job`, allowing this module to be invoked
+  by the Oban cron plugin.
   """
   @impl Oban.Worker
-  def perform(%Oban.Job{}) do
-    now = DateTime.utc_now()
-    grace_period = Application.get_env(:lightning, :max_run_duration) * -0.2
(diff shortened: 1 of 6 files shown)


lib/lightning/janitor.ex
```diff
@@ -0,0 +1,43 @@
...
+  @impl Oban.Worker
+  def perform(%Oban.Job{}) do
+    now = DateTime.utc_now()
+    grace_period = Application.get_env(:lightning, :max_run_duration) * -0.2
+    earliest_acceptable_start = Timex.shift(now, seconds: grace_period)
+
+    from(att in Attempt,
+      where: is_nil(att.finished_at),
+      # TODO: decide if this should be claimed_at or started_at
+      where: att.claimed_at < ^earliest_acceptable_start
```
stuartc, line 35: It should be claimed at I think.

Joe said "as soon as it's off the queue, it's a flight risk".

lib/lightning/janitor.ex
```diff
@@ -0,0 +1,43 @@
...
+  @doc """
+  Takes a type to either deletes completed Oban jobs, preserving
+  discarded/cancelled for inspection or marks orphaned Oban jobs as cancelled,
+  preserving them for inspection.
+  """
+  @impl Oban.Worker
+  def perform(%Oban.Job{}) do
+    now = DateTime.utc_now()
+    grace_period = Application.get_env(:lightning, :max_run_duration) * -0.2
+    earliest_acceptable_start = Timex.shift(now, seconds: grace_period)
```
stuartc, line 30: Can we use `DateTime.add(now, grace_period)` here, I'd like to avoid relying on Timex where we have built in alternatives.

stuartc (changes requested):
This looks all good @taylordowns2000, could you remove the TODO if you're in agreement over `claimed_at`, perhaps change Timex to DateTime.

And I think the query should be move to `Attempts.Query` and think a single test for it would be nice.

PR #1241, opened 2023-10-31
Claims immediate attempts first

### LINKED ISSUE #1238: Manual run Workorder prioritisation over queue
Currently all WorkOrders are executed in a FIFO fashion, for manual work orders we need to be able to 'skip the queue', or get priority.

In `Lightning.Attempts.Queue.claim/1`, attempts are allocated using a transaction lock, the order currently is based solely on time.

However manual workorders may end up waiting behind several thousand other workorders in busy moments. And an active user would be waiting indefinitely.

One proposal is to have a `priority` column that will ordinally make them be processed first. However there is a concern that having a priority column that allows more than "right now" or "not right now" could be unnecessary.

**Why is this important**

In the future, we want to be able to control the order of attempt execution by many other facets, and more importantly _dynamically_. For example, a critical project or user would get priority over a test project - for a small amount of time; e.g.

Pro User submits 10,000 workorders
Test User submits 100 workorders

We ideally would want the Pro user to be able to process 2-3x more workorders for a given timeframe, but for the Test user to not have to wait until the 10,000 workorders have been completed.

In addition a 3rd user coming in while the workorders are being processed, adding say 1,000 to the queue, we would split the work using a round robin pattern perhaps; this implies something more dynamic than a fixed set of priorities.

**But this isn't what we are implementing now**

Currently the logic is binary, interactive users shouldn't have to wait a long time for their results. So perhaps there is a `immediate` column that is a boolean is a better way to not introduce a field that shouldn't be used for other purposes.

**Is this absolutely the way to do it**

Probably, we don't want to introduce concepts we aren't absolutely sure about or things that we can't see its use far into the future.

jyeshe:
## Notes for the reviewer

On a claim from the attempts queue, it returns attempts with higher priority (`immediate`) before regular ones, expect when more than 1 is demanded and the amount immediate remaining ones is less than the demand. 

## Related issue

Resolves #1238 

## Review checklist

### DIFF
--- a/lib/lightning/attempts/attempt.ex
+++ b/lib/lightning/attempts/attempt.ex
@@ -55,6 +55,10 @@ defmodule Lightning.Attempt do
     field :started_at, :utc_datetime_usec
     field :finished_at, :utc_datetime_usec
 
+    field :priority, Ecto.Enum,
+      values: [immediate: 0, normal: 1],
+      default: :normal
+
     timestamps type: :utc_datetime_usec, updated_at: false
   end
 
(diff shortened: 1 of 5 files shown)


jyeshe:
Validated with: `mix test test/lightning/attempts_test.exs`

stuartc (approved):
This is great @jyeshe, super clean.

PR #1260, opened 2023-11-02
1256 worker usage

### LINKED ISSUE #1256: Update docker-compose and setup README for `ws-worker`
The goal here is that devs should be able to run via docker (using compose) _or_ run via mix (either running worker by hand) or running worker via Frank's GenServer.

For the `mix` pathway, some considerations:

1. They can start the worker server by hand. (Not great, but simple.)
2. They can start the worker server in a docker image. (Could run into networking issues on Mac or any environment where Docker stuff is inside a VM.)
3. They can start the worker server with something like `npx @openfn/ws-worker` - this could even be called inside Frank's module?

ToDo:

- [x] Update the docker-compose file for Lighting to use `ws-worker` if you're running only.
- [ ] Update the developer getting started section to include instructions on `ws-worker` setup
- [x] Consider updating @midigofrank 's GenServer to boot up `ws-worker` locally for the above.

taylordowns2000:
- [x] Update the docker-compose file for Lighting to use ws-worker if you're "running" only.
- [ ] Update the developer getting started section to include instructions on ws-worker setup
- [ ] Consider updating [@midigofrank](https://github.com/midigofrank) 's GenServer to boot up ws-worker locally for the above.

## Related issue

Fixes #1256 

## Review checklist


taylordowns2000:
good progress but not quite there @stuartc . i'm merging this down and tracking work on #1266 . as suggested, i've taken a new branch off @midigofrank 's work and while i'm having trouble with the difference between `esbuild` (not a genserver) and `RuntimeManager` (a genserver) it doesn't seem like it will be too hard to sort out down the line.

Top priority is that everyone is running the worker when they type `mix phx.server`, and this PR accomplishes that, so as discussed I'll merge. Hope to get Frank on #1266 as soon as we have a stable app running on main.

PR #1262, opened 2023-11-02
Validate attempt completion

### LINKED ISSUE #1251: Handle `attempt:complete` events from worker to set state
@stuartc , @josephjclark , @jyeshe , some decision making and implementation is needed around `attempt:complete` handling.

### What's happening now?

When `ws-worker` sends an `attempt:complete` message to the attempts channel, it looks like this:
```ex
Last message: %Phoenix.Socket.Message{
  topic: "attempt:29e73c3f-fb74-4441-b04e-293a9de5b445",
  event: "attempt:complete",
  payload: %{
    "final_dataclip_id" => "62a5f87e-5686-4a0d-bf95-88655ddee625",
    "message" => "fn is not defined",
    "reason" => "fail"
  },
  ref: "35",
  join_ref: "29"
}
```

This results in a GenServer error:

```
[error] GenServer #PID<0.1675.0> terminating
** (Postgrex.Error) ERROR 23502 (not_null_violation) null value in column "state" of relation "attempts" violates not-null constraint

    table: attempts
    column: state
```

## How to approach it

We need to look carefully at:

1. [lib/lightning/attempts/handlers.ex:184: complete_run](https://github.com/OpenFn/Lightning/blob/release-0.10.0/lib/lightning/attempts.ex#L184-L188)
2. [lib/lightning/attempts/handlers.ex:88: complete_attempt](https://github.com/OpenFn/Lightning/blob/release-0.10.0/lib/lightning/attempts.ex#L88-L97)
3. and the [handlers](https://github.com/OpenFn/Lightning/blob/release-0.10.0/lib/lightning/attempts/handlers.ex)

To decide how and where we map `attempt:complete` messages coming from the worker back to `:state` for attempts.

## User acceptance criteria

I'll test this by running an end-to-end test...
1. `openfn/lightning`: `iex -S mix phx.server` (with worker secret in `.env.dev`)
2. `openfn/kit/packages/ws-worker`: `WORKER_SECRET="shhhhhhh" pnpm start`
3. Start an attempt on Lightning
4. Ensure it's executed and all data gets sent back to Lightning.

jyeshe:
## Notes for the reviewer

- Adds validation for nil attempt state
- Extracts the state from payload `reason`

## Related issue

Fixes #1251 

## Review checklist

### DIFF
--- a/lib/lightning/attempts/attempt.ex
+++ b/lib/lightning/attempts/attempt.ex
@@ -17,6 +17,23 @@ defmodule Lightning.Attempt do
   alias Lightning.Workflows.Trigger
   # alias Lightning.Workflows.Node
 
+  @final_states [
+    :success,
+    :cancelled,
+    :failed,
+    :killed,
+    :crashed
+  ]
+
+  @doc """
+  Returns the list of final states for an attempt.
+  """
+  defmacro final_states do
+    quote do
+      unquote(@final_states)
+    end
+  end
+
   @type t :: %__MODULE__{
           __meta__: Ecto.Schema.Metadata.t(),
           id: Ecto.UUID.t() | nil,
@@ -40,15 +57,15 @@ defmodule Lightning.Attempt do
     many_to_many :runs, Run, join_through: AttemptRun
 
     field :state, Ecto.Enum,
-      values: [
-        :available,
-        :claimed,
-        :started,
-        :success,
-        :failed,
-        :killed,
-        :crashed
-      ],
+      values:
+        Enum.concat(
+          [
+            :available,
+            :claimed,
+            :started
+          ],
+          @final_states
+        ),
       default: :available
 
     field :claimed_at, :utc_datetime_usec
@@ -122,13 +139,12 @@ defmodule Lightning.Attempt do
   def complete(attempt, state) do
     attempt
     |> cast(%{state: state}, [:state])
-    |> change(finished_at: DateTime.utc_now())
-    |> validate_inclusion(:state, [:success, :failed, :killed, :crashed])
-    |> then(fn changeset ->
-      previous_state = changeset.data |> Map.get(:state)
-
+    |> validate_required([:state])
+    |> validate_inclusion(:state, @final_states)
+    |> then(fn %{data: %{state: previous_state}} = changeset ->
       if previous_state == :started do
         changeset
+        |> change(finished_at: DateTime.utc_now())
       else
         changeset
         |> add_error(:state, "cannot complete attempt that is not started")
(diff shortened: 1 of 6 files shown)


jyeshe:
Added mapping for RTM reason terms.

jyeshe:
Validation steps:
```
mix test test/lightning/attempts_test.exs
mix test test/lightning_web/channels/attempt_channel_test.exs
```

lib/lightning/attempts/attempt.ex
```diff
@@ -17,6 +17,23 @@ defmodule Lightning.Attempt do
   alias Lightning.Workflows.Trigger
   # alias Lightning.Workflows.Node
 
+  @final_states [
+    :success,
```
taylordowns2000, line 21: this is the only one that isn't in the past tense. why? would it make more sense to call it either:
1. `succeeded` or
2. `completed` ?

taylordowns2000 (commented):
looks great @jyeshe . have one question on the naming but otherwise looks cool. i assume this is working OK on your local setup?

jyeshe:
> looks great @jyeshe . have one question on the naming but otherwise looks cool. i assume this is working OK on your local setup?

It works, it doesn't raise the error and updates the attempt.

lib/lightning/attempts/attempt.ex
jyeshe: > this is the only one that isn't in the past tense. why? would it make more sense to call it either:
> 
> 1. `succeeded` or
> 2. `completed` ?

Just kept the use of `success`. Would you mind if we change it later when we stabilize the codebase?

taylordowns2000: lol, good idea!

taylordowns2000 (approved):

stuartc (commented):
@jyeshe I thought we discussed creating a handler or facade to manage the differences between our models and the messages from the worker? We have gotten away with the assuming the reason/state values are the same for a while (hence using `Attempt.complete`); but I'm weary of any further changes the worker will have that will require either data migrations or other changes to Lightnings data.

I agree we don't want to be changing all of these states from past to present tense.

My understanding here is that the worker reports in present tense, since it's the one that dealt with it at that very moment - and Lightning registers that as a "past event".

jyeshe:
> @jyeshe I thought we discussed creating a handler or facade to manage the differences between our models and the messages from the worker? We have gotten away with the assuming the reason/state values are the same for a while (hence using `Attempt.complete`); but I'm weary of any further changes the worker will have that will require either data migrations or other changes to Lightnings data.
> 
> I agree we don't want to be changing all of these states from past to present tense.
> 
> My understanding here is that the worker reports in present tense, since it's the one that dealt with it at that very moment - and Lightning registers that as a "past event".

For now a single function on the channel was enough. When we have multiple places to handle the protocol, it will be helpful to centralize in another place. Sure, no problem with different terms.

stuartc (approved):

PR #1269, opened 2023-11-03
Ignore broadcasted events not pushed to channel

jyeshe:
## Notes for the reviewer

`Lightning.Attempts.Events.AttemptUpdated` and `Lightning.Attempts.Events.RunStarted` are broadcasted but don't need to be sent to the socket so they are ignored on the channel.

## Validation steps

```
mix test test/lightning_web/channels/attempt_channel_test.exs
```

## Related issue

Fixes the warnings of ignored events.

## Review checklist


jyeshe:
If the events are not needed by any other PubSub subscriber  anymore, we can skip broadcasting them.

taylordowns2000 (approved):

stuartc:
@jyeshe thanks, I also saw this - we definitely need those events - we use them to follow attempts/stream logs to the front end.

But this particular issue is interesting because the channel _shouldn't_ be subscribing to any events, i.e. `handle_info` shouldn't be getting called - and it appears to only be happening in tests I think. 

I have a suspicion that it's because the channel has the same name as the Attempt events, `attempt:#{id}`. I'd be interested in knowing if we change the topic string in `Attempts.Events` to see if the issue goes away and we don't need to `handle_info` in the AttemptChannel.

jyeshe:
Hi @stuartc, @taylordowns2000 updated me that the LV would need to subscribe to them.
That's correct, Phoenix channels subscribe to the PubSub:
https://github.com/phoenixframework/phoenix/blob/cdf0eb83d51584e046770d95c6aaf53fd4c4ac95/lib/phoenix/channel/server.ex#L434

PR #1271, opened 2023-11-03
Refresh oath token on worker channel

### LINKED ISSUE #1246: Handle oAuth2 Refresh in RTM
When a Job is going to be executed it receives a credential (assuming it has one).

```
# lib/lightning/pipeline/runner.ex:119
Lightning.Credentials.maybe_refresh_token(run.job.credential)
```

What this does is, _if_ the credential is an oAuth2 credential (currently only GoogleSheets is supported) _and_ the token is due to expire soon (5mins), then a new access token is requested and then saved.

In the new worker processing model, the attempt channel does not use this functionality.

We need to fold in this behaviour and ensure it's tested (reasonably).

jyeshe:
## Notes for the reviewer

Uses the payload job to get the credential and refresh the token if it is about to expire (5 min check). 

Handled on `run:start` since `fetch:attempt` does not have payload.

## Validation steps
```mix test test/lightning_web/channels/attempt_channel_test.exs```

## Related issue

Fixes #1246 

## Review checklist


jyeshe:
## Validation steps
```mix test test/lightning_web/channels/attempt_channel_test.exs```

taylordowns2000 (approved):

Issue #1266, opened 2023-11-03
Move ws-worker out of devDependencies


taylordowns2000:
We'd like to optimize the developer experience for running a NodeJs worker server. Right now, we've got some boilerplate from how phoenix does esbuild, but we're side-stepping a lot of the cool stuff in there and simply using npm to install a `@openfn/ws-worker` module then running it from a genserver.

This ticket is to go back to a more `esbuild`-style approach, where the worker isn't installed by NPM, but instead installed by hand and kept separate from the project dependencies.

Notes:
1. esbuild is not a genserver. why are we using a genserver and starting the worker as a child process?
2. we don't support different architectures. (do we?)
3. running notes here...

taylordowns2000:
@midigofrank , @stuartc , feeling like we should wait on this one until https://github.com/OpenFn/kit/issues/426 is delivered.

taylordowns2000:
Based on [this esbuild](https://github.com/phoenixframework/esbuild/blob/main/lib/esbuild.ex#L220-L255) stuff, my guess is that we want a binary. The [`System.cmd`](https://github.com/phoenixframework/esbuild/blob/main/lib/esbuild.ex#L180-L181) here eventually expects to call that binary with (if necessary) some extra args.

But @stuartc may be targeting something that is still run using npm—he mentioned that he thought a zero-deps worker might be very hard to deliver and it's not necessary for our current DX.

PR #1286, opened 2023-11-06
Terminate worker socket process on start and stop

### LINKED ISSUE #1272: Close ws-worker port when shutting down the app
- If the app is started in dev mode, we run `npm exec @openfn/ws-worker` via a port.
- Later, this will be replaced with a binary and we'll bring in some esbuild-style goodness, but that's for later.
- For now, when a developer shuts down the app with `control-c`, the NodeJs worker app is left running... we're not sending the proper `Port.close` command, or SIGTERM, or whatever is required to stop the NodeJs worker app.
- It's a pain in the butt because the dev then has to do something like `kill $(lsof -n -i :2222 | grep LISTEN | awk '{print $2}')` to kill that old NodeJs process before they can reboot their app... or else they'll get an `Error: listen EADDRINUSE: address already in use :::2222`
- After a quick chat with @jyeshe and @josephjclark , it sounds like #1266 might take a while, but that handling the exit for this current implementation and shutting down the app we start with `npm exec` will be significantly easier.
- And it will make the DX _much_ better!

This issue is to do just that: if the app is started with `mix phx.server` in dev mode and we open a port for `npm exec @openfn/ws-worker`, make sure that we close that port when we stop the app.

jyeshe:
## Notes for the reviewer

Terminates the hanging/zombie socket process when closing/terminating Lightning app.

This process is not automatically terminated when the worker port process receives a `SIGTERM`.

The solution handles two cases:
1. Cleaning the zombie socket process on port 2222 when starting the app after closing it on IEx (which doesn't allow a safe cleanup on Lightning termination). 
2. Terminates the worker socket process on `Application.prep_stop`, when sending a `SIGTERM` to Lightning for example (case of stopping a container).

For the second item, the logic to get the `nodejs` pid follows this process tree:

```
$ pstree -p 132174
bash(132174)─┬─bash(132176)───bash(132177)
             └─npm exec @openf(132175)─┬─sh(132368)───node(132369)
```

Which is obtained with:
```
$ ps -s 132174 -o pid=
 132174
 132175
 132176
 132177
 132368
 132369
``` 

## Related issue

Fixes #1272

## Review checklist


taylordowns2000 (approved):
oooh, super tricky. ok, from my perspective this does the trick for the DX _for now_ and we'll need to handle this more gracefully in https://github.com/OpenFn/Lightning/issues/1266

Issue #895, opened 2023-06-21
Move enabled/disabled property from jobs to edges


taylordowns2000:
Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=385-7635&mode=design&t=GE0qZkJNsdMrtFey-4

Currently an edge comprises the following elements:
- A condition, which can take on values such as `:always`, `state.data.form.entries[0] == 'test application'`, or `state.data.form.user.submitted_by.crazy_long_path.email` == `'test@msf.org'`, depending on specific requirements.
- A label.
- An icon.

We need to extend this to now include an 'enabled' boolean to provide users with the ability to deactivate an edge. This deactivation prevents the edge from being considered when generating a 'plan' without entirely deleting it or setting it to 'false'.

This design pattern effectively separates the execution of 'jobs' from the decision of whether they should run. In the case of a condition, this responsibility lies with the 'edge,' and for 'enabled/disabled' status, it's determined by the component responsible for building the workflow 'plan.'"

~Noting that if an edge is disabled, all following edges will be disabled (and therefore, the jobs won't run).~ @benedictus-dev , @NickOpenFn , this isn't right. let's chat this morning.

- [x] Add checkbox for enabling/disabling the edge on the edge panel 
- [x] Remove checkbox for enabling/disabling the edge on the job panel 
- [x] Add a migration that will add a enabled column in the edge table `workflow edges` and remove the enabled column from the jobs table
- [x] Refactor all code areas that reference of `job.enabled` to `edge.enabled`
- [x] Update the project import via API feature in Lightning if necessary
- [x] ~If the trigger of the workflow is disabled, all edges are disabled~ (TD says this is not correct)
- [x] Fix up all tests

## Legacy app todo's:

- [ ] OpenFn/platform-app#1820

----

## Original spec: 

An edge has:
1. A condition (`:always`, `state.data.form.entries[0] == 'test application' || state.data.form.user.submitted_by.crazy_long_path.email == 'test@msf.org'`, `:on_success`, etc.)
2. A label
3. An icon

And now,

4. An `enabled` boolean which allows the user to turn off an edge (preventing it from being used when generating a `plan`) without deleting it entirely or setting it to `false`.

With this pattern, `jobs` are stupid. They're told to run, given state, etc. But they don't "decide" whether or not they run. That's the responsibility of the `edge` in the case of a condition or whatever builds a workflow `plan` in the case of `enabled/disabled`.

[...]

PR #1289, opened 2023-11-07
895 move enabled disabled property from jobs to edges

benedictus-dev:
## Related issue
This PR addresses issue #895, focusing on the migration of the 'enabled/disabled' property from the 'jobs' to the 'edges' within our workflow management system
Fixes #

1. Introduces an 'enabled' boolean to edges, allowing for a more granular control over workflow 
2. Removes the 'enabled/disabled' toggles from the job panel, aligning with our strategy to separate the execution of jobs from decision-making processes.
## Review checklist

### DIFF
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -465,11 +465,11 @@ defmodule Lightning.Projects do
       end
 
     Repo.transaction(fn ->
-      workflows = project_workflows_query(project) |> Repo.all()
+      triggers = project_triggers_query(project) |> Repo.all()
 
-      workflows
-      |> Enum.each(fn workflow ->
-        Lightning.Workflows.mark_for_deletion(workflow)
+      triggers
+      |> Enum.each(fn trigger ->
+        Lightning.Workflows.update_trigger(trigger, %{"enabled" => false})
       end)
 
       project =
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -291,6 +291,13 @@ defmodule Lightning.Projects do
     )
   end
 
+  def project_edges_query(project) do
+    from(e in Edge,
+      join: w in assoc(e, :workflow),
+      where: w.project_id == ^project.id
+    )
+  end
+
   def project_runs_query(project) do
     from(r in Run,
       join: j in assoc(r, :job),
@@ -447,6 +454,7 @@ defmodule Lightning.Projects do
   @spec export_project(:yaml, any) :: {:ok, binary}
   def export_project(:yaml, project_id) do
     {:ok, yaml} = ExportUtils.generate_new_yaml(project_id)
+   
 
     {:ok, yaml}
   end
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -291,8 +291,6 @@ defmodule Lightning.Projects do
     )
   end
 
-
-
   def project_runs_query(project) do
     from(r in Run,
       join: j in assoc(r, :job),
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -465,11 +465,11 @@ defmodule Lightning.Projects do
       end
 
     Repo.transaction(fn ->
-      jobs = project_jobs_query(project) |> Repo.all()
+      workflows = project_workflows_query(project) |> Repo.all()
 
-      jobs
-      |> Enum.each(fn job ->
-        Lightning.Jobs.update_job(job, %{"enabled" => false})
+      workflows
+      |> Enum.each(fn workflow ->
+        Lightning.Workflows.mark_for_deletion(workflow)
       end)
 
       project =
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -12,6 +12,8 @@ defmodule Lightning.Projects do
   alias Lightning.AttemptRun
   alias Lightning.Workflows.Trigger
   alias Lightning.Workflows.Job
+  alias Lightning.Workflows.Edge
+
   alias Lightning.Projects.ProjectUser
   alias Lightning.Repo
 
@@ -454,7 +456,6 @@ defmodule Lightning.Projects do
   @spec export_project(:yaml, any) :: {:ok, binary}
   def export_project(:yaml, project_id) do
     {:ok, yaml} = ExportUtils.generate_new_yaml(project_id)
-   
 
     {:ok, yaml}
   end
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -12,8 +12,6 @@ defmodule Lightning.Projects do
   alias Lightning.AttemptRun
   alias Lightning.Workflows.Trigger
   alias Lightning.Workflows.Job
-  alias Lightning.Workflows.Edge
-
   alias Lightning.Projects.ProjectUser
   alias Lightning.Repo
 
@@ -293,12 +291,7 @@ defmodule Lightning.Projects do
     )
   end
 
-  def project_edges_query(project) do
-    from(e in Edge,
-      join: w in assoc(e, :workflow),
-      where: w.project_id == ^project.id
-    )
-  end
+
 
   def project_runs_query(project) do
     from(r in Run,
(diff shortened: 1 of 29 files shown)


lib/lightning/workflows/edge.ex
```diff
@@ -39,6 +40,8 @@ defmodule Lightning.Workflows.Edge do
 
     field :condition, Ecto.Enum, values: @conditions
 
+    field :enabled, :boolean
```
taylordowns2000, line 43: i think this should default to true to keep parity with the old `job.ex`

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -682,6 +682,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
   @impl true
   def handle_info({"form_changed", %{"workflow" => params}}, socket) do
+    IO.puts("=====================When to trigger me ===============")
```
taylordowns2000, line 685: remove IO.inspect?

lib/lightning_web/live/workflow_live/workflow_params.ex
```diff
@@ -13,7 +13,13 @@ defmodule LightningWeb.WorkflowNewLive.WorkflowParams do
     Map.merge(current_params, form_params, fn _key, current, next ->
       case {current, next} do
         {current, next} when is_list(current) and is_map(next) ->
-          merge_list_params(current, next)
+          if Map.has_key?(form_params, "edges") &&
```
taylordowns2000, line 16: i think this is the code that's setting downstream edges. please ensure that if someone disbales an edge, only that edge gets disabled.

taylordowns2000 (changes requested):
hey @benedictus-dev , looking good! please address all of my inline change requests. 

the big thing (workflow_params, line 16) is that upon disabling an edge all downstream edges are getting disabled. it looks super deliberate so i'm guessing this was simply a miscommunication somewhere. please make sure that when a user disables an edge, only that edge gets disabled. other elements of the workflow should be left alone.

taylordowns2000 (changes requested):
1. please re-write these migrations to use a single migration that _moves_ the boolean from jobs to edges. this is fairly tricky. see details below.
2. please add tests (chat @midigofrank and @elias-ba on slack and rebase on release-0.10.0 once they've finished fixing the workflow params test.)

details on the migration:
- currently the enabled/disabled status lives on the jobs table.
- we need to determine what edge is pointing at each job, and _move_ the value of that boolean to the leading edge.
- the exception here is that no edges which lead from a trigger to a job may be disabled.

so to say it another way:
- first add the enabled column to edges
- then set the value with this logic:
  - if the edge has a `source_trigger_id`, set `edge.enabled = true`
  - if the edge does not have a source trigger, set the `edge.enabled` value to the `jobs.enabled` value for the `edge.target_job_id`
- then finally drop the `jobs.enabled` column

test/lightning/workflows/query_test.exs
```diff
@@ -60,21 +61,36 @@ defmodule Lightning.Workflows.QueryTest do
...
 
       _disabled_cronjob =
-        insert(:job, enabled: false, workflow: trigger.workflow)
+        insert(:job, workflow: trigger.workflow)
+
+      insert(:edge, %{
+        source_trigger: trigger,
+        target_job: _disabled_cronjob,
+        workflow: _disabled_cronjob.workflow,
+        enabled: false
```
taylordowns2000, line 80: this is impossible, i think!

taylordowns2000 (changes requested):
this works _great_ from a UX perspetive! nice job @benedictus-dev ! todo:

1. ~mark_for_deletion - is this doing what needs to be done? i thought we discussed setting the trigger to disabled.~ done
2. canonical_project - i see you've removed `enabled` from jobs. should i be added to edges?
3. `test/lightning/projects_test.exs` - i think this is related to number item 1. i think we discussed checking to see that triggers were disabled, rather than jobs.
4. `test/lightning/workflows/query_test.exs` - i'm not convinced that this enabled_cron_jobs_by_edge query is being tested properly with your changes. i'm also not sure it's being used anymore! how do we find a list of enabled cron workflows to create new attempts in the system?
5. see "impossible" comment on trigger-to-job edge with `enabled: false`. we should add validation here to stop this.
6. please remove remaining commented out code

test/lightning/workflows/query_test.exs
```diff
@@ -60,21 +61,36 @@ defmodule Lightning.Workflows.QueryTest do
...
       webhook_trigger = insert(:trigger, type: :webhook)
 
       _non_cron_job =
-        insert(:job, enabled: true, workflow: webhook_trigger.workflow)
+        insert(:job, workflow: webhook_trigger.workflow)
+
+      insert(:edge, %{
+        source_trigger: webhook_trigger,
+        target_job: _non_cron_job,
+        workflow: _non_cron_job.workflow,
```
stuartc, line 91: Variables prefixed with an underscore are a hint that the variable isn't supposed to be used: https://hexdocs.pm/elixir/main/naming-conventions.html#underscore-_foo

It was left there as a 'comment' of sorts, in tests they are permissible but we try and not have these kinds of variable names in the application code.

So in this situation, could you remove the `_` prefix from the variable.

lib/lightning/projects.ex
```diff
@@ -465,11 +465,11 @@ defmodule Lightning.Projects do
       end
 
     Repo.transaction(fn ->
-      jobs = project_jobs_query(project) |> Repo.all()
+      workflows = project_workflows_query(project) |> Repo.all()
```
taylordowns2000, line 468: this should be `triggers`

lib/lightning/workflows/query.ex
```diff
@@ -36,7 +36,7 @@ defmodule Lightning.Workflows.Query do
     from(e in Edge,
       join: j in assoc(e, :target_job),
       join: t in assoc(e, :source_trigger),
-      where: t.type == :cron and t.enabled,
+      where: t.type == :cron and t.enabled and e.enabled,
```
taylordowns2000, line 39: remove `and e.enabled`

lib/lightning/projects.ex
```diff
@@ -465,11 +465,11 @@ defmodule Lightning.Projects do
...
 
-      jobs
-      |> Enum.each(fn job ->
-        Lightning.Jobs.update_job(job, %{"enabled" => false})
+      workflows
+      |> Enum.each(fn workflow ->
+        Lightning.Workflows.mark_for_deletion(workflow)
```
taylordowns2000, line 472: with this PR, you've moved `:enabled` from Jobs to Edges. On this line you should go from:
```
Lightning.Jobs.update_job(job, %{"enabled" => false})
```
to:
```
Lightning.Triggers.update_trigger(trigger, %{"enabled" => false})
```
There may not be an `update_trigger` function and you may need to make one or call `Repo.update()`

test/lightning/setup_utils_test.exs
```diff
@@ -219,7 +217,7 @@ defmodule Lightning.SetupUtilsTest do
                });
              """
 
-      assert job_3.enabled
+      # assert job_3.enabled
```
taylordowns2000, line 220: remove commented code

test/lightning/setup_utils_test.exs
```diff
@@ -538,7 +534,7 @@ defmodule Lightning.SetupUtilsTest do
              fn(state => state);
              """
 
-      assert notify_upload_successful.enabled
+      # assert notify_upload_successful.enabled
```
taylordowns2000, line 537: remove commented code

test/lightning/workflows_test.exs
```diff
@@ -128,7 +130,6 @@ defmodule Lightning.WorkflowsTest do
 
       # Disabled Job
```
taylordowns2000, line 131: rename this comment; jobs can't be disabled

taylordowns2000 (changes requested):
looking really good @benedictus-dev . please address each of my final change requests and re-request a review. let's get this over the line!!!

taylordowns2000 (approved):

PR #1302, opened 2023-11-07
Fix failure alert tests

elias-ba:
## Related issue

Fixes #

## Review checklist

### DIFF
--- a/test/lightning/failure_alert_test.exs
+++ b/test/lightning/failure_alert_test.exs
@@ -1,126 +1,122 @@
 defmodule Lightning.FailureAlertTest do
-  use Lightning.DataCase, async: true
-  use Mimic
+  use LightningWeb.ChannelCase, async: true
 
-  alias Lightning.Pipeline
-  alias Lightning.{Attempt, AttemptRun}
-  alias Lightning.Invocation.{Run}
-
-  import Lightning.InvocationFixtures
-  import Lightning.JobsFixtures
-  import Lightning.ProjectsFixtures
-  import Lightning.AccountsFixtures
-
+  import Lightning.Factories
   import Lightning.Helpers, only: [ms_to_human: 1]
-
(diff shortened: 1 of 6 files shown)


test/lightning/failure_alert_test.exs
```diff
@@ -176,77 +175,77 @@ defmodule Lightning.FailureAlertTest do
...
+
+      push(socket, "run:complete", %{
+        "run_id" => run.id,
+        "output_dataclip_id" => Ecto.UUID.generate(),
+        "output_dataclip" => ~s({"foo": "bar"}),
+        "reason" => "normal"
+      })
+
+      # Wait a bit to make sure run:complete is handled
+      # TODO: Check with Taylor DOWNS and team if this is good enough
```
elias-ba, line 245: @taylordowns2000 please in the AM, would you mind checking this out ?

taylordowns2000: looks good. reason "normal" should be reason "failed", but otherwise it's great.

taylordowns2000 (approved):
@elias-ba , this looks great. i'm making a small change (reason normal to reason fail) but otherwise it seems to test the right stuff well!

taylordowns2000 (changes requested):
ahhh sorry @elias-ba ... i was wrong. it looks like this alerter is now bound to `run:complete` and it _should_ be bound to `attempt:complete`!

can you shift this around so that it notifies the user if an **attempt** fails, not a run?

elias-ba:
> ahhh sorry @elias-ba ... i was wrong. it looks like this alerter is now bound to `run:complete` and it _should_ be bound to `attempt:complete`!
> 
> can you shift this around so that it notifies the user if an **attempt** fails, not a run?

@taylordowns2000 this is now handled. Please review again and let me know

taylordowns2000 (approved):

PR #1304, opened 2023-11-08
New Manual Attempt components

### LINKED ISSUE #1245: Load the full log history of the "Attempt" when viewing Run History
**Background:**
When viewing the logs for the Run History, we display logs scoped only to certain CLI information and the specific job.
This information is limiting, and users require full logs to be shown.

**Changes when viewing logs for a job run:**
- We will now be displaying the full attempt log when a job is viewed.
- The job section will now be highlighted and scrolled into view, with a few rows displayed above (2 or three) so the user can easily see there is more info preceding the highlighted job section.

As discussed with @stuartc, the best way to drive this is by:
- Having all attempt logs displayed when called by attempt ID
- Adding a parameter to the call for the job ID which indicates that the job should be scrolled into view and highlighted.

**Areas to implement:**
- Job editor view
- Job detail page (accessed via history page)

**Notes:**
- Highlighted color to be defined from existing tailwind palette (for colors we will have a normal state, hover state, highlighted state and text selected state)

stuartc:
## Related issue

Fixes #1245 

## Review checklist

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -192,3 +192,9 @@ div[id^='tippy-'] {
 .collapsible-panel.collapsed .panel-header-title {
   transform: rotate(-90deg) !important;
 }
+
+div.line-log-level::before {
+  content: attr(data-log-level);
+  padding-left: 0.1em;
+  max-width: min-content;
+}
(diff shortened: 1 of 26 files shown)


taylordowns2000 (commented):
great job @stuartc !

PR #1244, opened 2023-10-31
1103 create workflow name before canvas

benedictus-dev:
## Notes for the reviewer
UI Enhancements: 
     The "Create" button remains disabled until a valid name is entered.
     Real-time validation checks for duplicate names, providing immediate feedback on each keypress. 
     A tooltip appears when users hover over the disabled "Create" button, guiding them on the required action.
    The modal design adheres strictly to our existing components, maintaining consistency across the platform.
    Users can easily close the modal by clicking "Cancel" or the "X" button, offering flexibility in navigation.

## Related issue 

Fixes #1103 

## Review checklist

- [yes] I have performed a **self-review** of my code
- [ yes] I have verified that all appropriate **authorization policies** have been implemented and tested
- [ ] If needed, I have updated the **changelog**
- [ ] Product has **QA'd** this feature

PR #1307, opened 2023-11-08
Navigate to edit workflow after save

### LINKED ISSUE #1280: Refreshing redirects to new workflow after save
1. Create a new workflow
2. Save it
3. Hit refresh
4. The user expects the page to refresh and for the newly saved workflow to still be there. Instead, it takes them to a new workflow with a placeholder name.

jyeshe:
## Notes for the reviewer

On refresh the new page is remounted (workflow is null) and the workflow is not on the query params.

The solution despite of that, was navigate to the edit mode after `save`, which makes sense in any case.

## Related issue

Fixes #1280 

## Review checklist


jyeshe:
Missing to prevent navigation when already on edit.

taylordowns2000:
@jyeshe , would you mind rebasing and marking this as not "draft" once it's ready for review?

elias-ba (changes requested):
Hey @jyeshe it feels like this is introducing a regression. When I open a workflow and I edit a job, I can no longer save it. Here is the error I am getting.
image

jyeshe:
> @jyeshe , would you mind rebasing and marking this as not "draft" once it's ready for review?

@taylordowns2000, @elias-ba 

I have rebased and there is still this problem of saving after refresh which I am working on. 
 
I was analyzing if it wouldn't be better fix it after #1244 however if we stay on the `:new` workflow mode there would be two things to fix:
1. After refresh the workflow is rendered as the initial one (only with 1 common job), as you've pointed out on the issue
2. Trying to save it gives an error of name uniqueness (the name is preserved but sets a new id).

I have switched back to `push_navigate` once with `push_patch` the panels stop working and I think switching to `:edit` mode is a coherent move. Working on the `Repo.update`'s changeset after refresh ...

taylordowns2000 (changes requested):
Getting some strange behavior here: https://www.loom.com/share/d5137003fb684d25bfe6197bc2d92f1c

Won't let me navigate to the new workflow

jyeshe:
> Getting some strange behavior here: https://www.loom.com/share/d5137003fb684d25bfe6197bc2d92f1c
> 
> Won't let me navigate to the new workflow

Maybe some cache @taylordowns2000 ? Could you try after a `mix do ecto.drop, ecto.setup` and clearing the cache. 
I can't reproduce it here and I have just rebased again the tests.

PR #1308, opened 2023-11-08
Partition log_lines

rorymckinley:
## Notes for the reviewer

Partitions the `log_lines` table by hash of the `attempt_id` plus some test fixes to deal with the removal of the PK on `id`.

On my branch there are currently 2 failing tests  that appear unrelated and I see that there is a feature branch addressing them - so I have left them for now. I have also looked at the other failing checks and they appear to be unrelated to the code changes in this branch.

I am not sure if this change warrants mention in the CHANGELOG?

## Related issue

Fixes #

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,11 +36,13 @@ and this project adheres to
 - Add plumbing to support the use of PromEx
   [#1199](https://github.com/OpenFn/Lightning/issues/1199)
 - Add warning text to PromEx config 
-  [#1222] (https://github.com/OpenFn/Lightning/issues/1222)
+  [#1222](https://github.com/OpenFn/Lightning/issues/1222)
 - Track and filter on webhook controller state in :telemetry metrics
   [#1192](https://github.com/OpenFn/Lightning/issues/1192)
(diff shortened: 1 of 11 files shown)


rorymckinley:
@stuartc @jyeshe There do not appear to be 'unit'-level tests for LogLine - once I am done with the table partitioning tasks, do you have any objections to me adding (at least) a test to cover `LogLine.new`? 

I imagine that this would not have been considered necessary while we were relying on baked-in Ecto/Postgres PK generation but, now that we no longer have that as a backstop, it may be worth adding some coverage?

jyeshe:
> @stuartc @jyeshe There do not appear to be 'unit'-level tests for LogLine - once I am done with the table partitioning tasks, do you have any objections to me adding (at least) a test to cover `LogLine.new`? 
> 
> I imagine that this would not have been considered necessary while we were relying on baked-in Ecto/Postgres PK generation but, now that we no longer have that as a backstop, it may be worth adding some coverage?

Good idea @rorymckinley

stuartc (approved):
Wonderful @rorymckinley! Thanks!

PR #1309, opened 2023-11-08
Update the broadcast events for attempts and runs on History page

### LINKED ISSUE #1284: Update the broadcast events for attempts and runs on History page
@elias-ba and @midigofrank point out that attempt and run events are not being handled properly by the new history page.

https://github.com/OpenFn/Lightning/pull/1281/commits/c413bd5c0ea7676f52914d7a0f3decca8252308f#diff-d83c666728f5ede80de9e0fa0c0576415368d56bac060201ef1452f2df53728aR118-R119

While we're at it, we can use this issue to track all the events we expect to handle:

1. new workorder should create a new row in the list (assuming it matches the current filters)
2. attempt:created should create a new attempt
3. attempt:updated should update an attempt
4. run:created should create a new run
5. run:updated should update a run

midigofrank:
## Notes for the reviewer

- Emits `Lightning.WorkOrders.Events.{WorkOrderCreated, WorkOrderUpdated, AttemptCreated, AttemptUpdated` events when the relevant resource is created or updated.  `Runs` also emit `AttemptUpdated`
- In order to test the broadcast in liveview, ~~I've had to `Process.sleep(2)`~~ to allow the liveview to process that event. @stuartc would like to get your thoughts on this. EDIT: I've found that calling `render(view)` also does the trick

## Related issue

Fixes #1284 

## Review checklist

### DIFF
--- a/lib/lightning/attempts/attempt.ex
+++ b/lib/lightning/attempts/attempt.ex
@@ -56,7 +56,10 @@ defmodule Lightning.Attempt do
     has_one :workflow, through: [:work_order, :workflow]
 
     has_many :log_lines, LogLine
-    many_to_many :runs, Run, join_through: AttemptRun
+
+    many_to_many :runs, Run,
+      join_through: AttemptRun,
+      preload_order: [asc: :finished_at]
 
     field :state, Ecto.Enum,
       values:
--- a/lib/lightning/attempts/attempt.ex
+++ b/lib/lightning/attempts/attempt.ex
@@ -59,7 +59,7 @@ defmodule Lightning.Attempt do
 
     many_to_many :runs, Run,
       join_through: AttemptRun,
-      preload_order: [asc: :finished_at]
+      preload_order: [asc: :started_at]
 
     field :state, Ecto.Enum,
       values:
--- a/lib/lightning/work_orders.ex
+++ b/lib/lightning/work_orders.ex
@@ -52,32 +52,32 @@ defmodule Lightning.WorkOrders do
   import Ecto.Query
   import Lightning.Validators
 
-  @pubsub Lightning.PubSub
-
   @type work_order_option ::
           {:workflow, Workflow.t()}
           | {:dataclip, Dataclip.t()}
           | {:created_by, User.t()}
 
-  # @doc """
-  # Create a new Workorder.
-  #
-  # **For a webhook**
-  #     create(trigger, workflow: workflow, dataclip: dataclip)
-  #
-  # **For a user**
-  #     create(job, workflow: workflow, dataclip: dataclip, user: user)
-  # """
+  @doc """
+  Create a new Workorder.
+
+  **For a webhook**
+      create_for(trigger, workflow: workflow, dataclip: dataclip)
+
+  **For a user**
+      create_for(job, workflow: workflow, dataclip: dataclip, user: user)
+  """
   @spec create_for(Trigger.t() | Job.t(), [work_order_option()]) ::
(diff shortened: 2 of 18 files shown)


taylordowns2000 (changes requested):
Looking great.

Small bug where the `attempt:complete` or `workorder:updated` message seems to be missed: https://www.loom.com/share/9c11359ce95649779a8d6b6c078aad47

lib/lightning/attempts/attempt.ex
```diff
@@ -56,7 +56,10 @@ defmodule Lightning.Attempt do
     has_one :workflow, through: [:work_order, :workflow]
 
     has_many :log_lines, LogLine
-    many_to_many :runs, Run, join_through: AttemptRun
+
+    many_to_many :runs, Run,
+      join_through: AttemptRun,
+      preload_order: [asc: :finished_at]
```
stuartc, line 62: Shouldn't this be `:started_at`? or perhaps `asc: :finished_at, asc: :started_at` because a fun can be either in progress or 'lost' where the finished at (afaik) isn't ever going to be set?

@taylordowns2000 care to comment?

lib/lightning/work_orders.ex
```diff
@@ -355,24 +387,4 @@ defmodule Lightning.WorkOrders do
     )
     |> Repo.one()
   end
-
-  def subscribe(project_id) do
-    Events.subscribe(project_id)
-  end
```
stuartc, line 361: Please put this back, I'd like to have callers having to reach into sub-modules for a context. 

You can also use `defdelegate` to forward the call to Events.

lib/lightning_web/live/run_live/workorder_component.ex
```diff
@@ -2,11 +2,32 @@ defmodule LightningWeb.RunLive.WorkOrderComponent do
...
+  #   Enum.map(assigns_sockets, fn {assigns, socket} ->
+  #     socket =
+  #       assign(socket, assigns)
+  #       |> assign(:work_order, updated_work_orders[assigns.id])
+
+  #     update(socket.assigns, socket)
+  #     |> then(fn {:ok, socket} -> socket end)
+  #   end)
+  # end
+
```
stuartc, line 30: Why is this no longer needed? Seems like an appropriate optimisation.

taylordowns2000 (changes requested):
noting the update to the component allows the live_view assigns to get "out of sync" and causes that funny "PENDING" bug. @midigofrank has (theoretically) solved this and will update

test/lightning_web/live/work_order_live_test.exs
```diff
@@ -900,6 +900,199 @@ defmodule LightningWeb.RunWorkOrderTest do
...
+        )
+
+      view |> element("#toggle_details_for_#{work_order.id}") |> render_click()
+
+      refute has_element?(view, "#attempt_#{attempt.id}")
+
+      Lightning.WorkOrders.Events.attempt_created(project.id, attempt)
+
+      # Force Re-render to ensure the event is included
+      render(view)
```
stuartc, line 952: :+1: yepper, the message is consumed in the LiveView process immediately but the view isn't rendered until you ask it to... under the hood things like `render_click|submit` call render as well.

stuartc (changes requested):
Nice @midigofrank, a few small comments but other than that this is great.

lib/lightning/attempts/attempt.ex
taylordowns2000: yes, great call @stuartc . @midigofrank , please make this `asc: :started_at`

PR #1343, opened 2023-11-09
New run & attempt complete messages (ws-worker@v0.1.9)

### LINKED ISSUE #1342: Handle new exit reasons, types, and attempt states
- Handle new `exit_reason` values
- Handle new `error_type` values
- Handle new `Attempt.State` values
- Handle new `Attempt.ErrorType` values
- Add xxx_at columns?
- Align with https://github.com/OpenFn/kit/pull/422

taylordowns2000:
## Read me:
1. This fixes #1342 and handles the new message structure for runs and attempts.
2. It can't be merged until `ws-worker@v0.1.9` is released - https://github.com/OpenFn/kit/pull/422
3. I have _not_ added `error_message`, but we could... see "TODO" in migration.
4. I want a stronger contract for the `{reason, error_type, error_message}` tuple we handle for completing runs and attempts. Right now, I'm pattern matching but it feels like I could be doing more.

## Review checklist


lib/lightning/attempts/attempt.ex
```diff
@@ -141,18 +145,26 @@ defmodule Lightning.Attempt do
...
         changeset
         |> change(finished_at: DateTime.utc_now())
+        |> change(error_type: error_type)
       else
         changeset
-        |> add_error(:state, "cannot complete attempt that is not started")
+        |> add_error(
+          :state,
+          "cannot complete attempt that is not started or has error"
+        )
```
jyeshe, line 167: "or has a previous error" would make it more clear that it's regardless of payload `error_type` value

josephjclark:
The new worker will actually be `0.2.0` to indicate the breaking change!

lib/lightning_web/live/run_live/components.ex
```diff
@@ -581,71 +548,56 @@ defmodule LightningWeb.RunLive.Components do
...
-  end
+  @spec run_icon(%{
+          :error_type => any(),
+          :reason => nil | <<_::32, _::_*8>>,
+          optional(any()) => any()
+        }) :: Phoenix.LiveView.Rendered.t()
+  # it's not really that complex!
+  # credo:disable-for-next-line
+  def run_icon(%{reason: reason, error_type: error_type} = assigns) do
+    [icon, classes] =
```
jyeshe, line 559: thank you for the mapping, makes it more understandable what the Heroicon need

jyeshe (commented):
It looks great! I am okay with well structured small tuples and couldn't find an issue or refactoring needed except for some wording. 

About the question on saving the `error_message` or not, would just try to narrow down the error messages to error types which is a string and as such seems to be flexible enough to describe variations of a reason.

lib/lightning/attempts/attempt.ex
```diff
@@ -20,10 +20,12 @@ defmodule Lightning.Attempt do
...
   @final_states [
     :success,
-    :cancelled,
     :failed,
+    :crashed,
+    :cancelled,
     :killed,
-    :crashed
+    :exception,
+    :lost
```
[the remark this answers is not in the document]
taylordowns2000: in this case, timeout is actually a `kill` because we know it took too long, we still have contact with the worker, and the worker shuts it down. lost is that we don't know what happened with the worker and it's now 20% longer than it should have timed out.

this is good fodder for a discussion next week at standup though :-)

PR #1345, opened 2023-11-10
Fix Incorrect Wording On The "Delete Job" Button Tooltip

### LINKED ISSUE #1313: Incorrect wording on the "Delete" tooltip
https://www.loom.com/share/bdf8e541b9684ce4a9f7fe80ba4cba93?sid=39d85564-2820-4509-a178-65e7792779de

Currently you can't delete the first job in the workflow (which is correct).
The delete button is disabled in this case, with correct wording.

You can delete the last jobs of an edge (which is correct).

You can't delete jobs that are in the middle of the workflow (not the first, nor the last jobs) (which is correct).
However, the tooltip when hovering on the button to delete it (which is correctly disabled) is incorrect since it's using the same wording as the tooltip when hovering on the disabled button of the first job.

Correct wording should be for any node that can't be deleted since it's in the middle of the workflow:
"This job cannot be deleted as it has downstream jobs associated with it."

elias-ba:
## Related issue

Fixes #1313 

## Review checklist


taylordowns2000 (changes requested):
1. please move changelog entry into the existing "changes" section. (it's below the breaking change warning)
2. this isn't working for me. see loom: https://www.loom.com/share/3a0a88d61f5b41038edca418e9ea10f3

stuartc (approved):

elias-ba:
@stuartc as discussed this morning, I have tried to reproduce the problem in tests but I couldn't do so. I think with this new approach we will no longer be dealing with that problem. We assign in the `apply_action` the count for the number of jobs in the workflow. Which is different than dealing with the URL params for the selected node. So I think you can merge this and we deal with the initial problem later.

stuartc (approved):

PR #1347, opened 2023-11-10
1254 remove work orders fk

rorymckinley:
## Notes for the reviewer

Removes FK blocking partitioning of `work_orders` and removes an unused method that can only correctly function if the FK is in place.

## Related issue

Fixes #

## Review checklist


rorymckinley:
@stuartc @jyeshe The test currently failing on this branch at the moment appears to be as the result of an unrelated timeout. I do not seem to have sufficient privileges to rerun that check so if one of you would be so kind as to rerun it, I would greatly appreciate it.

stuartc (approved):

PR #1370, opened 2023-11-13
End to end test with on_job_failure condition

### LINKED ISSUE #1143: RTM: Create End to End tests of Run Time Manager
Consider Dockerized Lightning and FLS in order to run integration tests to "do work"?

MVP:
- Porting the `end_to_end_test.exs` to work with the new RTM
- Ensure that this old runner test is covered _somehow_:

```ex
# todo - ensure that Rogerio tests this in https://github.com/OpenFn/Lightning/issues/1143
  test "scrub_result/1 removes :configuration from a map" do
    map = %{"data" => true, "configuration" => %{"secret" => "hello"}}
    assert Runner.scrub_result(map) == %{"data" => true}
  end
```

jyeshe:
## Notes for the reviewer

Validates the runs exit_reasons and log lines for a workflow execution involving a `on_job_failure` condition.

Additionally redacts and asserts the password contained in the output dataclip and in the log line.

## Related issue

Fixes #1143 

## Review checklist

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -34,7 +34,7 @@ config :lightning, LightningWeb.Endpoint,
   url: [host: "localhost", port: 4002],
   secret_key_base:
     "/8zedVJLxvmGGFoRExE3e870g7CGZZQ1Vq11A5MbQGPKOpK57MahVsPW6Wkkv61n",
-  server: false
+  server: true
 
 # In test we don't send emails.
 config :lightning, Lightning.Mailer, adapter: Swoosh.Adapters.Test
--- a/lib/lightning/invocation/dataclip.ex
+++ b/lib/lightning/invocation/dataclip.ex
@@ -51,8 +51,26 @@ defmodule Lightning.Invocation.Dataclip do
   def new(attrs \\ %{}) do
     change(%__MODULE__{id: Ecto.UUID.generate()}, attrs)
     |> change(attrs)
+    |> redact_password()
     |> validate()
   end
+
+  defp redact_password(
+         %Ecto.Changeset{
+           valid?: true,
+           changes: %{
+             body: %{"configuration" => %{"password" => _password}} = body
+           }
+         } =
+           changeset
+       ) do
+    body =
+      update_in(body, ["configuration", "password"], fn _any -> "***" end)
+
+    put_change(changeset, :body, body)
+  end
+
+  defp redact_password(changeset), do: changeset
 
   @doc false
   def changeset(dataclip, attrs) do
--- a/lib/lightning/invocation/dataclip.ex
+++ b/lib/lightning/invocation/dataclip.ex
@@ -59,13 +59,12 @@ defmodule Lightning.Invocation.Dataclip do
          %Ecto.Changeset{
            valid?: true,
            changes: %{
-             body: %{"configuration" => %{"password" => _password}} = body
+             body: body
            }
          } =
            changeset
        ) do
-    body =
-      update_in(body, ["configuration", "password"], fn _any -> "***" end)
+    body = Map.delete(body, "configuration")
 
     put_change(changeset, :body, body)
   end
--- a/lib/lightning/security.ex
+++ b/lib/lightning/security.ex
@@ -25,5 +25,5 @@ defmodule Lightning.Security do
     end
   end
 
-  def redact_password(changeset), do: changeset
+  def redact_password(changeset, _field), do: changeset
 end
--- a/lib/lightning/security.ex
+++ b/lib/lightning/security.ex
@@ -1,0 +1,29 @@
+defmodule Lightning.Security do
+  @moduledoc """
+  Security helpers.
+  """
+  alias Ecto.Changeset
+
+  @doc """
+  Redact a password inside a json on a changeset field.
+  If the field has a valid string value "password":"thesecret" becomes "password":"***".
+  """
+  @spec redact_password(Changeset.t(), atom()) :: Changeset.t()
+  def redact_password(%Changeset{valid?: true} = changeset, field) do
+    with {:ok, str_to_redact} <- Changeset.fetch_change(changeset, field),
+         true <- String.valid?(str_to_redact) do
+      redacted =
+        String.replace(
+          str_to_redact,
+          ~r/\"password\":\"\w+\"/,
+          "\\1\"password\":\"\*\*\*\""
+        )
+
+      Changeset.put_change(changeset, field, redacted)
+    else
+      _any -> changeset
+    end
+  end
+
+  def redact_password(changeset), do: changeset
+end
--- a/lib/lightning/security.ex
+++ b/lib/lightning/security.ex
@@ -11,7 +11,7 @@ defmodule Lightning.Security do
   @spec redact_password(Changeset.t(), atom()) :: Changeset.t()
   def redact_password(%Changeset{valid?: true} = changeset, field) do
     with {:ok, str_to_redact} <- Changeset.fetch_change(changeset, field),
-         true <- String.valid?(str_to_redact) do
+         true <- is_binary(str_to_redact) and String.valid?(str_to_redact) do
       redacted =
         String.replace(
           str_to_redact,
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -187,12 +187,6 @@ defmodule LightningWeb.AttemptChannel do
     end
   end
 
-  def handle_in("attempt:log", %{"message" => message}, socket)
-      when message == nil or message == [nil] do
-    {:reply, {:error, %{errors: %{message: ["This field can't be blank."]}}},
-     socket}
-  end
-
   def handle_in("attempt:log", payload, socket) do
     Attempts.append_attempt_log(socket.assigns.attempt, payload)
     |> case do
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -166,10 +166,15 @@ defmodule LightningWeb.AttemptChannel do
     end
   end
 
-  def handle_in("run:complete", payload, socket) do
+  def handle_in(
+        "run:complete",
+        %{"reason" => reason} = payload,
+        socket
+      ) do
     %{
       "attempt_id" => socket.assigns.attempt.id,
-      "project_id" => socket.assigns.project_id
+      "project_id" => socket.assigns.project_id,
+      "reason" => "#{map_rtm_reason_state(reason)}"
     }
     |> Enum.into(payload)
     |> Attempts.complete_run()
@@ -180,6 +185,12 @@ defmodule LightningWeb.AttemptChannel do
       {:ok, run} ->
         {:reply, {:ok, %{run_id: run.id}}, socket}
     end
+  end
+
+  def handle_in("attempt:log", %{"message" => message}, socket)
+      when message == nil or message == [nil] do
+    {:reply, {:error, %{errors: %{message: ["This field can't be blank."]}}},
+     socket}
   end
 
   def handle_in("attempt:log", payload, socket) do
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -48,14 +48,10 @@ defmodule LightningWeb.AttemptChannel do
 
   @impl true
   def handle_in("fetch:attempt", _, socket) do
-    # IO.write(:stderr, "fetch:attempt #{socket.assigns.attempt.id}\n")
-
     {:reply, {:ok, AttemptJson.render(socket.assigns.attempt)}, socket}
   end
 
   def handle_in("attempt:start", _, socket) do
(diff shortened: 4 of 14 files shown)


taylordowns2000 (changes requested):
This is looking great @jyeshe ! Would you please

- Update the ws-worker to v0.2.5 (rebase on main should do it)
- And add another job that runs parallel to the "flow job". Rename flow job to "path a" and call this new job "path b". Both should be triggered by the successful completion of the first job in the workflow.
- Then, please add a few assertions to make sure that the input data clip is correct for each of the runs generated in this attempt. This will help us close up another funky bug/maybe-bug

https://github.com/OpenFn/kit/issues/479

taylordowns2000 (changes requested):
code looking really good. test execution failing for me. two questions:
1. it fails on my machine at this line `{ps_out, 0} = System.cmd("ps", ["-s", "#{os_pid}", "-o", "cmd="])` - is that because i'm on a Mac M1/arm? (See below for output.)
2. it fails on [circle](https://app.circleci.com/pipelines/github/OpenFn/Lightning/5117/workflows/dd493ad8-b01e-44cf-b503-a5a6bf4b6c7c/jobs/26678/tests)... is that because we need to add some environment variables? or something else?

```
/bin/ps: illegal option -- s
usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
          [-g grp[,grp...]] [-u [uid,uid...]]
          [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
       ps [-L]
???

  0) LightningWeb.EndToEndTest: failure on setup_all callback, all tests have been invalidated
     ** (MatchError) no match of right hand side value: {"", 1}
     stacktrace:
       test/lightning_web/end_to_end_test.exs:312: anonymous fn/2 in LightningWeb.EndToEndTest.start_runtime_manager/1
       (elixir 1.15.4) lib/enum.ex:4204: Enum.predicate_range/5
       test/lightning_web/end_to_end_test.exs:305: LightningWeb.EndToEndTest.start_runtime_manager/1
       test/lightning_web/end_to_end_test.exs:2: LightningWeb.EndToEndTest.__ex_unit__/2
```

if not quite ready, can you mark this as "draft" until it's ready to go?

taylordowns2000:
> > code looking really good. test execution failing for me. two questions:
> > 
> > 1. it fails on my machine at this line `{ps_out, 0} = System.cmd("ps", ["-s", "#{os_pid}", "-o", "cmd="])` - is that because i'm on a Mac M1/arm? (See below for output.)
> > 2. it fails on [circle](https://app.circleci.com/pipelines/github/OpenFn/Lightning/5117/workflows/dd493ad8-b01e-44cf-b503-a5a6bf4b6c7c/jobs/26678/tests)... is that because we need to add some environment variables? or something else?
> > 
> > ```
> > /bin/ps: illegal option -- s
> > usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
> >           [-g grp[,grp...]] [-u [uid,uid...]]
> >           [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
> >        ps [-L]
> > ???
> > 
> >   0) LightningWeb.EndToEndTest: failure on setup_all callback, all tests have been invalidated
> >      ** (MatchError) no match of right hand side value: {"", 1}
> >      stacktrace:
> >        test/lightning_web/end_to_end_test.exs:312: anonymous fn/2 in LightningWeb.EndToEndTest.start_runtime_manager/1
> >        (elixir 1.15.4) lib/enum.ex:4204: Enum.predicate_range/5
> >        test/lightning_web/end_to_end_test.exs:305: LightningWeb.EndToEndTest.start_runtime_manager/1
> >        test/lightning_web/end_to_end_test.exs:2: LightningWeb.EndToEndTest.__ex_unit__/2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if not quite ready, can you mark this as "draft" until it's ready to go?
> 
> Hi @taylordowns2000 this PR has that failed assertion on the `input_dataclip` we discussed today. Should we wait for the RTM fix?

If you can get it passing on circleci (please **see my Qs above** about env variables and also about why it's failing on Mac?) without the dataclip_id assertion, please add a TODO next to that assertion and paste a link to https://github.com/OpenFn/kit/pull/484 - locking in green checks on this e2e test will help us prevent regressions in other areas.

config/test.exs
```diff
@@ -34,7 +34,7 @@ config :lightning, LightningWeb.Endpoint,
   url: [host: "localhost", port: 4002],
   secret_key_base:
     "/8zedVJLxvmGGFoRExE3e870g7CGZZQ1Vq11A5MbQGPKOpK57MahVsPW6Wkkv61n",
-  server: false
+  server: true
```
stuartc, line 37: I'm a little wary of enabling the server globally for testing.

This means that we would have the worker creating records without being in a test transaction right? We'd have to clean up before and after running the tests and they would have to be synchronous so we don't introduce flakiness.

What would be first prize is we start the server inside the test, and leverage the Ecto sandbox behaviour.

jyeshe: It's okay Stu, everything happens related to a workorder that is **not persisted** even created by a webhook with `server: true` and the sandbox tests are not affected by persisted records.

jyeshe:
Waiting for the fix on the worker for intermittent errors on `attempt:start` and `attempt:complete`:
```
%{"error_message" => "2a59c225-dbad-41ea-a710-edb23388545d: ENOENT: no such file or directory, open '/tmp/openfn/repo/node_modules/@openfn/language-common_1.6.2/package.json'", "error_type" => "CompileError", "final_dataclip_id" => "", "reason" => "crash"}
```

lib/lightning_web/channels/attempt_channel.ex
```diff
@@ -182,6 +187,12 @@ defmodule LightningWeb.AttemptChannel do
     end
   end
 
+  def handle_in("attempt:log", %{"message" => message}, socket)
+      when message == nil or message == [nil] do
+    {:reply, {:error, %{errors: %{message: ["This field can't be blank."]}}},
+     socket}
+  end
+
```
stuartc, line 195: I think we should be handling this in a changeset (and the ChangesetJSON module) like the rest of the `handle_in`s.

I'm curious about why this is needed - was it done before ff5f6f325231254511ba88a64a322d13fd27e247? That commit fixed coercing `nil` and `[nil]`. While it doesn't reject the message, it safely converts it to `null` - I don't think we want to stand in the way of controlling what is and isn't a valid message from the worker - we should either drop it or convert it. It's entirely possible a user tries to `console.log(a)` and a is `null`.

lib/lightning/invocation/dataclip.ex
```diff
@@ -51,9 +51,27 @@ defmodule Lightning.Invocation.Dataclip do
...
+             body: %{"configuration" => %{"password" => _password}} = body
+           }
+         } =
+           changeset
+       ) do
+    body =
+      update_in(body, ["configuration", "password"], fn _any -> "***" end)
+
+    put_change(changeset, :body, body)
+  end
```
stuartc, line 71: Is this enough? We shouldn't be receiving the configuration key, @josephjclark the worker removes `configuration` from the outgoing dataclip right?

Either way, we used to (we definitely did it in platform) drop the entire `configuration` key since there are a huge variety of configuration schemas and `password` is one of many keys that can be sensitive.

lib/lightning/security.ex
```diff
@@ -0,0 +1,29 @@
...
+          str_to_redact,
+          ~r/\"password\":\"\w+\"/,
+          "\\1\"password\":\"\*\*\*\""
+        )
+
+      Changeset.put_change(changeset, field, redacted)
+    else
+      _any -> changeset
+    end
+  end
```
stuartc, line 26: I missed the specification for this. while we definitely do want to scrub logs I think it's much more complicated that just looking for a password pattern - a bad actor could obfuscate this with little effort.

Take a look at `Lightning.Scrubber`, it takes a credential and looks for leaked values regardless of the key.

@taylordowns2000 care to add something here?

test/lightning_web/end_to_end_test.exs
```diff
@@ -1,173 +1,278 @@
...
+  alias Lightning.Attempt
+  alias Lightning.Attempts
+  alias Lightning.Attempts.Events
   alias Lightning.Invocation
+  alias Lightning.Repo
+  alias Lightning.WorkOrders
+  alias Lightning.Runtime.RuntimeManager
 
-  import Ecto.Query
+  require Attempt
```
stuartc, line 16: I think you meant `import Attempt`, but I'm guessing this wasn't supposed to be here.

test/lightning_web/end_to_end_test.exs
```diff
@@ -1,173 +1,278 @@
...
+        receive do
+          %Events.AttemptUpdated{attempt: %{state: state}}
+          when state in Attempt.final_states() ->
+            assert state == :success
+            true
+
+          _other_msg ->
+            false
+        end
+      end)
```
stuartc, line 61: How many messages can this process receive? In the failure case (where the `:success` state doesn't arrive), I guess the number of iterations is not relevant but `receive` waits indefinitely by default - looks like we would have to wait the entire test timeout for this. 

:man_shrugging: I guess it's fine... Can you think of ways this could be improved? If not then it's probably fine :sweat_smile:

test/lightning_web/end_to_end_test.exs
```diff
@@ -188,4 +293,38 @@ defmodule LightningWeb.EndToEndTest do
...
+  defp start_runtime_manager(_context) do
+    rtm_args = "node ./node_modules/.bin/worker -- --backoff 0.5/5"
+
+    Application.put_env(:lightning, RuntimeManager,
+      start: true,
+      args: String.split(rtm_args),
+      cd: Path.expand("../../assets", __DIR__)
+    )
+
+    {:ok, rtm_server} = RuntimeManager.start_link(name: E2ETestRuntimeManager)
```
stuartc, line 306: I think we can pass those options directly into `start_link/1`, the start args are merged into the Application env - if we do that and we want to start the worker elsewhere we won't have competing application envs.

test/lightning_web/end_to_end_test.exs
```diff
@@ -188,4 +293,38 @@ defmodule LightningWeb.EndToEndTest do
...
+
+    running =
+      Enum.any?(1..20, fn _i ->
+        Process.sleep(50)
+        %{runtime_port: port} = :sys.get_state(rtm_server)
```
stuartc, line 311: Should we not have a public interface for this?

stuartc (changes requested):
This looks fantastic @jyeshe! I left some comments about a few things.

lib/lightning_web/channels/attempt_channel.ex
jyeshe: I have got an error with message = [nil] on the payload, moved the validation to the changeset.

lib/lightning/invocation/dataclip.ex
jyeshe: For the same scope of previous assertion it would be enough. Even if it's removed on the worker I propose keep it here and improve it later with your scrubber.

lib/lightning/security.ex
jyeshe: There is an assertion needed for the log lines to check if any password is redacted. Please let me know if we wanna increase the scope.

test/lightning_web/end_to_end_test.exs
jyeshe: It's okay, it's for the macro to reuse the same list

test/lightning_web/end_to_end_test.exs
jyeshe: There are multiple messages, any 100 is just a big number to make it simple since it's a test setup only. I could use a stream but this is simpler and still readable.

test/lightning_web/end_to_end_test.exs
jyeshe: true, previously the other args were ignored with _args.

test/lightning_web/end_to_end_test.exs
jyeshe: I think it's fine for now, just for the test setup.

test/lightning_web/end_to_end_test.exs
jyeshe: has not worked though [cdf4773](https://github.com/OpenFn/Lightning/pull/1370/commits/cdf4773d0473cb23c2adb95f41db34d462624215) 
reverted on [c476be9](https://github.com/OpenFn/Lightning/pull/1370/commits/c476be939ea04ce95a54cae49fb91a96eeed55b8)

lib/lightning_web/channels/attempt_channel.ex
stuartc: Ah ok, since you rebased the other day this is no longer required (i.e. in the changeset validation as well).

The `LogMessage` Ecto type handles coercing from `nil|[nil]` into `"null"`.

stuartc (approved):

PR #1371, opened 2023-11-13
make-janitor-also-clean-up-runs-inside-an-attempt

### LINKED ISSUE #1348: Make janitor also clean up runs inside an attempt
- Right now runs look like they're in progress when they've really been lost.
- The janitor already marks attempts as lost here: https://github.com/OpenFn/Lightning/blob/1342-error-types/lib/lightning/janitor.ex#L43
- It **should** also mark any _unfinished_ runs inside its target attempt as lost.

`change(%{exit_reason: "lost"})`

Also maybe add logs...?

benedictus-dev:
## Notes for the reviewer
I've added a function mark_unfinished_runs_lost in the Lightning.Attempts module. This function is designed to check all runs associated with a 'lost' attempt and mark any unfinished ones as lost too

## Related issue

Fixes #1348 

## Review checklist

### DIFF
--- a/lib/lightning/attempts.ex
+++ b/lib/lightning/attempts.ex
@@ -18,9 +18,12 @@ defmodule Lightning.Attempts do
                 {:ok, Lightning.Attempt.t()}
   end
 
+  require Logger
+
   alias Lightning.Attempt
   alias Lightning.Attempts.Events
   alias Lightning.Attempts.Handlers
+
   alias Lightning.Repo
 
   import Ecto.Query
@@ -107,6 +110,8 @@ defmodule Lightning.Attempts do
     |> update_attempt()
   end
 
+  @spec complete_attempt(Attempt.t(), {atom(), binary() | nil, binary() | nil}) ::
+          {:ok, Attempt.t()} | {:error, Ecto.Changeset.t(Attempt.t())}
   def complete_attempt(attempt, {new_state, error_type, error_message}) do
     Attempt.complete(attempt, {new_state, error_type, error_message})
     |> case do
@@ -118,11 +123,8 @@ defmodule Lightning.Attempts do
     end
   end
 
-  # TODO - Implement this in https://github.com/OpenFn/Lightning/issues/1348
-  # def mark_unfinished_runs_lost(attempt) do
-  # for each run in this attempt call `complete_run` with exit_reason: "lost"
-  # end
-
+  @spec update_attempt(Ecto.Changeset.t(Attempt.t())) ::
+          {:ok, Attempt.t()} | {:error, Ecto.Changeset.t(Attempt.t())}
   def update_attempt(%Ecto.Changeset{data: %Attempt{}} = changeset) do
     attempt_id = Ecto.Changeset.get_field(changeset, :id)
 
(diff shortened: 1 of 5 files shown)


taylordowns2000 (approved):
this is great. thanks @benedictus-dev !

these defp's, i'd like to ask @stuartc or @jyeshe for their opinions... it's quite clean but i sometimes get worried about "over-factored" code (refactored beyond the point of usefulness). this is a concept that Stu introduced me to, and i'm always trying to stay on the lookout.

Stu, what do you think about this lines? Is this the right level of refactoring?

lib/lightning/attempts.ex
```diff
@@ -232,4 +233,30 @@ defmodule Lightning.Attempts do
...
+    end)
+  end
+
+  defp run_not_finished?(%{finished_at: nil, exit_reason: nil}), do: true
+
+  defp run_not_finished?(_), do: false
+
+  defp mark_run_as_lost(run) do
+    updated_run = Ecto.Changeset.change(run, exit_reason: "lost")
+    Repo.update!(updated_run)
```
jyeshe, line 260: We can consider it's finished after this update like the attempt by updating the `finished_at`

lib/lightning/attempts.ex
```diff
@@ -232,4 +233,30 @@ defmodule Lightning.Attempts do
...
   end
+
+  def mark_unfinished_runs_lost(%Attempt{id: attempt_id}) do
+    # Fetch all runs associated with the attempt
+    runs =
+      Repo.all(
+        from ar in Lightning.AttemptRun,
+          where: ar.attempt_id == ^attempt_id,
+          preload: [:run]
+      )
```
jyeshe, line 244: This works fine but would use a `join` to avoid a second query:
```
{"SELECT a0.\"id\", a0.\"attempt_id\", a0.\"run_id\", a0.\"inserted_at\" FROM \"attempt_runs\" AS a0 WHERE (a0.\"attempt_id\" = $1)"

and since the database already iterates through the runs, we can filter for the runs where the `exit_reason` is `nil`
```

jyeshe (changes requested):
@benedictus-dev this was really fast! Would tweak it a bit for stronger consistency.

lib/lightning/attempts.ex
jyeshe: Actually maybe better not, right @stuartc? The `finished_at` could be just for the record of the update but `updated_at` would be enough and it keeps some sign/mark that it has never reported back to Lightning.

stuartc: Hmm, I don't really have a clear opinion about using `finished_at` when using `:lost`. Given what I said this morning about it, I would agree with you and that we don't know that it's not finished...

However in order to definitively tell how long it took to become lost, using `finished_at` is more reliable than `updated_at` since Ecto will set that at any point and we need to be more vigilant about making sure we don't alter the record.

So given that risk, I'd say we should use `finished_at` for now when marking as lost.

@benedictus-dev could you move the changeset code for `mark_run_as_lost` to the `Attempt` model. There you will find a bunch of changeset functions that do this kind of stuff and _additionally_ perform state transition validation.

Also please avoid bang functions (`update!`) without considering all the places these functions could be called - if there is no specific reason please let it return an error rather than throwing an exception.

stuartc:
> this is great. thanks @benedictus-dev !
> 
> these defp's, i'd like to ask @stuartc or @jyeshe for their opinions... it's quite clean but i sometimes get worried about "over-factored" code (refactored beyond the point of usefulness). this is a concept that Stu introduced me to, and i'm always trying to stay on the lookout.
> 
> Stu, what do you think about this lines? Is this the right level of refactoring?

I agree, those specific defps are begging to be inside a case statement and/or a reduce call - or even in another module.

Like my comment about moving the changeset code to the Attempt model. The contexts are (at least _trying_) to be a thin layer between callers and the actual implementations.

jyeshe (approved):

taylordowns2000 (changes requested):
hey @benedictus-dev , looking better but please have a quick chat with the team tomorrow at standup about this. I think we may have got our wires crossed a tiny bit. The functions themselves look good and clean, but the naming and placement just got super weird. We shouldn't have an undocumented public function in the `Attempt` module called "mark_as_lost" that takes a _run_! (And does nothing to attempts!) I think it would be acceptable if it lived in the `Run` module, but I'm 95% sure this was just a miscommunication or a typo and I'm sure you guys can clear it up with a chat tomorrow 😄 

image

stuartc (approved):

PR #1392, opened 2023-11-15
Partition work_orders

rorymckinley:
## Notes for the reviewer

I tried to use as much of the existing functionality from `platform` as possible, to reduce the risk (given that w do add and drop tables). The SQL query that was part of `PartitionTableService.remove_empty` started misbehaving yesterday and I could not get to the bottom of it in a reasonable amount of time, so I decided to extract some of the logic into Elixir.

The new methods to support this were added as public methods - I felt that the improved granularity of testing outweighed the polluting of the public API :).

## Related issue

Fixes #

## Review checklist


lib/lightning/maintenance/partition_table_service.ex
```diff
@@ -0,0 +1,185 @@
...
+        |> Regex.run(range_expression)
+
+      [table, NaiveDateTime.from_iso8601!(to_as_string)]
+    end)
+    |> Enum.filter(fn [_table, to] -> to < bound end)
+    |> Enum.map(fn [table, _to] -> table end)
+  end
+
+  def drop_partition(parent, partition) do
+    unless valid_chars?(parent) && valid_chars?(partition) do
```
rorymckinley, line 135: String interpolation in queries makes me nervous :)- so I added some simple validation against the more obvious forms of injection (as I know that we can't bind params in DDL statements) just in case this method ends up being used someplace else.

lib/lightning/maintenance/partition_table_service.ex
```diff
@@ -0,0 +1,182 @@
...
+  require Logger
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"add_headroom" => %{"weeks" => weeks}}})
+      when is_integer(weeks) do
+    add_headroom(:all, weeks)
+  end
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"drop_older_than" => %{"weeks" => weeks}}})
```
rorymckinley, line 23: I could not get this method to work as originally implemented in platform so I modelled the arguments on the other `perform` variant.

test/lightning/maintenance/partition_table_service_test.exs
```diff
@@ -0,0 +1,495 @@
...
+          "#{parent}_default",
+          "#{parent}_monolith"
+        ]
+        |> Enum.sort()
+
+      assert associated_relations(all_relations(), parent) == expected
+    end
+  end
+
+  defp all_relations() do
```
rorymckinley, line 351: I ended up with a fair number of functions to manage the fixtures - so of which have overlaps. If the overlaps are resolved I am hoping that I could reduce the number of fixture methods.

However, I will need to extend (at least) the scenarios that cover `:all` use case as I partition the other tables - as such I was hoping to delay the cleanup until those scenarios have been catered for.

lib/lightning/maintenance/admin_tools.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Lightning.AdminTools do
+  def generate_iso_weeks(start_date, end_date) do
+    Date.range(start_date, end_date)
```
rorymckinley, line 3: When I was setting up tests for this class, I noticed that there was a small edge case that emerged when dates that are not mondays are passed in as arguments. I can't imagine that it was a big deal before, but the tweak to fix it was small, so I made an executive decision :).

test/lightning/maintenance/partition_table_service_test.exs
```diff
@@ -0,0 +1,495 @@
...
+      "#{parent}_2023_02",
+      "#{parent}_2023_03",
+      "#{parent}_default"
+    ]
+
+    assert Service.get_partitions("work_orders") |> Enum.sort() == expected
+  end
+
+  test "tables_to_add returns tables that do not already exist" do
+    now = DateTime.now!("Etc/UTC")
```
stuartc, line 78: `DateTime.utc_now()` :wink:

test/lightning/maintenance/partition_table_service_test.exs
stuartc: Thats cool, if the functions end up living in this test file thats all good imo.

lib/lightning/maintenance/partition_table_service.ex
stuartc: Ah interesting, this looks good to me though :)

lib/lightning/maintenance/partition_table_service.ex
```diff
@@ -0,0 +1,182 @@
...
+      when is_integer(weeks) do
+    upper_bound = Timex.shift(DateTime.utc_now(), weeks: weeks)
```
stuartc, line 25: Could you try using `DateTime.shift` instead of Timex - nothing against what you've done here - Timex was the gold standard for all things Date but the built in DateTime has come a long way since then. And we have had issues with Timex in compiled environments (like without having the entire app running, like in migrations on production); if you can achieve a passing test suite by just replacing the use of Timex then we should be good.

stuartc (commented):
Nice dude! This is awesome, I'll be wanting to test this against a backup of a real db before merging.

Are the migrations idempotent, i.e. if it borks out half way or we have to cancel it (on the unlikely case it takes too long) - will it leave behind a bunch of schemas that must be deleted before running it again? Not saying we _have_ to cater for this right now; but I'd like to know more about the expected sad path.

lib/lightning/maintenance/partition_table_service.ex
rorymckinley: @stuartc I have expunged all the uses of `Timex` from prod code, with one [exception](https://github.com/OpenFn/Lightning/pull/1392/files#diff-8fde6fc40ea0bdb0e15ecfaa85a407a45cbc6db9f74137f45d64bc7b8bfc1609R7) - I can't find an equivalent to `iso_week` on either `Date` or `DateTime` - if you think this is going to be problematic, I can try and handroll an equivalent?

rorymckinley:
This PR is being parked, pending a decision on the further partitioning of tables.

Amongst other things, this branch contains additional tests and small code tweaks (when compared to the original code) that may be useful when doing date-based partitioning.

rorymckinley:
This code needs to still be tested against a database with a significant number of records in the `work_orders` table to gauge performance.

taylordowns2000:
@rorymckinley , @stuartc , can i close this?

brandonjackson:
Closing due to inactivity @rorymckinley - please reopen if this is in error

PR #1393, opened 2023-11-15
1380 function clause error changeset json error

### LINKED ISSUE #1380: FunctionClauseError: no function clause matching in LightningWeb.ChangesetJSON.error/1
Sentry Issue: [LIGHTNING-7N](https://openfn.sentry.io/issues/4632139592/?referrer=github_integration)

```
FunctionClauseError: no function clause matching in LightningWeb.ChangesetJSON.error/1
  File "lib/lightning_web/controllers/changeset_json.ex", line 6, in LightningWeb.ChangesetJSON.error/1
  File "lib/lightning_web/channels/attempt_channel.ex", line 178, in LightningWeb.AttemptChannel.handle_in/3
  File "lib/phoenix/channel/server.ex", line 322, in Phoenix.Channel.Server.handle_info/2
  File "gen_server.erl", line 1077, in :gen_server.try_handle_info/3
  File "gen_server.erl", line 1165, in :gen_server.handle_msg/6
...
(1 additional frame(s) were not displayed)

(FunctionClauseError no function clause matching in LightningWeb.ChangesetJSON.error/1)
```

benedictus-dev:
## Notes for the reviewer
Update the CompleteRun handler in Lightning.Attempts.Handlers to handle
cases where a run is not found by returning a detailed error changeset.

## Related issue

Fixes #1380 

## Review checklist

### DIFF
--- a/lib/lightning/attempts/handlers.ex
+++ b/lib/lightning/attempts/handlers.ex
@@ -169,10 +169,16 @@ defmodule Lightning.Attempts.Handlers do
 
     defp update(complete_run) do
       Repo.transact(fn ->
-        with run = %Run{} <-
-               get_run(complete_run.run_id) || {:error, :not_found},
-             {:ok, _} <- to_dataclip(complete_run) |> Repo.insert() do
-          update_run(run, complete_run)
+        case get_run(complete_run.run_id) do
+          nil ->
+            # Return an error changeset when the run is not found
+            change(%Run{})
+            |> add_error(:run, "Run not found")
+
+          run ->
+            with {:ok, _} <- to_dataclip(complete_run) |> Repo.insert() do
+              update_run(run, complete_run)
+            end
         end
       end)
     end
--- a/lib/lightning/attempts/handlers.ex
+++ b/lib/lightning/attempts/handlers.ex
@@ -172,8 +172,10 @@ defmodule Lightning.Attempts.Handlers do
         case get_run(complete_run.run_id) do
           nil ->
             # Return an error changeset when the run is not found
-            change(%Run{})
-            |> add_error(:run, "Run not found")
+            %Run{}
+            |> change()
+            |> put_change(:exit_reason, "error")
+            |> put_change(:error_type, "Run not found")
 
           run ->
             with {:ok, _} <- to_dataclip(complete_run) |> Repo.insert() do


lib/lightning/attempts/handlers.ex
```diff
@@ -169,10 +169,18 @@ defmodule Lightning.Attempts.Handlers do
...
-               get_run(complete_run.run_id) || {:error, :not_found},
-             {:ok, _} <- to_dataclip(complete_run) |> Repo.insert() do
-          update_run(run, complete_run)
+        case get_run(complete_run.run_id) do
+          nil ->
+            # Return an error changeset when the run is not found
+            %Run{}
+            |> change()
+            |> put_change(:exit_reason, "error")
+            |> put_change(:error_type, "Run not found")
```
stuartc, line 178: I'm not sure this is the way to go about it. In the case of an error we don't return a Run changeset, we return a `CompleteRun` changeset, it's purpose is to stand in between the worker sending a bunch of data that represents data that is for a run and a dataclip. It can be thought of as a 'command'.

My suggestion is to add a `run: Run` field to the schema, we don't cast it - but set the field and then validate it.

```elixir
  # ... ^ new/1
  |> then(&assign_run/1)
end

defp assign_run(changeset) do
  with %{valid?: true} <- changeset,
        %Run{} = run <-
          get_field(changeset, :run_id) |> get_run() || :not_found do
    put_change(changeset, :run, run)
  else
    :not_found ->
      add_error(changeset, :run_id, "not found")

    changeset ->
      changeset
  end
end
```

And then further down the `update/1` function can assume the changeset is valid, and get the `run` off the changeset instead of finding it itself.

stuartc (changes requested):

lib/lightning/attempts/handlers.ex
benedictus-dev: The proposal is to be very granular with the fix changing only the flow of the complete run.
The change consists of just adding the changeset to the return value.

jyeshe (commented):
Just to keep the previous order of validations and I think this change is safer, going directly to the point by adding the validation only to the flow reported on the error (on `new/insertion` the validation won't be needed)

lib/lightning/attempts/handlers.ex
```diff
@@ -160,19 +160,25 @@ defmodule Lightning.Attempts.Handlers do
...
-               get_run(complete_run.run_id) || {:error, :not_found},
+        with %Run{} = run <- get_run(complete_run.run_id),
              {:ok, _} <- to_dataclip(complete_run) |> Repo.insert() do
           update_run(run, complete_run)
+        else
+          nil ->
+            {:error,
+             %Run{}
+             |> change()
+             |> add_error(:run_id, "not found")}
```
stuartc, line 178: Please don't return a Run changeset. As mentioned in my first review:

> In the case of an error we don't return a Run changeset, we return a CompleteRun changeset

The Run model does not have a `:run_id` key. You have access to the original `CompleteRun` changeset. Please change the `nil` case to:

```elixir
{:error, complete_run |> add_error(:run_id, "not found")}
```

stuartc (changes requested):

PR #1398, opened 2023-11-17
fix: StartRun handler throws an exception when the attempt doesn't exist

### LINKED ISSUE #1390: StartRun handler throws an exception when either the job or attempt doesn't exist
This has only occurred once, but I think so effort should be put into making the query in `StartRun.validate_job_reachable/1` more robust.

I suggest testing with different scenarios where the attempt and/or the job might not exist - or that the workflow for a given job is not associated with this attempt.

But `Repo.one()` may return `nil` for whatever reasons, so we should return an appropriately invalid changeset.

Sentry Issue: [LIGHTNING-7J](https://openfn.sentry.io/issues/4624930927/?referrer=github_integration)

```
Protocol.UndefinedError: protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Phoenix.LiveView.LiveStream, Postgrex.Stream, Range, Scrivener.Page, Stream, Timex.Interval
  File "lib/enum.ex", line 1, in Enumerable.impl_for!/1
  File "lib/enum.ex", line 166, in Enumerable.reduce/3
  File "lib/enum.ex", line 4387, in Enum.reduce/3
  File "lib/lightning/attempts/handlers.ex", line 47, in Lightning.Attempts.Handlers.StartRun.call/1
  File "lib/lightning_web/channels/attempt_channel.ex", line 150, in LightningWeb.AttemptChannel.handle_in/3
...

(Protocol.UndefinedError protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Phoenix.LiveView.LiveStream, Postgrex.Stream, Range, Scrivener.Page, Stream, Timex.Interval)
```

midigofrank:
## Notes for the reviewer
Was able to reproduce this by a non existent attempt

## Related issue

Fixes #1390 

## Review checklist

### DIFF
--- a/lib/lightning/attempts/handlers.ex
+++ b/lib/lightning/attempts/handlers.ex
@@ -99,14 +99,8 @@ defmodule Lightning.Attempts.Handlers do
           # Verify that all of the required entities exist with a single query,
           # then reduce the results into a single changeset by adding errors for
           # any columns/ids that are null.
-          from(a in Attempt,
-            where: a.id == ^attempt_id,
-            left_join: w in assoc(a, :workflow),
-            left_join: j in assoc(w, :jobs),
-            on: j.id == ^job_id,
-            select: %{attempt_id: a.id, job_id: j.id}
-          )
-          |> Repo.one()
+          attempt_id
+          |> fetch_existing_job(job_id)
           |> Enum.reduce(changeset, fn {k, v}, changeset ->
             if is_nil(v) do
               add_error(changeset, k, "does not exist")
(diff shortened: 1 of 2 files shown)


taylordowns2000 (changes requested):
Hey @midigofrank , I can't quite tell if these changes will make the error go away (that would be bad!) or make the _problem_ go away (that would be good!)

Why does this part of the code get called when an attempt doesn't exist? That seems bad. We wouldn't want to close this issue or make this error disappear until we've understood and solved that.

Apologies if I'm misunderstanding the PR!

stuartc (approved):

stuartc:
@taylordowns2000 this was from sentry, it _should_ never happen but it did. My vibe with this was to reject the message instead of blowing up.

PR #1406, opened 2023-11-20
940 workflow diagram order children consistently

### LINKED ISSUE #940: Workflow Diagram: order children consistently
A small issue that I don't really want to deal with now, but I'm logging to remember it.

When refreshing the page, the order of child nodes can vary. It essentially seems to be randomised in the layout. It's confusing to refresh the page and see the children move.

It doesn't always do it. It may only do it when the node name changes [citation needed]. But it is annoying and a bit unsettling when it happens.

benedictus-dev:
## Notes for the reviewer
1, I have updated the timestamp precision in the Job schema to **:naive_datetime_usec** .
2, This was necessary to ensure a stable order of child nodes in the workflow diagram, as the default second-level precision was not sufficient for jobs created or updated close together in time.
3,I ran some tests, and things look good on my end – the nodes stay put now. But, you know, another pair of eyes is always great.

## Related issue
This PR addresses the issue where the order of child nodes in the workflow diagram varied randomly upon page refresh, especially after performing update operations on the node names. This inconsistency in ordering was particularly noticeable and problematic when the names of the nodes were changed.

Fixes #940 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -481,7 +481,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
           |> Lightning.Repo.preload([
             :edges,
             triggers: Trigger.with_auth_methods_query(),
-            jobs: Workflows.jobs_ordered_subquery()
+            jobs: {Workflows.jobs_ordered_subquery(), [:credential]}
           ])
 
         socket |> assign_workflow(workflow) |> assign(page_title: workflow.name)
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -427,8 +427,6 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
   @impl true
   def mount(_params, _session, socket) do
-    IO.inspect(socket, label: "Testing Edit")
-
     {:ok,
      socket
      |> authorize()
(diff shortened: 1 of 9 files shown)


josephjclark:
I can't comment on the implementation but this looks like a good solution to me!

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -480,7 +480,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
           |> Lightning.Repo.preload([
             :edges,
             triggers: Trigger.with_auth_methods_query(),
-            jobs: [:credential]
+            jobs: Workflows.jobs_ordered_subquery()
```
stuartc, line 483: I don't think we should be moving the preload of the Jobs credential into a function that given it's name says it's just for ordering.

Anyone using that function may or may not expect credentials to be preloaded.

I would suggest one of the following:

1) Setting the `:preload_order` on the `has_many` of the Workflow model. See [here](https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-options)
2) Extend the returned query from `jobs_ordered_subquery` with a `preload(:credential)`, that will keep the credential preload out of that function.
3) Don't add a function, and have it all inline here. And we refactor/abstract these queries later.

stuartc (changes requested):
Functionally looks good! I see some merge conflicts, please rebase and fix them; and I've left a comment on the `jobs_ordered_subquery` function.

stuartc (approved):
Nice! Really like the `{query, [:more_preload]}` very clean.

PR #1414, opened 2023-11-21
Extra coverage for AttemptRun and Logline deletion

rorymckinley:
## Notes for the reviewer

A future PR is going to remove FK relationships that link `attempts` to `attempt_runs` and `log_lines`. These relationships are `ON DELETE CASCADE` so removing them raises the risk of records being orphaned. 

This PR provides additional coverage for cases where the code may have been relying on the FK relationship to do the deletion. 

UPDATE: While testing the effects of removing the FKs, I could not find any production code that actually consumes `Jobs.delete_job/1` and so this has just been removed.

## Related issue

Fixes #

## Review checklist


stuartc (commented):
Thanks @rorymckinley this looks good - make two small comments.
Thanks for spotting the disuse of `delete_job`, we delete jobs via workflow.

test/support/factories.ex
```diff
@@ -77,23 +77,45 @@ defmodule Lightning.Factories do
...
+  def attempt_with_dependencies_factory do
+    struct!(
+      attempt_factory(),
+      %{
+        created_by: build(:user),
+        work_order: build(:workorder),
+        dataclip: build(:dataclip),
+        starting_job: build(:job)
+      }
+    )
```
stuartc, line 101: I think it's preferable to use the `build` function to a create a new factory (from another one). While in this case it probably has exactly the same result but there may be other behavior we may be missing out on.

Consider writing this like:

```elixir
  def attempt_with_dependencies_factory do
    build(:attempt, 
        created_by: build(:user),
        work_order: build(:workorder),
        dataclip: build(:dataclip),
        starting_job: build(:job)
    )
  end
```

But tbh, I generally err on the side of avoiding adding variations that don't describe it's purpose - like could these fields (maybe no `created_by`) be put into the attempt factory? Or perhaps the repetition you're trying to avoid can be moved inside the test?

Would appreciate your thoughts, I'm just weary of new fixtures/factories that end up possibly living forever..

test/support/data_case.ex
```diff
@@ -57,4 +57,14 @@ defmodule Lightning.DataCase do
...
+
+  @doc """
+  Checks that the provided record is the only record of that type persisted in
+  the database.
+  """
+  def only_record_for_type?(expected_instance) do
+    %type{} = expected_instance
+
+    Lightning.Repo.all(type) |> Enum.map(& &1.id) == [expected_instance.id]
+  end
```
stuartc, line 69: Perhaps we put this in ModelHelpers? (Could say the same for the `errors_on` helper above), but yeah don't feel too strongly about it?

test/support/factories.ex
rorymckinley: @stuartc Thanks for the feedback! I originally had them in the individual tests, but then I ended up repeating the same helper function across 3 tests (`accounts`, `jobs` and `attempts/queue`) which is why I thought to it made sense to have them in a single place and I thought the factories may be the place to go. Happy to reverse tack, especially seeing as there is one less implementation now that I have removed the changes to `jobs_test.ex`.

My preference would be to have a factory that is 'batteries included`  along the lines of 'if you call this it will always produce something even if you have to modify that something to get what you want' but that is religion not science and I hear you re: factory sprawl - will ponder how strong my religion is :).

As for the use of `struct!` - I got that from the docs: https://hexdocs.pm/ex_machina/readme.html#overview - so I am happy to switch to exemplar you provided (I don't have a horse in this race)...although the example from the doc does seem DRYer?

stuartc: This is a tough question, you're right - we don't want factory sprawl; but we also don't want factories that are added by default to the main factories for one for a handful of cases that end up (not in this case) getting so dense people just make new ones in order to avoid breaking/understanding whats already there.

I'm happy to keep working on this, like we try and add a few more 'scenario' based factories - you can see I've played with the idea of extending factories (`with_[job|trigger|edge]`) forcing the scenarios to live in the tests but the factory implementation to live in the factories module.

Hmm interesting! Yeah I see the ex_machina library, yeah then whatever! :man_shrugging: hehe.

PR #1433, opened 2023-11-22
Use config port for test env

jyeshe:
## Notes for the reviewer

For test env, instead of using the PORT env var, it uses the config (port 4002).

## Related issue

If there is a Lightning instance running it gives an `eaddrinuse` on 4000 when running the tests.

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -183,15 +183,22 @@ url_scheme = System.get_env("URL_SCHEME", "https")
 
 # The webserver port will always prefer and environment variable _when_
 # given, otherwise it uses the existing config and lastly defaults to 4000.
+
 port =
-  (System.get_env("PORT") ||
-     Application.get_env(:lightning, LightningWeb.Endpoint)
-     |> Keyword.get(:http, port: nil)
-     |> Keyword.get(:port) ||
-     4000)
-  |> case do
-    p when is_binary(p) -> String.to_integer(p)
-    p when is_integer(p) -> p
+  case config_env() do
+    :test ->
+      Application.get_env(:lightning, LightningWeb.Endpoint)[:url][:port]
+
+    _dev_prod ->
+      (System.get_env("PORT") ||
+         Application.get_env(:lightning, LightningWeb.Endpoint)
+         |> Keyword.get(:http, port: nil)
+         |> Keyword.get(:port) ||
+         4000)
+      |> case do
+        p when is_binary(p) -> String.to_integer(p)
+        p when is_integer(p) -> p
+      end
   end
 
 # Use the `PRIMARY_ENCRYPTION_KEY` env variable if available, else fall back
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -33,7 +33,7 @@
         "zustand": "^4.3.7"
       },
       "devDependencies": {
-        "@openfn/ws-worker": "^0.2.9",
+        "@openfn/ws-worker": "^0.2.11",
         "@types/marked": "^4.0.8",
         "@types/react": "^18.0.15",
         "@types/react-dom": "^18.0.6",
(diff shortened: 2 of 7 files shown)


config/runtime.exs
```diff
@@ -183,15 +183,22 @@ url_scheme = System.get_env("URL_SCHEME", "https")
...
-     |> Keyword.get(:port) ||
-     4000)
-  |> case do
-    p when is_binary(p) -> String.to_integer(p)
-    p when is_integer(p) -> p
+  case config_env() do
+    :test ->
+      Application.get_env(:lightning, LightningWeb.Endpoint)[:url][:port]
+
+    _dev_prod ->
```
stuartc, line 192: I've noticed that in `test.exs` we set just `url` and in prod (with runtime) we have a mix of both `url` and `http`. 

Is there way we can conform both test and prod/runtime approaches?

In this case I'd like to avoid making a distinction between test, dev and prod in `runtime.exs` - since they all use the endpoint, could we not tweak `test.exs` to have matching configuration so that runtime doesn't have to make the distinction?

stuartc: If I change `test.exs:35` to have:

```
  http: [port: 4002],
```

Everything works. I think the issue here is only setting `:url` in `test.exs`. 

We do need to be careful though, since in production environments it's likely the URL and the HTTP settings will be different since the listen port will be different to the url one.

But for now does the one line change to `test.exs` make sense?

jyeshe: > I've noticed that in `test.exs` we set just `url` and in prod (with runtime) we have a mix of both `url` and `http`.
> 
> Is there way we can conform both test and prod/runtime approaches?
> 
> In this case I'd like to avoid making a distinction between test, dev and prod in `runtime.exs` - since they all use the endpoint, could we not tweak `test.exs` to have matching configuration so that runtime doesn't have to make the distinction?

The reason for the case is due to the order/priority of the source of the port config. On test it should come from `Application` config while on dev and prod it should consider the `PORT` env firstly.

jyeshe: > If I change `test.exs:35` to have:
> 
> ```
>   http: [port: 4002],
> ```
> 
> Everything works. I think the issue here is only setting `:url` in `test.exs`.
> 
> We do need to be careful though, since in production environments it's likely the URL and the HTTP settings will be different since the listen port will be different to the url one.
> 
> But for now does the one line change to `test.exs` make sense?

The issue is that the worker does not connect to 4002. With [97f7f82](https://github.com/OpenFn/Lightning/pull/1433/commits/97f7f82677afa5d00ac39079afc204074a9a123d) it listens to 4002 and the test results in timeout.

Opened an issue https://github.com/OpenFn/Lightning/issues/1435 to be able to start the worker in test mode and continue this PR.

jyeshe: Hey @stuartc may we merge this, it was needed to test your PR (in order to get the `port` from the config on MIX_ENV=test, not from the env).

stuartc:
@jyeshe between 9077ecd7c68778b5a47d8ae257c3e627802ca0f5 and 755cc07b7b22a5176f5f080825088aa5960e153b I believe this is now solved - could you pull from main and let me know if you still have issues?

jyeshe:
> @jyeshe between [9077ecd](https://github.com/OpenFn/Lightning/commit/9077ecd7c68778b5a47d8ae257c3e627802ca0f5) and [755cc07](https://github.com/OpenFn/Lightning/commit/755cc07b7b22a5176f5f080825088aa5960e153b) I believe this is now solved - could you pull from main and let me know if you still have issues?

@stuartc the result is:
14:53:33.381 [info] [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://localhost:4002/worker
because the port config from `config/test.exs` is ignored (this is what this PR fixes).

taylordowns2000 (approved):

PR #1453, opened 2023-11-23
1254 remove attempts fks

rorymckinley:
## Notes for the reviewer

Removes the FK constraint between `attempt_runs` & `attempts` as well as the FK between `log_lines` and `attempts`.  I added functionality to `Attempts` to replicate the `ON DELETE CASCADE` behaviour lost as a result of removing the FK.

I also simplified the tests in `AccountsTest` and `QueueTest` as much of what they were testing is now encapsulated in he methods added to `Attempts` and so the detailed tests now live in `AttemptsTest` (sorry, makes for a bit of a messy diff - happy to talk through the changes).

Note: I could not test the rollback behaviour for `Attempts.delete_for_user` as I could not find a way to trigger a rollback from a test.  Open to any suggestions as to how I can improve the situation.

I added mocking so that I could test the response when there is an error. 

## Related issue

Fixes #

## Review checklist


lib/lightning/attempts.ex
```diff
@@ -256,6 +258,69 @@ defmodule Lightning.Attempts do
...
+      )
+      |> Ecto.Multi.delete_all(
+        :attempts,
+        from(a in Attempt, where: a.id in ^attempt_ids)
+      )
+      |> transaction_handler.()
+
+    case result do
+      {:ok, %{attempts: deleted_attempts}} -> {:ok, deleted_attempts}
+      {:error, _, changeset} -> {:error, changeset}
```
rorymckinley, line 320: Match against all the elements of the `Ecto.Multi` error response tuple:

```elixir
{:error, _failed_operation, changeset, _changes_so_far} -> {:error, changeset}
```

taylordowns2000:
@rorymckinley , should this still be in "draft mode" or is it ready for review and merge?

PR #1455, opened 2023-11-23
New Attempt View

### LINKED ISSUE #1243: Allow users to view details of an attempt (attempt detail page)
## Overview:

Currently users can view the details of a run on a job (on the Run page), and have an overview of the workorder run (on the history page) - but do not have an overview of the attempt.

**To do:** 
1. Creation of the Attempt overview screen.
2. Update History table to access the attempt detail page. 

### 1. Attempt overview screen

Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=858-4291&mode=design&t=FAlI1MQiRCLpdd9N-11

Accessible via the History page (details below), the page needs to contain:

- Overview section at the top, with the following fields:
  - Workflow name
  - Attempt ID (styled correctly)
  - XXX (Chat to Nick to see what we should pop in here)
  - Exit Reason (styled correctly)
  
- Associated runs:
  - This is the Runs of the jobs that executed for the attempt
    - **We will have the following columns:**
      - Job name (including status of success/failure indicated by icon)
      - Job Run start time `Run started`
      - Job Run completion time `Run ended`
      - Job total runtime `Duration`
  -  If the user chooses to rerun a job, a new attempt page will open with the new attempt

If an attempt is busy _running_ we will have the associated run table fill with the exact same behaviour we have on the History Page currently.

**Log section:**
- This will display the full log of the attempt.

 
**Minor UX improvements:**
- Wherever we have a overview table with the details, let's add the dashed line below so people can scan it easily.
- On the Run table (same on the history page) let's have a hover state to indicate which row they hovering on

### 2. History page changes

Figma:  https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=901-4531&mode=design&t=FAlI1MQiRCLpdd9N-4

We need to make the attempt detail page accessible on the History page, to do that:
- When a Work order is expanded, if only a single attempt has run it still needs to display the attempt banner (with no "Show more attempts" hyperlink.
- Attempts now need to display ID's, which are hyperlinked (like the data input ID's):

stuartc:
Still **TODO**:

- [x] testing and cleanup
- [x] changing the tab bar icons
- [x] prevent clicking input or output without a run being selected
- [x] add link to new page from History page
- [ ] consider splitting up the very big (borderline integration) test
- [x] ensure all event handling is done
- [x] remove sleep in log stream

## Related issue

Fixes #1243

## Review checklist


config/test.exs
```diff
@@ -56,10 +61,8 @@ config :lightning,
   schemas_path: "test/fixtures/schemas",
   adaptor_icons_path: "test/fixtures/adaptors/icons"
 
-config :lightning, Lightning.Runtime.RuntimeManager, start: false
-
 # Print only warnings and errors during test
-config :logger, level: :warning
+config :logger, level: :debug
```
stuartc, line 65: Note to self, remove this!

lib/lightning_web/channels/attempt_channel.ex
```diff
@@ -64,17 +64,27 @@ defmodule LightningWeb.AttemptChannel do
...
 
   def handle_in("attempt:complete", payload, socket) do
-    %{
-      "reason" => reason,
-      "error_type" => error_type,
-      "error_message" => error_message
-    } =
+    params =
       payload
+      |> Map.get_and_update("reason", fn _ -> :pop end)
```
jyeshe, line 69: :metal:

jyeshe: only the `_` if you may ... in the future if we want to switch to `mix credo --strict` we won't to have to fix all these entries.

PR #1463, opened 2023-11-26
Dim disabled triggers and edges on the canvas

### LINKED ISSUE #1464: Dim/grey out triggers and edges on the canvas when they are disabled
This is a precursor to #898  (CC @NickOpenFn )

- Here, we are simply greying out the nodes and edges that are disabled on the canvas.
- There are no fancy downstream "is this node reachable from the default node?" projections.

image

taylordowns2000:
## Notes on style

Right now we open a different modal for the trigger and the first trigger-to-job edge. In the future, since enable/disable for that combo will only live on the trigger itself, it may make sense to show that trigger/edge combination in a single box with an enabled/disabled toggle.

Right now, I've opted to show the edge between the trigger and the first job as `disabled` if the trigger itself is disabled because the edge could never be reached with a disabled trigger. In other words, if you disabled the trigger, you'll see both the trigger and its only edge get dimmed.

## Notes on implementation

This isn't quite ready... I'm picking up two strange behaviours here:

1. When you click away from an edge onto the canvas (deselect _all_) it doesn't get unstyled.
2. When you change _another_ edge's enabled/disabled property, the disabled trigger-edge-combo looses its disabled style.

Broadly, I had a way harder time than I'd expect applying conditional styles based on trigger and edge data. Looks like there are a number of different places where styles are built and then sometimes overwritten. Maybe a refactor, just of the styling, with an objective to allow data-based styling (applying styles based on actual lightning props) via a single interface is needed. I see the beginnings of this approach were introduced with the `styleItem(...)` implementation. It's possible that I'm misunderstanding how to use that, since changes there didn't do the trick in all settings, but I'm hoping that a review of this PR will either help steer me in the right direction, or provide motivation for a new implementation here.

## Related issue

Fixes #1464

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to
 
 ### Added
 
+- Dimmed/greyed out triggers and edges on the canvas when they are disabled
+  [#1464](https://github.com/OpenFn/Lightning/issues/1464)
 - Async loading on the history page to improve UX on long DB queries
   [#1279](https://github.com/OpenFn/Lightning/issues/1279)
 
--- a/assets/js/workflow-diagram/edges/Edge.tsx
+++ b/assets/js/workflow-diagram/edges/Edge.tsx
@@ -21,7 +21,7 @@ const CustomEdge: FC<EdgeProps> = props => {
               transform: `translate(-50%, -50%) translate(${labelX}px,${labelY}px)`,
               background: 'white',
               pointerEvents: 'all',
-              ...labelStyles(selected),
+              ...labelStyles(selected, stepEdgeProps.data),
             }}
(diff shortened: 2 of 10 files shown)


taylordowns2000:
@josephjclark , i've requested a "pre-review" from you here. could we link up for fifteen minutes tomorrow to discuss? i found it harder than i expected to apply conditional styles to nodes and edges in the workflow, and i found myself begging for a single interface that would take the lightning attributes (stuff like `{ enabled, source, destination, condition, ...etc }`) and give me an opportunity to conditionally style each node or edge. hoping my flailing implementation (and the two bugs that it creates/highlights) will help us come up with a better pattern here.

PR #1476, opened 2023-11-28
fix: Initial credential creation  show after as  rather a value

### LINKED ISSUE #1118: Initial credential creation `changes` show after as `null` rather a value
When a credential is created, an `audit_event` is generated. It should display a hash of the credential body and include the initial label as you can see in the `updated` event below and after the `created` event in this image.

image

midigofrank:
## Related issue

Fixes #1118 

## Review checklist

### DIFF
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -124,9 +124,15 @@ defmodule Lightning.Credentials do
     changeset = %Credential{} |> change_credential(attrs) |> cast_body_change()
 
     Multi.new()
-    |> Multi.insert(:credential, changeset)
+    |> Multi.insert(
+      :credential,
+      changeset
+    )
     |> Multi.insert(:audit, fn %{credential: credential} ->
-      Audit.event("created", credential.id, credential.user_id)
+      Audit.event("created", credential.id, credential.user_id, %{
+        before: nil,
+        after: Map.take(changeset.changes, Credential.__schema__(:fields))
+      })
     end)
     |> Repo.transaction()
     |> case do
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -128,12 +128,7 @@ defmodule Lightning.Credentials do
       :credential,
       changeset
     )
-    |> Multi.insert(:audit, fn %{credential: credential} ->
-      Audit.event("created", credential.id, credential.user_id, %{
-        before: nil,
-        after: Map.take(changeset.changes, Credential.__schema__(:fields))
-      })
-    end)
+    |> derive_events(changeset)
     |> Repo.transaction()
     |> case do
       {:error, _op, changeset, _changes} ->
@@ -245,7 +240,8 @@ defmodule Lightning.Credentials do
 
   defp derive_events(
          multi,
-         %Ecto.Changeset{data: %Credential{}} = changeset
+         %Ecto.Changeset{data: %Credential{__meta__: %{state: state}}} =
+           changeset
        ) do
     case changeset.changes do
       map when map_size(map) == 0 ->
@@ -263,7 +259,7 @@ defmodule Lightning.Credentials do
           :audit,
           fn %{credential: credential} ->
             Audit.event(
-              "updated",
+              if(state == :built, do: "created", else: "updated"),
               credential.id,
               credential.user_id,
               changeset
--- a/lib/lightning_web/components/new_inputs.ex
+++ b/lib/lightning_web/components/new_inputs.ex
@@ -275,7 +275,7 @@ defmodule LightningWeb.Components.NewInputs do
       type="hidden"
       name={@name}
       id={@id}
-      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value) |> to_string()}
     />
     """
   end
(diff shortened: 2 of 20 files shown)


lib/lightning/credentials.ex
```diff
@@ -116,13 +116,19 @@ defmodule Lightning.Credentials do
...
       :credential,
-      Credential.changeset(%Credential{}, attrs |> coerce_json_field("body"))
+      changeset
     )
     |> Multi.insert(:audit, fn %{credential: credential} ->
-      Audit.event("created", credential.id, credential.user_id)
+      Audit.event("created", credential.id, credential.user_id, %{
+        before: nil,
+        after: Map.take(changeset.changes, Credential.__schema__(:fields))
+      })
```
stuartc, line 131: Could this be changed to use the `derive_events` function instead?

You might be able to match on `%{__meta__: %{state: :built}}` of the changeset.

test/lightning/credentials_test.exs
```diff
@@ -86,11 +86,22 @@ defmodule Lightning.CredentialsTest do
...
-             )
-             |> Repo.one!(),
+      assert audit_event =
+               from(a in Audit.base_query(),
+                 where: a.item_id == ^credential.id and a.event == "created"
+               )
+               |> Repo.one!(),
              "Has exactly one 'created' event"
+
+      _expected_change = %{
```
stuartc, line 96: Please don't use the `_` prefix on variables we use.

stuartc (changes requested):
Looks good @midigofrank, please take a look at my proposed change to how the event is built.

taylordowns2000 (changes requested):
hey @midigofrank , would you please:
1. ensure that `added_to_project` events are written alongside the `created` events if a credential is created _and_ added to a project all in a single event.
2. ensure that `removed_from_project` events are emitted when credentials are removed from projects.
3. check to see that we're storing the _crypted_ hash, not the hashed body in the audit trail.

re point 3, @jyeshe , i saw your comment on this issue last week and realize that in my initial issue description i used the wrong words! apologies. we aren't allowed to store any representation of the credential (base64 encoded or not) in the DB that could be accessed without the encryption key. so here, we're expected to see a crypted hash. users should be able to tell that the crypted hash changes, but not be able to read/decode the hash.

does that make sense? apologies for the initial confusing that I likely caused here!

taylordowns2000 (changes requested):
hey @midigofrank
1. i've set up a failing test here and i'd love for you to figure out why it doesn't work and propose a fix that retains the spirit of this change.
2. please implement your `created` module for webhook auth events and confirm that the secrets are being written to the audit trail in their crypted and base64 encoded format.

taylordowns2000 (approved):
@midigofrank , this works just as expected from the product perspective. great job. i see you've got a "WIP" commit in there and some commented out code. please let me know when it's done from your perspective. i've fixed up the test (thanks for your comment) and all looks good.

jyeshe (approved):
LGTM specially having the test loading the encrypted map 🙏

Two points I would analyse next:
1. Discuss about hashing when exporting the audit logs becomes a requirement or a task.
2. Macros are a cool stuff in Elixir but it's hard to maintain as you don't have the module source and the line to debug. Even though the code seems solid and there might have been a reason for using macros, I would try harder to use protocols to achieve polymorphism.

jyeshe:
> > Now looking at the final solution I see there is a encrypted dump of the credential body. The difference I have tried to understand in terms of requirements is that if it was asked to be hashed or not, on top of the encryption. If it's hashed it would be _not auditable_ without a credential history table. Moreover maybe we aren't doing it yet, and maybe never won't, but audit logs are generally exported to outside the database. When/if that will be the case, a hashed value would be safer because even if the encryption key is leaked by some client, their audit data is still safe, i.e the hacker could access no credential.
> 
> @jyeshe , the client doesn't have access to the encryption key, so i think we're good here. (they can never decrypt this data. only the app itself can.)
> 
> On this bit: "I would try harder to use protocols to achieve polymorphism." - I also see that you _approved_ the PR. Is this a rejection or an approval? Please feel free to take a stance here and let Frank know what you'd like him to do. (Maybe your position is that you'd like to merge this but only do so after a new issue is opened to remove the use of macros?)

Nice @taylordowns2000! About the macros that's the intended position. After merging we can investigate the protocols alternative.

lib/lightning_web/components/new_inputs.ex
```diff
@@ -275,7 +277,7 @@ defmodule LightningWeb.Components.NewInputs do
       type="hidden"
       name={@name}
       id={@id}
-      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value) |> to_string()}
```
stuartc, line 280: This got my attention, I can't figure out why it's needed. I mean it's likely harmless, but it's a bit puzzling why this test fails:

```
   test/lightning_web/live/credential_live_test.exs:567
     Assertion with == failed
     code:  assert Enum.count(audit_events) == 2
     left:  1
     right: 2
```

Why is the delete event not created is `nil|true|false` is not cast to a string?

stuartc: Fwiw, it might be a good idea to add `, default: false` to the ProjectCredential delete virtual field - I tried that but didn't solve this specific thing..

PR #1488, opened 2023-11-29
Fix flaky attempt run view

jyeshe:
## Notes for the reviewer

"Awaits" on the view async state instead on the element to render the change.

## Related issue

https://app.circleci.com/pipelines/github/OpenFn/Lightning/5568/workflows/53dbe327-fcf9-4970-8107-9dc4809b2422/jobs/29065

```
  1) test handle_async/3 lifecycle of an attempt (LightningWeb.AttemptLive.ShowTest)
     test/lightning_web/live/attempt_live/show_test.exs:43
     Assertion with =~ failed
     code:  assert view
            |> element("#run-input-#{run.id}")
            |> render_async()
            |> Floki.parse_fragment!()
            |> Floki.text() =~ ~s"{  \"x\": 1}"
     left:  "\n    Nothing yet...\n  "
     right: "{  \"x\": 1}"
     stacktrace:
       test/lightning_web/live/attempt_live/show_test.exs:113: (test)
```

## Review checklist

### DIFF
--- a/test/lightning_web/live/attempt_live/show_test.exs
+++ b/test/lightning_web/live/attempt_live/show_test.exs
@@ -78,9 +78,11 @@ defmodule LightningWeb.AttemptLive.ShowTest do
 
       Lightning.Attempts.start_attempt(attempt)
 
+      render_async(view)
+
       assert view
              |> element("#attempt-detail-#{attempt_id}")
-             |> render_async() =~ "Running",
+             |> render() =~ "Running",
              "has running state"
 
       refute view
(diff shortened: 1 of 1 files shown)


taylordowns2000 (approved):

stuartc:
@taylordowns2000 @jyeshe this wouldn't have fixed the issue, the log lines aren't done using `start_async` and so `render_async` etc won't actually wait for those to come back - they are regular tasks with a send back to the parent. I've realised we would need to use something _like_ the (but not exactly) `wait_for_assigns` helper used in the credential live tests.

I've taken a swing at this, and instead of waiting for messages and state - I've switched the two assigns that can be used to determine that the dataclips have arrived to be asyncs, and so we can lean in `render_async` to wait for them in the tests... seems to be more reliable so far.

PR #1493, opened 2023-11-30
Metadata not refreshing when credential is updated

### LINKED ISSUE #791: Metadata not refreshing when credential is updated
**Describe the bug** 
When I update a credential and click save,  the metadata doesn't get updated. 

**To Reproduce**

1. Go to a job and edit the credential
2. Click on save
3. See metadata doesn't get updated (ex: if you didn't have a credential before, then you add one and click save you still see the no_credential error)

**Expected behavior** 
The metadata should be refreshed. 

**Additional context**
we’d like a way for the metadata component to detect a change to the credential, as users can now save the job inspector form with a new credential without the form closing

midigofrank:
## Notes for the reviewer

- This PR uses the `project_credential_id` `value` in the form to get the selected `credential`. Initially, it was using the preloaded `credential` which doesn't change the form is updated

## Related issue

Fixes #791 

## Review checklist


taylordowns2000 (changes requested):
@midigofrank , this looks like it _might_ close #1531 but i don't think it yet closes #579 .

to test what happens when a credential is updated and verify that the metadata reloads properly, use dhis2.

1. create a dhis2 play credential (https://play.dhis2.org/) using 2.39.3.1 for example
2. check to see that the metadata loads properly
3. update the credential body so that it uses play 40.2.0, for example, and check to see that the metadata is properly reloaded.

taylordowns2000:
It’s not. Leave the API version blank when performing this test. Those are
not dhis2 api versions; they are actually different dhis2 servers and you
should be updating the hostUrl .

Do you see the difference between #1531 and #579? One is about updating the
body of a credential while the other is about updating a credential_id
field on a job. We need to fix both, and I think you’ve fixed #579 but not
#1531.

Am I right or misunderstanding this PR?

taylordowns2000 (approved):
tested and approved. thank you @midigofrank

PR #1499, opened 2023-11-30
Limit entries count on term workorders search

jyeshe:
## Notes for the reviewer

The workorder search with text term (text field value) parses log lines messages and dataclips which is an expensive operation. 
Therefore timeout might happen on non matching results but also when there are too many matches to count _EDIT_: or better too much data to parse (the count limit is not the final solution but decreases the chance of a timeout).

This PR also adds the limit on total entries counting to the customized `scrivener_ecto`:
https://github.com/OpenFn/scrivener_ecto/commit/c51f2138dc50aa21d854d1f82e2f707f6c258a7e
and moves the paginator module to Lightning in order to avoid managing the version on a separate repo (until the library accepts the proposed change). 

If prefered we can continue using: https://github.com/OpenFn/scrivener_ecto

## Related issue

Refs #1461 

## Review checklist

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -16,6 +16,8 @@ defmodule Lightning.Invocation do
   @workorders_search_timeout 30_000
   @workorders_count_limit 50
 
+  def get_workorders_count_limit, do: @workorders_count_limit
+
   @doc """
   Returns the list of dataclips.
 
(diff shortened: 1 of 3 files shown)


jyeshe:
Flaky addressed here: https://github.com/OpenFn/Lightning/pull/1501

taylordowns2000 (approved):
this is great @jyeshe . i've added some user-friendly help text here, but otherwise no changes: https://github.com/OpenFn/Lightning/pull/1499/files#diff-22230f2b83dbb80540c108098f13a1f28165b897fc29bb0065a5e0d78829179bR450-R457

PR #1512, opened 2023-12-01
Refactor and update AttemptViewerLive

### LINKED ISSUE #1474: Inspector run viewer enhancements
This builds on https://github.com/OpenFn/Lightning/issues/1243

- [x] Maybe try and have the step list on the manual run component
- [x] Highlighting on run on manual run component
- [x] Add hover to step list
- [x] Add padding to top of attempt/log viewer
- [x] Add back the output tab for the selected (or the ONLY?) run
- [x] Add a URL link to the attempt view page for the attempt that gets created
- [x] Add a URL link to the run view page for the run that gets created (or a way to navigate there from the step list)
- [x] Add a URL link for the Workorder (@stuartc , question here... i see that we have a filter that allows us to select just a single workorder in the list view. my guess is this is not good for performance, but does it give us the "workorder detail" page we want already? `projects/<blah>/runs?filters[workorder_id]=3e48c84a-02af-4ec2-94ea-1eb1f1912878`.)
- [x] Check to ensure that long log lines **_without word breaks_** don't break the x-scroll or overflow layout of the log viewer. (I think they do.)

stuartc:
All new streaming functionality that was introduced in the Attempt.Show page has been abstracted into macros.

AttemptViewerLive has been updated to use those macros and the components have been extended in various ways to allow extra classes to be added and certain behaviours to be changed.

## Related issue

Fixes #1474 

## Review checklist


stuartc:
@taylordowns2000 I've made a bunch of the changes requested, the dataclip type indicator and the link to the attempt view page (when looking at a manual run). I'm pretty sure the whitespace and breaking works nicely along with the scrolling. 

A few things though, I haven't implemented changing the run on the manual run component (not _yet_ at least, it would be a little different from the other implementation - so prioritised other things).

I haven't add the links to the workorders page, I'm battling to find space to put it... 

There isn't a lot of room here for much of anything.

Feedback so far?

taylordowns2000 (changes requested):
looks great. two things:
1. seems as though the order of the runs got flipped. i'm seeing step 3 show above step 1 (see image 1)
2. can you make sure the workorder id field (and link) still appears on the attempt detail page. (same as on the inspector page.)
3. the spacing on the input/output viewer gets super wonky when you collapse or expend the panels. can we fix this? (See image 2)

image 1
image

image 2
image

stuartc:
@taylordowns2000 I've made those changes, the dataclip views now fill the space vertically and horizontally...

The ordering thing is string, like I can't recreate it. Any change to the list of runs (even initially) is sorted in-memory. Is the list consistently out of order because if thats the case then the data is incorrect or the datetime isn't accurate enough.

In order to solve _that_ specific issue, the only reliable way to present it would be to compare it to the workflows ordering.

stuartc:
> Can you share the setup and navigation steps to validate the change? One example would be enough and the rest I can try to infer from the issue checklist.

Ah sorry dude, had to merge this for a demo. I'll see what I can do to demo this for you to get better feedback.

Issue #1442, opened 2023-11-22
Navigation when editing credential takes you to another settings section


NickOpenFn:

NickOpenFn:
Closed by Elias with #1064

PR #1520, opened 2023-12-03
Fix Credential Creation Page UI

### LINKED ISSUE #1064: UI broken for credential creation from within the canvas
Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=349-6393&mode=design

- [x] Update the credential creation to use a modal instead of the plain pages (see Figma)
- [x] Fix incorrect spelling of "Associate" (see Figma, incorrect spelling displayed with red text)
- [x] Add in button to create a credential on the settings credential overview page (see Figma)

## Old spec
html and css messed up when creating a credential from inside the job editor. (note that it works OK when _not_ inside a the job editor "new credential" modal.)

1. there's an extra box inside the box
2. the contents of that inner box spill over outside into the outer box

image

elias-ba:
## Notes for the reviewer

This PR updates the credential creation and editing UI, introducing modal components for improved user interaction. It removes outdated code and tests, reflecting the transition to a modal-based feature approach. The implementation aligns with the specified Figma design, enhancing the overall aesthetics and functionality. Additionally, the PR addresses and resolves issue #1442, fixing bugs in the credential management area. This comprehensive update streamlines the user experience and boosts the efficiency of credential handling in the application. For more details, view the Figma design [here](https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=349-6393&mode=design).

## Related issue

Fixes #1064 

## Review checklist


lib/lightning_web/live/credential_live/index.html.heex
```diff
@@ -52,15 +67,28 @@
...
+              action={:edit}
+              credential_type={@selected_credential_type}
+              credential={credential}
+              current_user={@current_user}
+              projects={@projects}
+              project={nil}
+              show_project_credentials={true}
+              can_create_project_credential={true}
+              return_to={~p"/credentials"}
+            />
```
stuartc, line 91: I'm not that keen on rendering components for every item, it seems a bit wasteful. I don't think we should do anything about it now given how infrequently its accessed but we should think about lowering the cost of rendering/diffing at every point; especially in lists.

stuartc (changes requested):
@elias-ba this is a solid piece of work, thank you!

I made some opinionated changes in a few places. I thought I broke the tests but it looks like some changes from main broke them first...

I tweak the test slightly but I'm getting some JS push failure on the same failing test. Would you mind fixing that and then we can merge?

lib/lightning_web/live/credential_live/index.html.heex
elias-ba: @stuartc I do agree. I also am not a big fan of this. Seems like the most reasonable thing to do now, is keep as it is and think about a better approach earlier January.

taylordowns2000 (changes requested):
@elias-ba , as far as i can tell the new interface looks perfect. great job. change request here is on the "new credential" link... seems like it's turned into a very big button with improper underlining and padding!

image

I'd keep it as a link (see below) unless you've got a very specific new design from @NickOpenFn .

image

taylordowns2000 (approved):

Issue #1519, opened 2023-12-03
Expand scrubber to take multiple credentials


taylordowns2000:
Previously, the `Lightning.Scrubber` would be started with a single credential (see `Lightning.Credentials.SensitiveValues`) and could be used to remove any occurrence of a credential body value (i.e., a _secret_) from an incoming log line.

Now that the worker handles a whole _attempt_, we must be able to start an instance of the scrubber (or design a new scrubber interface) that will scrub secrets as determined by any of the credentials used in an attempt.

In other words:
 - **Before:** we start the scrubber with a single credential and log lines are emitted (and scrubbed) for a single run.
 - **Now:** we need to tell the scrubber to check for occurrences of secrets from a list of credentials, as the password from run one might show up in the logs of run 2.

Note: As the scrubber is re-implemented to handle multiple credentials, please keep in mind the most efficient way to similarly scrub occurrences of secrets from `run_results`. The simplest (and likely least performant) way to do this would be to serialize the entire `run_result` and scrub that massive string before turning it back into a map and adding it to the DB. My sense is that this (adding the scrubber to run_results) should come later in a separate issue.

Verify that this is working properly by extending the end_to_end tests so that the _values_ of secrets are written to the logs independently from `console.log(state.configuration)`.

PR #1523, opened 2023-12-04
Scrub log lines using multiple credentials samples

jyeshe:
## Notes for the reviewer

1. Extends the Scrubber to accept updates to handle multiple credentials.
2. Applies the scrubbing to the log lines.
3. Handles scrubbing on Basic auth method that might be present on logs

## Related issue

Solves #1519 

## Review checklist

### DIFF
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -438,6 +438,23 @@ defmodule Lightning.Credentials do
     end
   end
 
+  def basic_auth_for(%Credential{body: body}) when is_map(body) do
+    usernames =
+      body
+      |> Map.take(["username", "email"])
+      |> Map.values()
+
+    password = Map.get(body, "password", "")
+
+    usernames
+    |> Enum.zip(List.duplicate(password, length(usernames)))
+    |> Enum.map(fn {username, password} ->
+      Base.encode64("#{username}:#{password}")
+    end)
+  end
+
+  def basic_auth_for(_credential), do: []
+
   @doc """
   Given a credential and a user, returns a list of invalid projects—i.e., those
   that the credential is shared with but that the user does not have access to.
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -472,10 +472,8 @@ defmodule Lightning.Credentials do
   end
 
   # TODO: this doesn't need to be Google specific. It should work for any standard OAuth2 credential.
-  @spec maybe_refresh_token(nil | Lightning.Credentials.Credential.t()) ::
-          {:error, :invalid_config}
-          | {:ok, Lightning.Credentials.Credential.t()}
-          | {:ok, nil}
+  @spec maybe_refresh_token(Lightning.Credentials.Credential.t()) ::
+          {:error, any()} | {:ok, Lightning.Credentials.Credential.t()}
   def maybe_refresh_token(%Credential{schema: "googlesheets"} = credential) do
     token_body = Google.TokenBody.new(credential.body)
 
@@ -494,7 +492,6 @@ defmodule Lightning.Credentials do
   end
 
   def maybe_refresh_token(%Credential{} = credential), do: {:ok, credential}
-  def maybe_refresh_token(nil), do: {:ok, nil}
 
   defp still_fresh(
          %{expires_at: expires_at},
--- a/lib/lightning/scrubber.ex
+++ b/lib/lightning/scrubber.ex
@@ -34,9 +34,7 @@ defmodule Lightning.Scrubber do
     end
 
     @spec scrub(state :: t(), data :: String.t()) :: String.t()
-    def scrub(nil, _state), do: nil
-
-    def scrub(string, {samples}) when is_binary(string) do
+    def scrub({samples}, string) when is_binary(string) do
       samples
       |> Enum.reduce(string, fn x, acc ->
         String.replace(acc, x, "***", global: true)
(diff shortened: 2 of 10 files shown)


jyeshe:
Handling now https://github.com/OpenFn/Lightning/issues/1525 ...

taylordowns2000 (changes requested):
@jyeshe , i've put back the base64 encoded requirement [here](https://github.com/OpenFn/Lightning/pull/1523/commits/a45cfd4c2e470ee323ff9c8309186570974bbb21) - can you check to ensure the scrubber handles this?

lib/lightning_web/channels/attempt_channel.ex
```diff
@@ -177,7 +184,10 @@ defmodule LightningWeb.AttemptChannel do
   end
 
   def handle_in("attempt:log", payload, socket) do
-    Attempts.append_attempt_log(socket.assigns.attempt, payload)
+    %{attempt: attempt, scrubber: scrubber} = socket.assigns
+    payload = Map.update(payload, "message", nil, &maybe_scrub(scrubber, &1))
+
+    Attempts.append_attempt_log(attempt, payload)
```
stuartc, line 190: Ergh, I can see why it's awkward here to use the changeset, it's hidden behind `append_attempt_log`. 

I am wary of trying to either scrub anything that _hasn't_ gone through our `LogMessage` casting, I think we should avoiding working on any data structure that isn't what the user will see?

Perhaps we could pass the scrubber pid into `append_attempt_log` as an optional argument, and then add a step to the multi before insert where we can access the changeset?

Or we could, use `LogMessage.cast/1` directly and take it's `{:ok, string()}` result, scrub that and then update payload with that before passing to the Attempts module.

Either way I think we should try and deal with strings here as much as possible when scrubbing. Maybe one day it would make sense to recursively scrub arbitrary structures but I'm not keen on making that change now.

In the future I do see us having a process that collects up log lines in a buffer and inserts them in batches, which would move the theme of scrubbing out of the AttemptChannel and into that process - just an idea of how we could improve this now or later.

lib/lightning/scrubber.ex
```diff
@@ -79,8 +96,10 @@ defmodule Lightning.Scrubber do
   Prepare a list of sensitive samples (strings) into a potentially bigger list
   composed of variations a sample may appear.
   """
-  @spec encode_samples(samples :: [String.t()]) :: [String.t()]
-  def encode_samples(samples) do
+  @spec encode_samples(samples :: [String.t()], basic_auth :: [String.t()]) :: [
+          String.t()
+        ]
+  def encode_samples(samples, basic_auth \\ []) do
```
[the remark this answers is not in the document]
jyeshe: Hey @stuartc , this is needed because the `username` and `email` are on the safe list of the `SensitiveValues` in order catch `Basic` auth method on request hearders.

It was caught by this additional assertion line:

```
MapSet.new([
          {"R/T", "Starting operation 1"},
          {"JOB", "#{expected_job_x_value}"},
          # Check to ensure that an inadvertantly exposed secret from job 2 is
          # still scrubbed properly in job 3.
          {"JOB", "quux is on the safelist"},
          {"JOB", "but *** should be scrubbed"},
          {"JOB", "along with its encoded form ***"}, <<<---
          {"R/T", "Expression complete!"}
        ])
```
when a run logs something like:
```
    "fn(state => {
      state.x = state.x * 3;
      console.log(state.x);
      console.log('quux is on the safelist')
      console.log('but immasecret should be scrubbed');
      console.log('along with its encoded form #{Base.encode64("quux:immasecret")}');
      return state;
    });"
```
where `quux:immasecret` is the `<username>:<password>` (same applies to a Basic auth with email and password so we have this combinations provided by `Credentials.basic_auth_for(%Credential{body: body})`)

lib/lightning/credentials.ex
```diff
@@ -494,7 +509,6 @@ defmodule Lightning.Credentials do
   end
 
   def maybe_refresh_token(%Credential{} = credential), do: {:ok, credential}
-  def maybe_refresh_token(nil), do: {:ok, nil}
```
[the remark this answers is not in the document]
jyeshe: There was a cleanup on the spec. It expects a credential and the nil is handled beforehand by the `handle_in("fetch:credential`

lib/lightning_web/channels/attempt_channel.ex
jyeshe: Done

taylordowns2000 (approved):

lib/lightning_web/channels/attempt_channel.ex
stuartc: Rad thanks!

stuartc (approved):
Ok cool this looks good to go.

@taylordowns2000 re: your changes here.

I'm not a fan of having the http basic auth stuff treated separately in the scrubber. I really would prefer that the scrubber doesn't know anymore than "filter out these strings". 
And with the new basic auth function in Credentials we now have a module for sensitive values, and a function in credentials and another argument on the scrubber - I know we can do more (I guess I mean less) here to conform all of this behaviour.

I'm going to merge this due to it's importance - but I'd very much like us to avoid making this more complicated than it already is, it's very difficult to follow all the different permutations and we now have two places that have lists of keys to use or not use.

jyeshe:
> Ok cool this looks good to go.
> 
> @taylordowns2000 re: your changes here.
> 
> I'm not a fan of having the http basic auth stuff treated separately in the scrubber. I really would prefer that the scrubber doesn't know anymore than "filter out these strings". And with the new basic auth function in Credentials we now have a module for sensitive values, and a function in credentials and another argument on the scrubber - I know we can do more (I guess I mean less) here to conform all of this behaviour.
> 
> I'm going to merge this due to it's importance - but I'd very much like us to avoid making this more complicated than it already is, it's very difficult to follow all the different permutations and we now have two places that have lists of keys to use or not use.

I understand your concern to avoid having multiple specific cases but the solution has only 2 different ways of processing in order to attend to 2 different requirements:
a. secret values
b. secret values combined with excluded ones from safe list

So whatever refactor we do there will be always these two parameters (for the given requirement). We might possibly need to give a more generic name but for now I think it clarifies the purpose.

PR #1527, opened 2023-12-05
Add scrivener to  mix.exs applications

benedictus-dev:
## Notes for the reviewer
Add scrivener to your mix.exs  applications list

## Related issue
Fixes #

## Review checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -39,7 +39,7 @@ defmodule Lightning.MixProject do
   def application do
     [
       mod: {Lightning.Application, [:timex]},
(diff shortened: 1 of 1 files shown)


stuartc:
@benedictus-dev thanks for this, but I'm curious as to why we need it now? I mean you are right, the documentation does state to add it to `:extra_applications` but we've done fine without it so far.

As far as I can tell Scrivener doesn't have it's own application, but _does_ require Logger to be started, which we already do ourselves.

What did you see/run into that made this necessary?

stuartc (approved):

PR #1530, opened 2023-12-05
Persist credential with typed values

### LINKED ISSUE #1518: Strings get sent to the worker where booleans should
Building on #1513, I'm seeing that all values (including booleans) are being saved to the DB as strings. When they get passed to the worker, they no longer match the required JSON schema specification and we're probably going to run into a bunch of issues with this down the line.

Put simply, we're sending
```json
{"requireSsl": "true"}
```
to the worker while the schema requires that we send
```json
{"requireSsl": true}
```

Please check all types that we handle in the config schema and ensure that they're sent to the worker properly.

jyeshe:
## Notes for the reviewer

The worker needs to receive credential body with types declared on schemas (e.g `ssl: true` instead of `ssl: "true"` for postgresql credential). 

There are 2 points where the casting for the schema type is applied:
1. When saving a credential (insert or update)
2. When fetching a credential (for those already saved)

It uses a dynamic changeset (from `SchemaDocument`) to cast the body field types.

## Related issue

Fixes #1518 

## Review checklist


taylordowns2000 (approved):
perfect, thanks @jyeshe . this passes UX review and i'm happy with the tests.

taylordowns2000:
@stuartc , a question for your code review... why was this not a problem in v1?

stuartc:
@taylordowns2000 It's because in platform the forms in react were correctly building boolean values and sending them straight to the model.
With LiveView, all parameters come through as strings, like any regular HTML form. We were never casting them using the schema. 
We were validating the form using the schema (in a nested component) but we never cast it again once we try and save a credential. 

When the credential gets sent to the worker it gets send _as is_. So if it's saved as a string, a string it is.

lib/lightning_web/channels/attempt_channel.ex
```diff
@@ -90,6 +90,7 @@ defmodule LightningWeb.AttemptChannel do
     %{attempt: attempt, scrubber: scrubber} = socket.assigns
 
     with credential <- Attempts.get_credential(attempt, id) || :not_found,
+         credential <- Credentials.cast_body(credential),
```
stuartc, line 93: I'm not sure this is necessary. If the credentials are persisted correctly we shouldn't cast them a second time. 

If we absolutely have to cast them again, I really think we shouldn't be replacing the `:body` key on the credential. I would prefer the cast body was the return value directly, and we send that out straight to the worker.

Originally I wanted these to be cast automatically, but it was more trouble than it was worth (at the time). I want to avoid situations where a model is changed and not persisted; we run the risk of downstream functions assuming everything is a certain shape and not getting the opportunity to know it was changed (or passed through `cast_body`).

Again I would prefer we not re-cast these, I don't believe this is the place for this code to have any say as to the validity or shape of something. All it fixes in my mind at this precise moment is previously (incorrectly) saved credentials appearing as if they are correct.

stuartc (changes requested):
This looks good, but I'm not so sure we want to add a layer to the outgoing credential.

We should be able to trust the persisted credential, in the case of existing bad credentials I'd rather fix them with a script than fix them as they go out.

lib/lightning_web/channels/attempt_channel.ex
jyeshe: Hey Stu, this is for existing credentials. I wanted to avoid touching/changing them persistently but I am working on the migration ...

test/lightning/credentials_test.exs
```diff
@@ -355,6 +256,245 @@ defmodule Lightning.CredentialsTest do
...
+      }
+
+      assert {:ok, %Credential{} = credential} =
+               Credentials.create_credential(valid_attrs)
+
+      assert credential.body ==
+               Map.merge(body, %{
+                 "port" => 5000,
+                 "ssl" => true,
+                 "allowSelfSignedCert" => false
```
taylordowns2000, line 317: please add one example string value here.

test/lightning/credentials_test.exs
```diff
@@ -355,6 +256,245 @@ defmodule Lightning.CredentialsTest do
     end
   end
 
+  describe "create_credential/1" do
+    test "suceeds with raw schema" do
+      valid_attrs = %{
+        body: %{"username" => "user", "password" => "pass"},
```
taylordowns2000: please add an `integer` and a `boolean` to this example raw credential and assert that they are not dumped as strings

taylordowns2000 (changes requested):

taylordowns2000 (approved):

PR #1536, opened 2023-12-06
Questions for credential typing

taylordowns2000:
@jyeshe , re: https://openfn.slack.com/archives/C05KZNPEJFN/p1701883685607299

Here's some food for thought. Would love to chat about this briefly tomorrow! @stuartc , would love your thoughts as well 🙏 

the thing i'd like to test here is that if we have to pass a bunch of strings because of liveview, when they come _back_ from the DB to the worker they're properly typed. the change to this test does that (along with a quick fix for `put_typed_body`) but i don't know if it's a sound approach.

please see/address all TODOs here


lib/lightning/credentials.ex
```diff
@@ -221,16 +230,19 @@ defmodule Lightning.Credentials do
...
+  # TODO - roger i think we need tests for this function
   defp put_typed_body(body, schema_name) do
     schema = get_schema(schema_name)
 
     with changeset <- SchemaDocument.changeset(body, schema: schema),
          {:ok, typed_body} <- Ecto.Changeset.apply_action(changeset, :insert) do
       updated_body =
         Enum.into(typed_body, body, fn {field, typed_value} ->
-          {to_string(field), typed_value}
+          # TODO - roger i removed the "to string here" and it seems to work better
```
jyeshe, line 244: we need to convert the keys resulting from `apply_action` from atom to string once the attributes comes from the form with string keys

test/lightning_web/channels/attempt_channel_test.exs
```diff
@@ -240,9 +240,15 @@ defmodule LightningWeb.AttemptChannelTest do
 
   describe "cast credential types" do
     defp set_postgresql_credential(_context) do
-      credential =
-        insert(:credential,
+      user = insert(:user)
+
+      # TODO - roger, if we want to check that the casting works could we use it
+      # here and verify that the output is valid for the worker? the below test
```
jyeshe, line 246: just a negative scope test, to show that we are not casting on the reply as @stuartc requested

taylordowns2000:
all addressed in the pr closing #1518

PR #1541, opened 2023-12-07
use Application.compile_env in endpoint module

taylordowns2000:
This PR does two things:
1. it uses Application.compile_env rather than Application.get_env in the endpoint module
2. it removes the default value, as we already set that in `runtime.exs:160` - felt weird to set it twice.

## Review checklist

### DIFF
--- a/lib/lightning_web/endpoint.ex
+++ b/lib/lightning_web/endpoint.ex
@@ -57,9 +57,7 @@ defmodule LightningWeb.Endpoint do
     parsers: [
       :urlencoded,
       :multipart,
-      # Increase to 10MB max request size only for JSON parser
-      {:json,
-       length: Application.get_env(:lightning, :max_dataclip_size, 10_000_000)}
+      {:json, length: Application.compile_env(:lightning, :max_dataclip_size)}
     ],
     pass: ["*/*"],
     json_decoder: Phoenix.json_library()


petermueller:
Hey!

Just found out about OpenFn and Lightning.

(at)akoutmos has a lovely library called [replug](https://github.com/akoutmos/replug) that was designed for this type of scenario, and we used when we worked at BridgeConnector together. I'd *hope* there's not some compile-time requirement from `Plug.Parsers` though. That would be annoying :-/

```elixir
# endpoint.ex
 plug Replug,
  plug: Plug.ParsersCorsica, expose_headers: ~w(X-Foo)},
  opts: {LightningWeb.PlugConfigs, :plug_parsers}

  plug Replug,
    plug: Plug.Parsers,
    opts: {LightningWeb.PlugConfigs, :plug_parsers}
    
 # plug_configs_or_something_like_that.ex
defmodule LightningWeb.PlugConfigs do
  def plug_parsers do
    [
      parsers: [
        :urlencoded,
        :multipart,
        # Increase to 10MB max request size only for JSON parser
        {
          :json,
          length: Application.get_env(:lightning, :max_dataclip_size)
        }
      ],
      pass: ["*/*"],
      json_decoder: Phoenix.json_library()
    ]
  end
end
```

jyeshe:
@taylordowns2000 automated and manual work order test passing.

taylordowns2000:
love the tests! thanks @jyeshe and thanks @petermueller for the suggestion. looks good from my end (added warnings_as_errors back) but would like to wait for @stuartc 's review to merge

stuartc (approved):
Good stuff! Ironically we can totally use `get_env` inside a plug, just not in the module body with the `plug` macro; anyhoo, this works nicely!

PR #1542, opened 2023-12-07
Remove get_env warning

jyeshe:
## Related issue

Fixes #

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -157,10 +157,11 @@ config :lightning,
        System.get_env("MAX_RUN_DURATION", "60000")
        |> String.to_integer()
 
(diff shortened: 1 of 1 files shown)


jyeshe:
Later we gonna need to decrease the size of `max_skip_body_length` to be able to validate it but I can tune it later when tests are needed.

taylordowns2000 (changes requested):
@jyeshe , i'll push a fix for this, but the code here wasn't causing any troubles. it's in endpoint.ex (unless you've discovered a different thing entirely.) please drop this. i'll grab you if needed for the endpoint.ex issue

Issue #755, opened 2023-04-03
Project usage dashboard


taylordowns2000:
**Figma overview page:** https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=1099-5268&mode=design&t=9AetCDKXOYdq9T4m-11

We need to improve the usability and value of the Dashboard when a user logs in.

Per the old description:

>Combining elements of both the "historical usage" and "projects list" pages from v1, this interface should give a user a quick overview of the health (are things working as expected?) and activity level (at what volume are these workflows running) of their project. It _could_ be achieved by fleshing out the cards on the workflow list with some key indicators.

----

## Header section

To achieve a portion of this, we will include a few cards in the header indicating some high level info for Work Orders (formerly attempts) and runs **over the last 30 days**:

- **Card 1:** Work orders (total count of work orders over last 30 days)
- **Card 2:** Runs (aggregate of total _completed_ (excluding pending/running) runs across all work orders) with the number of pending/running runs in parentheses.
- **Card 3:** Successful Runs, incl successful run % (successful state = `Success`) in parentheses
- **Card 4:** Workorders in a failed state, incl failed state % (failed state = `Failed`, `Crashed`, `Cancelled`, `Killed`, `Exception`, `Lost`) in parentheses

On the "workorders in a failed state" card, we need to include a link which opens the History page with the correct filters applied:
- `https://app.openfn.org/projects/fdfdf286-aa8e-4c9e-a1d2-89c1e6928a2a/runs?filters[body]=true&filters[cancelled]=true&filters[crashed]=true&filters[date_after]=XXXXXX&filters[date_before]=&filters[exception]=true&filters[failed]=true&filters[killed]=true&filters[log]=true&filters[lost]=true&filters[pending]=false&filters[running]=false&filters[search_term]=&filters[success]=false&filters[wo_date_after]=&filters[wo_date_before]=&filters[workflow_id]=`
_Note: The `[date_after]` filter in this context needs to populate with the date 30 days prior to today()._

---

## Workflow overview section:

To provide users with meaningful information of their workflows it is simpler to display the data in a table.

On the table, will need to include the following info:

| Name      | Last Work Order | Work Orders (30 days) | Work Orders in a failed state (30 days)
| --------- | ---------------- | ----------------------- | ----------------------------- |
| Workflow name | status of last **completed** work order  | Total Count of work orders (excl Pending or Running) | Total count of work orders in a failed state |
[...]

PR #1555, opened 2023-12-08
Dashboard with project and workflow stats

benedictus-dev:
## Notes for the reviewer

- Displays the dashboard as home screen
- Fetches the workflows data and aggregates it to get the project metrics (derived data)

The fetching and aggregation happens for every rendering and runs or even work orders amount might be more than 10mi so on next PR this data should be aggregated on the fly (increment runs and work orders count per day as they are created) so that it's always ready to be displayed and used in the future for rate limiting.   

## Related issue

Refs #755 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/dashboard_components.ex
+++ b/lib/lightning_web/live/workflow_live/dashboard_components.ex
@@ -270,16 +270,16 @@ defmodule LightningWeb.WorkflowLive.DashboardComponents do
       <.metric_card title="Work Orders in failed state">
         <:value><%= @metrics.work_order_metrics.failed %></:value>
         <:suffix>
-          <span class="mr-10">
-            (<%= @metrics.work_order_metrics.failure_percentage %>%)
-          </span>
+          (<%= @metrics.work_order_metrics.failure_percentage %>%)
+        </:suffix>
+        <:link>
           <.link
             navigate={~p"/projects/#{@project}/runs?#{%{filters: @filters}}"}
-            class="text-indigo-700"
+            class="text-indigo-700 hover:underline"
           >
             View all
           </.link>
-        </:suffix>
+        </:link>
       </.metric_card>
     </div>
     """
@@ -289,6 +289,7 @@ defmodule LightningWeb.WorkflowLive.DashboardComponents do
   slot :value, required: true
 
   slot :suffix, required: false
+  slot :link, required: false
 
   def metric_card(assigns) do
     ~H"""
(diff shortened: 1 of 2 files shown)


lib/lightning_web/live/workflow_live/components.ex
```diff
@@ -800,4 +800,61 @@ defmodule LightningWeb.WorkflowLive.Components do
...
+            <.metric_card
+              title="Work Orders in failed state"
+              body={Map.get(value, :failed_workorders)}
+              optional_text={
+                Map.get(value, :failure_percentage) |> (to_string |> Kernel.<>("%"))
+              }
+            />
+          <% _-> %>
+        <% end %>
+      <% end %>
```
stuartc, line 838: Why is this in a loop? We have a known number of cards, and looking at the case statement, every match requires different attributes.

Also I think putting the optional text in an expression might invalidate LiveViews diffing. I'd suggest changing `optional_text` to `suffix`. _or_ better yet, use a named slot for both the main and suffix section. So instead of passing in the body as text use a slot.

```heex
<.metric_card ...>
   <:value>{@metrics.failed_wordorders}</:value>
   <:suffix>({@metrics.failure_percentage}%)</:suffix>
</.metric_card>
```

Also please don't use the `assigns` variable (or any other local variable) inside a template string. See https://hexdocs.pm/phoenix_live_view/assigns-eex.html#pitfalls

lib/lightning/dashboard_metrics.ex
```diff
@@ -0,0 +1,161 @@
...
+          {completed, pending_running + run.count}
+
+        %{exit_reason: "running"} = run, {completed, pending_running} ->
+          {completed, pending_running + run.count}
+
+        run, {completed, pending_running} ->
+          {completed + run.count, pending_running}
+      end)
+
+    %{completed_runs: completed_runs, pending_running_runs: pending_running_runs}
```
stuartc, line 68: You may be able to replace this with a filter expression, and have a count for each exit reason: 

https://hexdocs.pm/ecto/Ecto.Query.API.html#filter/2

stuartc (commented):
This is a good start, left some notes.

lib/lightning_web/live/workflow_live/components.ex
benedictus-dev: Resolved. Thank you

elias-ba (approved):
@jyeshe @benedictus-dev this is really a great piece of work. it's looking good. Congratulations 🎉

jyeshe:
> @jyeshe @benedictus-dev this is really a great piece of work. it's looking good. Congratulations 🎉
> 
> PS: I am wondering if we can improve the coverage a little bit more ?

Glad you've appreciated it Elias. The tests were important but another task was prioritized. Now that we are settled with the Sync to Github I had some time left to implement the automated tests for fetching stats. The tests for UI on the year to come (:

jyeshe:
Adapted to the requirement change to let all query data to be bound to the last 30 days period.

taylordowns2000 (changes requested):
this is great. looks like the link for "107" here doesn't include the date filter, but otherwise all looks good!

jyeshe:
@taylordowns2000, filter updated.

taylordowns2000 (approved):

PR #1557, opened 2023-12-11
Derive dataclip from attempt in inspector

### LINKED ISSUE #1551: Populate dataclip in job inspector via URL parameter
This is work in preparation for #1524

Users need to be able to click through to the Job Inspector, where the inspector opens with the correct dataclip selected. This is valuable when:
- Sharing links
- Navigating from the History page for failed runs on workorders to see the input and job code alongside one another.

image

**For implementation:**
- Parameter `i={dataclip ID}` (short for input)
- If the Dataclip ID is included in the URL, it needs to be the selected item in the drop down
- If the Dataclip ID is included in the URL, the dataclip needs to be loaded in to the JSON IDE
- If the Dataclip ID is not in the last 3 run ID's (which is what the dropdown for the dataclips are populated with), it needs to be appended to the dropdown along with the other 3 last run (ideally first position, but last position is also ok)
- If a different existing dataclip is selected from the form, the `i={dataclip_id}` should update
- If the "create new dataclip" item is selected, the URL should set to `i=new`
- If the "save and run" button is pressed when the user has entered a new JSON blob for their dataclip, the URL should be set to the newly created dataclip's ID once it exists via `i={newly_created_dataclip}`

midigofrank:
## Notes for the reviewer
- uses the `attempt` and `job` parameters to fetch the dataclip for the attempt run

## Related issue

Fixes #1551 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to
   [#1217](https://github.com/OpenFn/Lightning/issues/1217)
 - Fix Credential Creation Page UI
   [#1064](https://github.com/OpenFn/Lightning/issues/1064)
+- Populate dataclip in job inspector via URL parameter
+  [#1551](https://github.com/OpenFn/Lightning/issues/1551)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


taylordowns2000:
@stuartc , yes this new implementation works. @midigofrank should have explained this/documented the change in his PR (Frank, please take note!!!) but we made this change in a discussion together and as long as the proper dataclip gets loaded when an attempt is provided we should be just fine. i'll test that and merge later today assuming it does.

taylordowns2000:
confirmed! thanks @midigofrank !

taylordowns2000 (approved):

PR #1560, opened 2023-12-11
Preload and keep the attempts workorder and project

### LINKED ISSUE #1550: Running jobs in inspector has output section flashing with re-renders

stuartc:
When manually running a job, the output pane would flash (white-out/rerender).

This was caused by the URLs not having the required preloads (`...workflow.project` in this case).

What was happening is that we _were_ preloading correctly, however when the attempt is updated a new version of it (without preloads) replaces it and the LiveView crashes.

It settles down when the Attempt is complete and it no longer updated.

I moved the `start_async` into the shared macro, and when the Attempt is loaded I pull the workflow and project off the attempt record and keep it separately, so when the Attempt is updated we still have the project etc in state. 

## Notes for the reviewer

Run a job manually, you should see any 'flashing' or the manual run component disappearing and reappearing quickly.

## Related issue

Fixes #1550

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to
   [#1487](https://github.com/OpenFn/Lightning/issues/1487)
 - Initial credential creation `changes` show `after` as `null` rather a value
   [#1118](https://github.com/OpenFn/Lightning/issues/1118)
+- AttemptViewer flashing/rerendering when Jobs are running
+  [#1550](https://github.com/OpenFn/Lightning/issues/1550)
 
 ## [v0.11.0] - 2023-12-06
 
@@ -102,7 +104,6 @@ and this project adheres to
   [#1254](https://github.com/OpenFn/Lightning/issues/1254)
 - Fix for missing data in 'created' audit trail events for webhook auth methods
   [#1500](https://github.com/OpenFn/Lightning/issues/1500)
-
 
 ## [v0.10.4] - 2023-11-30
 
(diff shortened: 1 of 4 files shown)


jyeshe:
@stuartc would you mind sharing the validation steps (setup + navigation) for `Run a job manually`? The history page only updates the pending icon on a `rerun` and on the workflow job, the `Save and Run` is disabled.

stuartc:
@jyeshe a manual run (not the best name given some of the recent changes) is created by editing a job (the code button on the job panel) - you select or create a new input and click "Save + Run" or press "Ctrl+Enter". On the right hand side the AttemptViewer component will be rendered which "follows" an attempt and shows the updated state.

image

jyeshe:
> @jyeshe a manual run (not the best name given some of the recent changes) is created by editing a job (the code button on the job panel) - you select or create a new input and click "Save + Run" or press "Ctrl+Enter". On the right hand side the AttemptViewer component will be rendered which "follows" an attempt and shows the updated state.
> 
> image

@stuartc In my case as I have mentioned the `Save & Run` is disabled like yours too (:
That's what I have done and with the demo's first job on OpenHIE workflow:
Screenshot from 2023-12-12 17-04-22
 
Besides the navigation could you provide the setup as part of the validation steps? After checking out the branch and cleaning up the database, should I install the demo and that's enough? With which job should I try please.

jyeshe (approved):
Code LGTM and there is no random blinking anymore. Not even on the transitions.

PR #1561, opened 2023-12-12
Close Placeholder Node When Another Node Or Edge Is Clicked

### LINKED ISSUE #1220: Job name creator breaking when open & selecting an existing job
https://www.loom.com/share/4ff4e4f5e65b4151a48ffe4ee6947e54?sid=ed0da86f-956a-48ed-a9ad-e5e790a51116

- [x] Close the name creator input when selecting a existing job
- [x] See if there is a underlying bug that is a concern on the job panel not working on closing the input field

Do this with #1218

elias-ba:
## Notes for the reviewer

This PR refactors the `onSelectionChange` function in `workflow-editor/index.ts` to support the possibility of changing selection after event of type `cancel-placeholder`. The approach is when firing the event `cancel-placeholder` we send the value `cancel-placeholder-${id}` where the `id` is the `id` of the node or edge that was previously selected. That way, in our `onSelectionChange` function we can conditionnaly handle that case by doing something like this:

```
onSelectionChange(id?: string) {
    (async () => {
      console.debug('onSelectionChange', id);

      await this.hasLoaded;

      if (id?.startsWith('cancel-placeholder')) {
        this.handleCancelPlaceholder(id);
      } else {
        this.handleUrlChange(id);
      }
    })();
  }
```

## Related issue

Fixes #1220 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to
   [#1064](https://github.com/OpenFn/Lightning/issues/1064)
 - Custom metric to track Attempt queue delay
   [#1556](https://github.com/OpenFn/Lightning/issues/1556)
+- Close Placeholder Job Node When Another Node / Edge Is Clicked
+  [#1220](https://github.com/OpenFn/Lightning/issues/1220)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


elias-ba:
Closing in favor of #1569

PR #1562, opened 2023-12-12
Handle dataclip insert error when body is not a map

jyeshe:
Avoids an error on dataclip changeset when it's not a map.

Steps to reproduce on main: 
- Open any job on the workflow
- Type "1" on the dataclip
- Save & Run


taylordowns2000 (changes requested):
please rename the `redact_password` function to `remove_configuration`. otherwise looks good.

jyeshe:
Fixes needed for other tests ...

jyeshe:
@taylordowns2000 next we can remove the double `change` on `Dataclip.new`. e0733aa shows that it causes 13 errors if replaced by a `cast` for the body:
https://app.circleci.com/pipelines/github/OpenFn/Lightning/5972/workflows/8f790c21-d2c7-4927-9a9e-e3a38fd4a802/jobs/31221
Nonetheless please let me know if you want this other fix/refactor included in this PR.

taylordowns2000 (approved):

taylordowns2000:
I don't understand the other proposed changes, but this PR looks good. Please either merge this in the AM or give me a little more explanation on the other problem and how you're proposing to solve it.

stuartc:
@jyeshe good catch here! You're absolutely right, amazed we didn't run into this before.

I made a few tweaks:

- Added some tests for the Manual WorkOrder schema.
- Added `is_map` validation there so we catch it early
- And because of the above, removed the extra pattern match for the `manual_run_submit` handler on WorkflowLive.Edit.
- Oh I also tried to remove that double `change` in `Dataclip` - really not sure why it was like that, but I replaced it with a single `change` and it seems to work. 

So the idea behind using `change` is for when we create things internally, vs. `cast` for anything that comes from outside. We can skip the casting for associated models etc because the callers are behind cast changesets already.

PR #1565, opened 2023-12-13
Expand workorder row when workorder_id is specified in the filter

### LINKED ISSUE #1515: Modify filter for Work Orders on history page
**Use case:**
~At the moment, you can view all Workflows, or individual attempts - but have no way to link and share all attempts that are related to a work order. This is extremely valuable for sharing and linking between users, and more importantly being able to share the links for failed work orders in notifications.~ (Edit. This is already delivered!)

We have planned for a nice and juicy Work Order detail page (#1465), but as a quick work around it would be great to ~have a parameter on the History Page URL that filters the page to a single work order~ (Edit: already delivered) and all it's attempts expanded.

To see how the existing filter works, click into the attempt viewer and click the link to the "work order" for that attempt. Note the URL.

**To do:**

- [x] Add in parameter to the URL that when received, filters the page to one work order~ (Edit, already done.)
- [ ] If this parameter is used, expand all the attempts within the work order (this could be an additional parameter)
- [ ] At the same time as this is done, update the Failed workflow notification mailers link (see below, the "here" hyperlink) to utilize the new URL.

midigofrank:
## Notes for the reviewer

- This PR uses the `workorder_id` filter param to control the logic.
- After the `workorders` `async` task is complete, it dispatches a `send_update` to the `WorkOrderOrderComponent` using the `workorder_id` from the `filters_changeset`

## Related issue

Fixes #1515 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to
   [#1064](https://github.com/OpenFn/Lightning/issues/1064)
 - Custom metric to track Attempt queue delay
   [#1556](https://github.com/OpenFn/Lightning/issues/1556)
+- Expand workorder row when a `workorder_id` is specified in the filter
+  [#1515](https://github.com/OpenFn/Lightning/issues/1515)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


jyeshe (approved):
Code looks good and testing manually it expands as well adding it to the filter.

stuartc (approved):

PR #1566, opened 2023-12-13
Hide project security setting tab from non-authorized users

### LINKED ISSUE #1477: Improve UI when non-accessible users access project security settings
Currently users that don't have access to the MFA page attached (assuming by role, this screenshot was created by an `Editor`) 

Possible solutions:
- Hide menu item for MFA on navigation and 404 if user tries to access it

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjRUQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--146d90643b27ee3ca882b6aa5ff088e188760ddd/image.png)

midigofrank:
## Related issue

Fixes #1477

## Review checklist

### DIFF
--- a/lib/lightning_web/live/project_live/settings.html.heex
+++ b/lib/lightning_web/live/project_live/settings.html.heex
@@ -42,6 +42,7 @@
               <span class="inline-block align-middle">Collaboration</span>
             </LightningWeb.Components.Common.tab_item>
             <LightningWeb.Components.Common.tab_item
+              :if={can_edit_project(assigns)}
               orientation="vertical"
               hash="security"
             >
@@ -473,7 +474,10 @@
           </.table>
         </LightningWeb.Components.Common.panel_content>
 
-        <LightningWeb.Components.Common.panel_content for_hash="security">
+        <LightningWeb.Components.Common.panel_content
+          :if={can_edit_project(assigns)}
+          for_hash="security"
+        >
           <div>
             <h6 class="font-medium text-black">Project Security</h6>
             <small class="block my-1 text-xs text-gray-600">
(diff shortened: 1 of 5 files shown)


elias-ba (changes requested):
Hey @midigofrank this is looking really great. I just could spot that when you enter the security tab URL in the browser (`http://localhost:4000/projects/4adf2644-ed4e-4f97-a24c-ab35b3cb1efa/settings#security`) it renders this blank page.

image

test/support/conn_case.ex
```diff
@@ -132,24 +132,19 @@ defmodule LightningWeb.ConnCase do
   Setup helper that creates a user adds them as project user with a given role and logs them them in
   """
   def setup_project_user(conn, project, role) do
-    user = Lightning.AccountsFixtures.user_fixture()
+    user = Lightning.Factories.insert(:user)
```
jyeshe, line 135: great you've replaced! importing would look like more with the rest of the code

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -473,7 +474,10 @@
           </.table>
         </LightningWeb.Components.Common.panel_content>
 
-        <LightningWeb.Components.Common.panel_content for_hash="security">
+        <LightningWeb.Components.Common.panel_content
+          :if={can_edit_project(assigns)}
```
jyeshe, line 478: :+1:

jyeshe (approved):
Even pasting the url it doesn't show the security option :clap: 
Small detail on the testing just to keep the same style with the rest of the code.

elias-ba (approved):
@midigofrank this is looking great. Nice job 🎉

taylordowns2000 (commented):
@midigofrank , @jyeshe , i've merged this down but want to note two things:
1. there's a commented out console.log that should be removed
2. i don't think this is the quickest fix for the desired UX. we want non admin users to be able to view this interface but for the MFA toggle to be `disabled`. they can _see_ whether it's enabled or not, but not modify it. is that harder to do than hiding the entire page? (hiding the page is _not_ what we want from a UX perspective.)

jyeshe:
> @midigofrank , @jyeshe , i've merged this down but want to note two things:
> 
> 1. there's a commented out console.log that should be removed
> 2. i don't think this is the quickest fix for the desired UX. we want non admin users to be able to view this interface but for the MFA toggle to be `disabled`. they can _see_ whether it's enabled or not, but not modify it. is that harder to do than hiding the entire page? (hiding the page is _not_ what we want from a UX perspective.) do we have some patterns elsewhere in the app where we make certain buttons or inputs `disabled`?

Hi @taylordowns2000 it was suggested to be hidden on the issue #1477 and I would choose this customization for non-admin as we are preventing to show options to users that they cannot/should not operate.

taylordowns2000:
Ah, but we need to actually *SHOW* this stuff to folks who can’t modify it.
No sweat, just a miscommunication. Follow up with Nick in the new years if
you’ve got questions. We’ll open a new issue!

Taylor Downs
CEO, OpenFn <https://www.openfn.org>

PR #1569, opened 2023-12-14
Conform all Diagram selections to the URL

### LINKED ISSUE #1220: Job name creator breaking when open & selecting an existing job

stuartc:
## Notes for the reviewer

In order to reproduce this _before_ these changes by clicking "+" on a Job node, and cancelling the placeholder (clicking "X"); the very next Node you click won't respond to being selected.

This is a set of fixes which changes selection behaviour in the WorkflowDiagram.

- The currently selected id is now derived only from the URL.
- The switching appears immediately because we pull the selection off the URL from the `navigate` event (i.e. before the liveview has to reply).
- This change appears to reduce render calls and fixes what looks like the node being selected twice.
- I've changed the logic for cancelling a placeholder, if you hit Esc, it cancels it.

Other changes include a CSS change that moves the opacity animation when LiveView is processing a click event (specifically when switching Nodes) to apply the animation only to the form and not the entire diagram.

## Related issue

Fixes #1220 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ and this project adheres to
   [#1550](https://github.com/OpenFn/Lightning/issues/1550)
 - Not able to create a new Job when clicking the Check icon on the placeholder
   [#1537](https://github.com/OpenFn/Lightning/issues/1537)
+- Improve selection logic on WorkflowDiagram
+  [#1220](https://github.com/OpenFn/Lightning/issues/1220)
 
 ## [v0.11.0] - 2023-12-06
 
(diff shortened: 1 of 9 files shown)


elias-ba (approved):
@stuartc this is gold. I played with the workflow diagram for about half an hour and it looks very smooth now. Thank you so much

PR #1573, opened 2023-12-14
Avoid error component when js expression is selected

### LINKED ISSUE #1498: Allow Javascript expressions as conditions for edges
**Figma:** https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=1086-3945&mode=design&t=9AetCDKXOYdq9T4m-11

We will now include a conditional match where the user can compare the output of the previous job against a Javascript expressions. 

**Edge panel:**
- Include new dropdown item "Matches a Javascript Expression" in the drop down list for conditions
- When selected, new input for "Condition label" (more on that below)
- Include lightweight code input (not sure which existing code input we should use here or just style a normal input) for the JS expression
- Add a link to a Javascript Expression checker under the input (open to whatever online tool seems best!)

**Canvas:**
- When the "Matches a Javascript Expression" is selected, we will display a label on the path which takes the text from the "Condition label" instead of an icon.
- There needs to be a max width, followed by ellipsis (I feel 16 char is probably the right length)
- If ellipsis are displayed, a tooltip should be shown on hover of the element.

**Acceptance criteria:**
- [ ] Option displayed in drop down list for conditions is correct
- [ ] The edges following job always runs when matching the regular expression
- [ ] The label is displayed
- [ ] New features included in documentation 

Noting:
- Will use same input as the input dataclip
- Will do basic validation that the query is valid, and no typeahead.
- In future will move to Monaco

jyeshe:
## Notes for the reviewer

The error component is rendered before the user starts filling the data since the changeset is initialized with required errors after selecting the condition `Matches a Javascript Expression`:
```
errors: [
          condition_label: {"can't be blank", [validation: :required]},
          condition_expression: {"can't be blank", [validation: :required]}
        ]
```
The fix changes the UI when editing an edge from this:
Screenshot from 2023-12-14 18-07-01
to this:
Screenshot from 2023-12-14 18-04-35

## Related issue

Closes #1498

## Review checklist

### DIFF
--- a/lib/lightning_web/components/new_inputs.ex
+++ b/lib/lightning_web/components/new_inputs.ex
@@ -303,7 +303,6 @@ defmodule LightningWeb.Components.NewInputs do
         ]}
         {@rest}
       />
-
       <div :if={Enum.any?(@errors)} class="error-space h-6">
         <.error :for={msg <- @errors}><%= msg %></.error>
       </div>
@@ -338,7 +337,7 @@ defmodule LightningWeb.Components.NewInputs do
     ~H"""
     <p
       data-tag="error_message"
-      class="mt-3 inline-flex items-center gap-x-1.5 text-xs text-danger-600 phx-no-feedback:hidden"
+      class="mt-3 inline-flex items-center gap-x-1.5 text-xs text-danger-600"
     >
       <.icon name="hero-exclamation-circle" class="h-4 w-4" />
       <%= render_slot(@inner_block) %>
--- a/lib/lightning_web/live/workflow_live/components.ex
+++ b/lib/lightning_web/live/workflow_live/components.ex
@@ -538,7 +538,7 @@ defmodule LightningWeb.WorkflowLive.Components do
         </div>
       <% end %>
       <%= if Phoenix.HTML.Form.input_value(@form, :source_trigger_id) do %>
-        <div class="max-w-xl text-sm text-gray-500 mt-2">
+        <div class="max-w-xl text-sm text-gray-500 mt-3">
           <p>This path will be active if its trigger is enabled.</p>
         </div>
       <% else %>
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -886,7 +886,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     end
   end
 
-  # Returns the index of the edge that is being edited on Path form
+  # Returns the inputs that have being edited on the form
   defp get_params_opts_for(workflow_attribute, params) do
     params
     |> Map.get(workflow_attribute, %{})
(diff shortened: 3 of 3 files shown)


jyeshe:
Last commit for showing the errors below the input fields:
Screenshot from 2023-12-14 18-54-00

lib/lightning_web/components/new_inputs.ex
```diff
@@ -338,7 +337,7 @@ defmodule LightningWeb.Components.NewInputs do
     ~H"""
     <p
       data-tag="error_message"
-      class="mt-3 inline-flex items-center gap-x-1.5 text-xs text-danger-600 phx-no-feedback:hidden"
```
elias-ba, line 341: @jyeshe I am wondering if this related to what I am describing here

https://www.loom.com/share/31b4520022034160bee2e7b3f25b5ee2?sid=0b78c88c-e622-4e86-8c95-dc5b474cfc39

elias-ba (changes requested):
@jyeshe this is really great. I attached a loom video and was curious if the line removed there was causing that validation to happen even when the textarea is not focused.

lib/lightning_web/components/new_inputs.ex
jyeshe: Hey @elias-ba this nice improvement. I think we have it now in a generalized way and  customizable for other forms.

jyeshe:
to be rebased ...

jyeshe:
@elias-ba @taylordowns2000 please find it here rebased is done (the main feature and rename refactor were merged)

lib/lightning_web/components/new_inputs.ex
elias-ba: Hey @jyeshe I just tested this again and it feels like we are no longer able to produce errors. I am wondering if this is what we want or I am missing something. See loom: https://www.loom.com/share/03543485934046bd8cc8d76d47390b0e?sid=b39eedfa-5212-4a44-81c9-cc310242986f

elias-ba (changes requested):
Hey @jyeshe I am wondering if my second loom video is relevant. But globally it all looks great. I just can't see errors rendered by the form when I type something invalid. Is this what we want ?

jyeshe:
> Hey @jyeshe I am wondering if my second loom video is relevant. But globally it all looks great. I just can't see errors rendered by the form when I type something invalid. Is this what we want ?

@elias-ba thanks for catching it. From the previous loom I have taken the suggestion to not show errors for fields that were not edited yet. But indeed, if the user deletes what has already typed on previous field, it would not show. So I have applied a rule for this form since it's already an improvement from the initial scope.

lib/lightning_web/components/new_inputs.ex
jyeshe: It would show now for fields that were edited and deleted. For fields that were not edited yet it applies the suggestion from previous loom.

elias-ba (approved):
Congratulations @jyeshe, this is looking great. :tada:

PR #1587, opened 2023-12-21
KeyError LightningWeb.RunLive.Index.handle_event/3

jyeshe:
## Notes for the reviewer

Validate if work order is present on toggle selection.

## Related issue

Fix for https://openfn.sentry.io/issues/4661748768/?project=6446500&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=2&utc=true

## Review checklist

### DIFF
--- a/lib/lightning_web/live/run_live/index.ex
+++ b/lib/lightning_web/live/run_live/index.ex
@@ -321,28 +321,24 @@ defmodule LightningWeb.RunLive.Index do
           "workorder_id" => workorder_id,
           "selected" => selected?
         },
-        %{assigns: assigns} = socket
+        socket
       ) do
+    %{page: page, selected_work_orders: selected_work_orders} = socket.assigns
+
     selected? = String.to_existing_atom(selected?)
+    workorder = Enum.find(page.entries, &(&1.id == workorder_id))
 
     work_orders =
-      if selected? do
-        workorder =
-          Enum.find(assigns.page.entries, fn wo ->
-            wo.id == workorder_id
-          end)
-
+      if selected? and nil != workorder do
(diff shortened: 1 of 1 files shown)


lib/lightning_web/live/run_live/index.ex
```diff
@@ -321,28 +321,24 @@ defmodule LightningWeb.RunLive.Index do
...
+    workorder = Enum.find(page.entries, &(&1.id == workorder_id))
 
     work_orders =
-      if selected? do
-        workorder =
-          Enum.find(assigns.page.entries, fn wo ->
-            wo.id == workorder_id
-          end)
-
+      if selected? and nil != workorder do
```
elias-ba, line 332: Wondering if we need this double condition as `nil` evaluates to false. I mean couldn't we write the condition like this: `if selected? and workorder do` ?

elias-ba (approved):
Hey @jyeshe this is looking great, thanks a lot.

PS: please see my comment on the `nil != workorder` clause and see if it is relevant enough to you for change.

lib/lightning_web/live/run_live/index.ex
jyeshe: It's not a double condition, it's just a more readable expression `if selected? and nil != workorder do` instead of `if selected? and nil != workorder do` calls the attention to the desired check.

elias-ba: > `nil != workorder` is not a double condition, it just composes a more readable expression where `if selected? and nil != workorder do` instead of `if selected? and workorder do` calls the attention to the desired check.

Thanks for clarifying this Roger. Looks good

PR #1598, opened 2024-01-08
key-bindings for save/retry getting overriden when user focuses on the job editor

### LINKED ISSUE #1596: Work Orders getting accidentally duplicated when "rerun from here" is pressed
Check out this video: https://www.loom.com/share/0926d3aa804745a7b6614cfe34fcb088

To recap the desired behavior of the inspector view:
1. When an admin accesses the inspector with a run loaded (as I do in the video) they're most likely _debugging_ and will want to add `console.log` statements and re-run the work order multiple times from whatever step they're on.
2. Keep a close eye on the buttons in the bottom right corner... there are two of them. A "primary action button" (which in this video is supposed to "Rerun from here") and a "save" button (which has a helpful red-dot-indicator that will tell us if the job code has been saved).
3. Pressing `Cntrl-S`/`⌘-S` will save the job code.
4. Pressing `Cntrl-Enter`/`⌘-Enter` will press the primary action button. (In this case, that should be to "Rerun from here".)
5. There is also a "secondary action button" that can be reached via the split-button-select-list or by pressing `Cntrl-Shift-Enter` but that button is not pressed in this video. The secondary action would create a new work order, rather than only creating a new run for the existing work order.
6. Try to get a sense for what I'm doing here... I want to rerun this work order lots of times while I'm debugging and modifying the job code... all to try to _eventually_ process this input dataclip (in this case, a Stripe transaction) properly. I'd like to keep running this thing dozens of times after making small adjustments, and eventually see a successful run. When that's done, I'll click back to the history view and see that this work order is in a "success" state, even if there are two dozen runs (previously called _attempts_) associated with it.

The bug I'm seeing here is three-fold:
1. In this video it sometimes appears that the job code is not saved when I press/click "Rerun from here". This is not good. We always want to "save _and_ run" the job rather than running it without saving.
2. In this video we also see that despite always clicking/pressing "Rerun from here" a _new work order_ gets generated a few times. I thought this might be happening when the job code was edited, but it seems to be less predictable than that.
3. _EDIT:_ After some more digging, it appears that at some point in this video a new (duplicate) dataclip was generated. That shouldn't be happening and feels linked to the duplicate work order.

midigofrank:
## Notes for the reviewer

- added commands in the monaco editor
- introduced a `common.js` for storing common functions.

## Related issue

Fixes #1596

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to
   [a6e4ada](https://github.com/OpenFn/Lightning/commit/a6e4adafd558269cfd690e7c4fdd8f9fe66c5f62)
 - Fixing attempt/run language for "skipped" tooltip on inspector.
   [fd7dd0c](https://github.com/OpenFn/Lightning/commit/fd7dd0ca8128dfba2902e5aa6a2259e2073f0f10)
+- Fixed key-bindings for save/retry getting overrided when user focuses on the monaco job editor
+  [#1596](https://github.com/OpenFn/Lightning/issues/1596)
 
 ## [2.0.0-rc1] - 2023-01-05
 
--- a/assets/js/common.js
+++ b/assets/js/common.js
@@ -1,0 +1,15 @@
+// this has been added here because we still can't find a better place for it
+// it's being used in both the hooks and the job editor
+export function retryOrCreateWorkOrder(buttonElement) {
+  if (buttonElement.getAttribute('type') == 'submit') {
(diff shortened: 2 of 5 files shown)


elias-ba (approved):
This is looking great @midigofrank 🎉

taylordowns2000 (changes requested):
almost there! good progress but saves aren't coming through: https://www.loom.com/share/bc92c54ea4794cb5b3b3ae0550329794

taylordowns2000 (approved):

PR #1603, opened 2024-01-09
3 instances of state to status

taylordowns2000:
@NickOpenFn , we'd already been calling it "Status" in the history page filter. This amends the run viewer and inspector.


taylordowns2000:
approved by @NickOpenFn

PR #1614, opened 2024-01-11
add search by ID on the history page

### LINKED ISSUE #1468: Improve Search on History page
Extend queries that can be searched:
- [x]  Attempt/Run Logs
- [x]  Dataclip Body
- [x]  `workorder_id` (work order id)
- [x]  `attempt_id` (run id)
- [x]  `run_id` (step run id)

Implementation:

- [x] Update search bar to change "Both" to "Any"
- [x] Update search params to search across Workorder, Attempt, or Run UUID when "Any" is selected
- [x] Add "ID" selector to search bar for searching across only Workorder, Attempt, or Run UUID

image

Note that we'll need a new toggle here (see image above) to search only on ID or to search "Any" rather than "Both"

## Later?
- [ ]  Workflow names
- [ ]  etc...
- [ ] Error state
- [ ] No search results (another card exists for this)

midigofrank:
## Notes for the reviewer

- Builds on top of the existing `search_fields` `query`

### TODO

- [x] Do away with the fragment, use ecto functions `ilike/2, like/2, type/2`
- [x] Use `like/2` instead of `ilike/2` when matching `id` because `ids` are case sensitive

## Related issue

Fixes #1468 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to
 
 - Dashboard with project and workflow stats
   [#755](https://github.com/OpenFn/Lightning/issues/755)
+- Add search by ID on the history page
+  [#1468](https://github.com/OpenFn/Lightning/issues/1468)
 
 ### Changed
 - Give project editors and viewers read only access to project settings instead
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -8,7 +8,7 @@
       "name": "lightning",
       "license": "LGPLv3",
       "dependencies": {
-        "@heroicons/react": "^2.0.16",
+        "@heroicons/react": "^2.1.1",
         "@monaco-editor/react": "^4.4.5",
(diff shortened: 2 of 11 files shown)


taylordowns2000 (changes requested):
@midigofrank , this is great but i've realized that the buttons aren't behaving optimally.

- see my "todo".
- note that when you first go to this page, both the "input" and the "logs" search buttons are selected. that's nice, as these are meant to be a multi-select.
- then, note that when you click on a button, it _only_ selects that one search attribute.
- can you re-write this interface so that you can multi-select the 3 buttons, searching with any combination of the three?
- if _none_ are selected, the "search" button should be disabled
- if _all_ are selected, we should search all.
- there should only be 3 buttons when you're done... you can remove the commented out code in my last commit.

taylordowns2000 (approved):

PR #1617, opened 2024-01-11
Scrub run_result dataclip on dataclip viewer

### LINKED ISSUE #1606: Scrub dataclips during presentation
Currently, we may expose secrets in the run results in the output dataclip. We will be implementing a rule that limits what the runtime writes to the run_result, but would like an additional failsafe of scrubbing this before displaying it to the user on the presentation layer:

Per the Slack conversation from @taylordowns2000: https://openfn.slack.com/archives/C0394D0GBUN/p1704798081537769?thread_ts=1703149940.085049&cid=C0394D0GBUN

>just finished the call with Mtuchi and Joe around secrets getting stored in run_result dataclips. We think that (if the additional overhead isn’t too high) it makes sense to scrub dataclips during presentation (i.e., after the liveview gets the JSON from the db, after it converts it to a string, and finally right before it renders it in the code-viewers on run_viewer, attempts_live, and the dataclip viewer.
In addition to this, we’ll handle the actual object in the runtime, but it’s a very good failsafe to scrub the strings right before they’re sent to the HTML view. What do you think? (Roger, is this a terrible idea from a performance perspective? Could we hook up the existing scrubber (using the secrets list from an associated attempt) to scrub dataclips with type run_result once they’re converted to text and are about to be viewed as HTML?)

Notes:
- Scrubbing should occur after the liveview receives the JSON from the database.
- Scrubbing should take place after the JSON is converted to a string.
- Integration with the existing scrubber should be implemented to utilize the secrets list from the associated attempt for scrubbing dataclips with type run_result.

Acceptance requirement:
- [x] All sensitive information in the run_result dataclips during the presentation layer should be scrubbed (duh!)
- [x] Need to ensure that the page load speed and performance overhead doesn't affect the user or application performance.
- [ ] Areas to check where run result is displayed:
  - [x] Inspector `Output and Logs` `Output` section
  - [x] Run detail page `Output` (https://demo.openfn.org/projects/4adf2644-ed4e-4f97-a24c-ab35b3cb1efa/attempts/e419d937-0d2c-4df1-b4f0-0e8ffee9140a?r=1f20434d-40b2-4ec8-9a25-f07e54cfda36#output)

jyeshe:
## Notes for the reviewer

Scrub output dataclip when it is a `run_result` and has an associated credential.
Applies the scrubbing to the Job inspector and Attempt output/dataclip viewer.

## Related issue

Fixes #1606 

## Review checklist


stuartc (approved):
@jyeshe nice one!

test/lightning_web/live/attempt_live/streaming_test.exs
```diff
@@ -0,0 +1,139 @@
...
+
+      # foo: "bar" is not scrubbed because it is from a following job executed on run3
+      expected_lines = [
+        ~S("integer":***),
+        ~S("another_no":***),
+        ~S("third_no":12***34),
+        ~S("map":{"list":[{"any-key":"some-***s"}]}),
+        ~S("bool":true),
+        ~S("foo":"bar")
+      ]
```
jyeshe, line 130: @taylordowns2000 ready with logic for the selection of credentials! validated here.

taylordowns2000 (approved):

PR #1624, opened 2024-01-15
Fix get_env/3 deprecation warning by using compile_env/3 instead

elias-ba:
## Notes for the reviewer

This PR changes `Application.get_env/3` to `Application.compile_env/3` to avoid the deprecation warning we see when we launch the server using: `iex -S mix phx.Server`

image

## Related issue

Fixes #

## Review checklist

### DIFF
--- a/lib/lightning_web/endpoint.ex
+++ b/lib/lightning_web/endpoint.ex
@@ -59,7 +59,8 @@ defmodule LightningWeb.Endpoint do
       :multipart,
       # Increase to 10MB max request size only for JSON parser
       {:json,
-       length: Application.get_env(:lightning, :max_dataclip_size, 10_000_000)}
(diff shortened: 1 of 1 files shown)


taylordowns2000 (changes requested):
let's keep the conversation here around this https://github.com/OpenFn/Lightning/pull/1541

stuartc:
@elias-ba I second what @taylordowns2000 said, while you're right to want to change this - I think it was implemented with the expectation we can change it at runtime; which we would lose if we used the compile envs (we'd have to build the image again in order to change it).

elias-ba:
@taylordowns2000 @stuartc thank you so much for clarifying this to me. It looks like this is under your radars which is cool 😄

PR #1635, opened 2024-01-17
do not persist the active tab setting to localStorage

midigofrank:
## Related issue

Fixes #

## Review checklist


taylordowns2000 (changes requested):
putting this back to status as "enqueued" isn't a valid result: `<:label>Result</:label>`

PR #1641, opened 2024-01-18
Add moduledoc to remaining modules

jyeshe:
Solves `mix credo --strict --all` Code Readability suggestions for `@moduledoc`s


taylordowns2000 (changes requested):
looks good but please also update `.credo.exs` to now _ENFORCE_ @moduledoc via CI. do this by removing `exit_status` from https://github.com/OpenFn/Lightning/blob/main/.credo.exs#L108

taylordowns2000 (approved):

PR #1643, opened 2024-01-18
Soft delete jobs

### LINKED ISSUE #1570: Don't hide deleted Jobs on history page from past attempts (soft deletion of Jobs)
We have an issue where deleted jobs now no longer display on the attempt, which is incorrect from an audit perspective and the fact that an attempt should in essence be immutable.

There are many suggestions on improving this (such as versioning), but the fastest solution to solve the immediate issue would be to having soft deletion.

When a job is soft deleted:
- It won't display in the canvas
- The job won't be sent to the worker and processed when a Work Order is run
- The history page will display attempts of all **run** jobs of an attempt, even if it has been deleted.
  - If the job has been deleted, they should be able to access the Job detail page by selecting the job (they would need to be able to see the input/output/logs)
  - The user should not be able to access the job editor page via the link done in #1524 (naming not defined yet, hence the reference to the task)

elias-ba:
## Related issue

Fixes #1570 

## Review checklist


lib/lightning/helpers.ex
```diff
@@ -60,6 +60,10 @@ defmodule Lightning.Helpers do
     data |> Map.from_struct() |> json_safe()
   end
 
+  def json_safe(%{microsecond: _} = map) do
```
elias-ba, line 63: @stuartc and this too. I am plucking out the microsecond tuple to allow json_safe not to break

PR #1652, opened 2024-01-19
Fix http endpoint for docker compose

### LINKED ISSUE #1653: Fix Run via Docker
Following these README [steps](https://github.com/OpenFn/Lightning?tab=readme-ov-file#run-via-docker), the `docker compose` with default dockerfile (`Dockerfile-dev`) listens to 4000 but doesn't accept browser or worker connections:
```
Connected to localhost (127.0.0.1) port 4000 (#0)
> GET / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Recv failure: Connection reset by peer
```
and
```
worker-1    | [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://web:4000/worker
worker-1    | [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://web:4000/worker
worker-1    | [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://web:4000/worker
worker-1    | [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://web:4000/worker
```  
Even though it listens to 4000:
```
$ netstat -nat | grep LISTEN
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN     
tcp        0      0 127.0.0.1:631           0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:4000            0.0.0.0:*               LISTEN     
tcp6       0      0 ::1:631                 :::*                    LISTEN     
tcp6       0      0 :::80                   :::*                    LISTEN     
tcp6       0      0 :::4000  
```

**Version number** 
v2.0.0-rc3

- [x] Yes
- [ ] No

**To Reproduce** 

Follow these README [instructions](https://github.com/OpenFn/Lightning?tab=readme-ov-file#run-via-docker) and check that localhost:4000 is not available on browser.

jyeshe:
## Notes for the reviewer

The change is that it requires to listen to "0.0.0.0" for `localhost`. 

Please find error details on related issue.

## Related issue

Fixes #1653 

## Review checklist


taylordowns2000 (commented):
i don't want to approve this explicitly and will defer to @stuartc . i know there was some problem related to the loopback address (127 vs localhost?) and i don't know enough about this stuff to say whether this impacts that, or vice versa.

stuartc (approved):
This is fine for now, but there is a reason we default to `127.0.0.1`. If someone is on an open network, port 4000 is visible to _anyone_, whereas with only the loopback address only the same machine running Lightning can access it. 

So while Docker containers would need at least the containers IP - or `0.0.0.0`, making that the default for when running on an interface that is routable from a wifi network isn't ideal.

We should be able to use `LISTEN_ADDRESS` in both prod and dev modes. I'll make a follow up issue.

PR #1658, opened 2024-01-22
Build Backend functionality for Zero-persistence

### LINKED ISSUE #1212: Zero-persistence: Build Backend functionality for Zero-persistence
Once users enable zero-persistence from the settings section(#1209), we need to ensure that the data is not persisted (duh!).

**Worker:**
Extend worker to support Zero-persistence:
- https://app.zenhub.com/workspaces/kit-64d36072fcd9fb4aa00a3db9/issues/gh/openfn/kit/562

**Lightning:**
- Admin settings, need to save retention period and one of `n` (3 right now) options for zero-persistence
- In the websocket handler, decide to save or not save `dataclips`.
- When we send a plan to the worker, add an attribute to request that the workers _NOT_ send back `run_results` 
- We are still broadcasting logs as they arrive. (NO CHANGE TO LOGS!)
- "tidy" dataclip _after_ either the dataclip has been 'used':
    - either after 'fetch:dataclip'
    - or after job is finished/orphaned/missing
 - relax dependency on `input_dataclip_id`
    
Still tbd:
- Define what "tidy" means:
    - i.e. we are keeping the dataclip records, but not the body?
    - 

Acceptance criteria:

- [ ] See new attribute in `plan` that instructs worker to send/not-send `run_results`
- [ ] After a `http_request` dataclip is used in a workorder, if I view that dataclip I can see that the `body` of the webhook request is no longer visible. (redacted, tidied, something)
- [ ] After a workorder is executed, if i view the workorders in a proejct where this setting is enabled (via the DB, there's no frontend yet!) there are no `run_results` created for that work order.
- [ ] When running with an old version of the worker which doesn't yet know about "zero persistence" we will receive `run_results` over the wire, but we will NOT persist them.

midigofrank:
## Notes for the reviewer

- [x] assign `retention_policy` to the attempt channel `socket` on join
- [x] Add `include_run_results` flag to the worker plan. This is controlled by `assigns.retention_policy != :erase_all`
- [x] Wipe dataclip on `fetch:dataclip`
- [x] Drop `input_dataclip_id`, `output_dataclip` and `output_dataclip_id` fields from the payload `if assigns.retention_policy != :erase_all`
- [x] Fix regression on the attempt viewer when there's no `input_dataclip`
- [x] Test that if zero-persistence is enabled, the dataclips are not saved
- [x] *REMEMBER* to drop the `retention_policy` field from the migration before merge. This has already been added in #1662 

## Related issue

Fixes #1212

## Review checklist

### DIFF
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -245,6 +245,20 @@ defmodule Lightning.Projects do
     end)
   end
 
+  @spec project_retention_policy_for(Attempt.t()) ::
+          :retain_all | :erase_all | :erase_successful_only
+  def project_retention_policy_for(%Attempt{work_order_id: wo_id}) do
+    query =
+      from(wo in WorkOrder,
+        join: wf in assoc(wo, :workflow),
+        join: p in assoc(wf, :project),
+        where: wo.id == ^wo_id,
+        select: p.retention_policy
+      )
+
+    Repo.one(query)
+  end
+
   def project_attempts_query(project) do
     from(att in Attempt,
       join: wo in assoc(att, :work_order),
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -10,7 +10,7 @@ defmodule LightningWeb.AttemptChannel do
   alias Lightning.Repo
   alias Lightning.Scrubber
   alias Lightning.Workers
-  alias LightningWeb.AttemptJson
+  alias LightningWeb.RunWithOptions
 
   require Jason.Helpers
   require Logger
@@ -50,12 +50,11 @@ defmodule LightningWeb.AttemptChannel do
   end
 
   @impl true
-  def handle_in("fetch:attempt", _, socket) do
+  def handle_in("fetch:attempt", _, %{assigns: assigns} = socket) do
     {:reply,
      {:ok,
-      AttemptJson.render(socket.assigns.attempt,
-        include_run_results:
-          include_run_results?(socket.assigns.retention_policy)
+      RunWithOptions.render(assigns.attempt,
+        output_dataclips: include_output_dataclips?(assigns.retention_policy)
       )}, socket}
   end
 
@@ -221,7 +220,7 @@ defmodule LightningWeb.AttemptChannel do
     )
   end
 
-  defp include_run_results?(retention_policy) do
+  defp include_output_dataclips?(retention_policy) do
     retention_policy != :erase_all
   end
 
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -32,7 +32,8 @@ defmodule LightningWeb.AttemptChannel do
          id: id,
          attempt: attempt,
          project_id: project_id,
-         scrubber: nil
+         scrubber: nil,
+         include_run_results: true
        })}
     else
       {:error, :not_found} ->
@@ -49,7 +50,11 @@ defmodule LightningWeb.AttemptChannel do
 
   @impl true
   def handle_in("fetch:attempt", _, socket) do
-    {:reply, {:ok, AttemptJson.render(socket.assigns.attempt)}, socket}
+    {:reply,
+     {:ok,
+      AttemptJson.render(socket.assigns.attempt,
+        include_run_results: socket.assigns.include_run_results
+      )}, socket}
   end
 
   def handle_in("attempt:start", _, socket) do
--- a/lib/lightning_web/channels/attempt_channel.ex
+++ b/lib/lightning_web/channels/attempt_channel.ex
@@ -6,6 +6,7 @@ defmodule LightningWeb.AttemptChannel do
 
   alias Lightning.Attempts
   alias Lightning.Credentials
+  alias Lightning.Projects
   alias Lightning.Repo
   alias Lightning.Scrubber
   alias Lightning.Workers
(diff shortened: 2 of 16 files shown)


lib/lightning/projects.ex
```diff
@@ -245,6 +245,20 @@ defmodule Lightning.Projects do
     end)
   end
 
+  @spec project_retention_policy_for(Attempt.t()) ::
+          :retain_all | :erase_all | :erase_successful_only
+  def project_retention_policy_for(%Attempt{work_order_id: wo_id}) do
```
taylordowns2000, line 250: note that we need to change this name so it's not confused with `retention_period`

lib/lightning/projects.ex
```diff
@@ -245,6 +245,20 @@ defmodule Lightning.Projects do
...
+    query =
+      from(wo in WorkOrder,
+        join: wf in assoc(wo, :workflow),
+        join: p in assoc(wf, :project),
+        where: wo.id == ^wo_id,
+        select: p.retention_policy
```
taylordowns2000, line 256: same here. will consider names here later

lib/lightning_web/channels/attempt_channel.ex
```diff
@@ -49,7 +51,12 @@ defmodule LightningWeb.AttemptChannel do
 
   @impl true
   def handle_in("fetch:attempt", _, socket) do
-    {:reply, {:ok, AttemptJson.render(socket.assigns.attempt)}, socket}
+    {:reply,
+     {:ok,
+      AttemptJson.render(socket.assigns.attempt,
```
taylordowns2000, line 56: @stuartc , is this how we expect to send an attribute like `include_run_results` to the worker in the plan? I'd expect this to be next to other attributes like `allow_console_logging`, etc.

lib/lightning_web/channels/attempt_json.ex
```diff
@@ -7,15 +7,16 @@ defmodule LightningWeb.AttemptJson do
   alias Lightning.Workflows.Job
   alias Lightning.Workflows.Trigger
 
-  def render(%Attempt{} = attempt) do
+  def render(%Attempt{} = attempt, include_run_results: include_run_results) do
```
taylordowns2000, line 10: i don't love this @midigofrank . we will be passing _lots_ of options to the worker based on project settings. they may include:

1. `include_run_results`
2. `allow_console_log`
3. `other_things_i_cant_remember_right_now`

I wonder if whether or not zero-persistance was enabled should be stamped onto the attempt itself?

taylordowns2000 (changes requested):
this is fantastic @midigofrank . i'd like @stuartc 's guidance on how we build the plan we're sending to the ws-worker.... see inline comments. otherwise all good.

lib/lightning_web/channels/attempt_json.ex
midigofrank: Good point. I was trying to imitate `render(conn, assigns)`.

The only challenge I can see is that the `Attempt` is the `plan`.
What if we merge them in later?, like:
```elixir
plan = AttemptJson.render(attempt)
Map.merge(plan, opts)
```

stuartc: I'm glad you guys are talking about this. It's a moment where we are departing from Attempt into what the worker refers (at least in spirit) as a Plan.

I'm happy with adding a field to the payload we send, like: 

```
{ ...,  // current "AttemptJson"
  options: { 
    output_dataclips: false | true,
    ... // room for others
  }
}
```

However this is (from a Lightning perspective) not an `Attempt` anymore, even though it's 99% the same.

I would suggest renaming `AttemptJson` to `PlanJson` or `RunPlanJson` (because Attempts are going to be called Runs everywhere soon).

The module would take an Attempt model and any other arguments and return the new payload.

I think @josephjclark should chime in here about potential contract changes; what does a plan look like in the worker? Would it make sense for us to use that word in Lightning given the example above?

josephjclark: Well, first off, as Stu says, I expect the "attempt" coming from Lightning to have an `options` key with a bunch of attempt-specific configuration on it.

I am perfectly happy with this setup. I don't care how Lightning assembles this Attempt object, but the JSON that is sent to me right now is convenient, effective and scalable. I'm not in a rush to break that contract.

It's worth noting that I actually internally split the Attempt it into two objects - an ExecutionPlan and Options.

An Attempt/Run is a Lightning abstraction. It doesn't exist in the engine or runtime - although the ExecutionPlan is broadly equivalent.

However, if Lightning had the ability to take a WorkOrder and generate an ExecutionPlan (with `options`), we could a) send that verbatim to the worker, and b) download it locally to run in the CLI (maybe the CLI doesn't need the options?).

I don't care about a) but I think b) is neat, so I would vote in favour.

josephjclark: Here is the Attempt -> ExecutionPlan conversion code: https://github.com/OpenFn/kit/blob/main/packages/ws-worker/src/util/convert-attempt.ts

Here is the ExecutionPlan typedef (it's a big fragmented): https://github.com/OpenFn/kit/blob/df38b06b50665131cb2ec0c842af544e6c74ed21/packages/runtime/src/types.ts#L56

And here are the attempt options so far (not much to look at actually): https://github.com/OpenFn/kit/blob/df38b06b50665131cb2ec0c842af544e6c74ed21/packages/ws-worker/src/types.d.ts#L76

midigofrank: @taylordowns2000 I've updated the changes

taylordowns2000 (approved):

PR #1660, opened 2024-01-22
Do not delete jobs that have been run

### LINKED ISSUE #1570: Don't hide deleted Jobs on history page from past attempts (soft deletion of Jobs)

elias-ba:
### Summary

This PR introduces robust safeguards to prevent the deletion of steps that have already been executed in a workflow. The changes are designed to enhance data integrity and user experience in the workflow management process. The key enhancements include:

### 1. Database Constraint Enforcement

We've added a database-level constraint to the job_id foreign key. This ensures that any attempt to delete a step (job) that has associated runs (executions) will trigger an Ecto constraint error, thereby preserving the integrity of our execution history.

### 2. UI-Level Safeguards

The 'Delete Step' button is now conditionally disabled for steps with associated runs. This change makes the user interface more intuitive, preventing accidental deletion attempts on executed steps.

### 3. Enhanced User Feedback

For scenarios where deletion is attempted despite the disabled 'Delete Step' button (e.g., forced actions or bypassing UI constraints), we've implemented an informative flash message. This message will inform the user that steps with associated runs cannot be deleted, ensuring clear communication and enhancing user understanding of the system's behavior.

### Rationale

These changes are crucial to maintain the integrity of the workflow execution history. By preventing the deletion of executed steps, we ensure that historical data remains intact for audit trails, compliance, and analysis purposes. Additionally, the UI enhancements contribute to a more user-friendly and error-resistant experience, guiding users away from potentially disruptive actions.

## Related issue

Fixes #1570 

## Review checklist


taylordowns2000 (approved):

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1202,7 +1213,14 @@ defmodule LightningWeb.WorkflowLive.Edit do
     case type do
       :jobs ->
         socket
-        |> assign(selected_job: value, selected_trigger: nil, selected_edge: nil)
+        |> assign(
+          has_steps: has_steps?(value.id),
```
stuartc, line 1217: I think this should be either preloaded, a virtual field or something that would apply to all jobs when they are loaded from the database. This create an n+1 situation, which each job is making a new query to the db everytime _any_ change occurs.

stuartc (changes requested):
Good stuff here @elias-ba, but the `has_step` function and how it's used is open to superfluous queries to the db - lets find a way to merge or preload the state of the job for deleting that applies to all of the jobs at once (at least).

Ideally we want to avoid hitting the DB during `apply_params` entirely.

lib/lightning_web/live/workflow_live/edit.ex
elias-ba: Done

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -703,6 +707,14 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
          socket
          |> put_flash(:error, "You can't delete the first step of a workflow.")}
+
+      :has_steps ->
+        {:noreply,
+         socket
+         |> put_flash(
+           :error,
+           "You can't delete a step that has already been ran."
+         )}
```
stuartc, line 717: I don't think we should do anything about this now, but I think we're at the point where this logic needs to be moved out into something that acts as the "source of truth" for all this behaviour, this is business logic that exists only in one place.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1049,6 +1061,10 @@ defmodule LightningWeb.WorkflowLive.Edit do
     |> Enum.any?()
   end
 
+  defp has_steps?(job) do
+    !Enum.empty?(job.steps)
+  end
```
stuartc, line 1066: Just a tip with these kind of comparisons, when it comes to Ecto - if the relationship is not loaded we'll get a pattern match error here; because `Ecto.Association.NotLoaded` doesn't have the Enumerable protocol on it. So we're making an assumption here that the caller has loaded everything correctly.

If we were to abstract these kinds of functions (see comment up from this), we would use pattern matching to handle these cases.

stuartc (changes requested):
@elias-ba great work here dude! If you don't mind (and if you agree) could you move the tooltip message to into the function - I think it makes the component less cluttered (if nothing else).

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -418,12 +432,10 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
-  defp deletion_tooltip_message(has_multiple_jobs) do
-    if has_multiple_jobs do
-      "You can't delete a step that other downstream steps depend on."
-    else
-      "You can't delete the only step in a workflow."
-    end
+  defp deletion_tooltip_message(conditions) do
+    Enum.find_value(conditions, fn {condition, message} ->
+      if condition, do: message
+    end)
```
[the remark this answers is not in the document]
elias-ba: Done, thanks for the tip

lib/lightning_web/live/workflow_live/edit.ex
elias-ba: Okay got it, I am noting it somewhere so that we can chat about it in one of our 1-1s.

lib/lightning_web/live/workflow_live/edit.ex
elias-ba: Yes you are right. For now the way I am making sure this a no problem is that I preload the steps anytime we load the jobs. See here:
- https://github.com/OpenFn/lightning/blob/fe313af42bf763083b723b6f7a7d26671c1d49fc/lib/lightning_web/live/workflow_live/edit.ex#L648
- https://github.com/OpenFn/lightning/blob/fe313af42bf763083b723b6f7a7d26671c1d49fc/lib/lightning_web/live/workflow_live/edit.ex#L1266

This might not be the sexiest way to do it (and I am happy to change it by pattern matching the `Ecto.Association.NotLoaded`) but it might be the best thing to do for now, given that all of this is temporary solution. Let me know what you think.

taylordowns2000 (changes requested):
fix changelog please! otherwise good to go

taylordowns2000 (approved):

stuartc (approved):

PR #1662, opened 2024-01-23
Add zero persistence project setting

### LINKED ISSUE #1209: Zero-persistence: Add Retention options on the settings page
**Designs:**
- Data retention: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=643-888&mode=design&t=6IdtumAfCQZ8SEMe-11
- Zero-persistance enabled: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=643-994&mode=design&t=6IdtumAfCQZ8SEMe-11

------

- If checkbox for the "Retain data..." is selected, then show "Retention Period..." section with dropdown 
- If checkbox for the "Zero-persistance..." is selected:
  -  Then show section with dropdown disabled (whole div), and the retention period DDL item being set to "Not stored"
  -  Display a warning when enabled (yellow panel)

jyeshe:
## Related issue

Fixes #1209

## Review checklist


lib/lightning/attempts.ex
```diff
@@ -126,6 +126,20 @@ defmodule Lightning.Attempts do
     |> Repo.one()
   end
 
+  @doc """
+  Clears the body and request fields of the dataclip associated with the given attempt.
+  """
+  @spec wipe_dataclip_body(Attempt.t()) :: :ok
+  def wipe_dataclip_body(%Attempt{} = attempt) do
```
stuartc, line 133: Perhaps this should be named `wipe_dataclips`, since we're wiping more than the body, and the plural form tells us it's going to do it on potentially more than one given the `update_all`.

lib/lightning/projects.ex
```diff
@@ -245,6 +245,20 @@ defmodule Lightning.Projects do
     end)
   end
 
+  @spec project_retention_policy_for(Attempt.t()) ::
+          :retain_all | :erase_all | :erase_successful_only
```
stuartc, line 249: Can we move these into a type that is defined elsewhere, so any changes will propagate to any other functions?

lib/lightning_web/channels/run_with_options.ex
```diff
@@ -7,6 +7,12 @@ defmodule LightningWeb.AttemptJson do
   alias Lightning.Workflows.Job
   alias Lightning.Workflows.Trigger
 
+  @spec render(Attempt.t(), Keyword.t() | map()) :: map()
+  def render(%Attempt{} = attempt, options) do
```
stuartc, line 11: Can we add a type onto options please. It could arguably be a map at this point, but either way we should have some kind of safeguard as to the shape of the options we can pass to the worker.

lib/lightning_web/components/new_inputs.ex
```diff
@@ -17,13 +17,23 @@ defmodule LightningWeb.Components.NewInputs do
...
 
   def button(assigns) do
+    custom_classes =
+      if assigns[:cancel] do
+        "bg-neutral-100 hover:bg-neutral-300 text-neutral-950"
+      else
+        "bg-primary-600 hover:bg-primary-700 text-white"
+      end
+
+    assign(assigns, :class, fn class -> class ++ custom_classes end)
```
stuartc, line 35: Just a tip here, since LiveView can interpolate lists into attrs. Instead of creating a new assign you can add to the class attr below with `..., @custom_classes, @class`.

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -579,6 +589,16 @@ defmodule LightningWeb.ProjectLive.Settings do
     """
   end
 
+  def checked?(changeset, id) do
```
stuartc, line 592: Could this be `defp` instead?

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -579,6 +589,16 @@ defmodule LightningWeb.ProjectLive.Settings do
...
+    case Ecto.Changeset.fetch_field(changeset, :retention_policy) do
+      {_changes_data, ^id} ->
```
stuartc, line 594: I don't understand the use of the term `id` here, is it not just `value`?

Also I think we can safely use `fetch_field!/2` here, that would also tighten this up to: 

```suggestion
Ecto.Changeset.fetch_field!(changeset, :retention_policy) == value
```

test/lightning/attempts_test.exs
```diff
@@ -656,4 +656,30 @@ defmodule Lightning.AttemptsTest do
...
+
+      %{attempts: [attempt]} =
+        work_order_for(trigger, workflow: workflow, dataclip: dataclip)
+        |> insert()
+
+      assert dataclip.body
+      assert dataclip.request
+      refute dataclip.wiped_at
+
+      :ok = Attempts.wipe_dataclip_body(attempt)
```
stuartc, line 676: Please extend the test to cover the effects of `update_all` that are used in the implementation.

stuartc (changes requested):
Good stuff here! Keen to get this in, I made a few notes but this is very thorough, thanks!

lib/lightning_web/live/project_live/settings.ex
jyeshe: From the UI side, `id` here would mean the `input_id` or input field while the value comes from the changeset. But you're right about using `fetch_field!`, we don't need to know anymore if the value comes from changes or data.

lib/lightning_web/components/new_inputs.ex
jyeshe: Yes, thanks! We use on other parts of the code ... here it concatenates as the custom classes are defined internally (have opted for the smallest change but could avoid string concatenation though adding a new `@custom_classes` assign to put it as an element of the class list)

midigofrank: I've updated the code to use a different `assign` called `visual_class`. This assign will be responsible of holding all styles related to `color` and so it can be swapped by a different set entirely.
I'm not sure if `visual_class` is the right name for it though. Maybe I should call it `color_class` instead 🤔 ?

test/lightning/attempts_test.exs
midigofrank: @stuartc I've renamed the function to `wipe_dataclips`, but at the moment it only clears the given attempt's `dataclip`. Should I update the logic to update the associated `steps` dataclips' as well?

taylordowns2000 (changes requested):
1. ~using the link to navigate to "data storage settings" doesn't work, requires a refresh to bind to the project settings hash section.~ edit, must use `href` instead of `navigate` to get the urlHash to work
2. i'm still able to rerun using the checkbox selector. what happens when this happens? (we discussed that preventing the "select all" functionality is out of scope for this issue, but blocking the rerun of a work order—or at least throwing a big bad error!—is definitely i scope for this issue.)

--- update ---

ok, i fixed (1) and see that (2) is actually because `http_request` dataclips are not being wiped. they need to be treated just like the `saved_input` dataclips that are created by hand. everything else looks good

--- update 2 ---

the failure to wipe dataclips was not an issue introduced by this PR, it was from a bug on main. i've fixed that in https://github.com/OpenFn/Lightning/pull/1696

now there's just one spooky behaviour left: if i click "inspect" on step 3 of a run with no saved dataclips, on the left panel it shows me the wiped input dataclip for step 1! i'm guessing i know why this was done... is it hard to have that picklist not be selecting anything? (we don't want, "create new", we want nothing in this case.) @midigofrank , if you get this in the AM, please take a crack. `git merge hotfix-1695` and then see what's up

lib/lightning/invocation/dataclip.ex
```diff
@@ -73,7 +73,11 @@ defmodule Lightning.Invocation.Dataclip do
   end
 
   defp remove_configuration(changeset) do
-    add_error(changeset, :body, "must be a map")
+    if changeset |> get_change(:body) |> is_nil() do
+      changeset
+    else
+      add_error(changeset, :body, "must be a map")
+    end
   end
```
stuartc, line 81: Can we just combine the two cases? I think it's less than obvious now given the "fall-through" function calls `get_change`.

This can be wrapped up in a `case changeset |> get_change(:body) do ...` and then checking the type returned from that.

lib/lightning_web/channels/run_channel.ex
```diff
@@ -176,10 +175,10 @@ defmodule LightningWeb.RunChannel do
   def handle_in("step:complete", payload, socket) do
     %{
       "run_id" => socket.assigns.run.id,
-      "project_id" => socket.assigns.project_id
+      "project_id" => socket.assigns.project_id,
+      "wipe?" => socket.assigns.retention_policy == :erase_all
     }
     |> Enum.into(payload)
-    |> maybe_drop_dataclip(socket.assigns.retention_policy)
     |> Runs.complete_step()
```
stuartc: Instead of extending the payload that goes into `complete_step`, I think we should have a `complete_step(map(), rentention_policy() \\ :retain_all)` (I'm assuming what the options are for policies here). 

_or_ we get the projects retention policy from here?

Basically we have a function inside `Runs` that is there to handle these kinds of decisions...

Not sure what we can/should do right now - but I don't love extending string maps that change behaviour based on information we have control of (i.e. project settings vs params sent from the worker)

stuartc (changes requested):

taylordowns2000 (changes requested):
hey @midigofrank , after addressing stu's changes, please test this stuff. it was super close before but the last round of commits has lost the proper behavior on:
1. rerun from history page when zero-persistence is enabled
2. input/output in run viewer when zero-persistence is enabled
3. input/output in inspector when zero-persistence is enabled
4. the left panel (select/create new dataclip) when zero-persistence is NOT enabled

taylordowns2000 (approved):

stuartc (approved):
Nicely done guys! :+1:

PR #1663, opened 2024-01-23
Add storing webhook request headers in Dataclips for use in jobs

### LINKED ISSUE #1638: Change how http requests are persisted and given to worker
Background: https://community.openfn.org/t/access-to-request-header/367/13

@stuartc , given that you and @josephjclark are aligned on this and your decision (made independently, AFAIK!) also lines up with the vote in the community, I think we should move forward with this change.

When creating an `http_request` dataclip in the DB via a webhook request to the trigger URL for a workflow, we should store the  dataclip as:
```js
{
  data: {...}, // the body of the http request
  request: {
    headers: {...}, // the headers of the http request
    ... // anything else we might want to stick in here later
  }
}
```
this will allow us to _REMOVE_ the confusing stuff around state_assembly (see [here](https://github.com/OpenFn/Lightning/blob/main/lib/lightning_web/live/workflow_live/manual_workorder.ex#L60-L64)) and drop this conditional [here](https://github.com/OpenFn/Lightning/blob/main/lib/lightning_web/channels/attempt_channel.ex#L136-L139).

This is a small but important change... happy to wait to make sure your whole team is in alignment!

## Still to decide

We're aligned on what initial state should look like for the worker, but we disagree on what we do in the DB when we save a new `http_request` dataclip.

### Option A
One proposal would be to have a column for `body` and another for `request` and then continue to `assemble_state_from_http_request_dataclips` separately from how we assemble state from `run_result` and `saved_input` dataclips.

### Option B
Another proposal would be to store the request body and request headers as keys in the dataclip body, then assemble state _the same way_ for all current dataclips: `saved_input`, `http_request`, `run_result`.

Whether we go with "A" or "B", when we _display_ these `saved_input`, `http_request`, `run_result` dataclips to the users, our user base is adamant that we need to display them all _as if they were valid state_. We solve this problem in v1 by injecting a little extra stuff when we show the user the dataclip... so when you look at it via the UI you actually see `{"data": #{dataclip.body}}`

We want to avoid the cumbersome explanation below, which makes working with `http_request` dataclips _super different_ than working with all other dataclips.

image

stuartc:
## Notes for the reviewer

- Migration to add a `request` jsonb column to `dataclips`.
- `Attempts.get_dataclip_body` renamed to `get_input`. 
- `get_input` moves all serialisation to Postgres.
- Input returns `{data: <body>, request: <request>}`.

## Related issue

Fixes #1638 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to
 
 ### Added
 
+- Store webhook request headers in Dataclips for use in jobs.
+  [#1638](https://github.com/OpenFn/Lightning/issues/1638)
+
 ### Changed
 
 ### Fixed
--- a/lib/lightning/attempts.ex
+++ b/lib/lightning/attempts.ex
@@ -87,13 +87,41 @@ defmodule Lightning.Attempts do
   end
 
   @doc """
-  Returns a tuple with {type, string} so that initial state can be created for
-  the worker. See LightingWeb.AttemptChannel.handle_in("fetch:dataclip", _, _)
+  Returns only the dataclip request as a string
+  """
+  def get_dataclip_request(%Attempt{} = attempt) do
+    from(d in Ecto.assoc(attempt, :dataclip),
+      select: type(d.request, :string)
+    )
+    |> Repo.one()
+  end
+
+  @doc """
+  Returns an Attempts dataclip formatted for use as state.
+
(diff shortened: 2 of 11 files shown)


taylordowns2000 (approved):
perfect. i _love_ the `jsonb_build_object` and how simple `handle_in("fetch:dataclip", ...)` becomes

Issue #639, opened 2023-02-13
Add a rate limiter to webhook endpoints


amberrignell:
### Details

Using the data from load testing, implement a rate limiter using `mnesia` for webhook endpoints

1. There should be an instance rate-limit for `superusers` to worry about
2. There should be a project-level rate-limit for `admins` and `owners` to worry about

### Implementation notes

Testing will be a bit funky for this—it's almost load testing (we'll need to make a _large_ number of requests to a single webhook endpoint)

PR #1668, opened 2024-01-25
Add limiters extensions

jyeshe:
## Notes for the reviewer

Adds rate limiting and runtime limiting extensions. The actual modules are defined on compile time and are stubbed for now.

Plus allows extending the router as well.

## Related issue

Refs #639 

## Review checklist


lib/lightning/extensions/runtime_limiter.ex
```diff
@@ -0,0 +1,33 @@
...
+          function: fun(),
+          attrs: Keyword.t()
+        }
+
+  @spec check_limits(Context.t()) ::
+          :ok | {:error, RuntimeLimiting.limit_error(), message()}
+
+  def check_limits(context) do
+    adapter().check_limits(context)
+  end
```
stuartc, line 24: Should these functions not use `@impl true` instead of having their own specs?

I see the contract for `check_limits` is different from the behaviour, dialyzer doesn't pick it up because of the specs, a little puzzling because I would assume it would moan about either being different or think the function is missing.

Removing the `@behaviour` has no effect on dialyzer.

lib/lightning_web/controllers/webhooks_controller.ex
```diff
@@ -1,30 +1,41 @@
...
 
   @spec create(Plug.Conn.t(), %{path: binary()}) :: Plug.Conn.t()
   def create(conn, _params) do
-    case conn.assigns.trigger do
-      nil ->
-        conn |> put_status(:not_found) |> json(%{"error" => "Webhook not found"})
+    with %Workflows.Trigger{enabled: true, workflow: %{project_id: project_id}} =
+           trigger <- conn.assigns.trigger,
+         :ok <-
+           RateLimiter.limit_request(conn, %Context{project_id: project_id}, []) do
```
stuartc, line 16: I think we should 'hide' the Context struct behind the `limit_request` function, we can cast either a keyword list or map into this struct and validate/throw errors if needed.

It's not a deal breaker right now, but I can see these kinds of contracts finding their way into lots of parts of the app - and we'd do well to close the contract down a bit so we can lower the mental overhead of wrapping the desired functionality.

lib/lightning_web/live/workflow_live/index.ex
```diff
@@ -5,23 +5,37 @@ defmodule LightningWeb.WorkflowLive.Index do
...
   def render(assigns) do
     ~H"""
     <LayoutComponents.page_content>
+      <:banner>
+        <%= if assigns[:banner] do %>
+          <%= TagEngine.component(
+            @banner.function,
+            @banner.attrs,
+            {__ENV__.module, __ENV__.function, __ENV__.file, __ENV__.line}
+          ) %>
```
stuartc, line 36: :+1: Nice one! Bonus points for passing in the compile context..

lib/lightning_web/router.ex
```diff
@@ -168,16 +171,23 @@ defmodule LightningWeb.Router do
...
-  #   pipe_through :api
-  # end
+  scope "/" do
+    @routing_config Application.compile_env(
+                      :lightning,
+                      Lightning.Extensions.Routing,
+                      []
+                    )
+    @services_opts @routing_config |> Keyword.get(:session_opts, [])
+    @services_routes @routing_config |> Keyword.get(:routes, [])
```
stuartc, line 181: Super nice! I love this.

mix.lock
```diff
@@ -58,6 +58,7 @@
   "jumper": {:hex, :jumper, "1.0.2", "68cdcd84472a00ac596b4e6459a41b3062d4427cbd4f1e8c8793c5b54f1406a7", [:mix], [], "hexpm", "9b7782409021e01ab3c08270e26f36eb62976a38c1aa64b2eaf6348422f165e1"},
   "junit_formatter": {:hex, :junit_formatter, "3.3.1", "c729befb848f1b9571f317d2fefa648e9d4869befc4b2980daca7c1edc468e40", [:mix], [], "hexpm", "761fc5be4b4c15d8ba91a6dafde0b2c2ae6db9da7b8832a55b5a1deb524da72b"},
   "libcluster": {:hex, :libcluster, "3.3.3", "a4f17721a19004cfc4467268e17cff8b1f951befe428975dd4f6f7b84d927fe0", [:mix], [{:jason, "~> 1.1", [hex: :jason, repo: "hexpm", optional: false]}], "hexpm", "7c0a2275a0bb83c07acd17dab3c3bfb4897b145106750eeccc62d302e3bdfee5"},
+  "lightning_extensions": {:git, "https://github.com/OpenFn/lightning_extensions.git", "89f9069a088ad723e0eb9b12603c7941ce9750af", [ref: "89f9069"]},
```
stuartc, line 61: I think we can remove this.

stuartc (changes requested):
This is really nice dude, really thorough and a great foundation!

Only really a comment on the behaviours and the specs, and a hanging dep on `lightning_extensions`. Other than that, this is great! Thank you!

lib/lightning/extensions/runtime_limiter.ex
jyeshe: I remember a case where dialyzer haven't checked properly only with `@impl true` but it's been a while. Let's see.

lib/lightning_web/live/workflow_live/index.ex
```diff
@@ -62,15 +76,25 @@ defmodule LightningWeb.WorkflowLive.Index do
...
+      NewWorkflowForm.validate(%{}, socket.assigns.project.id)
+    )
+    |> then(fn socket ->
+      case RuntimeLimiter.check_limits(%Context{
+             project_id: socket.assigns.project.id
+           }) do
+        :ok ->
+          {:ok, socket}
+
+        {:error, :too_many_runs, %{position: position} = component} ->
```
taylordowns2000, line 94: @jyeshe , this strikes me as out of place. if i'm understanding the approach right, we'd apply an `ActionLimiter` here. The user wants to build a new workflow (not run it) and we'd need to check if they are allowed to do so, based on the number of workflows allowed by their plan.

the `too_many_runs` error is what i'd expect to see if someone tried to create a `WorkOrder+Run` or a `Run` for an existing work order.

jyeshe: The banner is shown here independently of any action. As a notification/reminder to the user that his plan has reached a limit where `:too_many_runs` is the reason/kind of limit reached.

That's the difference between `check_limits` and `limit_internal` that receives an action as a parameter. I will rename the later to `limit_action`.

jyeshe: Done @taylordowns2000, the rename and added a handling for multiple future reasons.

stuartc (approved):
Great stuff! I've just rebased this, will merge when CI says 'yay'.

PR #1679, opened 2024-01-29
Update Attempt and Run detail page for zero persistence

### LINKED ISSUE #1211: Zero-persistence: Attempt and Run detail page
From the History page, you can access the run detail page for each job that was attempted within a run workflow.
If someone has zero-persistence enabled, then they cannot view the details of the input and the output data on these pages.

When someone tries to access these pages, there needs to be content indicating that there is no data and the reason why. 

If the user has the correct admin access to edit the settings, they can link off directly.

If they don't have the admin access, the wording will change and inform them to speak to one of their account administrators. On hover, we can display the list of admins for simplicity for the user.

- Input data: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=1497-466&mode=design&t=0k7X53XaId2o3C67-11
- Output data: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=1497-482&mode=design&t=0k7X53XaId2o3C67-11
- No admin access: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=1497-493&mode=design&t=0k7X53XaId2o3C67-11
- 

Acceptance Criteria:

- [x] Whether looking at the "Inspector" or the "Run Viewer", if the project has zero-retention enabled you cannot see `run_result` dataclips in accordance with the figma.
- [x] When you look at the `http_request` dataclip for such a proejct, you should see the redacted dataclip.
- [x] When you look at the "Inspector" or the "Run Viewer" for a project _WITHOUT_ zero-persistence and there is no dataclip, it tells you that there isn't a dataclip (either `YET` or `EVER`) but does not inaccurately tell you that you probably have zero-persistence set.

midigofrank:
### TODO
- [ ] The manual run form body isn't cleared after step has finished running

## Related issue

Fixes #1211 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to
   [#1209](https://github.com/OpenFn/Lightning/issues/1209)
 - Wipe dataclip after use when zero-persistence is enabled
   [#1212](https://github.com/OpenFn/Lightning/issues/1212)
+- Show appropriate message when a wiped dataclip is viewed
+  [#1211](https://github.com/OpenFn/Lightning/issues/1211)
 
 ### Changed
 
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -49,6 +49,7 @@ defmodule Lightning.Invocation do
   def list_dataclips_for_job(%Lightning.Workflows.Job{id: job_id}) do
     Query.last_n_for_job(job_id, 5)
     |> Query.select_as_input()
+    |> where([d], is_nil(d.wiped_at))
     |> Repo.all()
   end
 
--- a/lib/lightning/policies/project_users.ex
+++ b/lib/lightning/policies/project_users.ex
@@ -26,6 +26,7 @@ defmodule Lightning.Policies.ProjectUsers do
           | :create_webhook_auth_method
           | :create_project_credential
           | :edit_webhook_auth_method
+          | :edit_data_retention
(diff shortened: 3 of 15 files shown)


taylordowns2000 (approved):
this works for me. i will adjust language in final PR to main. ready for @stuartc or @jyeshe to code review

stuartc (approved):
I'm unsure of the naming going on here, I think there's a bit too much going on in the viewer components; while it makes sense why it's all together I think we can separate things a bit.

I think the function name `step_dataclip_viewer` may be ambiguous next to `dataclip_viewer`. The checks in place like `dataclip_wiped?` etc probably deserve to live elsewhere.

I don't have a fully formed recommendation at the moment, and overall this is good to go. But I think down the line we would want to move things around and possibly invert the responsibilities (i.e. we pass onto a step in, or a struct specially for this that has the state precomputed).

Either that or we use slots to keep `@can_edit_data_retention` and `@admin_contacts` out of this component.


```
<.step_dataclip_viewer ...>
   <:wiped>
      <.wiped_message ... />
   </:wiped>
   ...?
</.step_dataclip_viewer>
```

This would allow us to use a LiveComponent that fetches the contacts and the permissions only when rendering the wiped message.

Basically this components interface is very large and is tied to a lot of attrs.

PR #1683, opened 2024-01-30
Replace access control test for Run show

### LINKED ISSUE #1646: Replace negative test for run_live show URL
The run live `:show` URL was no longer being used and was removed in this commit alongside: https://github.com/OpenFn/Lightning/pull/1645/files#diff-0d2174d557f95f0bf8356930daa8a2e91380ae2dc36400d9f39ef8f56cc9b8beL840-L875

This test should be re-created, but instead of checking for a run show URL, it should take an attempt show URL and ensure that users without access to that project get the proper `not_found`

stuartc:
## Related issue

Fixes #1646 

## Review checklist

### DIFF
--- a/test/lightning_web/live/project_live_test.exs
+++ b/test/lightning_web/live/project_live_test.exs
@@ -75,8 +75,7 @@ defmodule LightningWeb.ProjectLiveTest do
     end
 
     test "saves new project", %{conn: conn} do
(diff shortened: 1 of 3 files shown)


taylordowns2000 (approved):
thank you! love the raise on attempt not built also :-)

PR #1685, opened 2024-01-31
Disable selecting workorders that have wiped dataclips in the history page

### LINKED ISSUE #1210: Zero-persistence: History Page
When the user has Zero-persistence enabled, we need to disable certain functionality on the history page since the data won't exist.
- Figma: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=331-5765&mode=design&t=6IdtumAfCQZ8SEMe-11

Items:
- [x] Disable the "Rerun" button (grey it out) and display a tooltip
- [x] Disable the click for the Dataclip and display a tooltip
- [x] Disable the checkbox for the Workorder and display a tooltip that you cannot rerun the workorder
- [x] Tooltip to link to settings page
- [x] If non-admin, then the tooltip needs to be slightly tweaked: https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=706-4166&mode=design&t=6IdtumAfCQZ8SEMe-11

midigofrank:
## Notes for the reviewer
- Preloads the `:dataclip` on the searched `work orders`
- Checks in the preloaded dataclip has been wiped
- There's an edge case which hasn't been handled. When all the workorders in the page have been disabled, the select all checkbox doesn't select anything.
- ~~The authorization logic has been copied from #1679 . Which ever gets merged first, will mean we get a conflict on the other.~~

## Related issue

Fixes #1210

## Review checklist

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -385,12 +385,8 @@ defmodule Lightning.Invocation do
     )
   end
 
-  def search_workorders_having_dataclips_query(
-        %Project{} = project,
-        %SearchParams{} = params
-      ) do
-    project
-    |> search_workorders_query(params)
+  def exclude_wiped_dataclips(work_order_query) do
+    work_order_query
     |> join(:inner, [workorder: wo], assoc(wo, :dataclip), as: :dataclip)
     |> where([dataclip: d], is_nil(d.wiped_at))
   end
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -385,6 +385,16 @@ defmodule Lightning.Invocation do
     )
   end
 
+  def search_workorders_having_dataclips_query(
+        %Project{} = project,
+        %SearchParams{} = params
+      ) do
+    project
+    |> search_workorders_query(params)
+    |> join(:inner, [workorder: wo], assoc(wo, :dataclip), as: :dataclip)
+    |> where([dataclip: d], is_nil(d.wiped_at))
+  end
+
   defp base_query(project_id) do
     from(
       workorder in WorkOrder,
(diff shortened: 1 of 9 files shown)


lib/lightning/invocation.ex
```diff
@@ -385,6 +385,16 @@ defmodule Lightning.Invocation do
...
 
+  def search_workorders_having_dataclips_query(
+        %Project{} = project,
+        %SearchParams{} = params
+      ) do
+    project
+    |> search_workorders_query(params)
+    |> join(:inner, [workorder: wo], assoc(wo, :dataclip), as: :dataclip)
+    |> where([dataclip: d], is_nil(d.wiped_at))
+  end
```
stuartc, line 396: It looks like the two places that call this function were also the only two places that called `search_workorders` - why not extend the existing function? Where would we not want to have this behaviour?

stuartc (commented):
Nice one on the component tests, it's nice to see this being so thorough! 

Could I get your opinion on the new search workorders function you added on top of the old one; make a change if you agree. Either way will merge this straight after that.

lib/lightning/invocation.ex
midigofrank: There's a 3rd place where `search_workorders` is being used, loading workorders for the whole history page.

stuartc: Ah all right, perhaps rearranging this to be like:

```
project
|> search_workorders_query(params)
|> exclude_wiped_dataclips()
```

How does that sound?

midigofrank: Hmm, yeah, that looks better. Let me update it

stuartc (approved):

PR #1686, opened 2024-01-31
1684 impact tracking persistence

### LINKED ISSUE #1684: Setup plumbing for managing impact tracking submissions
- Allow the generation and persistence of a unique instance identifier
- Persist metrics as they were submitted

rorymckinley:
## Notes for the reviewer

This PR includes two changes each in a separate commit to make it easier to follow:
- Implements configuration for impact tracking
- Stores the impact tracking metrics in the DB (follow-up work will allow reports that failed to be submitted to be retried)

There is also a commit for some test cleanup.

## Related issue

Fixes #1684 

## Review checklist

### DIFF
--- a/lib/lightning/impact_tracking/client.ex
+++ b/lib/lightning/impact_tracking/client.ex
@@ -6,8 +6,17 @@ defmodule Lightning.ImpactTracking.Client do
   """
   use Tesla, only: [:post], docs: false
 
+  alias Lightning.ImpactTracking.ResponseProcessor
+
   def submit_metrics(metrics, host) do
-    build_client(host) |> post("/api/metrics", metrics)
+    build_client(host)
+    |> post("/api/metrics", metrics)
+    |> elem(1)
+    |> ResponseProcessor.successful?()
+    |> then(fn
+      true -> :ok
+      false -> :error
+    end)
   end
 
   defp build_client(host) do
--- a/lib/lightning/impact_tracking/report.ex
+++ b/lib/lightning/impact_tracking/report.ex
@@ -1,0 +1,16 @@
+defmodule Lightning.ImpactTracking.Report do
+  @moduledoc """
+  Report submitted to ImpactTracker
+
+  """
+  use Ecto.Schema
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  schema "impact_tracking_reports" do
+    field :data, :map
+    field :submitted, :boolean
+    field :submitted_at, :utc_datetime_usec
+
+    timestamps()
+  end
+end
--- a/lib/lightning/impact_tracking/response_processor.ex
+++ b/lib/lightning/impact_tracking/response_processor.ex
@@ -1,0 +1,9 @@
+defmodule Lightning.ImpactTracking.ResponseProcessor do
+  @moduledoc """
+  Utility module to abstract deaing with some of the Tesla plumbing
+
+  """
+  def successful?(%Tesla.Env{status: status}) do
+    status >= 200 && status < 300
+  end
+end
--- a/lib/lightning/impact_tracking/worker.ex
+++ b/lib/lightning/impact_tracking/worker.ex
@@ -10,15 +10,37 @@ defmodule Lightning.ImpactTracking.Worker do
     max_attempts: 1
 
   alias Lightning.ImpactTracking.Client
+  alias Lightning.ImpactTracking.Configuration
+  alias Lightning.ImpactTracking.Report
+  alias Lightning.Repo
 
   @impl Oban.Worker
   def perform(_opts) do
     if Application.get_env(:lightning, :impact_tracking)[:enabled] do
+      find_configuration()
+
       host = Application.get_env(:lightning, :impact_tracking)[:host]
 
-      Client.submit_metrics(%{}, host)
+      metrics = %{}
+
+      Client.submit_metrics(metrics, host) |> create_report(metrics)
     end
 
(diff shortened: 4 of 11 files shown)


lib/lightning/impact_tracking/response_processor.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Lightning.ImpactTracking.ResponseProcessor do
```
rorymckinley, line 1: Pulled this into a module of its own (rather than exposing the functionality as a public method on `ImapctTracking.Client`) to make it easier to test directly.

lib/lightning/impact_tracking/client.ex
```diff
@@ -6,11 +6,24 @@ defmodule Lightning.ImpactTracking.Client do
...
   def submit_metrics(metrics, host) do
-    build_client(host) |> post("/api/metrics", metrics)
+    result =
+      build_client(host)
+      |> post("/api/metrics", metrics)
+      |> elem(1)
+      |> ResponseProcessor.successful?()
+      |> indicate_success
+
+    {result, metrics}
```
jyeshe, line 19: if we don't intend to change metrics and only submitting would be fair, we could return only the result (without the result variable)

lib/lightning/impact_tracking/client.ex
```diff
@@ -6,11 +6,24 @@ defmodule Lightning.ImpactTracking.Client do
...
+  alias Lightning.ImpactTracking.ResponseProcessor
+
   def submit_metrics(metrics, host) do
-    build_client(host) |> post("/api/metrics", metrics)
+    result =
+      build_client(host)
+      |> post("/api/metrics", metrics)
+      |> elem(1)
+      |> ResponseProcessor.successful?()
+      |> indicate_success
```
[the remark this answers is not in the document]
jyeshe: or better we could use a `then` here to avoid the indirection like this:
|> ResponseProcessor.successful?()
|> then(fn 
  true -> :ok
  false -> :error
end)

lib/lightning/impact_tracking/worker.ex
```diff
@@ -10,15 +10,33 @@ defmodule Lightning.ImpactTracking.Worker do
...
 
-      Client.submit_metrics(%{}, host)
+      Client.submit_metrics(%{}, host) |> create_report()
     end
 
     :ok
   end
+
+  defp find_configuration do
+    Configuration |> Repo.one() || %Configuration{} |> Repo.insert()
```
jyeshe, line 31: I think you meant `Repo.insert!()` to have the same return type. An idiomatic suggestion would be:
```
with nil <- Repo.one(Configuration) do
  Repo.insert!(%Configuration{})
end
```

test/lightning/impact_tracking/worker_test.exs
```diff
@@ -1,12 +1,15 @@
...
-  alias Lightning.ImpactTracking.{Client, Worker}
+  alias Lightning.ImpactTracking.{Client, Configuration, Report, Worker}
 
   import Mock
   import Lightning.ApplicationHelpers, only: [put_temporary_env: 3]
 
   @host "https://foo.bar"
+  @metrics %{}
+  @ok_response {:ok, @metrics}
+  @error_response {:error, @metrics}
```
jyeshe, line 12: this is a nit pick and probably you have used it in multiple places before. Now that these are used only on the mocking template functions, would just set the return there to be `{:ok, @metrics}` and `{:error, @metrics}`

lib/lightning/impact_tracking/report.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Lightning.ImpactTracking.Report do
+  @moduledoc """
+  Report submitted to ImpactTracker
+
+  """
+  use Ecto.Schema
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  schema "impact_tracking_reports" do
+    field :data, :map
```
jyeshe, line 10: I imagine you've picked a generic term as it probably will save other kind of data besides metrics, right?

test/lightning/impact_tracking/worker_test.exs
```diff
@@ -1,12 +1,15 @@
...
+  use Lightning.DataCase, async: false
 
-  alias Lightning.ImpactTracking.{Client, Worker}
+  alias Lightning.ImpactTracking.{Client, Configuration, Report, Worker}
 
   import Mock
   import Lightning.ApplicationHelpers, only: [put_temporary_env: 3]
 
   @host "https://foo.bar"
+  @metrics %{}
```
jyeshe, line 10: what you think about testing to save a non empty map? would be cool to see some example close to actual metrics.

jyeshe (commented):
Nice test coverage, thank you @rorymckinley ! Few suggests ...

lib/lightning/impact_tracking/report.ex
rorymckinley: Yes - the idea is that `data` will contain everything that is sent to the ImpactTracker - so the metrics as well as any metadata that may be required (e.g. the time the metrics were generated)

test/lightning/impact_tracking/worker_test.exs
rorymckinley: Watch this space :) - the next Lightning PR will contain the actual metrics.

jyeshe (approved):
:clap:

taylordowns2000 (approved):

PR #1687, opened 2024-01-31
Hide inspector rerun button when the selected step has a wiped dataclip

### LINKED ISSUE #1639: Zero-Persistence: Block "rerun from here" button on Inspector
If a run or a RunStep has no input dataclip, you cannot rerun it.

This issue:
1. Ensures that the left side panel looks OK after you've created a new dataclip and run a workorder with it. (The data in the dataclip will _change_ after it's created, and you'll see a redacted copy in the left panel. This is probably a no-op.)
2. Ensures that if you navigate to the inspector from a specific step in a run on the history and there is NO `input_dataclip` associated with that run, you're not able to "Rerun from here".
3. Ensures that if you navigate to the inspector from a specific step in a run on the history and there is a REDACTED `input_dataclip` associated with that run, you're not able to "Rerun from here".
3. Ensures that the left panel says "no dataclip here" just like the panel on the history page interface and the "input/output" tabs on the inspector.

Note that the rerun button is greyed out, but the secondary button (create new workorder) may be available if/when you create a new `saved_input` via the left panel.

midigofrank:
## Notes for the reviewer

- Every time an `attempt` is followed in the `inspector`, we subscribe to the `Attempt Events`. This is there to help update the dataclip body copy if a user follows an attempt before it was started / while running. Since most users Inspect failed / completed steps, the live view will mostly subscribe to events that may never arrive.
- ~~In the test, I have a `Process.sleep(5)` to give an allowance for liveview to process an event.~~ Replaced with `render(view)` 

## Related issue

Fixes #1639 

## Review checklist

### DIFF
--- a/test/lightning_web/live/workflow_live/editor_test.exs
+++ b/test/lightning_web/live/workflow_live/editor_test.exs
@@ -781,6 +781,91 @@ defmodule LightningWeb.WorkflowLive.EditorTest do
       assert has_element?(view, "button:disabled", "Create New Work Order"),
              "create new workorder button is disabled"
     end
+
+    test "selected dataclip viewer is updated correctly if dataclip is wiped and step completed",
+         %{
+           conn: conn,
+           project: project,
+           workflow: %{jobs: [job_1, _job_2 | _rest]} = workflow
+         } do
+      unique_val = "random" <> Ecto.UUID.generate()
+
+      input_dataclip =
+        insert(:dataclip,
+          project: project,
+          type: :saved_input,
+          body: %{"foo" => unique_val}
+        )
+
+      %{attempts: [attempt]} =
+        insert(:workorder,
+          workflow: workflow,
+          dataclip: input_dataclip,
+          state: :running,
+          attempts: [
+            build(:attempt,
+              dataclip: input_dataclip,
+              starting_job: job_1,
+              state: :started,
+              steps: []
+            )
+          ]
+        )
+
+      {:ok, view, _html} =
+        live(
+          conn,
+          ~p"/projects/#{project}/w/#{workflow}?#{[s: job_1.id, a: attempt.id, m: "expand"]}"
+        )
+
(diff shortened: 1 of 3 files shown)


test/lightning_web/live/workflow_live/editor_test.exs
```diff
@@ -724,6 +724,148 @@ defmodule LightningWeb.WorkflowLive.EditorTest do
...
+          attempt_id: attempt.id,
+          project_id: project.id
+        })
+
+      assert_received %Lightning.Attempts.Events.StepCompleted{
+        step: %{id: ^step_id}
+      }
+
+      # wait for 5 milliseconds to give liveview some time to process the event
+      Process.sleep(5)
```
stuartc, line 862: Could this not be replaced with `assert_receive`?

Or I may be thinking of this backwards and you're waiting for changes from the `handle_info` call; in which case waiting _shouldn't_ be necessary.

If it still doesn't work, then calling 

```
view |> render()
refute view |> has_element?("#manual-job-#{job_1.id}", unique_val)
```

should be enough, the prior render does a GenServer.call which flushes out the mailbox to answer the next call.

Or add `view |> render()` above the `html = ...`

stuartc (changes requested):
Looks fine, but yeah this LiveView is so gnarly now - not just from these changes but all of us plying on more logic. We'll address this in the coming weeks.

Only thing is that we should try super hard to add `Process.sleep` calls even if they are super short, they can be brittle and need to be made longer just for a single environment leaving everyone waiting just because CI is slow etc.

test/lightning_web/live/workflow_live/editor_test.exs
midigofrank: Aaah, yes. I forgot about the `render(view)` function. 
Brilliant!!

stuartc (approved):

PR #1705, opened 2024-02-03
fix permissions on webhook auth methods

### LINKED ISSUE #1692: Editor can modify inbox security
As Editor (editor@openfn.org), I have access to set up inbox security (both at the workflow level and in project settings:webhook security) while according to user roles, I shouldn't have.

taylordowns2000:
- This fixes #1692 , ensuring that only owners and admins can edit webhook auth methods for a project.
- It _also_ collapses "create_webhook_blah" and "edit_webhook_blah" and _adds_ _deletion_ prevention by using `write_webhook_auth_methods`.

On this ⬆️ , I'd _really_ when possible I'd really like to collapse down our permissions. For instance, the only two permissions we should have on workflows are `write` and `read` - if you can create one, you can edit one, and you can delete one. In some cases, I bet we'll separate delete out, but in many cases it will be much cleaner to simply use `write_workflows`.

A second thing here, I'd like to remove `edit_job` and replace it with `create_workflows`. Especially as we've now got a big workflow changeset (soon JSON!?) it makes sense to have `write_workflow` and have that control whether you can create/edit/delete triggers, paths, and steps. This will come in later PRs.

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,11 @@ and this project adheres to
 
 ### Fixed
 
+- Fixed permissions issue that allowed viewer/editor to modify webhook auth
+  methods. These permissions only belong to project owners and admins
+  [#1692](https://github.com/OpenFn/Lightning/issues/1692)
 - Fixed bug that was duplicating inbound http_requests, resulting in unnecessary
-  data storage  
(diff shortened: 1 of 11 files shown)


taylordowns2000:
closing this in favor of #1708

PR #1708, opened 2024-02-03
Permissions update

### LINKED ISSUE #1692: Editor can modify inbox security

### LINKED ISSUE #1703: Don't allow editor to set up github sync

### LINKED ISSUE #1704: Don't allow viewer to initiate github sync

taylordowns2000:
Better to put this all in one PR as it's all permissions stuff and each branches off the previous. This PR:

1. Ensures that only owners and admins can edit webhook auth methods for a project to fix #1692 
2. Collapses `:create_webhook_auth` and `:edit_webhook_auth` into a single `:write_webhook_auth` permission
3. Removes unused permissions and collapses `run` and `rerun` into one thing. I'd really like to simplify our permissions. The only two permissions we should have on workflows are `write` and `readonly`: If you can create one, you can edit one, and you can delete one. In some cases, I bet we'll want to separate `delete` out, but in many cases it will be much cleaner to simply use `write_thing` and `read_thing`. More is _not_ better when it comes to programming and communicating these permissions to end users on the docs site and in training
4. Stops `:editor` from managing github repo connections to fix #1703 
5. Stops `:viewer` from initiating github syncs to fix #1704 

❗❗ **I'd like a _careful_ review** 🙏  **of how I'm blocking these github actions with permissions.** Is this the correct way to do it? I both disable the button with a permissions check, and then also check again on the `handle_event`.

## Review checklist

### DIFF
--- a/test/lightning_web/live/project_live_test.exs
+++ b/test/lightning_web/live/project_live_test.exs
@@ -2,6 +2,7 @@ defmodule LightningWeb.ProjectLiveTest do
   use LightningWeb.ConnCase, async: false
 
   alias Lightning.Repo
+  alias Lightning.Name
 
   import Phoenix.LiveViewTest
   import Lightning.ProjectsFixtures
@@ -1313,22 +1314,11 @@ defmodule LightningWeb.ProjectLiveTest do
       end)
     end
 
-    test "all project users can see the project webhook auth methods", %{
-      conn: conn
-    } do
+    test "all project users can see the project webhook auth methods" do
       project = insert(:project)
       auth_methods = insert_list(4, :webhook_auth_method, project: project)
 
-      for project_user <-
-            Enum.map([:editor, :admin, :owner, :viewer], fn role ->
-              insert(:project_user,
-                role: role,
-                project: project,
-                user: build(:user)
-              )
-            end) do
-        conn = log_in_user(conn, project_user.user)
-
+      for conn <- project_user_conns(project, [:editor, :admin, :owner, :viewer]) do
         {:ok, _view, html} =
           live(
             conn,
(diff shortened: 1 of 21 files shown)


test/lightning_web/live/project_live_test.exs
```diff
@@ -1311,22 +1312,11 @@ defmodule LightningWeb.ProjectLiveTest do
...
-            Enum.map([:editor, :admin, :owner, :viewer], fn role ->
-              insert(:project_user,
-                role: role,
-                project: project,
-                user: build(:user)
-              )
-            end) do
-        conn = log_in_user(conn, project_user.user)
-
+      for conn <- project_user_conns(project, [:editor, :admin, :owner, :viewer]) do
```
stuartc, line 1319: A little nitpick, could we name this like `build_conn_for_user_roles`?

Or similar, something that implies it builds things

Or split it between roles as the enum, and the build conn function be more generic?

test/lightning_web/live/project_live_test.exs
```diff
@@ -1384,7 +1365,7 @@ defmodule LightningWeb.ProjectLiveTest do
                |> element("form#choose_auth_type_form_#{modal_id}")
                |> has_element?()
 
-        credential_name = "#{project_user.role}credentialname"
+        credential_name = Name.generate()
```
stuartc, line 1368: We should move this into a exmachina sequence..

stuartc (commented):
This is great dude! Two small things, feel free to make the changes or not, the changes are solid, thank you!

test/lightning_web/live/project_live_test.exs
taylordowns2000: cool. tracked here: https://github.com/OpenFn/Lightning/issues/1713

PR #1712, opened 2024-02-05
fix inspector view stuck at processing when manual run has crashed

### LINKED ISSUE #1711: Inspector view and RunDetail page hang when a run crashes BEFORE the first step
Create a job with an expression: `fn(state 321321323{{`, run it and watch it crash before any steps return. There are two bad behaviors to fix:
1. Inspector view _hangs_ on the "processing" mode...
2. RunDetail shows input/output as "Nothing yet" rather than "crashed, no input/output available"

midigofrank:
## Notes for the reviewer

- [x] Inspector view hangs on the "processing" mode...
- [x] RunDetail shows input/output as "Nothing yet" rather than "crashed, no input/output available"
- [x] Tests

## Related issue

Fixes #1711

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ and this project adheres to
   [#1703](https://github.com/OpenFn/Lightning/issues/1703)
 - Fixed permissions issue that allowed viewers to initiate syncs to github
   [#1704](https://github.com/OpenFn/Lightning/issues/1704)
+- Fixed inspector view stuck at processing when following a crashed run
+  [#1711](https://github.com/OpenFn/Lightning/issues/1711)
 
 ## [v2.0.0-rc8] - 2024-01-30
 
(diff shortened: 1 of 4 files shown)


stuartc (approved):
All good here!

stuartc:
This is all good from me, but @midigofrank @taylordowns2000 the edit page and more specifically the logic required to determine the state of a run and it's dataclips availability is getting out of hand. There are a lot of places used to determine how to render things that are compounded if expressions. This is a fix for an issue that could be a result of having all this logic being spread around the views.

I think we need to make an issue to collect up all the states and variations into a struct that can used to handle changes to the run/dataclip. I've mentioned this before that we need some kind of presenter pattern.

taylordowns2000:
Great call @stuartc . We're already seeing little things pop up: https://github.com/orgs/OpenFn/projects/3/views/1?pane=issue&itemId=52278822

I'll do a UX review here and merge for now, but after this sprint we should consider collecting these all together.  (Also, we should talk about making this all React... at some point!)

PR #1716, opened 2024-02-05
Runtime scheduler extension

### LINKED ISSUE #1715: Allow customizing the runtime scheduling/queue
Some entry point to customize the `Lightning.Run.Queue` is needed to allow claiming runs for a specific project.

In order to have a single way to define extensions, a solution might involve moving the existing "adaptor" to the extensions interface set.

jyeshe:
## Notes for the reviewer

Adds an extension following the same pattern of the other customization parts under `Lightning.Extensions`.

## Related issue

Fixes #1715 

## Review checklist


stuartc (changes requested):
It's great to see the Run Adaptor getting moved out, but I have some reservations about the naming.

1. I don't think this 'scheduler/ing' is the best term for what this is, I think the behaviour should be called `RunQueue`.
2. While the current adaptor is the 'default', I don't think it should called (or placed) in anything that states that. The actual implementation is a fifo queue, so I think that we should use a structure like this:

```
module Lightning.Extensions.RunQueue do
  ^ the behaviour
  also includes the implementation picker (1)
  def claim(...) do
    Lightning.Config.run_queue().claim()
  end
end
  
Lightning.Extensions.FifoQueue
  ^ the current queue/runtime_scheduler
  
```

(1) I'm much more fond of this pattern since it cuts down on a level of indirection, it's difficult following the 3 different files required by splitting the behaviour into a different module. Merging the "entrypoint" for calls to an adaptor with the behaviour is common in many libraries (Oban being the first thing that comes to mind).

```
RunQueue          <- "proxy" service & behaviour
FifoRunQueue      <- The default for Lightning
```

If you disagree can we perhaps adopt a pattern that more clearly splits the behaviours and the implementation.

For example:
```
RunQueue          <- "Proxy" service
RunQueue.Adaptor  <- behaviour
FifoRunQueue      <- The default for Lightning
```

taylordowns2000 (approved):
@jyeshe , this looks good to me, but i'd love to confirm that the default behaviour is not somehow scoped by project. default should be FIFO for the whole instance. people could build queue mechanisms (like round robin by project, for instance) elsewhere. is this right? this seems to be the concept here, but the test "claims a run from a project queue" got me worried that queues might be somehow related to projects.

jyeshe:
> @jyeshe , this looks good to me, but i'd love to confirm that the default behaviour is not somehow scoped by project. default should be FIFO for the whole instance. people could build queue mechanisms (like round robin by project, for instance) elsewhere. is this right? this seems to be the concept here, but the test "claims a run from a project queue" got me worried that queues might be somehow related to projects.

@taylordowns2000 that's right, the default implementation of the queue is not scoped by project. The name of the test case was just an abstraction as there is no actual queue implemented. I have generalize the `RuntimeLimiter` here already.

stuartc (approved):
Nice! :ok_hand:

PR #1747, opened 2024-02-12
Broadcast created run for trigger

jyeshe:
## Notes for the reviewer

In order to track the runs count we need to know that a run was created for a trigger (same way it's broadcast for manual).
Also adds the `project_id` to the `RunCreated` to allow the tracking per project.

## Related issue

Refs #1754 

## Review checklist


taylordowns2000 (approved):
this makes sense. no reason to not broadcast runs from triggers. thanks @jyeshe .

PR #1802, opened 2024-02-26
close #1720

### LINKED ISSUE #1720: Add "ALLOW_DEMO_RESET" env variable, default FALSE
As far as "foot guns" go, this is a big one. I'd prefer it if most lightning instances could not be reset to a demo state with a single command.

- Use the Vapor API
- Make sure that if nothing gets set, this thing is `false`. (It's a special case when someone would actually want a "destroy everything" button... we happen to use it for our public sandbox which gets a nightly cleaning.)
- Expect `Lightning.Demo.reset_demo()` throws an exception if someone attempts it without this ENV flag
- Expect the mix task to throw an exception also

taylordowns2000:
## Validation Steps

- Try to call `Lightning.Demo.reset_demo` before setting `IS_RESETTABLE_DEMO` to `yes` in your `.env.dev` - nothing should happen.
- (If you want to reset!!!) set that environment variable to `yes` and then run the same command - you should get your database reset to the demo state

## Notes

This fixes #1720 but I have _not_ done anything special with Vapor. The default I'm proposing here is that this is not allowed anywhere unless the env is specified. If we want something more sophisticated later, I'd want to understand the value of it before doing too much. (AFAIK, the only consumer for this is the company that's hosting demo.openfn.org 😉 and devs that should think twice before calling the command locally anyway.) In the current implementation, adding to you `.env.dev` feels right to me.

## Review checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -186,3 +186,8 @@ PURGE_DELETED_AFTER_DAYS=7
 # By default, impact tracking metrics will be reported to https://impact.openfn.org. Use the
 # below if you wish to change that.
 # USAGE_TRACKER_HOST=https://impact.openfn.org
+
+# OpenFn.org hosts a public sandbox that gets reset every night. If you'd like to
+# make your instance "resettable" (a highly descrutive action—this destroys all
+# data in your instance) you can set the following environment variable to "yes"
+# IS_RESETTABLE_DEMO=no
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to
 
 ### Changed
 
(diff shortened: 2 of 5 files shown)


stuartc:
I understand your point about hiding the flag to allow demo resets. But if that's the case we should (and we don't see this option belonging to the application) the. We shouldn't put it in the application.

I.e. we read the value straight from the environment. Skip putting it in the application entirely,

Vapor is designed to replace placing ENVs into the application. So we either use Vapor or we read the env variable directly in runtime.

taylordowns2000:
Have opened this issue for Vapor: https://github.com/OpenFn/lightning/issues/1810

PR #1829, opened 2024-02-28
1819 tweak version chip code (part of 'Extend version that ImpactTracker reports)

rorymckinley:
## Validation Steps

Given the nature of the change (removing redundant code) and the nature of the code (introspecting branch/image/commit) , I think the only cost-effective test (other than the automated tests) is to confirm that the version is correctly displaying at the bottom of the LH nav bar in the UI. On this branch it should read 'v2.0.5'

## Notes for the reviewer

This PR forms part of a larger body of work to extract the functionality that decides what version to show so that it can be accessible by the UsageTracker code. As preparation for this, I wanted to ensure that the critical elements of the code had coverage. This, in turn, lead to me discovering a redundant condition (see comments).

I have grouped the changes into commits for easier review and I will collapse those prior to merging in.

## Related issue

Fixes #

## Review checklist

### DIFF
--- a/lib/lightning_web/live/components/common.ex
+++ b/lib/lightning_web/live/components/common.ex
@@ -25,12 +25,6 @@ defmodule LightningWeb.Components.Common do
            "Docker image tag found: '#{image}' tagged release build from #{commit}",
            :release}
 
-        # If running in docker and tag doesn't match :vsn, display image tag.
-        image != nil and image != vsn and image != "edge" ->
-          {image,
-           "Detected image tag that does not match application version #{vsn}; image tag '#{image}' built from #{commit}",
-           :warn}
-
         # If running in docker and tag doesn't match :vsn, display commit.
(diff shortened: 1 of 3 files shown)


lib/lightning_web/live/components/common.ex
```diff
@@ -25,12 +25,6 @@ defmodule LightningWeb.Components.Common do
            "Docker image tag found: '#{image}' tagged release build from #{commit}",
            :release}
 
-        # If running in docker and tag doesn't match :vsn, display image tag.
```
rorymckinley, line 28: Line 17 of this file already has behaviour to cover the case where `image == "edge"` so this clause is effectively identical to the clause below ` image != nil and image != vsn`.

jyeshe (approved):
Missing only to reference the issue on the PR description. Thanks for the extra test case.

taylordowns2000 (approved):

PR #1830, opened 2024-02-28
Show flash error when editing stale project credentials

### LINKED ISSUE #1795: RuntimeError: you are attempting to change relation :project_credentials of
Sentry Issue: [LIGHTNING-9R](https://openfn.sentry.io/issues/5005615730/?referrer=github_integration)

```
RuntimeError: you are attempting to change relation :project_credentials of
Lightning.Credentials.Credential but the `:on_replace` option of this relation
is set to `:raise`.

By default it is not possible to replace or delete embeds and
associations during `cast`. Therefore Ecto requires the parameters
given to `cast` to have IDs matching the data currently associated
to Lightning.Credentials.Credential. Failing to do so results in this error message.

If you want to replace data or automatically delete any data
not sent to `cast`, please set the appropriate `:on_replace`
option when defining the relation. The docs for `Ecto.Changeset`
covers the supported options in the "Associations, embeds and on
replace" section.

However, if you don't want to allow data to be replaced or
deleted, only updated, make sure that:

  * If you are attempting to update an existing entry, you
    are including the entry primary key (ID) in the data.

  * If you have a relationship with many children, all children
    must be given on update.


  File "lib/ecto/changeset/relation.ex", line 246, in Ecto.Changeset.Relation.on_replace/2
  File "lib/ecto/changeset/relation.ex", line 417, in Ecto.Changeset.Relation.reduce_delete_changesets/5
  File "lib/ecto/changeset/relation.ex", line 112, in Ecto.Changeset.Relation.cast/5
  File "lib/ecto/changeset.ex", line 1192, in Ecto.Changeset.cast_relation/4
  File "lib/lightning/credentials/credential.ex", line 46, in Lightning.Credentials.Credential.changeset/2
...

(RuntimeError you are attempting to change relation :project_credentials of
Lightning.Credentials.Credential but the `:on_replace` option of this relation
is set to `:raise`.

By default it is not possible to replace or delete embeds and
associations during `cast`. Therefore Ecto requires the parameters
given to `cast` to have IDs matching the data currently associated
to Lightning.Credentials.Credential. Failing to do so results in this error message.

If you want to replace data or automatically delete any data
not sent to `cast`, please set the appropriate `:on_replace`
option when defining the relation. The docs for `Ecto.Changeset`
covers the supported options in the "Associations, embeds and on
replace" section.

However, if you don't want to allow data to be replaced or
deleted, only updated, make sure that:

  * If you are attempting to update an existing entry, you
    are including the entry primary key (ID) in the data.

  * If you have a relationship with many children, all children
    must be given on update.

)
```

jyeshe:
## Validation Steps

1. Setup demo
2. Run Lightning
3. Go to User credentials (under user profile)
4. Create a credential 
5. Open another browser window
6. Click on Edit credential on both windows
7. Associate the credential to a project and save on one window.
8. Edit only the body on the other window and save
9. A flash error should be displayed informing that the credential was edited by another session (stale data)

## Notes for the reviewer

This issue happens on a scenario of concurrent updates on a credential involving two sessions. It handles the error when saving a credential with stale/outdated project credentials list.

The solution would be easier by changing the default `on_replace=:raise` to `:update` but has_many for a good reason does not support it.

## Related issue

Fixes #1795 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Show flash error when editing stale project credentials
+  [#1795](https://github.com/OpenFn/lightning/issues/1795)
+
 ## [v2.0.7] - 2024-02-29
 
 ### Fixed
@@ -50,8 +53,6 @@ and this project adheres to
 
 - Fix series of sentry issues related to OAuth credentials
   [#1799](https://github.com/OpenFn/lightning/issues/1799)
-- Show flash error when editing stale project credentials
-  [#1795](https://github.com/OpenFn/lightning/issues/1795)
 
 ## [v2.0.5] - 2024-02-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ and this project adheres to
 
 - Fix series of sentry issues related to OAuth credentials
   [#1799](https://github.com/OpenFn/lightning/issues/1799)
+- Show flash error when editing stale project credentials
+  [#1795](https://github.com/OpenFn/lightning/issues/1795)
 
 ## [v2.0.5] - 2024-02-25
 
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -173,8 +173,6 @@ defmodule Lightning.Credentials do
   def update_credential(%Credential{} = credential, attrs) do
     changeset = credential |> change_credential(attrs) |> cast_body_change()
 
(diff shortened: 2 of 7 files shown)


taylordowns2000 (approved):
this looks good, but i'm honestly a bit surprised by the error itself. credentials only have one owner, and they can only ever be edited by one user. seems very odd that someone would actually open multiple browser windows to modify their own credential twice. i wonder if this is only happening on demo.openfn.org where you could have a bunch of different people all using the same public login.

Issue #1790, opened 2024-02-23
Adjust exported metrics


rorymckinley:
Adjust exported metrics to align with requirements

PR #1852, opened 2024-03-01
Polling metric - count available runs

rorymckinley:
## Validation Steps

Start Lightning up locally with the following env vars:

```
PROMEX_ENABLED=true                                                                                     
PROMEX_UPLOAD_GRAFANA_DASHBOARDS_ON_START=false                                                          
PROMEX_DATASOURCE_ID=promex                                                                             
PROMEX_METRICS_ENDPOINT_AUTHORIZATION_REQUIRED=no
```

If you have a sizeable number of runs (50 or so) in your local database, the easiest way to observe changes in this metric is to update the state of all of the runs to `available` and then navigate to  `/metrics` - the `lightning_run_queue_available_count` should have a non-zero value which will decrease as the runs are caimed.

## Notes for the reviewer

This poilling metric counts the number of runs in the available state.

## Related issue

#1790 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to
 ## [Unreleased]
 
 ### Added
+- Polling metric to track number of available runs.
+  [#1790](https://github.com/OpenFn/lightning/issues/1790)
 
 - Allows limiting creation of new runs and retries.
   [#1754](https://github.com/OpenFn/Lightning/issues/1754)
--- a/lib/lightning/runs/prom_ex_plugin.ex
+++ b/lib/lightning/runs/prom_ex_plugin.ex
@@ -12,6 +12,7 @@ defmodule Lightning.Runs.PromExPlugin do
   alias Lightning.Repo
   alias Lightning.Run
 
+  @available_count_event [:lightning, :run, :queue, :available]
   @average_claim_event [:lightning, :run, :queue, :claim]
   @stalled_event [:lightning, :run, :queue, :stalled]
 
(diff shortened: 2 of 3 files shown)


jyeshe (commented):
Code looks good to me. Is there a maximum frequency for the intended polling?

rorymckinley:
> Code looks good to me. Is there a maximum frequency for the intended polling?

@jyeshe I am not sure if you mean in terms of the underlying telemetry libraries (:telemetry & promex) or the application within Lightning, so to be on the safe side, I will attempt to answer both:

1) Wrt :telemetry/promex itself: The internet is silent on any constraints that may be imposed by the libraries. I am happy to dig through the code should you so wish.
2) Wrt Lightning: At the moment the period is defined by `METRICS_RUN_QUEUE_METRICS_PERIOD_SECONDS` which defaults to 5 seconds (an entirely arbitrary value intended to invite feedback :)) but there are no limitations applied to this value. Therefore, the period (and the resulting frequency) can be set to any value at the operator's discretion. Should you wish me to add a constraint, I can certainly add an issue to do so.

jyeshe:
My concern is about the latency if we have multi-tenant. How long would it take for 1 million records for example? In a scenario where multiple users create lots of work orders, more than the runtime can execute them.

jyeshe (approved):
@rorymckinley it would be great to profile the query on the scenario of accumulated runs (not the priority now possibly) and add a constraint or cache the counting if needed.

rorymckinley:
> @rorymckinley it would be great to profile the query on the scenario of accumulated runs (not the priority now possibly) and add a constraint or cache the counting if needed.

@jyeshe Ah, yes, I understand your concern now. Performance of the metric itself will definitely need to be something that we keep a close eye on as the number of records grows.  Once the metrics have stabilised, I agree that we should define the parameters within which the metrics (as implemented)  make sense. I will add an issue to see if we can get a 'degradation curve' for the metric generation queries.

stuartc (approved):

PR #1906, opened 2024-03-18
Ameliorate small-screen usability issues on history and inspector

taylordowns2000:
This fixes https://github.com/OpenFn/lightning/issues/1908

## Validation Steps

1. Navigate to the history page to see how the table now has a nested card for each run within a workorder.
2. Navigate to the inspector and collapse/expand the panels, execute runs, and _change_ to inspect different jobs from the steps list.
3. Navigate to the run detail page to see how the step icons are now standardized across all 3 interfaces.

## Review checklist


taylordowns2000:
@stuartc , this is reviewed and ready to go from the product side and i've deployed it to demo so we have it for the ICT4D conference (elias and ayo!)

still 6 broken tests, and i'm not sure i'll get to them tonight. if you're up before me tomorrow and want to fix one or two as part of your code review, i'd very much appreciate it! 🙏

lib/lightning_web/live/run_live/workorder_component.ex
```diff
@@ -273,28 +277,48 @@ defmodule LightningWeb.RunLive.WorkOrderComponent do
...
                     <% :started -> %>
-                      started @ <.timestamp timestamp={run.started_at} />
+                      started @
+                      <.timestamp
+                        tooltip_prefix="Run started at"
+                        timestamp={run.started_at}
+                      />
                     <% _state -> %>
-                      <%= run.state %> @ <.timestamp timestamp={run.finished_at} />
+                      finished @
```
taylordowns2000, line 302: @stuartc , this isn't on `main`, but only appears to be an issue when the run crashes. i can replicate it on `v2.0.9-pre5` if i set a job to run with only `fn(state => this is a bbbad execption)`. I'm looking at this code here and can't quite figure out why it would behave any different from what's on `main`.

stuartc: Thanks!

stuartc: Ok I've pushed some fixes, looks like all but ☝️ this one is fixed.

PR #1889, opened 2024-03-13
Make resettable demo true for dev and test

jyeshe:
## Validation Steps

```
$ mix ecto.reset; mix run -e "Lightning.Demo.reset_demo"
```

## Notes for the reviewer

Makes resettable demo true for dev and test environments (enables it without editing `.env`)

## Related issue

Refs #1859

## Review checklist

- [x] I have performed a **self-review** of my code
- [x] I have verified that all appropriate **authorization policies** (`:owner`, `:admin`, `:editor`, `:viewer`) have been implemented and tested
- [x] If needed, I have updated the **changelog**
- [ ] Product has **QA'd** this feature

PR #1907, opened 2024-03-18
Remove all defaults from .env.example

### LINKED ISSUE #1904: Remove defaults from .env.example
### Details

Currently developers are expected to use `.env.example` to seed various defaults in the application.

This practice creates a few issues namely 'configuration drift', e.g. when new config is introduced or when defaults change resulting in surprise errors and local .env files overriding the value respectively.

- - -

### Implementation notes

1. All env variables that `.env.example` should be commented out with their defaults updated to reflect the current default.
2. The application starts in both test and dev modes without issue.
3. The worker keys are fixed/defaulted in `dev.exs` and `test.exs`.
4. We still use `Vapor` but it's only responsibility is to cast worker settings from base64 to PEM.

~~In the coming days we will remove Vapor, as its implementation hides this behaviour and has no responsibility beyond base64 decoding (it no longer enforces the existence of the worker env vars).~~

_Update_ Vapor has been removed since it was setting the worker keys to `nil` with no easy way to default to the application config it no longer served any purpose, replaced by `dotenvy`.

stuartc:
## Validation Steps

_How can a reviewer validate your work?_

- Delete any `.env{.test|.dev}` files, and run the application and tests.

## Notes for the reviewer

The pattern for environment `.env` files has changed to `.{ENV}.env` (previously was `.env.{ENV}`, please rename any files that you have.

The defaults for the database user have been conformed to the application config, so if previously your Postgres username as `hello`, you will need to set a `POSTGRES_USER=hello` env variable in `.env` (or `.dev.env` etc).

Or recreate your db/server with the docker instructions. 

## Related issue

Fixes #1904
Supercedes #1889 

## Review checklist


.env.example
```diff
@@ -69,11 +64,21 @@ ORIGINS=//localhost:*
 
 # You can configure the max run duration for jobs in milliseconds. This should
 # be lower than the pod termination grace period if using Kubernetes.
-WORKER_MAX_RUN_DURATION_SECONDS=60
-WORKER_MAX_RUN_MEMORY_MB=500
-WORKER_CAPACITY=4
+# WORKER_MAX_RUN_DURATION_SECONDS=60
 
-MAX_DATACLIP_SIZE_MB=10
+# TODO: these aren't specified in Runtime, do they belong to the worker process?
```
taylordowns2000, line 69: yes! interesting point... we _SHARE_ config between worker and app. should we document that in lightning?

config/config.exs
```diff
@@ -35,8 +35,13 @@ config :lightning, Lightning.Extensions,
   usage_limiter: Lightning.Extensions.UsageLimiter,
   run_queue: Lightning.Extensions.FifoRunQueue
 
+# TODO: don't use this value in production
```
taylordowns2000, line 38: what does this todo mean and when must it be done?

taylordowns2000 (approved):
approved, with two open questions. thanks!

jyeshe (approved):
bb97084a works on umbrella app

.env.example
stuartc: Ah ok, so this is a bit of an unusual case - this _doesn't_ matter if you're deploying in a cluster/different images; and only really applies in self-contained environments (like dev and test).

So in this exact example, we're implicitly passing through env vars to the rtm process;
So we can make our own defaults inside the rtm process that pass down env vars to the rtm child process, which has a downside of overriding any default changes in the cli.
Or we leave it alone and expect if someone wants to change these settings to use env vars.
Or lastly we parse the env vars into lightnings config (under the `:workers` key) and conditionally set them in the child process env; which has little gain - but is more obvious and kinda 'self-documenting'.

config/config.exs
stuartc: Ah yes thanks for spotting this; will message you about it.

stuartc: I've made an issue over here: https://github.com/OpenFn/lightning/issues/1911

PR #1781, opened 2024-02-21
push`.github/workflows/{pull,deploy}.yml` files to connected github branch

midigofrank:
## Validation Steps
This PR aims to commit `.github/workflows/{pull,deploy}.yml` files when the user clicks to connect a given branch.

### Testing
To test this locally, you need to have the `GITHUB_APP_ID` and `GITHUB_CERT` envs set. 
-  To visually test this, head to your project settings page under the sync to github tab.
- There's an [end to end test for this here](https://github.com/OpenFn/lightning/pull/1781/files#diff-e57c90e701f594d5212cc8fd12cb852afc88bc5f0b539334c2e734c0b8b2a638). 

## Notes for the reviewer

- Adds a `@moduletag :integration` on the end to end test. Also configures the `test_helper.exs` to `exclude: [:integration]` 
- Uses a hardcoded `github installation id` for the end to end test. Maybe a security risk??

### TODO

- [x] Maybe write an actual end to end test that validates that existing non workflow files are not deleted. I think this is very important
- [x] Replace `midigofrank/openfn` with `OpenFn/github_integration_testing` as well as the `installation id` in the end to end test

## Related issue

Fixes #1046

## Review checklist

- [x] I have performed a **self-review** of my code
- [x] I have verified that all appropriate **authorization policies** (`:owner`, `:admin`, `:editor`, `:viewer`) have been implemented and tested
- [x] If needed, I have updated the **changelog**
- [x] Product has **QA'd** this feature

PR #1912, opened 2024-03-19
Allow users to connect to their GitHub accounts

### LINKED ISSUE #1894: Allow users to connect their accounts to Github
See v1:
- [buildOauthConsentUrl](https://github.com/OpenFn/platform-app/blob/main/assets/js/components/UserEdit.js#L80)
- https://github.com/OpenFn/platform-app/blob/main/assets/js/helpers/index.js#L21
- Connect to github [figma](https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=2086%3A2303&mode=design&t=OEZgJslyfupUtLUE-1)
- Remove from github [figma](https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=2090%3A2409&mode=design&t=OEZgJslyfupUtLUE-1)

User Settings (button color wrong here)

Image

Image

midigofrank:
### Demo:
https://github.com/OpenFn/lightning/assets/39288959/caa3f05f-49c3-4f1a-9a41-1f4573383def
### Required ENVS:
- `GITHUB_APP_CLIENT_ID`
- `GITHUB_APP_CLIENT_SECRET`
### Checklist

- [x] User can connect
- [x] User can disconnect 

## Notes for the reviewer

- Adds a new column, `github_oauth_token` to the `users` table. This field is encrypted at rest, with the help of `Lightning.Encrypted.Map`
- Adds a new route, `/oauth/:provider/callback` and controller `OauthController`. I've decided NOT to use the existing route and controller `OidcController` because this new logic deals with users who are logged in. Maybe I should merge them in together? 🤷‍♂️ 
- Github supports both non-expiry and expiry tokens. I have handled both cases
- I have defined a global test adapter for `tesla`, `config :tesla, adapter: Lightning.Tesla.Mock`. I think I should have done this in https://github.com/OpenFn/lightning/pull/1781 instead of defining only one for the `GithubClient`. Consequently, I have been forced to `stub` this with the `Hackney adapter` so that the `Bypass` tests work. 

## Related issue

Fixes #1894 

## Review checklist

- [x] I have verified that all appropriate **authorization policies** (`:owner`, `:admin`, `:editor`, `:viewer`) have been implemented and tested. 
- [x] The callback urls have been updated in the `prod` and `demo` Github Apps Setting

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -68,12 +68,24 @@ github_app_id =
 github_app_name =
   System.get_env("GITHUB_APP_NAME") ||
     Application.get_env(:lightning, :github_app, [])
-    |> Keyword.get(:app_id, nil)
+    |> Keyword.get(:app_name, nil)
+
+github_app_client_id =
+  System.get_env("GITHUB_APP_CLIENT_ID") ||
+    Application.get_env(:lightning, :github_app, [])
+    |> Keyword.get(:client_id, nil)
+
+github_app_client_secret =
+  System.get_env("GITHUB_APP_CLIENT_SECRET") ||
+    Application.get_env(:lightning, :github_app, [])
+    |> Keyword.get(:client_secret, nil)
 
 config :lightning, :github_app,
   cert: decoded_cert,
   app_id: github_app_id,
-  app_name: github_app_name
+  app_name: github_app_name,
+  client_id: github_app_client_id,
+  client_secret: github_app_client_secret
 
 if start_rtm = System.get_env("RTM") do
   unless start_rtm in ["true", "false"] do
--- a/lib/lightning/version_control/version_control.ex
+++ b/lib/lightning/version_control/version_control.ex
@@ -8,8 +8,8 @@ defmodule Lightning.VersionControl do
 
   import Ecto.Query, warn: false
 
-  alias Lightning.Repo
   alias Lightning.Accounts.User
+  alias Lightning.Repo
   alias Lightning.VersionControl.Events
   alias Lightning.VersionControl.GithubClient
   alias Lightning.VersionControl.GithubError
--- a/lib/lightning/version_control/version_control.ex
+++ b/lib/lightning/version_control/version_control.ex
@@ -275,6 +275,8 @@ defmodule Lightning.VersionControl do
     ])
   end
 
+  @spec save_oauth_token(User.t(), map(), notify: boolean()) ::
+          {:ok, User.t()} | {:error, Ecto.Changeset.t()}
   def save_oauth_token(%User{} = user, token, opts \\ [notify: true]) do
     token =
       token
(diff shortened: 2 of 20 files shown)


taylordowns2000 (approved):

config/runtime.exs
```diff
@@ -62,12 +62,24 @@ github_app_id =
...
+
+github_app_client_id =
+  System.get_env("GITHUB_APP_CLIENT_ID") ||
+    Application.get_env(:lightning, :github_app, [])
+    |> Keyword.get(:client_id, nil)
+
+github_app_client_secret =
+  System.get_env("GITHUB_APP_CLIENT_SECRET") ||
+    Application.get_env(:lightning, :github_app, [])
+    |> Keyword.get(:client_secret, nil)
```
jyeshe, line 75: If mandatory on prod, would move to prod section and use a `Keyword.fetch!`. Unless we have some instances where the Github feature would be turned off (possibly on a cluster or set of nodes for free users). cc @taylordowns2000

lib/lightning/version_control/version_control.ex
```diff
@@ -123,6 +127,197 @@ defmodule Lightning.VersionControl do
...
+        false
+    end
+  end
+
+  @doc """
+  Fecthes the access token for the given `User`.
+  If the access token has expired, it `refreshes` the token and updates the `User` column accordingly
+  """
+  @spec fetch_user_access_token(User.t()) ::
+          {:ok, String.t()} | {:error, :expired | map()}
```
jyeshe, line 216: The spec is a bit different `{:error, :expired}`. Could you double check the error handling?

lib/lightning/version_control/version_control.ex
```diff
@@ -8,11 +8,15 @@ defmodule Lightning.VersionControl do
...
   import Ecto.Query, warn: false
 
+  alias Lightning.Accounts.User
   alias Lightning.Repo
+  alias Lightning.VersionControl.Events
   alias Lightning.VersionControl.GithubClient
   alias Lightning.VersionControl.GithubError
   alias Lightning.VersionControl.ProjectRepoConnection
 
+  defdelegate subscribe(user), to: Events
```
jyeshe, line 18: :+1:

jyeshe (commented):
@midigofrank this is great. Found some opportunities for fail fast, to know directly the root cause of some errors and there is one drift on spec for error handling. Wondering why dialyzer has not caught it :thinking:

config/runtime.exs
jyeshe: I have talked with Taylor and the code is correct. The feature might not be used on prod on certain instances.

jyeshe (changes requested):
Hey @midigofrank I wasn't clear about the timing for this changes (considering the latest planning). I would appreciate if you had a look specially at the error spec and on the fail fast when `client_id` is `nil`. These are important to not affect our MTTR in the future. The other one is a quick/simple style thing. Thank you!

jyeshe (approved):
This is great! 🌟

PS: The Github disconnect is on a parallel thread on slack (so ignoring the CI error for this PR)

PR #1914, opened 2024-03-19
Add indexes to improve search performance

### LINKED ISSUE #1898: We add tsvector/gin indexes on logs and dataclips in order to improve search performance.

stuartc:
## Validation Steps

After running the migrations:

- Using `psql`, run `\d dataclips` and find the resulting: `"dataclips_body_index" gin (body)`
- All existing log lines should have their `search_vector` column filled.
- Any new log lines should also have their search column filled.

## Notes for the reviewer

I have not reviewed the queries in the History page as to how they would use the indexes yet, chances are there are changes required there to leverage them.

Searching log lines is a known quantity since it is just text, but with dataclips it could be tricky. The GIN index assumes containment queries, but unlike trigger matches (which used json containment queries) users might need to be educated and/or the UI handle conercing and validation of text inputs into containment queries.

## Related issue

Fixes #1898

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ and this project adheres to
   [#1836](https://github.com/OpenFn/lightning/issues/1836)
 - Added ability to remove project collaborators
   [#1837](https://github.com/OpenFn/lightning/issues/1837)
+- TSVector index to log_lines, and gin index to dataclips
+  [#1898](https://github.com/OpenFn/lightning/issues/1898)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to
 
 ### Added
 
+- TSVector index to log_lines, and gin index to dataclips
+  [#1898](https://github.com/OpenFn/lightning/issues/1898)
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


taylordowns2000 (approved):
@stuartc , i've reviewed and hooked up the tsvector search for log lines [here](https://github.com/OpenFn/lightning/pull/1914/commits/b886e0d04dea344334960e44782268a9ef15c73f).

For the dataclips... hrmmm.... In v1 we did this:

```ex
  @doc """
  Converts jsonb to text and does a VERY EXPENSIVE partial string search over the result.
  """
  def with_string_search(query, %{"string" => string}) do
    soft_search_string = "%" <> string <> "%"

    from(r in query,
      where: fragment("?::text LIKE ?", r.body, ^soft_search_string)
    )
  end
```
And that's almost exactly how we do it in v2 right now:
```ex
ilike(type(dataclip.body, :string), ^"%#{search_term}%")
```
I have [changed](https://github.com/OpenFn/lightning/pull/1914/commits/740b5936b9b1175226a28dffa2daae3d9c06692a) the `ilike` to `like` because we don't want case insensitivity for bodies. This should help our performance a small bit.

What's next???

The most powerful difference between v1 and v2, then, is that v1 slowly stepped back through time, delivering results for the first few days to the UI, then the next few days, then the next few days, etc., etc.

I think we should get this into main now and see if the DB timeouts decrease before considering a shift to a new progressive search result loading thing like v1.

Thoughts?

jyeshe (approved):
Excelent solution. Dataclips searching to be explored more thoroughly using jsonb operators.

Issue #1939, opened 2024-03-26
Allow dataclip body to be searched as text (tsvector)


christad92:
We recently added a `gin` index to `dataclips.body`, however the search query is treating the body as a string.

In order to use a gin index on a JSONB type, the query should use a json operator (`@>`, `@@`, `@?`...).

So in order to search dataclips as text (and be performant), we need to:

1. Cast the json as a tsvector
  Attention should be paid as to the dictionary choice to ignore JSON syntax.
2. Store the vector in a new column (`body_search_vector`?)
3. Add a GIN index to the new column.
4. Change the history page query to use the search vector instead of casting the body as a string.

**NOTE**: re: dictionary choice, we need to make sure that this accommodates a variety of searching patterns - please can we document at least 5 varying  json document structure and/or search string variations. With tsvector it's entirely possible that compound words (camelCase) and other "not English words" might be ignored.

PR #1940, opened 2024-03-26
Fix nightly history delete query timeout

### LINKED ISSUE #1937: DataRetention cron can fail if the DB query takes too long
Sentry Issue: [LIGHTNING-AY](https://openfn.sentry.io/issues/5095397800/?referrer=github_integration)

```
DBConnection.ConnectionError: tcp recv: closed (the connection was closed by the pool, possibly due to a timeout or because the pool has been terminated)
  File "lib/ecto/adapters/sql.ex", line 1047, in Ecto.Adapters.SQL.raise_sql_call_error/1
  File "lib/ecto/adapters/sql.ex", line 945, in Ecto.Adapters.SQL.execute/6
  File "lib/ecto/multi.ex", line 865, in Ecto.Multi.apply_operation/4
  File "lib/ecto/multi.ex", line 844, in Ecto.Multi.apply_operation/5
  File "lib/enum.ex", line 2528, in Enum."-reduce/3-lists^foldl/2-0-"/3
...

(DBConnection.ConnectionError tcp recv: closed (the connection was closed by the pool, possibly due to a timeout or because the pool has been terminated))
```

taylordowns2000:
This PR does a few things:
1. @stuartc , it drops that `dataclips.body` as I think that may have been slowing down the dataclip delete. (Per #1939 , we don't want this index anymore.)
2. It grants a longer timeout to this nightly query. You can pass query timeout options directly to `Repo` and `Multi` commands... this is exactly what we want if there is a query which won't happen often. This multi is used inside an `Enum.each`, so it will run serially, rather than in parallel and we won't be in danger of blocking all our DB connections.
3. It uses the new `Dotenvy` style for configuring useful DB options. (This was useful for performance tweaking in v1.) CC @stuartc again here. Is this how you're envisioning Dotenvy gets used?

Fixes #1937 

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -288,7 +288,20 @@ end
 
 database_url = System.get_env("DATABASE_URL")
 
-config :lightning, Lightning.Repo, url: database_url
+config :lightning, Lightning.Repo,
+  url: database_url,
+  pool_size: env!("DATABASE_POOL_SIZE", :integer, 10),
+  timeout: env!("DATABASE_TIMEOUT", :integer, 15_000)
+
+if System.get_env("DATABASE_QUEUE_TARGET") do
+  config :lightning, Lightning.Repo,
+    queue_target: env!("DATABASE_QUEUE_TARGET", :integer, 10_000)
+end
+
+if System.get_env("DATABASE_QUEUE_INTERVAL") do
+  config :lightning, Lightning.Repo,
+    queue_interval: env!("DATABASE_QUEUE_INTERVAL", :integer, 1000)
+end
 
 if config_env() == :prod do
   unless database_url do
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -291,17 +291,9 @@ database_url = System.get_env("DATABASE_URL")
 config :lightning, Lightning.Repo,
   url: database_url,
   pool_size: env!("DATABASE_POOL_SIZE", :integer, 10),
-  timeout: env!("DATABASE_TIMEOUT", :integer, 15_000)
-
-if System.get_env("DATABASE_QUEUE_TARGET") do
-  config :lightning, Lightning.Repo,
-    queue_target: env!("DATABASE_QUEUE_TARGET", :integer, 10_000)
-end
-
-if System.get_env("DATABASE_QUEUE_INTERVAL") do
-  config :lightning, Lightning.Repo,
-    queue_interval: env!("DATABASE_QUEUE_INTERVAL", :integer, 1000)
-end
+  timeout: env!("DATABASE_TIMEOUT", :integer, 15_000),
+  queue_target: env!("DATABASE_QUEUE_TARGET", :integer, 50),
+  queue_interval: env!("DATABASE_QUEUE_INTERVAL", :integer, 1000)
 
 if config_env() == :prod do
   unless database_url do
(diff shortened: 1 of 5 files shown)


config/runtime.exs
```diff
@@ -306,7 +319,6 @@ if config_env() == :prod do
     # TODO: determine why we see this certs verification warn for the repo conn
     # ssl_opts: [log_level: :error],
     url: database_url,
-    pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
```
taylordowns2000, line 309: Why was this only being configured in `:prod`?

priv/repo/migrations/20240326155113_drop_jsonb_gin_index_from_dataclips.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Lightning.Repo.Migrations.DropJsonbGinIndexFromDataclips do
+  use Ecto.Migration
+
+  def up do
+    drop index(:dataclips, [:body])
+  end
+
+  def down do
+    create index(:dataclips, [:body], concurrently: true, using: :gin)
+  end
```
jyeshe, line 10: Probably won't be used but for the `down` we would need:
```
@disable_ddl_transaction true
@disable_migration_lock true
```

jyeshe (approved):
A detail in the case of a rollback and dropping the index might need to disable the ddl transaction.

stuartc (changes requested):
Dropping looks good, I made some comments about `dotenvy`.

As for the performance thing, I'd like to be proven wrong but I can't imagine why the index would be slowing down the deletion here (not unless there's perhaps a full vacuum being run at the same time).

priv/repo/migrations/20240326155113_drop_jsonb_gin_index_from_dataclips.exs
stuartc: @taylordowns2000 @jyeshe is correct, can you put those in - however unlikely their use will be.

config/runtime.exs
```diff
@@ -288,7 +288,20 @@ end
...
+
+if System.get_env("DATABASE_QUEUE_TARGET") do
+  config :lightning, Lightning.Repo,
+    queue_target: env!("DATABASE_QUEUE_TARGET", :integer, 10_000)
+end
+
+if System.get_env("DATABASE_QUEUE_INTERVAL") do
+  config :lightning, Lightning.Repo,
+    queue_interval: env!("DATABASE_QUEUE_INTERVAL", :integer, 1000)
+end
```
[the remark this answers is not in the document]
taylordowns2000: `database_url = env!("DATABASE_URL", :string?)` throws:
```
** (RuntimeError) Application environment variable DATABASE_URL not set
    (dotenvy 0.8.0) lib/dotenvy.ex:155: Dotenvy.env!/2
    /Users/taylor/build/lightning/config/runtime.exs:289: (file)
    (stdlib 5.0.2) erl_eval.erl:750: :erl_eval.do_apply/7
    (stdlib 5.0.2) erl_eval.erl:494: :erl_eval.expr/6
    (stdlib 5.0.2) erl_eval.erl:136: :erl_eval.exprs/6
```
So I'm going to leave it as it was before. I don't _love_ setting queue target to `nil` _or_ shadowing the elixir defaults, but happy to do so for now

PR #1946, opened 2024-03-28
Create Snapshots for Workflows

### LINKED ISSUE #1822: Create Workflow Snapshots from change records
In prior designs we had the concept of both Snapshots and Revisions. Revisions being individual models at a given point in time and Snapshots being a collection of Revisions building up a particular set of revisions.

For simplicity of implementation we are going to persist the entire workflow in a JSONB column using Ecto’s embedding fields. The major tradeoff here is two operations are very sensitive to this structure:

* Migrations \
In the inevitable event of changing the structure of this JSONB field we would need to perform a potentially expensive operation.
* Reading inside nested data \
The structure needs to expose easily indexable values, as several parts of the application will point directly to things like a jobs name.

Create a WorkflowSnapshots table that has at least the following field

* id
* workflow_id
* snapshot \
The JSONB column

### The Embed

Special attention needs to be paid to how we persist workflow snapshots. Since Snapshot will be read by several places; below being a non-exhaustive:

* Every work order and run in the History page
* Whenever viewing a Run
* Viewing a workflow/job code for anything in the past.

The shape of the data, and the indexability is critical as the History page will need to retrieve job names out of a nested structure.

### Building Snapshots

Carbonite (may) only keep the changes for a given table/transaction, in which case the starting point/baseline state for a workflow needs to be seeded.

#### Approach

1. Collect up all Carbonite change records 
2. Take the workflow in its current state (or the last snapshot) and apply the changes.
3. Persist it
    1. Optionally ‘tagging’ the snapshot with a checksum or number.

We _want_ to only create a snapshot when the changed workflow is used, allowing us to minimize storage creating snapshots every time a user presses Cmd+S.

We have two approaches here:

1. **Pruning**
Where we create a new snapshot with every save, and prune periodically removing snapshots that are not associated with any runs/steps or work orders.
    - Simplest as pruning can happen out of band, and doesn’t intersect with snapshots that are associated with existing records which makes it less sensitive to race conditions.
    - May cause avoidable pressure on the DB especially when it comes to GIN indexes being built against a ‘medium’ sized JSON object.
    - Gets records out of Carbonite immediately lessening dependency (slightly) 
2. **Compaction**
Where we leave the Carbonite change records to accumulate and reduce over them at the time a WorkOrder is requested.
    - Slightly more complex but is more efficient - we don’t leave snapshots which are unused and potentially almost exactly the same as the one before.
    - Requires some kind of transaction lock ensuring that only one request for a snapshot gets to compact a new snapshot.
    - Reduces the risk of records getting added and potentially ignored, as with pruning every Workflow affecting operation would be _required_ to not only capture the changes but also create a new snapshot.

_Neither of these approaches have been decided on, however **Compaction** is likely the first one we will attempt._

### LINKED ISSUE #1823: Create migration for existing Workflows
When building the WorkflowSnapshot feature we need to ensure that existing WorkOrders, Runs and Steps are associated with a snapshot.

An initial snapshot must be created, and all existing WorkOrder models must be link to this snapshot.

### LINKED ISSUE #1824: Associate Models with Workflow Snapshots
There are three tables that are currently associated with a Workflow (or it's dependants).

- WorkOrders
  - Workflow
  - Trigger
- Runs
  - Job (as `starting_job`)
- Steps
  - Job

All of these relationships must be extended to include a `snapshot_id` to link the foreign key (`job_id`, `trigger_id` etc) to the record in the snapshot instead of directly model.

### LINKED ISSUE #1826: Associate a workorder/run with the current snapshot
When a WorkOrder/Run is created, it must be associated with the current snapshot.

## Notes

When a workflow is saved, we just add changes to the changes table.
When we Save and Run a workflow, we proactively create a new snapshot.
It is important not to rely on the 'compaction' functionality in the queue as there will be caching to avoid over querying for changes to make a new snapshot.

stuartc:
## Validation Steps

- Create or update a workflow
- Inspect the database for a `workflow_snapshot` record reflecting the up to date state of the workflow.

## Notes for the reviewer

This PR currently ignores transactional changes (Ala Carbonite), and focusses just on creating a snapshot for a given workflow.

Currently the demo setup script does not assemble workflows the same way the UI does and there are 'empty' snapshots after running the demo script.

The provisioning API does not invoke the 'save workflow' function, and currently does not result in a snapshot being created.

## Related issue

Fixes #1822 
Fixes #1823 
Fixes #1824 
Fixes #1826 
Fixes #1824

## Review checklist


lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -834,7 +842,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
       %{assigns: %{changeset: changeset}} =
         socket = socket |> apply_params(next_params)
 
-      Lightning.Repo.insert_or_update(changeset)
+      Workflows.save_workflow(changeset)
```
taylordowns2000, line 845: phew!

lib/lightning/workflows/snapshot.ex
```diff
@@ -0,0 +1,192 @@
...
+  defp get_latest_query(workflow) do
+    from(s in __MODULE__,
+      where: s.workflow_id == ^workflow.id,
+      order_by: [desc: s.inserted_at],
+      limit: 1
+    )
+  end
+
+  @doc """
+  Get the latest snapshot for a workflow, or create one if it doesn't exist.
```
taylordowns2000, line 163: I really like this. I'm assuming that it will be called when one of those "oh dang we really need a snapshot now" actions is performed:
1. manually initiate a github sync (must send _current_ version when button is pressed)
2. execute a run
3. something else maybe?

taylordowns2000 (changes requested):
This looks awesome @stuartc . My thoughts on your two concerns:

> - The demo script doesn't create workflows in a very clean way resulting currently in a blank snapshot

We should break this out into a separate issue but _not_ block merging because of it.

> - The provisioning API doesn't use the Workflow functions and therefore currently doesn't create snapshots.

This we should fix before merging. The provisioning API (i think [`import_document`](https://github.com/OpenFn/lightning/blob/1822_create_snapshots/lib/lightning/projects/provisioner.ex#L32-L45)) is so critical because of the GitHub sync. Given how shaky that part of the application has been I'm nervous about introducing any sort of "oh it's different there but we'll fix it later" stuff. Updating that API to use snapshots would be a win for snapshots _and_ for the GitHub sync.

That's my 2¢

lib/lightning/workflows/edge.ex
```diff
@@ -47,7 +47,7 @@ defmodule Lightning.Workflows.Edge do
 
     field :delete, :boolean, virtual: true
 
-    timestamps()
+    timestamps(type: :utc_datetime)
```
jyeshe, line 50: It has not failed but on the database it's still without time zone:
Screenshot from 2024-04-04 12-29-18

jyeshe (changes requested):
@stuartc this is clean and straight to the point. I'm wondering about two details:
1. If we want to migrate the timestamps now to utc;
2. What if we could associate the runs with snapshots?

lib/lightning/workflows/edge.ex
stuartc: That's the expected type, Ecto's `utc_datetime` types in migrations don't use a timezone option in column types. The application/repo handles the tz conversion.

It's a tricky path to deviate from, having the db set to another tz, or have Ecto store as a different one. So we're using the defaults provided by Ecto, everything in UTC via Ecto and if we want to do logic/presentation in a different tz it all happens in queries or frontend functions.

stuartc:
> @stuartc this is clean and straight to the point. I'm wondering about two details:
> 
> 1. If we want to migrate the timestamps now to utc;
> 2. What if we could associate the runs with snapshots?

Thanks! 

1. everything is already 'in utc' via Ecto (addressed in the above comment).
2. that's the plan, see: #1826

lib/lightning/workflows/snapshot.ex
stuartc: Yep thats the case

PR #1948, opened 2024-03-28
Handle error translations for jobs

### LINKED ISSUE #1935: Fix validation message when job name is duplicated in a workflow
The error/validation message for duplicate job names in a workflow is misleading as it reads "The email already exists". We should change this to reflect the error by setting it to "Job names should be distinct". 

See image below
Image

elias-ba:
## Notes for the reviewer

This PR addresses error messages for the Job model. It proposes to generate custom generic error messages for the Ecto changesets associated to the Job model then translate those messages using the errors.po file. This will allow a better management of the error messages and it can be replicated for all the other form models of the app.

## Related issue

Fixes #1935

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to
 
 ### Changed
 
+- Translate job error messages using errors.po file
+  [#1935](https://github.com/OpenFn/lightning/issues/1935)
+
 ### Fixed
 
 ## [v2.2.1] - 2024-03-27
--- a/assets/js/workflow-diagram/nodes/Node.tsx
+++ b/assets/js/workflow-diagram/nodes/Node.tsx
@@ -30,7 +30,7 @@ type LabelProps = React.PropsWithChildren<{
 
 function errorsMessage(errors: ErrorObject): string {
   const messages = Object.entries(errors).map(([key, errorArray]) => {
-    return `${key} ${errorArray.join(', ')}`;
+    return `${errorArray.join(', ')}`;
   });
 
   return messages.join(', ');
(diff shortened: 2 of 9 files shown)


christad92 (changes requested):
Thanks @elias-ba. 
Here are a few comments: 

- Let's trim spaces around the job names so that we are not saving `New Job` and `New Job ` as two different job names. 
- There is a similar case in the editor when the code editor is empty. It returns "step can't be empty". Please let's change this to "Code editor cannot be empty"

Screenshot 2024-03-29 at 13 30 59

jyeshe (approved):
This fixes the issue as described on #1935. We would just need to be aware that the "has already been taken" is inferred as a duplicate email. 
https://github.com/OpenFn/lightning/blob/abe037a059c9abb06714cd26764d4531e6847b9c/priv/gettext/en/LC_MESSAGES/errors.po#L4-L5
Maybe open a more generic issue @christad92?

taylordowns2000 (approved):
all of @christad92 's change requests have been met. thanks, elias.

christad92:
> This fixes the issue as described on #1935. We would just need to be aware that the "has already been taken" is inferred as a duplicate email.
> 
> https://github.com/OpenFn/lightning/blob/abe037a059c9abb06714cd26764d4531e6847b9c/priv/gettext/en/LC_MESSAGES/errors.po#L4-L5
> 
> 
> Maybe open a more generic issue @christad92?

Yeah, I'd propose we use a better `msgid` like "email already taken"  or something like that. I'll take a look at the dictionary to see what suggestions I'd propose.

PR #1966, opened 2024-04-03
Allow multiple projects per repo

### LINKED ISSUE #1811: GitHub Sync should allow multiple projects per repo
## Context

1. Right now, the "Install/Reconfigure Github Sync" adds a single set of workflows that use the openfn Github action to a target repository+branch.
2. To allow multiple repo<>project connections to be made in the _same_ repo (allowing for a common "`staging` to test" and "`main` to prod" pattern) we should modify the Lightning installation code, the workflows, and the Github actions.

### Modify PRC:
- [x] #1892 

### Modify OpenFn Github "actions"
- [x] Update the [gtihub deploy action here](https://github.com/OpenFn/cli-deploy-action/blob/main/action.yml) so that it can take a _variable_ `config.json`—the file that points to a `project.yaml` and to a `projectState.json`
- [x] Update the [github pull action here](https://github.com/OpenFn/cli-pull-action/blob/main/action.yml) so that it can take a _variable_ `config.json` also.

### Modify auto-generated workflows
- [x] Update the auto-generated [`pull.yml`](https://github.com/OpenFn/demo-openhie/blob/main/.github/workflows/pull.yml) so that it handles a specific BRANCH and allows the user to _either_ specify a path to an existing `config.json` or generates a `config.json` for the user which is named `project-name.json` and points to a `./project.yaml` and a `project-name-state.json`
- [x] Update the auto-generated [`deploy.yml`](https://github.com/OpenFn/demo-openhie/blob/main/.github/workflows/deploy.yml) so that it handles a specific BRANCH and allows the user to _either_ specify a path to an existing `config.json` or generates a `config.json` for the user which is named `project-name.json` and points to a `./project.yaml` and a `project-name-state.json`

image

## For validation
- You should be able to set up two different OpenFn projects "my-test" and "my-prod" and have them connected to the same github repository.
- You should see 4 workflows (two for "my-test" and two for "my-prod").
- You will add two different config files to your repo: `test-config.json` and `prod-config.json`. These config files will be referenced in the `pathToConfig` attribute of the `project_repo_connection`.
- When you initiate sync, you should see one `project.yaml` and two project state files: `my-prod-state.json` and `my-test-state.json`.
- On commit to the `staging` branch, you should see an update to the "my-test" OpenFn project.
- On commit to the `main` branch, you should see an update to the "my-prod" OpenFn project.
- When staging is _merged_ to the `main` branch, you should see that both OpenFn projects now have identical configuration.

Image

### LINKED ISSUE #1892: Github "sync to branch" doesn't target branch—uses default
**_Please note_** that we are currently using a "repository dispatch event" and we must change to use a "workflow dispatch event".

1. When the user clicks "Sync to branch", a repository dispatch event is triggered.
2. No target branch is provided, so it initiates a sync to the default branch (usually `main`) despite the user having set up the workflows on the `staging` branch, for example.
3. Use [this documentation](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event) to specify the workflow dispatch event for a target branch.
4. To accomplish this, we'll need to know/store the workflow ID for the "pull" workflow for a specific PRC. That data can be kept on the PRC table.

https://github.com/OpenFn/Lightning/blob/main/lib/lightning/version_control/version_control.ex#L118

### LINKED ISSUE #1895: Github - allow user to create Project Repo Connection ONLY from existing installations
- Require the user to set up personal github access first - https://github.com/OpenFn/lightning/issues/1894
- Use the current user's github oauth token to see which installations they have access to.
- Give them three two dropdowns... installations, repos, branches.
- Give them a text input to specify the path to their config file - #1811 
- If they don't see what they want, allow them to hyperlink off to `https://github.com/apps/openfn/installations/select_target`
- When they come back to OpenFn, they can refresh those lists.

## If the user hasn't yet connected their account to Github
https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=2062%3A1533&mode=design&t=OEZgJslyfupUtLUE-1
Image

## Once the user has connected their account to Github
https://www.figma.com/file/2Uvr2jOaQh6rQNR1cHlLKM/App?type=design&node-id=2090%3A2467&mode=design&t=OEZgJslyfupUtLUE-1

Image

midigofrank:
### Required ENVs

- `GITHUB_APP_ID`
- `GITHUB_APP_NAME`
- `GITHUB_APP_CLIENT_ID`
- `GITHUB_APP_CLIENT_SECRET`
- `GITHUB_CERT`
- `REPO_CONNECTION_SIGNING_SECRET`

Check the [README](https://github.com/OpenFn/lightning/pull/1966/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) on how to install `libsodium`

#### Connecting 
Users can connect their project to a github repo which they have access to (on github).
Only an `admin` or `owner` can add the github connection. 
Screenshot 2024-04-04 at 22 07 29

When you click `connect branch`, the following will happen:

- [x] `.github/workflows/openfn-pull.yml` is committed into the **default branch** of the chosen repo
- [x] `.github/workflows/openfn-{{projectId}}-deploy.yml` is committed into the selected branch
- [x] `openfn-{{projectId}}-config.json` is committed into the selected branch
- [x] `ProjectRepoConnection` record is created with an `access_token`
- [x] `OPENFN_{{projectId}}_API_KEY` repo secret is added to the repo whose value is the `access_token` above.
- [x] If you choose _Sync to GitHub immediately upon setup_ , "[skip actions]" is added to the commit message getting in the selected branch and `initiate sync` is called. [skip actions] is included so that the deploy workflow is not triggered.
- [x] If you choose _Deploy from GitHub immediately upon setup_ , "[skip actions]" is NOT added to the commit message getting in the selected branch and `initiate sync` is NOT called. This triggers the deploy workflow in github

#### Reconnecting 
After successful connection, incase your the github repo was not properly setup, you get an option to `reconnect`. 
A properly setup repo meets the following criteria:
- [x] has `.github/workflows/openfn-pull.yml` in the **default branch** 
- [x] has a `.github/workflows/openfn-{{projectId}}-deploy.yml` in the **target branch**
- [x] has a `openfn-{{projectId}}-config.json` in the **target branch**
- [x] has an `OPENFN_{{projectId}}_API_KEY` repo secret

Only an `admin` or `owner` can `reconnect`. 
The reconnect button will only be shown if and only if the acting user has access to the installation, regardless of their role.

#### Disconnecting / Remove Integration
Only an `admin` or `owner` can `disconnect`.

You don't need github oauth setup for your account in order to perform this action. But if you have set it up, the actions below will occur, if you don't they'll just get ignored
- [x] `.github/workflows/openfn-{{projectId}}-deploy.yml` is deleted from the selected branch
- [x] `openfn-{{projectId}}-config.json` is deleted from the selected branch
- [x] `OPENFN_{{projectId}}_API_KEY` repo secret gets removed

Each delete operation will have it's own commit message. We will therefore have **2 commits** 

#### Initialize Sync
This sends a workflow dispatch event to github which in turn triggers the `gh action pull`

Project `editor`, `admin` and `owner`(s) are allowed to initialize sync 

#### API

- [x] Authenticate API requests made using `Bearer repo.connection.access_token`

## Notes for the reviewer

- Dropped the `user_id` column and adds `access_token` to `ProjectRepoConnection`
- Renamed `current_user` assign to `current_resource` on the api route to account `ProjectRepoConnection` as a resource that can used to authorize. Here's the specific change: https://github.com/OpenFn/lightning/pull/1966/commits/c7210f888d8870cacabfc6fc05eec2991534ba4f
- `ProjectRepoConnection` access tokens are prefixed with `prc_`. This prefix is what we use to differentiate it at the authorization plug. Also these tokens are generated the same way we do for `api user token`, i.e. `Joken.sign(%{project_id: "someid"})`
- Deleted the github installation end to end test: https://github.com/OpenFn/lightning/pull/1966/commits/6a80d2b7e503c6d879f1a0bcca1cad5c0401db29 . This because the introduction of oauth tokens makes it a little harder to maintain. For instance, whose account shall we use for the token? What do you think? Am I overcomplicating it?

## Related issue

Fixes #1811
Fixes #1892 
Fixes #1895

## Review checklist

- [x] Releases are created for https://github.com/OpenFn/cli-pull-action/pull/1 and https://github.com/OpenFn/cli-deploy-action/pull/1
- [x] Update the `pull` and `deploy` files in the project to point to the openfn actions. They're currently point to `midigofrank/cli-{pull, deploy}-action`
- [ ] Generate `PRC_SIGNING_SECRET` for `prod` and `demo`


taylordowns2000 (approved):

lib/lightning/version_control/project_repo_connection.ex
```diff
@@ -26,20 +25,69 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
...
+    else
+      changeset
+    end
+  end
+
+  defp generate_access_token(project_id) do
+    Joken.generate_and_sign!(default_claims(skip: [:exp]), %{
+      "project_id" => project_id
+    })
+  end
```
stuartc, line 87: Tell me more about this token, how would we rotate it and where is the signing secret? If it's using the default in `config.exs` we need to enforce using one from an env variable (like we do for the worker)

lib/lightning_web/components/github_components.ex
```diff
@@ -0,0 +1,38 @@
...
+    default:
+      "text-center py-2 px-4 shadow-sm text-sm font-medium rounded-md focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-primary-500 bg-primary-600 hover:bg-primary-700 text-white"
+
+  attr :user, Lightning.Accounts.User, required: true
+
+  def connect_to_github_link(assigns) do
+    ~H"""
+    <.link
+      id={@id}
+      href={"https://github.com/login/oauth/authorize?" <> build_query_params()}
```
stuartc, line 18: can you put this into an assign, I'm not sure that LiveView will optimise this - causing it to be rerendered every time.

lib/lightning_web/controllers/user_auth.ex
```diff
@@ -172,10 +173,12 @@ defmodule LightningWeb.UserAuth do
 
   def authenticate_bearer(conn, _opts) do
     with {:ok, bearer_token} <- get_bearer(conn),
-         user when not is_nil(user) <-
-           Accounts.get_user_by_api_token(bearer_token) do
-      update_last_used(bearer_token)
-      assign(conn, :current_user, user)
+         %_{} = resource <- get_current_resource(bearer_token) do
```
stuartc, line 176: How about? Either way the `%_{}` underscore here doesn't serve any purpose..

```
%{__struct__: type} = resource <- ... do
  if type == User do
    update_last_used(bearer_token)
  end
  
  ...
```

lib/lightning_web/controllers/user_auth.ex
```diff
@@ -190,6 +193,16 @@ defmodule LightningWeb.UserAuth do
...
   end
 
+  defp get_current_resource(token) do
+    case token do
+      "prc_" <> _rest ->
+        Lightning.VersionControl.get_repo_connection_for_token(token)
+
+      _user_token ->
+        Accounts.get_user_by_api_token(token)
+    end
```
stuartc, line 203: Hmm I'm not in love with the `prc_` prefix on all tokens, right? It's `prc_<JWT>` hey?

stuartc (changes requested):
She, this is massive... looks like a fantastic set of work...

I've left a few comments - I haven't looked at everything, but the `prc_` prefix isn't great. I think we should pay the cost of two queries instead of having _all_ tokens need custom parsing to decode.

Also we need to separate out the `default_signer` for joken - the default is just `secret` meaning these can be forged really easily.

I would really like us to use `Lightning.Config` (and Mox) to be the call site for getting the GitHub app config, and allow us to ditch `put_temporary_env` for testing. But I get it if we're running short on time.

lib/lightning_web/live/profile_live/edit.ex
```diff
@@ -9,7 +9,10 @@ defmodule LightningWeb.ProfileLive.Edit do
 
   @impl true
   def mount(_params, _session, socket) do
-    VersionControl.subscribe(socket.assigns.current_user)
+    if connected?(socket) do
+      VersionControl.subscribe(socket.assigns.current_user)
+    end
+
```
stuartc, line 15: 👌

stuartc (approved):

PR #1967, opened 2024-04-03
sort projects in picker by name

taylordowns2000:
this sorts the projects in the project picker by name


christad92 (approved):
Thanks for this! tested and it works.

elias-ba (approved):
Looking good 👏🏽

PR #1970, opened 2024-04-04
Change all System.get_env calls to use Dotenvy

### LINKED ISSUE #1968: Ensure we use dotenvy everywhere in runtime.exs
### Details

We switched to using dot envy a few weeks ago, but we didn't roll it out in all the different places where environment variables are used.

If you developers have bumped into issues where either their environment variables or dot files are applied inconsistently.

We need to ensure that everything inside runtime.exs uses `env!` and not `System.get_env`.

### Implementation notes

There is a call to `System.get_env` in `repo.ex`, we need to verify that value is not clobbered
or set differently when a regular exported env is set _and_ a .env file is also provided.

stuartc:
## Validation Steps

_How can a reviewer validate your work?_

Run the app using various env configuration that you are aware of, ideally tested on staging with Github and Google integrations enabled.

## Related issue

Fixes #1968 

## Review checklist

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -34,7 +34,7 @@ config :lightning,
        :is_resettable_demo,
        env!(
          "IS_RESETTABLE_DEMO",
-         :boolean,
+         &Utils.ensure_boolean/1,
          Application.fetch_env!(:lightning, :is_resettable_demo)
        )
 
@@ -110,7 +110,7 @@ config :lightning, Lightning.Runtime.RuntimeManager,
   start:
     env!(
       "RTM",
-      :boolean,
+      &Utils.ensure_boolean/1,
       Utils.get_env([:lightning, Lightning.Runtime.RuntimeManager, :start])
     )
 
@@ -173,7 +173,7 @@ config :lightning, :oauth_clients,
 config :lightning,
   schemas_path:
     env!(
-      "SCHAMAS_PATH",
+      "SCHEMAS_PATH",
       :string,
       Utils.get_env([:lightning, :schemas_path], "./priv")
     )
@@ -256,7 +256,7 @@ config :lightning,
 
 config :lightning,
        :init_project_for_new_user,
-       env!("INIT_PROJECT_FOR_NEW_USER", :boolean, false)
+       env!("INIT_PROJECT_FOR_NEW_USER", &Utils.ensure_boolean/1, false)
 
 # To actually send emails you need to configure the mailer to use a real
 # adapter. You may configure the swoosh api client of your choice. We
@@ -320,8 +320,10 @@ if config_env() == :prod do
     """
   end
 
-  maybe_ipv6 = if env!("ECTO_IPV6", :boolean, false), do: [:inet6], else: []
-  disable_db_ssl = env!("DISABLE_DB_SSL", :boolean, false)
+  maybe_ipv6 =
+    if env!("ECTO_IPV6", &Utils.ensure_boolean/1, false), do: [:inet6], else: []
+
+  disable_db_ssl = env!("DISABLE_DB_SSL", &Utils.ensure_boolean/1, false)
 
   config :lightning, Lightning.Repo,
     ssl: not disable_db_ssl,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1,3 +1,4 @@
+alias Lightning.Config.Utils
 import Config
 import Dotenvy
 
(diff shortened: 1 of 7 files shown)


config/runtime.exs
```diff
@@ -121,49 +137,54 @@ config :lightning, :workers,
...
+    client_id: env!("SALESFORCE_CLIENT_ID", :string, nil),
+    client_secret: env!("SALESFORCE_CLIENT_SECRET", :string, nil)
   ]
 
 config :lightning,
   schemas_path:
-    System.get_env("SCHEMAS_PATH") ||
-      Application.get_env(:lightning, :schemas_path) || "./priv"
+    env!(
+      "SCHAMAS_PATH",
```
rorymckinley, line 176: @stuartc Typo here.

config/runtime.exs
```diff
@@ -18,30 +26,36 @@ source!([
...
-if is_resettable_demo = System.get_env("IS_RESETTABLE_DEMO") do
-  config :lightning,
-         :is_resettable_demo,
-         is_resettable_demo == "yes"
-end
+config :lightning,
+       :is_resettable_demo,
+       env!(
+         "IS_RESETTABLE_DEMO",
+         :boolean,
```
jyeshe, line 37: .env.example needs to be updated with `IS_RESETTABLE_DEMO=false` because IS_RESETTABLE_DEMO=no is evaluated to true:
```
$ export IS_RESET=no; iex
iex(1)> Mix.install [{:dotenvy, "0.8.0"}]
:ok
iex(2)> Dotenvy.env!("IS_RESET", :boolean, false)
true
```
or force the use of with `:boolean!`

jyeshe (changes requested):
Hi Stu, docker compose continued working and DATABASE_URL on mix also works. However the demo is not anymore:
`$ IS_RESETTABLE_DEMO=true mix run -e "Lightning.Demo.reset_demo"`
resulted in:
```
[info] [SRV] ✔ Worker public-actors-start listening on 2222
[info] [SRV] ❯ Connecting to Lightning at ws://localhost:4000/worker
[info] [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://localhost:4000/worker
[info] [SRV] ❯ connect ECONNREFUSED 127.0.0.1:4000
```

rorymckinley (changes requested):
@stuartc Marking this as 'Request changes' for the `SCHAMAS_PATH` typo.

The DATABASE_URL change works for me - I could run both IEx and Phoenix and the code picked up the URL from the .env file(s).

Other than that, tested PromEx and UsageTracking stuff as much as I could localhost (at a guess more than 90% of the functionality). I also eyeballed the env vars and checked that ENV changes correctly propagated. I can't pick up any regressions - nicely done!

config/runtime.exs
stuartc: Good catch! I've changed it to use `ensure_boolean/1` which accepts `true|false|yes|no` - which also rejects other arbitrary values and don't consider the existance of the var as `true`.

config/runtime.exs
stuartc: :+1:

stuartc:
> Hi Stu, docker compose continued working and DATABASE_URL on mix also works. However the demo is not anymore: `$ IS_RESETTABLE_DEMO=true mix run -e "Lightning.Demo.reset_demo"` resulted in:
> 
> ```
> [info] [SRV] ✔ Worker public-actors-start listening on 2222
> [info] [SRV] ❯ Connecting to Lightning at ws://localhost:4000/worker
> [info] [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://localhost:4000/worker
> [info] [SRV] ❯ connect ECONNREFUSED 127.0.0.1:4000
> ```

Oh interesting... thanks for finding this. I need to find out what was the previous behaviour, did the server start or did the RTM not start. My hunch is that server would start - which it _shouldn't_ imo; but also then the RTM shouldn't start as well. Will dig in.

jyeshe:
> > Hi Stu, docker compose continued working and DATABASE_URL on mix also works. However the demo is not anymore: `$ IS_RESETTABLE_DEMO=true mix run -e "Lightning.Demo.reset_demo"` resulted in:
> > ```
> > [info] [SRV] ✔ Worker public-actors-start listening on 2222
> > [info] [SRV] ❯ Connecting to Lightning at ws://localhost:4000/worker
> > [info] [SRV] ✘ CRITICAL ERROR: could not connect to lightning at ws://localhost:4000/worker
> > [info] [SRV] ❯ connect ECONNREFUSED 127.0.0.1:4000
> > ```
> 
> Oh interesting... thanks for finding this. I need to find out what was the previous behaviour, did the server start or did the RTM not start. My hunch is that server would start - which it _shouldn't_ imo; but also then the RTM shouldn't start as well. Will dig in.
> 
> _edit_ @jyeshe, found the issue - was an interesting one. The reset _was_ working correctly **but** the log levels weren't being set correct and actually resulted in a crash if you tried to set it via an env variable... 🙌 could you take it for another spin?

tricky one, it works! thanks @stuartc

jyeshe (approved):

PR #1984, opened 2024-04-15
limit activation of workflows

### LINKED ISSUE #1992: Check workflow limits before adding workflow
This ensures that we have a pathway to check thunderbolt.limit_action when enabling workflows triggers

midigofrank:
## Notes for the reviewer

- Renames the action to `:activate_workflow` instead of `:new_workflow`
- Checks if the workflow is activated before making call to the limiter.
- `:activate_workflow` is added to the api provisioning flow as well. 
- Introduces a `WorkflowUpdated` event which emitted when the workflow is created/updated. Also when the workflow is marked for deletion  

## Related issue

Fixes #1992 

## Review checklist

### DIFF
--- a/lib/lightning/extensions/usage_limiting.ex
+++ b/lib/lightning/extensions/usage_limiting.ex
@@ -12,11 +12,11 @@ defmodule Lightning.Extensions.UsageLimiting do
   defmodule Action do
     @moduledoc false
     @type t :: %__MODULE__{
-            type: :new_run | :new_workflow | :new_user,
+            type: :new_run | :activate_workflow | :new_user,
             amount: pos_integer()
           }
 
-    defstruct type: nil, amount: 1
+    defstruct type: nil, amount: 1, changeset: nil
   end
 
   defmodule Context do
--- a/lib/lightning/extensions/usage_limiting.ex
+++ b/lib/lightning/extensions/usage_limiting.ex
@@ -16,7 +16,7 @@ defmodule Lightning.Extensions.UsageLimiting do
             amount: pos_integer()
           }
 
-    defstruct type: nil, amount: 1, changeset: nil
+    defstruct type: nil, amount: 1
   end
 
   defmodule Context do
--- a/lib/lightning/projects/provisioner.ex
+++ b/lib/lightning/projects/provisioner.ex
@@ -12,17 +12,15 @@ defmodule Lightning.Projects.Provisioner do
   import Ecto.Query
 
   alias Lightning.Accounts.User
-  alias Lightning.Extensions.UsageLimiting.Action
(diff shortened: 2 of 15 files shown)


stuartc (approved):
Thanks @midigofrank, I'll merge this - but I think we should move the behaviour inside these helpers into the workflows context. 

I'm not sure what I was thinking putting this in here - but I think these functions have enough responsibility to be part of `Lightning` and not helpers anymore (or at least the limiting part); perhaps not inside `save_workflow` etc but this is business domain logic and should be in Lightning's contexts. 

Lets do something about it in the near future.

jyeshe (changes requested):
Good job @midigofrank ! The save workflow is working well. Got an error trying to save a job change when the limit has already been reached:
```
[error] GenServer #PID<0.1410.0> terminating
** (Ecto.Query.CastError) deps/ecto/lib/ecto/association.ex:760: value `""` cannot be dumped to type :binary_id in query:

from c0 in Lightning.Credentials.Credential,
  join: p1 in Lightning.Projects.ProjectCredential,
  on: p1.credential_id == c0.id and p1.id == ^"",
  select: c0

    (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir 1.16.2) lib/enum.ex:1826: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto 3.10.3) lib/ecto/repo/queryable.ex:211: Ecto.Repo.Queryable.execute/4
    (ecto 3.10.3) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
    (ecto 3.10.3) lib/ecto/repo/queryable.ex:151: Ecto.Repo.Queryable.one/3
    (lightning 2.4.0) lib/lightning_web/live/workflow_live/editor_pane.ex:42: LightningWeb.WorkflowLive.EditorPane.update/2
    (phoenix_live_view 0.20.5) lib/phoenix_live_view/utils.ex:498: Phoenix.LiveView.Utils.maybe_call_update!/3
    (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
    (phoenix_live_view 0.20.5) lib/phoenix_live_view/diff.ex:681: anonymous fn/4 in Phoenix.LiveView.Diff.render_pending_components/6
```

jyeshe:
@midigofrank It might be an opportunity to add the limiter to the create workflow button on the dashboard page. It would prevent this error and anticipate to the user that the workflow cannot be created.

jyeshe (approved):

PR #1986, opened 2024-04-15
Custom metric to track number of finalised runs

### LINKED ISSUE #1790: Adjust exported metrics

rorymckinley:
## Validation Steps

- Start a local lightning instance running and open an IEx session
- Browse to `http://localhost:4000/metrics` - if you search for `lightning_run_queue_finalised_count` on the page, it will most likely have a value of 0.
- In the IEx session:

```
alias Lightning.Repo
alias Lightning.Run

runs = Run |> Repo.all()
number_of_runs = runs |> Enum.count() # if this is zero, you will need to run the demo setup.
finished_at  = DateTime.add(DateTime.utc_now(), 30)

import Ecto.Changeset
# NOTE: Once you have executed the below code move on to the next step of the instructions immediately. 
# If you find that it you do not have enough time, you can set the `finished_at` value further into the future.
for run <- runs do
  run |> change(%{finished_at: DateTime.add(DateTime.utc_now(), 30)}) |> Repo.update!()
end
```

- Return to `http://localhost:4000/metrics` - if you refresh the page you should see the value of `lightning_run_queue_finalised_count` change to match the `number_of_runs` from the IEx session above, and then return to zero.

## Notes for the reviewer

This change introduces a new metric that tracks the number of runs finished in the last n seconds where n also corresponds to the period of the metric polling job.

## Related issue

Fixes #1790 

## Review checklist

### DIFF
--- a/lib/lightning/runs/prom_ex_plugin.ex
+++ b/lib/lightning/runs/prom_ex_plugin.ex
@@ -12,8 +12,11 @@ defmodule Lightning.Runs.PromExPlugin do
   alias Lightning.Repo
   alias Lightning.Run
 
+  require Run
+
   @available_count_event [:lightning, :run, :queue, :available]
   @average_claim_event [:lightning, :run, :queue, :claim]
+  @finalised_count_event [:lightning, :run, :queue, :finalised]
   @stalled_event [:lightning, :run, :queue, :stalled]
 
   @impl true
@@ -116,7 +119,7 @@ defmodule Lightning.Runs.PromExPlugin do
     Polling.build(
       :lightning_run_queue_metrics,
       period_in_seconds * 1000,
-      {__MODULE__, :run_queue_metrics, [run_age_seconds]},
+      {__MODULE__, :run_queue_metrics, [run_age_seconds, period_in_seconds]},
       [
         last_value(
           [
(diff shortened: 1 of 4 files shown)


jyeshe (commented):
Hi @rorymckinley is there anything I should do to enable the `/metrics` endpoint? As it's not mandatory anymore I am running `mix phx.server` without the .env

lib/lightning/runs/prom_ex_plugin.ex
```diff
@@ -137,17 +140,29 @@ defmodule Lightning.Runs.PromExPlugin do
...
-      check_repo_state(pid)
+  def run_queue_metrics(run_age_seconds, consideration_window_seconds) do
+    if repo_pid = Process.whereis(Repo) do
+      check_repo_state(repo_pid)
 
       trigger_run_claim_duration(run_age_seconds)
       trigger_available_runs_count()
+
+      trigger_finalised_runs_count(
+        DateTime.utc_now(),
```
jyeshe, line 163: since trigger is private and now is used for the threshold it can be inferred that count_finalised_runs will work based on current time. I would stay with two options:
```
A. count_finalised_runs receives the threshold_datetime
B. count_finalised_runs receives the window_seconds

jyeshe (changes requested):
Hey @rorymckinley. It worked with the `Lightning.Runs.PromExPlugin.count_finalised_runs(DateTime.add(DateTime.utc_now(), 30), 40)` after reseting the demo. I have a small simplification refactor request.

lib/lightning/runs/prom_ex_plugin.ex
rorymckinley: @jyeshe Thanks for the suggestion. I pass in `now` for the same reason that I made the `count_finalised_runs` method public rather private: To allow for tests that allow for higher-fidelity and better readability while removing the risk of a test that may flicker in the future. 

By exposing this interface, I can test the primary logic by passing in the same fixed datetime that was used to generate my fixtures and this means that I do not have to add in a more error-prone time tolerance to deal with the fact that the now() used by the production code may have drifted a nontrivial amount in comparison with the now() used to set up my fixtures.

Of course, I am working on the assumption that there is no cheap, non-brittle way to lock the value of now() for an Elixir test (e.g. something similar to Ruby's Timecop). If this assumption is incorrect, I am happy to modify this method to align with your suggestions above.

However, if my assumption is correct, I am hoping that I can convince you that the testing benefits outweigh the cost of the messier method signature?

rorymckinley: @jyeshe Firstly, my apologies for the above screed :) on closer inspection of the tests I realised that the change you requested will result in far less impact to test fidelity than I expected. In practical terms, the impact is probably zero!

I have made the change as requested in a new commit.

jyeshe (approved):
:100:

stuartc (approved):

PR #1989, opened 2024-04-15
Allow customized menu items

### LINKED ISSUE #1988: Allow customizing menu items
Umbrella apps might want to customize the menu items and currently they are static.

On `LightningWeb.LayoutComponents/menu_items/1` this can be safely done preserving the default menu items whenever custom ones are not set.

jyeshe:
## Validation Steps

Setup an umbrella app and set the config:
```
config :lightning, :menu_items,
  <some_assign_set>: [
    {"/some_path", &<Heroicons.some_icon>/1, "Menu Name", <active_menu_item_atom>}
```

## Notes for the reviewer

This change preserves the default menu items when custom ones are not set.

## Related issue

Fixes #1988 

## Review checklist

### DIFF
--- a/lib/lightning_web/components/layout_components.ex
+++ b/lib/lightning_web/components/layout_components.ex
@@ -5,7 +5,39 @@ defmodule LightningWeb.LayoutComponents do
   import PetalComponents.Dropdown
   import PetalComponents.Avatar
 
+  @menu_items Application.compile_env(:lightning, :menu_items, [])
+
   def menu_items(assigns) do
+    custom_menu_items =
+      @menu_items
+      |> Enum.filter(fn {assign_set, _items} ->
+        not is_nil(assigns[assign_set])
+      end)
+      |> Enum.flat_map(fn {_assign_set, items} -> items end)
+
+    if Enum.empty?(custom_menu_items) do
+      default_menu_items(assigns)
+    else
+      assigns = assign(assigns, custom_menu_items: custom_menu_items)
+
+      ~H"""
+      <div class="mt-4">
+        <%= for {to, icon, text, menu_item} <- @custom_menu_items do %>
+          <Settings.menu_item to={to} active={@active_menu_item == menu_item}>
+            <%= Phoenix.LiveView.TagEngine.component(
+              icon,
+              [class: "h-5 w-5 inline-block"],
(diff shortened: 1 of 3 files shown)


lib/lightning_web/components/layout_components.ex
```diff
@@ -106,6 +118,16 @@ defmodule LightningWeb.LayoutComponents do
...
+      <%= for {to, icon, text, menu_item} <- @project_menu_items do %>
+        <Settings.menu_item to={to} active={@active_menu_item == menu_item}>
+          <%= Phoenix.LiveView.TagEngine.component(
+            icon,
+            [class: "h-5 w-5 inline-block mr-1"],
+            {__ENV__.module, __ENV__.function, __ENV__.file, __ENV__.line}
+          ) %>
+          <span class="inline-block align-middle"><%= text %></span>
+        </Settings.menu_item>
+      <% end %>
```
stuartc, line 130: Do we really need the icon to be dynamic like this? It feels like a slippery slope going this route.

We need to do some more abstraction here that simplifies this, @midigofrank any ideas around how we would improve the existing implementation so that we don't have to do this kind of dynamic stuff?

I'm thinking using slots would clean this up a lot, and decouple the component from using `Application.get_env` which feels like a performance liability here, ideally these extra menu items get passed in via attrs or slots...

This component was written a long long time ago, and could benefit from some `attr` specs with defaults - I see we call this whole component with `{assigns}` which is generally a no no.

stuartc (changes requested):
We're going to have to find a better way to do this, this is a little too 'quick and dirty' imo - and given how this component is rendered _everywhere_ we need to make sure it makes sense going forward and isn't going to cause extraneous re-renders.

lib/lightning_web/components/layout_components.ex
jyeshe: To be directly configurable it needs to be dynamic. We can try encapsulating with modules like an Extension but the code is very localized and explicit on what it does so I avoid something fancier.

jyeshe: Another way when we don't want to inject dependencies would be making everything composable and `LayoutComponents.menu_items` would call set A, B or C ...

stuartc (approved):
I'm not in love with this approach, but its fine for now until we get over the line for extensions.

PR #2000, opened 2024-04-19
Disable add collaborators button when limit is reached

midigofrank:
This PR  the `Add collaborators` button if the limit is reached

Depends on: https://github.com/OpenFn/lightning/pull/1984

## Review checklist

### DIFF
--- a/test/lightning_web/live/project_live_test.exs
+++ b/test/lightning_web/live/project_live_test.exs
@@ -2168,11 +2168,12 @@ defmodule LightningWeb.ProjectLiveTest do
         )
         |> render_submit()
 
-        updated_html = render(view)
-        # users are shown in the page
-        for new_user <- [admin, editor, viewer] do
-          assert updated_html =~ new_user.last_name
-        end
+        assert_patch(
+          view,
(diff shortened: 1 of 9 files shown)


jyeshe (commented):
@stuartc we've double checked yesterday that no database transaction needs to be performed when consuming the event. However like the other events, they might be lost if the GenServer restarts.

The effect would be to allow more collaborators than the would plan accepts.

jyeshe:
@midigofrank could you check the umbrella app tests?

test/lightning_web/live/project_live_test.exs
```diff
@@ -2395,6 +2392,8 @@ defmodule LightningWeb.ProjectLiveTest do
         refute html =~ "You are not authorized to perform this action"
         assert html =~ "Collaborator removed successfully!"
 
+        assert_patch(view, ~p"/projects/#{project}/settings")
```
jyeshe, line 2395: :pray:

jyeshe (approved):
:rocket:

stuartc (approved):

PR #2002, opened 2024-04-19
Don't include deleted workflows in the provisioning API

### LINKED ISSUE #2001: CLI: Filter deleted workflows when generating state.json file
When a user runs openfn pull on the CLI to generate state.json and spec.yaml files for a project, the state.json lists out all workflows (deleted and non-deleted) whilst spec.yaml only lists the jobs that are not deleted. This causes an error during github to openfn sync (See below) 

Image

We want state.json to only contain workflows that are not deleted, making it listing the same workflows in spec.yaml. 

PS: We should also ensure to write a test that locks in deleted/non-deleted workflows and making sure that only non-deleted workflows are generated.

# Repro steps

* Sign up to Lightning at (app.openfn.org)[https://app.openfn.org] (you can also use demo.opefn.org with credential `editor@openfn.org / welcome123` - but be aware that the demo server resets daily and work will be lost)
* Create a workflow
* Delete that workflow
* Locally, use the CLI and run `openfn pull <project-id>` to download your project. You can get the project id from the URL on the dashboard
* Running `openfn deploy` in the CLI should result in the above error

If you look at the downloaded project.yaml and state.json files, you should notice that the deleted workflow is present in state.json, but not in project.yaml.

# Background

Using Lightning, users can create multiple Workflows for a given project. These workflows are used for data integration and workflow automation.

You can use our [CLI tooling](https://docs.openfn.org/documentation/cli) to download a local copy of a project. This creates a project.yaml file on your system with all the workflows contained within. It also creates a state.json which tracks information unique to a deployment, like UUIDs. We use two files because a workflow can be deployed to multiple projects, and so some workflow information is objective (the name of a step), while some is specific to a project (like a step id).

# Implementation Notes

The code in `packages/deploy/src/stateTransform.ts` is responsible for generating the state.json file. I suspect you'll need to add logic in here somewhere to ignore deleted workflows.

Any fix should include a unit test, which fails on main, and proves that your solution works.

midigofrank:
## Validation Steps

1. Delete a workflow in lightning project
2. Perform a `cli pull` for the project. i.e `openfn pull <project_id`. The workflow deleted in step 1 above won't be included in your `state.json`
3. Perform a `cli deploy`. The operation should complete successfully. Initially, it was failing with an `Internal Server Error` as seen in sentry: https://openfn.sentry.io/issues/4884904031/events/244bfa2438984080b07040f442b98811/

## Notes for the reviewer
- Adds a filter to the `workflows assoc` in the project schema. 

## Related issue

Fixes #2001 

## Review checklist

### DIFF
--- a/lib/lightning/projects/project.ex
+++ b/lib/lightning/projects/project.ex
@@ -40,7 +40,7 @@ defmodule Lightning.Projects.Project do
     has_many :project_credentials, ProjectCredential
     has_many :credentials, through: [:project_credentials, :credential]
 
-    has_many :workflows, Workflow, where: [deleted_at: nil]
+    has_many :workflows, Workflow
     has_many :jobs, through: [:workflows, :jobs]
 
     timestamps()
--- a/lib/lightning/projects/project.ex
+++ b/lib/lightning/projects/project.ex
@@ -40,7 +40,7 @@ defmodule Lightning.Projects.Project do
     has_many :project_credentials, ProjectCredential
     has_many :credentials, through: [:project_credentials, :credential]
 
-    has_many :workflows, Workflow
+    has_many :workflows, Workflow, where: [deleted_at: nil]
     has_many :jobs, through: [:workflows, :jobs]
 
     timestamps()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to
 
 - Fix github cli deploy action failing to auto-commit
   [#1995](https://github.com/OpenFn/lightning/issues/1995)
+- Fix proviosning API includes deleted workflows in project state
+  [#2001](https://github.com/OpenFn/lightning/issues/2001)
 
 ## [v2.4.1-pre] 2024-04-18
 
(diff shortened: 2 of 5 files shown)


lib/lightning/projects/project.ex
```diff
@@ -40,7 +40,7 @@ defmodule Lightning.Projects.Project do
     has_many :project_credentials, ProjectCredential
     has_many :credentials, through: [:project_credentials, :credential]
 
-    has_many :workflows, Workflow
+    has_many :workflows, Workflow, where: [deleted_at: nil]
```
taylordowns2000, line 43: Whoa. I wasn't expecting this to be the solution. Isn't the trouble also that when individual _jobs_ inside a workflow have been deleted they show up inside the `state.json`?

taylordowns2000 (changes requested):
@midigofrank , my understanding of this issue was that the CLI was failing when we returned workflows that had individual `jobs` that were deleted.

can you check with @christad92 ? maybe we need to filter out both? (the test is that the `project.yaml` and the `state.json` should be COMPATIBLE for projects where there are workflows that have been deleted or where there are jobs within existing workflows that have been deleted.

taylordowns2000 (approved):

PR #2030, opened 2024-04-25
Run deploy github workflows one at a time

### LINKED ISSUE #2044: Conflict occurs during github sync when 2  workflows are triggered by the same commit
## Description
When a github project has more than 1 workflow in the same branch, a commit into that branch will trigger the workflows to run `cli deploy`

The `deploy action` commits changes from lightning into the repository on completion. 

Given that these workflows were running concurrently, only one succeeds. The rest fail to commit because they get out of date with the git history.
 
## Acceptance criteria
Given 2 workflows in a repo, only one should run at a time

taylordowns2000:
## Description
this is just a concept:
1. needs testing
2. _maybe_ (probably) also needs the workflow to pull/merge previously made changes... see @mtuchi 's latest error when he reruns the workflow.

- we want workflow a to run
- then we want workflow b to run
- they are both triggered by the same commit, but they do different things because we have multiple unrelated openfn projects in the same repo.
- how do we stop the conflicts?

## Related Issue
Fixes #2044

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to
 
 ### Fixed
 
+- Don't let two deploy workflows run at the same time to prevent git collisions
+  [#2044](https://github.com/OpenFn/lightning/issues/2044)
 - Stopped sending emails when creating a starter project
   [#2161](https://github.com/OpenFn/lightning/issues/2161)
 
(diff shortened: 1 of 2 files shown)


taylordowns2000:
you can test this by modifying the workflow in place @mtuchi . the "fix" is just for automatically generated workflows from lightning.

midigofrank:
@mtuchi I have tested that this actually runs the workflows one at a time. When you get some time, do modify the github workflows in the project with this option and see the how it works

mtuchi (approved):

Issue #2102, opened 2024-05-17
Ironing out Generic Oauth before full release


taylordowns2000:
- [ ] ~#2100~ (edit, moved to new issue)
- [x] put the "create credential" option ON TOP of the "create client" option in the dropdown. (it is what 99% of people will do.)
- [x] mark the "create client" option as `(advanced)` in the dropdown
- [ ] #2114
- [x] #2104
- [x] remove the "not working?" prompt. it's really confusing. i'd recommend only showing it once you come _back_ to a previously saved credential, if that's possible, and then say, "All good, but if your credential stops working, you may need to re-authorize **here**."

- [x] #2105
- [x] Refresh token before sending to worker
- [ ] #2113
- [ ] ~Adaptor /Step  configuration form dies after 10 seconds and the page refreshes. you lose your work (likely NOT introduced here, already on main? if so, please track this in a NEW issue @christad92 )~ edit, tracked on: #1882 and #2007

<img width="747" alt="image" src="https://github.com/OpenFn/lightning/assets/8732845/272e15c4-a57d-4baa-b8f4-16f264d16564">

christad92:
@taylordowns2000 I am not seeing the duplication of clients on my end and I've been able create credentials on my end. 

On docs, we will have a draft ready for clients and creating credentials from clients. 

Not clear about others.

taylordowns2000:
all tracked individually, closing

PR #2111, opened 2024-05-17
🔐 Restrict OAuth Client Creation to Superusers with Feature Flag

### LINKED ISSUE #2104: feature flag OUT the oauth client creation. for now, only allow `superuser` to create clients.

elias-ba:
## Related issue

Re #2102 
Fixes #2104

## Review checklist

### DIFF
--- a/lib/lightning_web/live/components/credentials.ex
+++ b/lib/lightning_web/live/components/credentials.ex
@@ -348,7 +348,7 @@ defmodule LightningWeb.Components.Credentials do
               disabled={@disabled}
             >
               <%= name %><span
-                :if={Map.get(option, :badge)}
+                :if={Map.get(option |> IO.inspect(), :badge)}
                 class="ml-2 inline-flex items-center rounded-md bg-gray-50 px-1.5 py-0.5 text-xs font-medium text-gray-600 ring-1 ring-inset ring-gray-500/10"
               ><%= Map.get(option, :badge) %></span>
             </a>
--- a/lib/lightning_web/live/components/credentials.ex
+++ b/lib/lightning_web/live/components/credentials.ex
@@ -309,6 +309,7 @@ defmodule LightningWeb.Components.Credentials do
   attr :id, :string, required: true
   attr :options, :list, required: true
   attr :disabled, :boolean, default: false
+  attr :badge, :any, default: false
   slot :inner_block, required: true
 
   def options_menu_button(assigns) do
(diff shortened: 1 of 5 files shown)


taylordowns2000 (changes requested):
1. looks like oauth client creation is still possible from the project credentials interface, though the button has been altered (image 1)

image

lib/lightning_web/live/components/credentials.ex
```diff
@@ -323,18 +324,7 @@ defmodule LightningWeb.Components.Credentials do
...
-          fill="currentColor"
-          aria-hidden="true"
-        >
-          <path
-            fill-rule="evenodd"
-            d="M5.23 7.21a.75.75 0 011.06.02L10 11.168l3.71-3.938a.75.75 0 111.08 1.04l-4.25 4.5a.75.75 0 01-1.08 0l-4.25-4.5a.75.75 0 01.02-1.06z"
-            clip-rule="evenodd"
-          />
-        </svg>
+        <Heroicons.chevron_down class="ml-1 h-5 w-5" />
```
stuartc, line 327: Nice, thank you for removing inline svgs 🙌

lib/lightning_web/live/components/credentials.ex
```diff
@@ -357,7 +347,10 @@ defmodule LightningWeb.Components.Credentials do
               phx-click={show_modal(target)}
               disabled={@disabled}
             >
-              <%= name %>
+              <%= name %><span
+                :if={Map.get(option, :badge)}
+                class="ml-2 inline-flex items-center rounded-md bg-gray-50 px-1.5 py-0.5 text-xs font-medium text-gray-600 ring-1 ring-inset ring-gray-500/10"
+              ><%= Map.get(option, :badge) %></span>
             </a>
```
stuartc, line 354: I think this interface is kinda confused, this is best encapsulated using slots.

The `options` assign for creating n number of a tags is exactly where slots are best. 

Also I see the button is only rendered in one place, with one set of options. I don't see the need for it to be so generic vs a composed function component that reuses other function components (if any).

stuartc (changes requested):
I see there are bunch of tests failing, but also the `options_menu_button` component is too complex given its use case here. I think make it a single component that takes no arguments or if there is some kind of other use case I'm not aware of, let me know and we can chat about how to leverage slots etc.

PR #2117, opened 2024-05-20
Fix Duplicated Credential Types

### LINKED ISSUE #2100: Generic Oauth clients can appear multiple times in the credential type selector
Here, there should only be one "Salesforce Sandbox" credential type. Occurs on `v2.5.0-pre`

There are two ways of recreating this bug:
1. Adding a new credential through the canvas
- Attempt to create a new credential from the workflow canvas by clicking "New Credential"
- Create a new credential from one of your Oauth clients
- Now try to add a new credential by clicking "new credential" 

2. This can also be recreated by switching selecting a new credential type in the list of available credentials on the form.

image

And sometimes it really goes wild 😎 

image

elias-ba:
## Related issue

Fixes #2100 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/credential_live/credential_form_component.ex
+++ b/lib/lightning_web/live/credential_live/credential_form_component.ex
@@ -413,6 +413,7 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
           projects={@projects}
           users={@users}
           on_save={@on_save}
+          allow_credential_transfer={@allow_credential_transfer}
           return_to={@return_to}
         />
       </.modal>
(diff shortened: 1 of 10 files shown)


lib/lightning_web/live/credential_live/credential_form_component.ex
```diff
@@ -529,10 +538,7 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
                   </div>
                 </fieldset>
               </div>
-              <div
-                :if={@action == :edit and @allow_credential_transfer}
-                class="space-y-4"
-              >
+              <div :if={@action == :edit and true} class="space-y-4">
```
taylordowns2000, line 541: elias to change back to `@allow_credential_transfer`

taylordowns2000 (changes requested):
one small change on allow/true

lib/lightning_web/live/credential_live/credential_form_component.ex
elias-ba: Thanks this is done.

PR #2118, opened 2024-05-20
Improve performance and memory consumption mailer digest

jyeshe:
## Validation Steps

Test cases on `test/lightning/digest_email_worker_test.exs`.

## Notes for the reviewer

Decrease processing and memory usage on email digest.

Other features will be investigated for the interval between 1:00PM of weekly digest and 1:10PM of a OOM crash.

## Related issue

Refs #2121

## Review checklist

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -352,7 +352,38 @@ defmodule Lightning.Invocation do
     |> Repo.paginate(params)
   end
 
+  def count_workorders(%Project{} = project, search_params) do
+    project
+    |> count_workorders_query(search_params)
+    |> Repo.aggregate(:count)
+  end
+
   def search_workorders_query(
+        %Project{id: project_id},
+        %SearchParams{status: status_list} = search_params
+      ) do
+    status_filter =
+      if SearchParams.all_statuses_set?(search_params) do
+        []
+      else
+        status_list
+      end
+
+    base_query_with_preload(project_id)
+    |> filter_by_workorder_id(search_params.workorder_id)
+    |> filter_by_workflow_id(search_params.workflow_id)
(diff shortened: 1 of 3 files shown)


stuartc (approved):
Nicely done! Could you add a line to the changelog linking to the PR and say you've fixed a potential memory usage issue.

lib/lightning/invocation.ex
```diff
@@ -347,54 +347,42 @@ defmodule Lightning.Invocation do
...
   end
 
-  def count_workorders(%Project{} = project, search_params) do
-    project
-    |> count_workorders_query(search_params)
-    |> Repo.aggregate(:count)
-  end
-
-  def search_workorders_query(
+  def search_workorders_without_wiped_dataclips(
```
stuartc, line 356: Hmm, I'm not 100% getting why this gets a function to its own here.
I'm usually adverse to making function specifically for different cases like this, it can lead to divergence between the two differ variants.

I wasn't involved a whole bunch with the creation of this query stuff, and I don't actually have a strong enough stake in it (nor do I have a solid idea on what the alternative would be).

But the change in `run_live` [here](https://github.com/OpenFn/lightning/pull/2118/commits/5858e459b84d19b2639e96aa13471b124be99a98#diff-505fd6e66663185ea709cefb203e65dca8da3b7d0e1613b1dc1a778b82ca87ddL492) make me pause.

Happy to go with what Taylor thinks on this.

jyeshe: Taking to another PR once there is a fix needed on existing query for workorders without wiped dataclips.

taylordowns2000 (approved):
great. thank you! (And thanks for moving the refactor into a separate PR.)

PR #2127, opened 2024-05-21
use Monaco for the log viewer

### LINKED ISSUE #1863: Implement logs with Monaco editor
Implementing logs with Monaco allows us to better manage performance issues with large logs and line numbers when referencing logs. 

For line numbers, users often rely on logs for debugging errors in their jobs but this becomes harder during collaboration if there is no way to reference sections of the logs. By adding line numbers, will make it easier for users to trace and reference log outputs.

Image

midigofrank:
## Notes for the reviewer

- [x] Split the logs by `\n` so as to get the correct line numbers when highlighting step logs
- [x] Move the `source` from the line numbers to the content
- [x] Pass the store as a prop to avoid duplicating the lines
- [x] Style the highlighter to match the previous one
- [x] Make the main viewer to fill the rest of remaining space in the screen. Or just make it bigger 🤷‍♂️ . This is quite tricky because we also don't want to affect the dataclip viewers. `absolute` positioning seems like the way o go, but what would be the height? 🤷 

## Related issue

Fixes #1863

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to
 
 ### Added
 
+- Replace LiveView Log Viewer component with React Monaco
+  [#1863](https://github.com/OpenFn/lightning/issues/1863)
+
 ### Changed
 
 ### Fixed
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -218,40 +218,6 @@ div.line-num::before {
   max-width: min-content;
 }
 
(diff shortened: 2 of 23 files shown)


josephjclark (commented):
Hey @midigofrank, I was just looking at how you're using the monaco hooks and it looks like you're making a bit of a meal of nulls? I think you should be able to simplify some of that stuff - see comments.

If it doesn't make sense or doesn't work just give me a shout and I'll get on a call with you

stuartc (approved):

PR #2143, opened 2024-05-24
Stop users from creating credentials with deprecated types

### LINKED ISSUE #2142: Prevent users from creating new `googlesheets` and `salesforce_oauth` credentials
- This is the first logical step to deprecating these old OAuth2 credential types.
- Once we're done here, we can start migration credentials from the old to the new.
- Then when that's done we can remove all the old code.

taylordowns2000:
## Validation Steps

1. Create a `salesforce_oauth` or `googlesheets` credential on `main`.
2. Move to this branch.
3. Check that you can still see, use, edit your previous credentials.
4. Check that you **cannot** create a new `salesforce_oauth` or `googlesheets` credential.

## Notes for the reviewer

This is the first step to getting rid of the old credential types. Please note the TODO, but know that we're fine without a perfect solution right now.

Fixes #2142 

## Review checklist

### DIFF
--- a/lib/lightning_web/live/credential_live/credential_form_component.ex
+++ b/lib/lightning_web/live/credential_live/credential_form_component.ex
@@ -595,42 +595,12 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
         {name |> Phoenix.HTML.Form.humanize(), name, nil, nil}
       end)
 
-    oauth_clients_from_env =
-      Application.get_env(:lightning, :oauth_clients)
-
     schemas_options
     |> Enum.concat([{"Raw JSON", "raw", nil, nil}])
-    |> handle_oauth_item(
-      {"GoogleSheets", "googlesheets", ~p"/images/oauth-2.png", nil},
-      get_in(oauth_clients_from_env, [:google, :client_id])
-    )
-    |> handle_oauth_item(
-      {
-        "Salesforce",
-        "salesforce_oauth",
-        ~p"/images/oauth-2.png",
-        nil
-      },
-      get_in(oauth_clients_from_env, [:salesforce, :client_id])
-    )
+    # TODO: Gracefully determine which adaptors define custom credential types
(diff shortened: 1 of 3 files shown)


lib/lightning_web/live/credential_live/credential_form_component.ex
```diff
@@ -595,44 +595,14 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
...
-        "Salesforce",
-        "salesforce_oauth",
-        ~p"/images/oauth-2.png",
-        nil
-      },
-      get_in(oauth_clients_from_env, [:salesforce, :client_id])
-    )
+    # TODO: Gracefully determine which adaptors define custom credential types
+    # and which adaptors use OAuth2 credential types.
+    |> List.delete({"Googlesheets", "googlesheets", nil, nil})
```
elias-ba, line 602: @taylordowns2000 what is this for ?

elias-ba (changes requested):

lib/lightning_web/live/credential_live/credential_form_component.ex
taylordowns2000: It removes the Googlesheets option from the list of available credential types.

elias-ba: Oh Waouh ! then shouldn't this be handled in the adaptor side ? Like from the configuration schema ?

taylordowns2000: No. Do you remember how `append_if_missing` worked? It was for this reason. I'll call you quick to explain

elias-ba: Thanks @taylordowns2000 for the call. All clear for me now.

elias-ba (approved):

christad92 (approved):

Issue #2008, opened 2024-04-19
Jump to steps on a different branches (i.e., merge)


christad92:
When building a workflow, user want to be able to reuse a step that has been created in a different branch. For example when a workflow has a series of steps to manage an error can be designed so that it can be reused instead of being duplicated or when a user wants to skip some steps in a workflow.

 - [x] Implement the UI for connecting two steps by dragging outputs to inputs with validations (see rules)
 - [x] Allow downstream jobs to handle throughput of input state from other steps

Rules: 
- Disable cyclical flows/branches i.e. A branch should not go upstream on its path thereby causing loops

Designs: https://www.figma.com/file/4hGWdO4sajnjcOwE81j8BK/App-(new)?type=design&node-id=4253-739&mode=design&t=dtwY2FmzuHXrL4hY-0

How users currently manage this: 
- duplicating the step
- creating a step that calls a different workflow via webhook (see below)

![Image](https://github.com/OpenFn/lightning/assets/5230530/24c25db7-dc24-43f8-825e-86c5b1ca19f2)

https://www.loom.com/share/45396a36dc4345c9bce3b41b896eaffa

stuartc:
https://reactflow.dev/examples/nodes/add-node-on-edge-drop

stuartc:
![Image](https://github.com/OpenFn/lightning/assets/144796/676a690c-d411-4a31-9b76-a23e0d291cef)

josephjclark:
implementation notes:

* don't autolayout on edge creation
* add a toolbar on the chart with zoom and layout controls
* cross fingers on dagre

Issue #2006, opened 2024-04-19
Add zoom control panel to workflow canvas.


christad92:
EPIC: #2021 
We need to implement a zoom control panel that would allow users to lock, zoom out and zoom in when using the canvas. With this we should set the minimum zoom level to 6.

React flow offers a the useStore() that allow us to manipulate zoom controls via UI. 

![Image](https://github.com/OpenFn/lightning/assets/5230530/d95b0185-2191-4c77-980b-6b5202d61100)

**Might be useful:** 
- https://reactflow.dev/examples/interaction/contextual-zoom
- https://reactflow.dev/api-reference/hooks/use-store

christad92:
Fixed in #2160

PR #2160, opened 2024-05-30
Workflow Diagram: "edge merging" (new edge creation)

josephjclark:
Allow new edges to be created in the workflow diagram. Also enables the removal of edges (removing an edge should not remove the steps at either end)

Edges can only be created between two steps:
- If they would not create a circular workflow
- If  no edge exists

# Changes in this PR

* Use dagre layout instead of hierachy
* Add a new hover button for creating edges
* Create new edges on drop
* Visual indicator for valid drop targets
* Don't allow edges that would create loops, or duplicate edges, or connection to triggers
* Show an error message if an edge cannot be created

## Issue

#2008 #2006

# TODO

* [x] Turns out edges  are still created if you drop on an invalid node!
* [x] Allow edges to be removed
* [x] Get save working properly
* [ ] Don't auto layout
* [x] edges don't connect to the node now
* [x] tweak layout because it's a bit spacey
* [ ] Maybe we should style the source edge and links while dragging
* [ ] it would be neat for the edge to not be drawn over any nodes (the source or target), ie, the line goes behind or is clipped by the source and target handles
* [x] Better feedback when hovering the connector over an invalid node (a message would be ideal)
* [x] Use curved edges generally
* [x] Layout doesn't behave nicely on adding of new nodes (improved but a dagre bug is problematic)
* [x] Add control bar (zoom fit,  disable autofit, zoom in/out, maybe layout)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+assets/vendor
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -213,7 +213,7 @@ export default React.forwardRef<HTMLElement, WorkflowDiagramProps>(
           fitViewOptions={{ padding: FIT_PADDING }}
           {...connectHandlers}
         >
-          <Controls showInteractive={false} position="top-left">
+          <Controls showInteractive={false} position="bottom-left">
             <ControlButton
               onClick={() => {
                 setAutofit(!autofit);
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -185,7 +185,7 @@ export default React.forwardRef<HTMLElement, WorkflowDiagramProps>(
       }
     }, [flow, ref]);
 
-    const connectHandlers = useConnect(model, setModel);
+    const connectHandlers = useConnect(model, setModel, store);
 
     return (
       <ReactFlowProvider>
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -1,5 +1,7 @@
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import ReactFlow, {
+  Controls,
+  ControlButton,
   NodeChange,
   ReactFlowInstance,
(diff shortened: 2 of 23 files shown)


josephjclark:
Removal of edges is going to have to happen from the lightning side I think - just like how we remove steps. I'll need help with that.

josephjclark:
Encountered a big problem with the layout.

We are using the dagre algorithm to run layouts now. This can handle simple graphs in a way that d3's tree layout cannot (to be fair I haven't actually tested this..)

The problem with dagre is that when you add nodes, it has  tendency to switch nodes orders around more or less at random. Looks like d3 is doing this in production too - maybe not quite as badly as dagre on this branch.

Now we can fix this by setting `disableOptimalOrderHeuristic` on the dagre layout, like: `Dagre.layout(g, { disableOptimalOrderHeuristic: true });`. The layout is much, much more stable like this.

This option was exposed in fairly recent PR: https://github.com/dagrejs/dagre/pull/263. I've reproduced this locally by hacking the code, so I know it works.

But here's the fun bit!  Due to what looks like a bug in the release, the options object doesn't actually get passed through to the layout algorithm. It basically looks like half the fix PR got merge conflicted out, or something. It's totally borked.

Dagre is basically in maintenance mode and only gets very sporadic updates. I don't expect much of a response if we turn to the community.

So here are our options:
* We fork dagre into the openfn org and make the changes we need. This is likely to be fine and won't take long.
* We use someone else'se fork - there is at least one on npm. We have to vet it to be sure what's in it.
* We copy/paste a build of dagre into the lightning repo and support it like a lib file. We could copy source or a dist file (it'll get built into our eventual production distribution anyway).
* We live with the annoying ordering thing
* We look for an alternative to dagre

What do you reckon @stuartc ? I say FORK IT

josephjclark:
An observation: the highlight rings clip over the edge of the node label, which is untidy. Also I think they're too heavy

josephjclark:
Just for the record, if we use d3 hierarchy, it seems to only use the first edge to work out the hierarchy. So it basically won't layout at all, it assumes a single parent

josephjclark:
show the error message under the adaptor name

josephjclark:
So we just need to support edge removal and this is good to go.

I am going to take a look at a better autofit in another PR tomorrow - I have a feeling I can turn it around quickly. But it should not block this release.

Suggest we stick with autolayout for now

stuartc (approved):

PR #2162, opened 2024-05-30
Add CORS support

### LINKED ISSUE #2157: Configure CORS to allow requests from different origins
### Details
Given that the app runs on a subdomain while the marketing website runs on the root domain, doing registrations from the marketing website won't be possible because they're essentially different origins.

This issue makes sure that we can configure lightning to accept requests from `*.openfn.org` or `www.openfn.org`

stuartc:
## Validation Steps

Run the app with the env `CORS_ORIGIN=https://foo.com`.

Then do a preflight request:

```
curl -vvvv 'http://localhost:4000/api/users/register' \
  -X 'OPTIONS' \
  -H 'Accept: */*' \
  -H 'Access-Control-Request-Headers: content-type' \
  -H 'Access-Control-Request-Method: POST' \
  -H 'Origin: https://foo.com' \
  -H 'Sec-Fetch-Mode: cors'
```

Look for a `Access-Control-Allow-Origin` header in the response which has the domain specified in `Origin` on the request.

Repeat with a different domain:

```
curl -vvvv 'http://localhost:4000/api/users/register' \
  -X 'OPTIONS' \
  -H 'Accept: */*' \
  -H 'Access-Control-Request-Headers: content-type' \
  -H 'Access-Control-Request-Method: POST' \
  -H 'Origin: https://anythingelse.net' \
  -H 'Sec-Fetch-Mode: cors'
```

Verify that a `...Allow-Origin` header is not in the response.

## Notes for the reviewer

The default is set it `*`.
The environment variable can be set to more than one domain using commas.

## Related issue

Fixes #2157 

## Review checklist


midigofrank:
Hey @stuartc , don't you think we should use this library instead? It seems to have recent activity compared to CORSPlug

https://github.com/whatyouhide/corsica/

config/runtime.exs
```diff
@@ -344,11 +344,14 @@ config :lightning, Lightning.Repo,
...
   queue_interval: env!("DATABASE_QUEUE_INTERVAL", :integer, 1000)
 
-url_scheme = env!("URL_SCHEME", :string, "https")
 host = env!("URL_HOST", :string, "example.com")
 port = env!("PORT", :integer, 4000)
 url_port = env!("URL_PORT", :integer, 443)
 
+config :lightning,
+  cors_origin:
+    env!("CORS_ORIGIN", :string, "*") |> String.split(",") |> List.wrap()
```
midigofrank, line 353: Should we maybe we should default to the `endpoint url` instead?

stuartc: Oh sorry, to add on why `cors_plug` vs `corsica`, we wanted to be able to configure this from `runtime.exs` which I couldn't find a way to do that easily with corsica, I now see in the docs you can give the `origin` key an MFA [here](https://hexdocs.pm/corsica/Corsica.html#t:origin/0) so we could use it... 

If `cors_plug` gives us even a hint of an issue, happy to try our corsica.

midigofrank: Aaah, makes sense

PR #2171, opened 2024-06-04
Hide deprecated adaptors from version picklist

### LINKED ISSUE #2169: Filtering deprecated adaptors from dropdown list
**Description:**

The current dropdown list for selecting adaptors includes both active and deprecated options. This can be confusing for users and lead to selecting unsupported adaptors. 

**Proposed Solution:**

Implement a filtering mechanism to exclude deprecated adaptors from the dropdown list. 

**Additional Considerations:**

* If there's a need to retain visibility of deprecated adaptors for informative purposes, consider a separate option to "Show Deprecated Adaptors". This would allow users to view all options but clearly differentiate between active and deprecated choices.
* Provide a clear indication to the user if a selected adaptor is deprecated. This can be done through a visual cue (e.g., dimmed text) or a tooltip message.

**Benefits:**

* Improved user experience by presenting a clear and concise list of supported adaptors.
* Reduced risk of users selecting deprecated and potentially unsupported adaptors.

taylordowns2000:
Fixes #2169 

## Validation Steps

- notice `language-salesforce@4.4.0` and `@5.0.1` in your adpator list on `main`.
- clear your local adaptor cache: `rm -rf /tmp/openfn/worker/repo`.
- switch to this branhc
- boot up lightning
- notice that `4.4.0` and `5.0.1` no longer appear

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to
 
 ### Changed
 
+- Don't show deprecated adaptor versions in the adaptor version picklist (to be
+  followed by some graceful deprecation handling/warning in
+  [later work](https://github.com/OpenFn/lightning/issues/2172))
+  [#2169](https://github.com/OpenFn/lightning/issues/2169)
+
 ### Fixed
 
 ## [v2.5.4] - 2024-05-31
--- a/lib/lightning/adaptor_registry.ex
+++ b/lib/lightning/adaptor_registry.ex
@@ -257,7 +257,10 @@ defmodule Lightning.AdaptorRegistry do
       repo: details["repository"]["url"],
       latest: details["dist-tags"]["latest"],
       versions:
-        Enum.map(details["versions"], fn {version, _detail} ->
+        Enum.reject(details["versions"], fn {_version, detail} ->
+          detail["deprecated"]
+        end)
+        |> Enum.map(fn {version, _detail} ->
           %{version: version}
         end)
     }


mtuchi (commented):
@taylordowns2000 Tested locally everything seems fine ✅ , 
But i think it would be nice if we can have unit test in `test/lightning/adaptor_registry_test.exs`.  Do you think we should have a test like this 👇🏽 ? 
```
test"filter deprecated adaptors from the adaptor lists" do ...
  //assert deprecated version of an adaptor will be filtered 

end
```

taylordowns2000:
> @taylordowns2000 Tested locally everything seems fine ✅ , But i think it would be nice if we can have unit test in `test/lightning/adaptor_registry_test.exs`. Do you think we should have a test like this 👇🏽 ?
> 
> ```
> test"filter deprecated adaptors from the adaptor lists" do ...
>   //assert deprecated version of an adaptor will be filtered 
> 
> end
> ```

i took a look, but as far as i can tell we don't have any tests for the response processing from NPM. an additional set of tests (including the one you suggest) is a _great_ idea and we should do it in https://github.com/OpenFn/lightning/issues/2172

stuartc:
> > @taylordowns2000 Tested locally everything seems fine ✅ , But i think it would be nice if we can have unit test in `test/lightning/adaptor_registry_test.exs`. Do you think we should have a test like this 👇🏽 ?
> > ```
> > test"filter deprecated adaptors from the adaptor lists" do ...
> >   //assert deprecated version of an adaptor will be filtered 
> > 
> > end
> > ```
> 
> i took a look, but as far as i can tell we don't have any tests for the response processing from NPM. an additional set of tests (including the one you suggest) is a _great_ idea and we should do it in #2172

We have tests that the registry requests all modules, and then a small test that language-common is queried, for which we provide a stubbed response in a fixture... So we have the bits to test this - but it's better tested more thoroughly; happy to move the testing out to a bit later. But @mtuchi is right, we definitely must test this.

taylordowns2000:
All makes sense. I've linked these comments to the new issue. Thank you both!!!!

Issue #2027, opened 2024-04-24
Allow users to delete steps with associated runs


taylordowns2000:
This was enabled by the work on snapshots (merged April 24th) and we should now let users delete steps in workflows whether or not they have associated activity.

![Image](https://github.com/OpenFn/lightning/assets/8732845/db843ed4-4d13-4610-8503-e8035f506735)

Issue #1825, opened 2024-02-28
Allow UI to reference a Workflow Snapshot instead of actual model when viewing runs


stuartc:
When viewing a run that was executed against a snapshot, the UI should show _that_ workflow and it's steps with the Job/Credential/Trigger/Edge that was used at the time of execution.

In addition, the WorkerChannel must use a WorkflowSnapshot as it's source of truth.

A proposed design would be to have the `WorkflowSnapshot` or similar module, act as an interface to retrieve information about it.

Every run is associated with a snapshot, so when we are viewing a run in a run view and on the inspector, we should see the snapshot ID. 

Clicking run or retry for a run on the history page should do nothing different from the current behaviour -- take the latetst snapshot and run it against the associated input dataclip.

When a user is about to retry a run from the history page, we should show a tool tip saying `This run will be executed with the latest snapshot of this workflow`.

Designs here: https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App-(new)?node-id=4332-1242&t=DSaaMAGAa6cV5deY-0

This is a precursor to #1843

PR #2176, opened 2024-06-04
Workflow Snapshots UI

### LINKED ISSUE #1843: Implement viewonly interface for viewing snapshots for previous runs
When inspecting runs, we want to be able to preview the associated canvas and editor interface in a read only mode. 

On the Canvas, the user should see an alert that let's them know that they are currently viewing the workflow in a view only mode and they can switch to the latest version of the workflow if they want to make any changes. (See design)
The editor is also presented in a view only mode such that the user is able see the job code, input data and output data but can switch to the latest snapshot to enable them to run/modify job code.

https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App-(new)?node-id=4332-1242&t=DSaaMAGAa6cV5deY-0

elias-ba:
## Description

This PR introduces support for Workflow Snapshots in the UI, enhancing some sort of version control and historical accuracy within the workflow canvas. Two main areas are updated: the Workflow Canvas and the History Page.

#### In the Workflow Canvas

A kind of versioning system is implemented, tracked through the `workflow lock_version` property. The version is passed via the URL parameter `v` to load specific snapshots.

Snapshots are rendered in the canvas according to the version specified in the URL. If a snapshot is not the latest version, the UI switches to read-only mode. A toggle allows switching to the latest version for editing. Only the latest version can be edited; previous versions are view-only.

#### In the History Page

Runs now display using snapshots. If a run was executed with a non-latest version, rerunning it will use the latest workflow version. A tooltip on the rerun button informs users about this behavior.

When inspecting a run, it loads in the Workflow Canvas with its specific version. Users can toggle to the latest version using the same switch as in the Workflow Canvas.

## Validation Steps

1. **Create a New Workflow**: Start by creating a new workflow in the app. Verify that the workflow is tagged as the latest version.
2. **Edit Your Workflow:** Make various edits to your workflow such as adding nodes, edges, and editing the job. Confirm that the workflow remains tagged as the latest version.
3. **Run Your Workflow**: Execute your workflow by running it. Ensure the run completes successfully.
4. **Make Additional Changes**: Continue editing the workflow by making further modifications. The workflow should still be tagged as the latest version.
5. **Run Your Workflow Again**: Execute the workflow once more. Confirm the second run completes successfully.
6. **Navigate to the History Page**: Go to the history page to view your workflow runs. Ensure all runs are displayed correctly.
7. **Inspect the First Run**: Open the first run and inspect it in detail. Verify it is tagged as non-latest with a shortened UUID of the snapshot ID.
8. **Read-Only Mode Verification**: Check that the UI is in read-only mode for the inspected run.
9. **Toggle Button for Workflow Snapshot Version**: Locate the workflow snapshot version switcher toggle button. Ensure the toggle button is present and functional.
10. **Switch to the Latest Version**: Use the toggle button to switch to the latest version of the workflow. Observe the content changes and confirm the UI transitions to edit mode.
11. **Perform Additional Edits**: Make further edits to the workflow. Execute the workflow again to ensure the changes are properly integrated.

## Related issue

Fixes #1843 #1825 #2027 

## Review checklist

### DIFF
--- a/lib/lightning/workflows/snapshot.ex
+++ b/lib/lightning/workflows/snapshot.ex
@@ -13,6 +13,7 @@ defmodule Lightning.Workflows.Snapshot do
 
   alias Lightning.Projects.ProjectCredential
   alias Lightning.Repo
+  alias Lightning.Workflows.WebhookAuthMethod
   alias Lightning.Workflows.Workflow
 
   @type t :: %__MODULE__{
@@ -54,7 +55,7 @@ defmodule Lightning.Workflows.Snapshot do
       field :type, Ecto.Enum, values: [:webhook, :cron]
       field :has_auth_method, :boolean, virtual: true
 
-      many_to_many :webhook_auth_methods, Lightning.Workflows.WebhookAuthMethod,
+      many_to_many :webhook_auth_methods, WebhookAuthMethod,
         join_through: "trigger_webhook_auth_methods",
         on_replace: :delete
 
--- a/lib/lightning_web/live/run_live/show.ex
+++ b/lib/lightning_web/live/run_live/show.ex
@@ -13,11 +13,40 @@ defmodule LightningWeb.RunLive.Show do
 
   on_mount {LightningWeb.Hooks, :project_scope}
 
+  attr :run, :map, required: true
+  attr :workflow, :map, required: true
+
+  defp snapshot_version(assigns) do
+    %{run: run, workflow: workflow} = assigns
+
+    snapshot_version =
+      if run.snapshot.lock_version == workflow.lock_version do
+        "latest"
+      else
+        String.slice(run.snapshot.id, 0..6)
+      end
+
+    assigns =
+      assign(assigns, snapshot_version: snapshot_version)
+
+    ~H"""
+    <LightningWeb.Components.Common.snapshot_version_chip
+      id="run-workflow-version"
+      version={@snapshot_version}
+      tooltip={
+        if @snapshot_version == "latest",
+          do: "This run is based on the latest version of this workflow.",
+          else:
+            "This run is based on a snapshot of this workflow that was taken on #{Lightning.Helpers.format_date(run.snapshot.inserted_at)}"
+      }
+    />
+    """
+  end
+
   @impl true
   def render(assigns) do
-    no_step_selected? = is_nil(assigns.selected_step_id)
-
-    assigns = assign(assigns, :no_step_selected?, no_step_selected?)
+    assigns =
+      assigns |> assign(:no_step_selected?, is_nil(assigns.selected_step_id))
 
     ~H"""
     <LayoutComponents.page_content>
@@ -30,22 +59,9 @@ defmodule LightningWeb.RunLive.Show do
             </span>
             <div class="mx-2"></div>
             <.async_result :let={run} assign={@run}>
-              <% snapshot_version =
-                if run.snapshot.lock_version ==
-                     @workflow.lock_version,
-                   do: "latest",
-                   else: String.slice(run.snapshot.id, 0..6)
-
-              tooltip =
-                if snapshot_version == "latest",
-                  do: "This run is based on the latest version of this workflow.",
-                  else:
-                    "This run is based on a snapshot of this workflow that was taken on #{Lightning.Helpers.format_date(run.snapshot.inserted_at)}" %>
-              <LightningWeb.Components.Common.snapshot_version_chip
-                id="run-workflow-version"
-                version={snapshot_version}
-                tooltip={tooltip}
-              />
+              <%= if run do %>
+                <.snapshot_version run={run} workflow={@workflow} />
+              <% end %>
             </.async_result>
           </:title>
         </LayoutComponents.header>
--- a/lib/lightning/runs.ex
+++ b/lib/lightning/runs.ex
@@ -241,6 +241,8 @@ defmodule Lightning.Runs do
     |> Repo.transaction()
     |> tap(fn result ->
       with {:ok, %{runs: {_n, runs}}} <- result do
+        # TODO: remove the requirement for events to be hydrated with a specific
+        # set of preloads.
         runs
         |> Enum.map(fn run -> Repo.preload(run, :snapshot) end)
         |> Enum.each(&Events.run_updated/1)
--- a/test/lightning_web/channels/run_channel_test.exs
+++ b/test/lightning_web/channels/run_channel_test.exs
@@ -1,11 +1,11 @@
 defmodule LightningWeb.RunChannelTest do
-  alias Lightning.Workflows
   use LightningWeb.ChannelCase, async: true
 
   alias Lightning.Extensions.UsageLimiting.Context
   alias Lightning.Invocation.Dataclip
   alias Lightning.Invocation.Step
   alias Lightning.Workers
+  alias Lightning.Workflows
 
   import Ecto.Query
   import Lightning.TestUtils
--- a/test/lightning_web/live/run_live/components_test.exs
+++ b/test/lightning_web/live/run_live/components_test.exs
@@ -1,9 +1,9 @@
 defmodule LightningWeb.RunLive.ComponentsTest do
-  alias Lightning.Workflows
   use LightningWeb.ConnCase, async: true
 
   import Phoenix.LiveViewTest
 
+  alias Lightning.Workflows
   alias LightningWeb.RunLive.Components
   alias LightningWeb.RunLive.Index
 
--- a/test/lightning_web/live/work_order_live_test.exs
+++ b/test/lightning_web/live/work_order_live_test.exs
@@ -1,5 +1,4 @@
 defmodule LightningWeb.WorkOrderLiveTest do
-  alias Lightning.Workflows
   use LightningWeb.ConnCase, async: true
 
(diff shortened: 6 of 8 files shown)


lib/lightning_web/live/workflow_live/job_view.ex
```diff
@@ -183,6 +192,29 @@ defmodule LightningWeb.WorkflowLive.JobView do
...
+        id="workflow-snapshot-version-chip"
+        phx-hook="Tooltip"
+        data-placement="bottom"
+        aria-label={
+          if @version == "latest",
+            do: "This is the latest version of the workflow.",
+            else:
+              "You are viewing a snapshot of this workflow that was made on #{Helpers.format_date(@snapshot_date)}"
+        }
+        class={"inline-flex items-center rounded-md bg-#{if @version == "latest", do: "blue-100", else: "yellow-100"} px-2 py-1 text-xs font-medium text-#{if @version == "latest", do: "blue-800", else: "yellow-800"}"}
```
stuartc, line 208: Tailwind can't detect partial/computed styles, so this should be:

`#{if @version == "latest", do: "bg-blue-100", else: "bg-yellow-100"}`

It may have worked by coincidence because the style was generated somewhere else.

christad92 (commented):
Good progress so far. here are a few things I noticed: 

1. Can't open a snapshot from the run view
2. After switching from the snapshot to latest, the message "You're in a read-only mode" should be cleared off as I am no longer in a read-only/snapshot view. 
3. When I switch from snapshot to latest, the input/editor is not updated. 
4. I can't see the snapshot ID on the canvas.

christad92 (commented):
Thanks @elias-ba 

Here are few things we need to fix: 
- Attempting to move from Run view to canvas fails with an error
- I was able to edit and save on a snapshot without switching to latest version. You can recreate this by opening a snapshot on the canvas or inspector then pressing cms + S. Also, on the canvas, I was able to add new steps and press cmd + S. The expected behavior here should be that I get an error like "Cannot save in snapshot mode, switch to latest"
- When I switch to latest on the canvas, I should not be able to switch back to the snapshot. We can't associate the canvas with a run/input/output like the inspector where you'd see the related data. 
- When I make a new work order/retry after switching from a snapshot, I should not be able to switch back. The toggle should be hidden as well. 
- Apply style to all tooltips on the history page. 


I'd love for us to review some copies: 

**When I hover on the snapshot id/latest tag in the run view, the message should be :** 
- for old versions: This run is based on a snapshot of this workflow that was taken on {{date}}
- for latest: This run is based on the latest version of this workflow.

**When I hover on the tags in the canvas/editor:** 
for old: You are viewing a snapshot of this workflow that was taken on {{date}}
for latest: This is the latest version of this workflow

**When I am retrying a run:** 
Rerun workflow from this step with the latest version of this workflow

taylordowns2000 (changes requested):
three regressions here as of `379bb108aa8d7eabd324d448ce308a3ce75c5289` on Jun 17 825am UTC:
1. clicking on the workflow when a run is latest results in a 500 (images 1 and 2)
2. pressing "Save" on the inspector removes the current _run_ (logs, input, etc.) from the inspector
3. don't let users add auth methods to historical snapshots. right now it lets you attempt to create auth methods and then crashes with `no function clause matching in LightningWeb.WorkflowLive.Edit.apply_action/3` (image 3) but this behaviour should be prevented anyway when in "edit only" mode.
4. when viewing a snapshot on the _**Canvas**_ there is no "Run" loaded (this UI will come later, but it doesn't exist.) Therefore, it is impossible to "toggle between" latest and a snapshot. Once you click the "Go to latest" button (which cannot be a toggle, it must be a one-way button) there is no "going back" to a snapshot, because it would beg the question "which one?"

### image 1
image

### image 2
image

### image 3
image

elias-ba:
Review change requests checklist

From @christad92 
- [x] Attempting to move from Run view to canvas fails with an error
- [x] I was able to edit and save on a snapshot without switching to latest version. You can recreate this by opening a snapshot on the canvas or inspector then pressing cms + S. Also, on the canvas, I was able to add new steps and press cmd + S. The expected behavior here should be that I get an error like "Cannot save in snapshot mode, switch to latest"
- [x] When I switch to latest on the canvas, I should not be able to switch back to the snapshot. We can't associate the canvas with a run/input/output like the inspector where you'd see the related data.
- [x] When I make a new work order/retry after switching from a snapshot, I should not be able to switch back. The toggle should be hidden as well.
- [x] Apply style to all tooltips on the history page.
- [x] When I hover on the snapshot id/latest tag in the run view, the message should be : for old versions: This run is based on a snapshot of this workflow that was taken on {{date}}, for latest: This run is based on the latest version of this workflow.
- [x] When I hover on the tags in the canvas/editor: for old: You are viewing a snapshot of this workflow that was taken on {{date}}, for latest: This is the latest version of this workflow
- [x] When I am retrying a run: Rerun from this step with the latest version of this workflow

From @taylordowns2000 
- [x] clicking on the workflow when a run is latest results in a 500
- [ ] pressing "Save" on the inspector removes the current run (logs, input, etc.) from the inspector
- [x] don't let users add auth methods to historical snapshots. right now it lets you attempt to create auth methods and then crashes with no function clause matching in LightningWeb.WorkflowLive.Edit.apply_action/3 but this behaviour should be prevented anyway when in "edit only" mode.
- [x] when viewing a snapshot on the Canvas there is no "Run" loaded (this UI will come later, but it doesn't exist.) Therefore, it is impossible to "toggle between" latest and a snapshot. Once you click the "Go to latest" button (which cannot be a toggle, it must be a one-way button) there is no "going back" to a snapshot, because it would beg the question "which one?" looks like the top-right corner of this page has been missed/omitted.

christad92 (approved):
Thanks!! This works and it's good.

assets/js/editor/Editor.tsx
```diff
@@ -313,7 +314,13 @@ export default function Editor({
...
         loading={<div className="text-white">Loading...</div>}
         value={source || DEFAULT_TEXT}
-        options={options}
+        options={{
+          ...options,
+          readOnly: disabled,
+          readOnlyMessage: {
+            value: disabledMessage,
+          },
+        }}
```
stuartc, line 323: @josephjclark could I get your opinion on this? I'm concerned this might defeat the render/props checking causing re-renders?

lib/lightning/runs/handlers.ex
```diff
@@ -37,7 +37,8 @@ defmodule Lightning.Runs.Handlers do
     def call(run, params) do
       with {:ok, attrs} <- new(run, params) |> apply_action(:validate),
            {:ok, step} <- insert(attrs) do
-        run = Runs.get(attrs.run_id, include: [:workflow])
+        run = Runs.get(attrs.run_id, include: [:workflow, :snapshot])
+        step = Repo.preload(step, :snapshot)
```
stuartc, line 41: @elias-ba we need to come back to this stuff and push the responsibiliy of preloading down to the subscribers of the event.

lib/lightning/workflows/snapshot.ex
```diff
@@ -52,6 +52,11 @@ defmodule Lightning.Workflows.Snapshot do
       field :cron_expression, :string
       field :enabled, :boolean
       field :type, Ecto.Enum, values: [:webhook, :cron]
+      field :has_auth_method, :boolean, virtual: true
```
stuartc, line 55: @elias-ba I'm not quite able to work out how this is used, I see it's in the workflow_params, but I can't find anything that sets it?

assets/js/editor/Editor.tsx
josephjclark: Yeah this is probably fine. You're right to ask and this pattern is usually a no-no with react components for that reason.

But often less so with options. I expect Monaco will deep diff the options before triggering a re-render, which case we're golden.

josephjclark: Having said that! I also see that `options` is held in a `useState`, and it would likely be cleaner to drive the state options properly rather than overriding here

lib/lightning/workflows/snapshot.ex
elias-ba: Because we are using a snapshot to build the changeset in the workflow Canvas, when viewing a workflow that has auth methods set to its trigger, the UI would break without this.

stuartc: I guess what I'm saying is, I can't see anything that sets it... so affect it will always be `nil`. If it's not ever being set we should make an effort to remove it in another PR perhaps?

lib/lightning_web/live/run_live/show.ex
```diff
@@ -27,6 +57,12 @@ defmodule LightningWeb.RunLive.Show do
             <span class="pl-2 font-light">
               <%= display_short_uuid(@id) %>
             </span>
+            <div class="mx-2"></div>
+            <.async_result :let={run} assign={@run}>
+              <%= if run do %>
+                <.snapshot_version run={run} workflow={@workflow} />
+              <% end %>
+            </.async_result>
```
stuartc, line 65: fyi @elias-ba I rewrote this to use a component, please avoid assigning variables inside heex templates. I also changed `JobView.job_edit_view` and moved the editor disabled data into assigns.

stuartc (approved):

PR #2185, opened 2024-06-07
Adds Project and Account hooks

### LINKED ISSUE #2029: Allow umbrella apps to handle user registration and project creation
Lightning already has `UserRegistered` and `ProjectCreated` events but some entities are expected to be inserted on the same database transactions.

jyeshe:
## Validation Steps

1. Register a new user (Lightning form)
2. Login as super user and create another super user
3. Still as a super user, create a regular user
4. Still as a super user, create a project

## Notes for the reviewer

These hooks (`ProjectHook` and `AccountHook`) inside Project and Account contexts allow umbrella apps to handle project and user creation on the same DB which means that all set of operations that need to be atomic can be handled without relying on async events.

As `ProjectHook.handle_create_project/1`, `AccountHook.handle_create_user/1` and so on are called in the context of a DB transaction can be reused by an umbrella app to override and/or extend these hooks.

## Related issue

Resolves #2029
Refs https://github.com/OpenFn/thunderbolt/issues/146

## Review checklist

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,7 +33,9 @@ config :lightning, LightningWeb.Endpoint,
 config :lightning, Lightning.Extensions,
   rate_limiter: Lightning.Extensions.RateLimiter,
   usage_limiter: Lightning.Extensions.UsageLimiter,
-  run_queue: Lightning.Extensions.FifoRunQueue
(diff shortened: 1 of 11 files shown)


stuartc (approved):
👌

PR #2186, opened 2024-06-10
Reject unsupported content-types on webhooks

### LINKED ISSUE #2180: Don't throw an error when the content type is unsupported
Sentry Issue: [LIGHTNING-6Z](https://openfn.sentry.io/issues/4578553444/?referrer=github_integration)

```
Protocol.UndefinedError: protocol Jason.Encoder not implemented for %Plug.Conn.Unfetched{aspect: :body_params} of type Plug.Conn.Unfetched (a struct), Jason.Encoder protocol must always be explicitly implemented.

If you own the struct, you can derive the implementation specifying which fields should be encoded to JSON:

    @derive {Jason.Encoder, only: [....]}
    defstruct ...

It is also possible to encode all fields, although this should be used carefully to avoid accidentally leaking private information when new fields are added:

    @derive Jason.Encoder
    defstruct ...

Finally, if you don't own the struct you want to encode to JSON, you may use Protocol.derive/3 placed outside of any module:

    Protocol.derive(Jason.Encoder, NameOfTheStruct, only: [...])
    Protocol.derive(Jason.Encoder, NameOfTheStruct)
. This protocol is implemented for the following type(s): Any, Atom, BitString, Date, DateTime, Decimal, Ecto.Association.NotLoaded, Ecto.Schema.Metadata, Float, Integer, Jason.Fragment, Jason.OrderedObject, Lightning.RedactedMap, List, Map, NaiveDateTime, Time
  File "lib/jason.ex", line 213, in Jason.encode_to_iodata!/2
  File "/app/deps/postgrex/lib/postgrex/type_module.ex", line 947, in Lightning.PostgrexTypes.encode_params/3
  File "lib/postgrex/query.ex", line 75, in DBConnection.Query.Postgrex.Query.encode/3
  File "lib/db_connection.ex", line 1403, in DBConnection.encode/5
  File "lib/db_connection.ex", line 1503, in DBConnection.run_prepare_execute/5
...
```

stuartc:
## Validation Steps

1. Run your local server
2. Get a web hook url for any workflow
3. Run a curl command:

```sh
curl \
 -X POST \
 -H "Accept: */*" \
 -H "Accept-Encoding: gzip, deflate" \
 -H "Content-Length: 4834" \
 -H "Content-Type: text/xml" \
 $WEBHOOK_URL
```

You should see: `{"error":"Unsupported Media Type"}` as the response.

## Notes for the reviewer

Users that send data without a supported mime type cause exceptions to be raised because the controller tries to encode the request body (which is `Unfetched`).

Instead of throwing an exception, we should return a 415 error.

## Related issue

Fixes #2180

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to
   [#2182](https://github.com/OpenFn/lightning/pull/2182)
 - Update email notification for changing retention period
   [#2066](https://github.com/OpenFn/lightning/issues/2066)
+- Return 415s when Webhooks are sent Content-Types what are not supported.
+  [#2180](https://github.com/OpenFn/lightning/issues/2180)
 
 ### Fixed
 
--- a/lib/lightning_web.ex
+++ b/lib/lightning_web.ex
@@ -22,7 +22,6 @@ defmodule LightningWeb do
 
   def router do
     quote do
-      # , helpers: false
       use Phoenix.Router
 
       import Plug.Conn
--- a/lib/lightning_web/controllers/webhooks_controller.ex
+++ b/lib/lightning_web/controllers/webhooks_controller.ex
@@ -9,7 +9,23 @@ defmodule LightningWeb.WebhooksController do
   alias Lightning.Workflows
   alias Lightning.WorkOrders
 
-  require OpenTelemetry.Tracer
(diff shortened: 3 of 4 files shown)


elias-ba (approved):
This is looking good @stuartc 🎉. Thanks

PR #2194, opened 2024-06-11
Use erlang 26.2.5

taylordowns2000:
Bump to the latest patch of Erlang OTP 26: https://www.erlang.org/patches/otp-26.2.5


stuartc (approved):
Nice! I double checked and the CI image we use has 26.2.5 as it's OTP version.

Issue #2014, opened 2024-04-22
Update default comment in job editor to help users get started


christad92:
Currently new users get lost on the editor page when they as they don't know which language to code in or how to organize their code. Templates is an expensive way of solving this so we are adopting a much simple approach of adding multiline comments to  users know what to do. 

```
// Use pure JavaScript to add operations to your
// step. Click Docs to see list of operations or visit
// https://bit.ly/OFNJWG for adaptor documentation
// and example job codes.
```

jyeshe:
would add also this snippet:
```
fn(state => {
  return state
})
```
It's elegant but it's exactly the same as js arrow functions:
`const greetings = (a) => {return console.log("Hello " + a)};`

PR #2196, opened 2024-06-12
update default step text

josephjclark:
## Details

This is another attempt to address #2014 - but that issue doesn't even scratch the surface of the problems users face getting started.

## What was wrong with the last fix?

* It wrongly claimed that jobs are written in "pure" JS (the pure bit in particular I have issues with)
* It used a URL shortener to link to our own docs. But URL shorteners are poor UX and bad security practice
* It had bad english eg "job codes"
* It had some weird stuff about the Docs panel
* Also its full of jargon

I'm sorry but basically it just hadn't been thought through with care and rigour.

Also, doing any of this stuff in a comment only lasts as long as the comment. Once the user deletes it, it's gone forever.

## Why is this fix so much better

* First of all this "fix" is terrible and doesn't address the issue in question
* It removes the URL shortener and is honest about the content it links to
* It is good english
* It is accurate

## What next?

Onboarding the user into something as complex as openfn job writing is really hard and won't be solved by a comment If there is to be any attempt to address this problem it needs coordinated, thoughtful effort.

Here are some brief thoughts about things we might look at:

* Having a Help or Getting Started tab, with a short and friendly onboarding guide
* Don't show four large, complex and intimidating panels when the user first opens the inspector
* Putting a leading example in the Docs panel, one per adaptor
* Similarly, put a default example in each adaptor and insert that into the template (doesn't work when the adaptor changes)
* Show a permanent banner with a link to the job writing guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to
   [#2066](https://github.com/OpenFn/lightning/issues/2066)
 - Return 415s when Webhooks are sent Content-Types what are not supported.
   [#2180](https://github.com/OpenFn/lightning/issues/2180)
+- Updated the default step text
 
 ### Fixed
 
--- a/assets/js/editor/Editor.tsx
+++ b/assets/js/editor/Editor.tsx
@@ -10,7 +10,7 @@ import { initiateSaveAndRun } from '../common';
 import dts_es5 from './lib/es5.min.dts';
 
 export const DEFAULT_TEXT = `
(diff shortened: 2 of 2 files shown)


taylordowns2000 (approved):
I love it. @christad92 , what do you think?

christad92 (commented):
Happy for us to make the fix.

christad92 (approved):

stuartc (approved):

PR #2198, opened 2024-06-13
Fix Docs Panel

### LINKED ISSUE #2195: Docs panel vanishes in Inspector
I _think_ this may have to do with the resize detector/hooks. Apologies for not being able to add audio to the loom. Here are the details:

1. Headline - on the inspector you can "lose" the docs/metadata panel and you can't get it back without some special hacking.
2. To reproduce, go to the inspector on app.openfn.org or on `main` and with a browser zoom level similar to mine in the loom (standard zoom on Chrome on a laptop with a 1680x1050 display) try out the various docs panel controls to (a) change orientation and (b) expand/collapse. Note that the panel vanishes under certain conditions.
3. To get the panel back, collapse and expand the editor pane and it will figure out the proper spacing/sizing.

https://www.loom.com/share/30e2928754904a12ae1ec6983c2f54ca

josephjclark:
* Tweak the margins and scrollbars inside the panel so that it looks a little better in small windows (it's always a small window when it opens by default)
* Restore `overflow: hidden` from #2154, which seemed to get dropped somewhere

Closes #2195

# QA notes

This needs really thorough testing!

* Check the docs panel retains a reasonable size when the left, right and center columns are expanded
* Check the docs panel retains a reasonable size when expanded and collapsed, in horizontal and vertical layout
* Check the docs panel retains a reasonable size while the browser window is resized and zoomed

Note that if the docs panel is very narrow, some tabs and controls will be hidden from view. This is expected and there is nothing we can do about it.

### DIFF
--- a/assets/js/job-editor/JobEditor.tsx
+++ b/assets/js/job-editor/JobEditor.tsx
@@ -166,7 +166,7 @@ export default ({
         </div>
         <div
           className={`${
-            showPanel ? 'flex flex-1 flex-col z-10 overflow-auto' : ''
+            showPanel ? 'flex flex-1 flex-col z-10 overflow-hidden' : ''
           } ${vertical ? 'pt-2' : 'pl-2'} bg-white`}
         >
           <div
--- a/assets/js/job-editor/JobEditor.tsx
+++ b/assets/js/job-editor/JobEditor.tsx
@@ -155,7 +155,7 @@ export default ({
     <>
       <div className="cursor-pointer"></div>
       <div className={`flex h-full flex-${vertical ? 'col' : 'row'}`}>
-        <div className="flex-1 rounded-md">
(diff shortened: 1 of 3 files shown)


assets/js/job-editor/JobEditor.tsx
```diff
@@ -155,7 +155,7 @@ export default ({
     <>
       <div className="cursor-pointer"></div>
       <div className={`flex h-full flex-${vertical ? 'col' : 'row'}`}>
-        <div className="flex-1 rounded-md">
+        <div className="flex-1 rounded-md overflow-hidden">
```
josephjclark, line 158: This is the actual fix from #2154

christad92 (approved):
Works as expected! LGTM

stuartc (approved):

PR #2204, opened 2024-06-17
Fix race condition when syncing from openfn while deploy actions are running

### LINKED ISSUE #2179: Race condition when triggering github sync from openfn while deploy actions are running
Will add more details later..

@mtuchi ran into a situation where one of his deploy workflows could not complete because his branch was out of date with the remote branch. Know that git error which tells you to pull from the remote first?

He said he did a push to his branch and then triggered a sync from openfn. My gut tells me that the sync from lightning did a commit before one of his workflows could complete. I think this is possible given that the `deploy` workflows are currently running sequentially, while the `pull` ones aren't.

Maybe we should have the `pull` actions running in the `openfn-deployment` group as well so that we avoid all these race conditions.

midigofrank:
## Validation Steps
In order to test this properly, your github project should have more than 1 connection to openfn. Let's go with 2

- Trigger the deploy actions by committing and pushing to your branch then immediately go to Lightning then click sync to branch.
- You'll notice that your `pull job` will run `3rd`, that is after the 2 deploy  actions have ran.

## Related issue

Fixes #2179 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Fix race condition when syncing from openfn while deploy actions are running
+  [#2179](https://github.com/OpenFn/lightning/issues/2179)
+
 ## [v2.6.2] - 2024-06-13
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


midigofrank:
Oops!
https://taurit.pl/github-canceling-since-a-higher-priority-waiting-request-exists/

midigofrank:
This doesn't fix the issue. It will cancel pending workflows. See: https://github.com/midigofrank/openfn/actions/runs/9542614901

PR #2215, opened 2024-06-24
Reverted behaviour on "Rerun from here" to select the Log tab.

### LINKED ISSUE #2202: Two regressions "snap-to-logs" for Inspector view
## Regressions
1. When clicking **"Retry from here"** or "**Create new workorder**" from the inspector, the right-side panel should snap to the logs view.
2. When clicking **"Inspect Step"** from the History page, the right side-panel should snap to the logs view.

Both of these did this _before_ version 2.6, but were lost in the TabSelector refactor.

Notes:
1. Right now it stays on whatever tab you had open before. (So the default is "Run". See first image.)
4. See second image for what it should look like after you click run, no matter which tab was open before: Run, Log, Input, Output.

### Clicking Run
image

image

### History to Logs
image

stuartc:
## Validation Steps

- When inspecting an existing Run, ensure you have any tab other than 'Logs' selected.
- Click "Retry from here"
- The Logs tab should be selected

## Related issue

Fixes #2202

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to
 
 ### Changed
 
+- Reverted behaviour on "Rerun from here" to select the Log tab.
+  [#2202](https://github.com/OpenFn/lightning/issues/2202)
+
 ### Fixed
 
 ## [v2.6.3] - 2024-06-19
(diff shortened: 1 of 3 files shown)


taylordowns2000 (changes requested):
(we spoke offline, just noting for GH in case this gets lost)

point 1 of 2 was done perfectly. second regression needs addressing

taylordowns2000:
@stuartc , fyi i didn't like overriding the default behaviour without actually thinking through the implications. i'm using the current (working nicely) implementation where a link with `#log` at the end will focus the logs tab.

from run view:
image

from history
image

PR #2218, opened 2024-06-25
Provide link to `/src` from Inspector

taylordowns2000:
## Validation Steps

View adaptor docs for adaptors without any docs:
image

View adaptor docs for adaptors with docs:
image

## Notes for the reviewer

@josephjclark , documentation is getting better every day, but in my typically workflow I still _almost always_ go to the source. There's something nice and soothing about knowing exactly what this thing I'm about to call really is. What do you think?

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to
 
 ### Added
 
+- Link to adaptor `/src` from inspector.
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


josephjclark:
@taylordowns2000 how about we just ALWAYS render the "you can check the external docs HERE" line at the top of the the docs.

Then if we have docs, we'll render them, if not we'll show an error message

christad92 (approved):

Issue #2221, opened 2024-06-25
Workflow diagram: contextual min zoom


josephjclark:
The workflow diagram has hard-coded min and max zoom levels, which means that you can only ever zoom in or out so far.

I would love a smart min zoom, where you can always zoom out to the point where the chart fits the canvas, but no further (well maybe 10% further)

That would mean that in small workflows it's impossible to zoom out so much that you can't see anything, but in big workflows we can guarantee we'll fit everything on screen (even if you can't really see it). And all without supporting "useless" zoom levels.

We should be a able to work this out, because we can calculate (at least approximately) the total size of the chart  from the positional data. So if we know that the chart requires 1000x1000 pixels, we can scale the zoom level accordingly. We would need to work out what zoom:1 means in terms of chart size, and from there we should be able to figure out at least a roughty correct min zoom level.

theroinaochieng:
This was completed and merged on this PR: https://github.com/OpenFn/lightning/pull/2222

PR #2222, opened 2024-06-25
workflow-diagram: Reduce min zoom

### LINKED ISSUE #2214: Allow users to be able to fit their workflows into their view
When building workflows with many steps, the workflow tends to extend the beyond viewport of the screen and our current minimum zoom level does not allow us to zoom out to fit the workflow to the screen size of the device. Users want to be able to see the whole workflow by zooming out without needing to pan around. 

See the video below: 

https://www.loom.com/share/5cf61fa7dddf47f6a5a0b9fc901dee4b?sid=1df7e069-5434-4794-9873-ee6b8c670c3b

josephjclark:
This PR simply reduces the min-zoom level, letting you zoom further out on big diagrams

This is a simple fix and won't work in all cases. It probably also zooms too far out so maybe is annoying?

I've raised this as a better solution: https://github.com/OpenFn/lightning/issues/2221

Closes #2214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to
 
 - Reverted behaviour on "Rerun from here" to select the Log tab.
   [#2202](https://github.com/OpenFn/lightning/issues/2202)
+- Reduce the minimum zoom in the workflow diagram
+  [#2214](https://github.com/OpenFn/lightning/issues/2214)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


josephjclark:
@christad92 this diff is trivial from a technical perspective. Could you run some QA and make sure it seems OK from a product POV?

christad92 (approved):
This works and looks good to me. 
image

stuartc (approved):

PR #2230, opened 2024-06-26
Dynamic zoom

### LINKED ISSUE #2174: Workflow Diagram: Add dynamic zoom
In the workflow diagram, when you add a new node, the current behavior is to zoom to fit the whole chart.

This is super annoying on large charts if you've zoomed into a corner of it, because you'll be zoomed out. In order to edit the placeholder you have to zoom right back in again.

Workflows don't even have to be particularly big before this becomes a problem.

We could add buttons somewhere to zoom to fit and toggle auto zoom - but these feel clumsy. They're hard to communicate (what icon would you use?) and it's fiddly to keep adjusting settings.

I think the answer is a dynamic zoom. When you add a node, we zoom the diagram to fit whatever steps you can see now, PLUS the new step.

These kind of behaviours are fiddly to implement, but if we can at least reduce the amount of zoom movement on add, we'll greatly improve the user experience.

josephjclark:
This PR introduces a smart auto-layout feature to the workflow diagram.

In brief, the changes are:

* When adding a placeholder node, instead of fitting the whole chart, we just fit the currently visible nodes PLUS the placeholder
* When removing a placeholder node, we automatically try to fit to the currently visible nodes (I don't think this works very well)
* The "toggle auto-fit" controller has been removed from the toolbar in the bottom left - the whole concept of this feature is to make that button redundant.
*  When the window is resized, instead of zooming to fit the whole chart, try to just fit the visible nodes (to my surprise, this works well!)

As a rule, I am trying to only move the view if I really have to.

Despite a day of sweating over this, I am not yet sold. Detailed QA is needed.

## Still To do

I think that we should NOT try to fix the view when a placeholder is removed.

In some cases it works great (it's quite good on a small new workflow), but in others its terrible (it's particularly bad on larger workflows)

## Notes for the reviewer

Before this PR, when running a layout, there is a little function that calculates the total size of the chart and zooms the view to fit those dimensions. We do this programmatically and sync it to layout, so that layout and zoom both happen at the same time. The result is pretty slick!

The problem is that users lose their context whenever adding or removing placeholder. This is a particular problem in bigger workflows - you can add a new node and bosh, you'll be zoomed out so far you can't even see.

Here, I've added a bunch of logic to the layout function to try and work out the best dimension to fit the view to. To do this, we:
* Build a list of all the currently visible nodes (with an offset/margin so that partially visible nodes can be included or dropped, depending)
* If the currently visible nodes or the placeholder are NOT going to be visible in the current view in the new layout, we prepare a new zoom-fit
* We fit to the bounds of all the nodes we think are visible (plus the placeholder)
* And basically the same thing when a placeholder is removed - we attempt to re-fit the view. The idea here is to reclaim any space that the placeholder's view created, but the results aren't great

One difficulty with this is that we'll always fit to what the chart thinks is a perfect view of the visible nodes. This often results in small and arbitrary and occasionally even quite harmful zooms.

There's also new logic added for resizing. When a resize drag starts, we take a snapshot of what nodes are currently visible, and as the user resizes we constantly zoom-fit on the chart to try and keep those nodes in view. It doesn't matter that this isn't perfect or that the canvas size changes - it just locks the view in a _similar_  place to it was before the resize started. And that's a heck  of a lot better than resetting the zoom.

Rather crudely, we reset this "resize drag" state after 3 seconds. I'm sure we could be smarter about this but this seems reasonably effective.

## QA Notes

This one needs thorough and careful QA.

We need to make sure that:
* The new behaviour is not annoying to users
* The new behaviour preserves a user's context within the workflow
* The new behaviour does not change the view unless it really has to

Here's the kind of thing I'm testing

* On a new workflow, add a bunch of nodes without touching the zoom level
  * Make the workflow deep and wide
  * The view should happily expand to fit the newly created nodes
* On an existing, large-ish workflow, zoom in to one area and add new nodes
  * The view should expand just enough to fit the new node but shouldn't blow up your  context
 * Try all this at extremely high and low zooms
 * Try creating a very very deep chain of nodes, child after child, zooming in occasionally
 * Try creating  lots of children on one node. Note that the new node often comes up on the left, which is an annoying layout problem, but more importantly note that the view will fit horizontally.
 * Try cancelling a placeholder node and observe what the view does
   * Try this at different zoom levels on small and large workflows
  * On a large workflow, zoom in, then resize the browser window. Try constantly dragging the chart around and watch it update with broadly the same context. If you zoom to a different place and resize again, the chart should try and zoom to this new position
* Try adding and removing placeholder nodes in the middle of the workflow, where they should not affect the zoom at all. Generally, adding  placeholders is quite well behaved (minimal  view changes), but removing them is quite bad (too many view changes)

## Related issue

Fixes #2174

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to
 
 ### Changed
 
+- In the workflow diagram, smartly update the view when adding new nodes
+  [#2174](https://github.com/OpenFn/lightning/issues/2174)
+- In the workflow diagram, remove the "autofit" toggle in the control bar
+
 ### Fixed
 
 ## [v2.7.0] - 2024-06-26
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -244,7 +244,6 @@ export default React.forwardRef<HTMLElement, WorkflowDiagramProps>(
           onEdgeClick={handleEdgeClick}
           onInit={setFlow}
           deleteKeyCode={null}
-          minZoom={0.2}
           fitView
           fitViewOptions={{ padding: FIT_PADDING }}
           minZoom={0.2}
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -6,6 +6,8 @@ import ReactFlow, {
   ReactFlowInstance,
   ReactFlowProvider,
   applyNodeChanges,
+  getRectOfNodes,
+  Rect,
 } from 'reactflow';
 import { useStore, StoreApi } from 'zustand';
 import { shallow } from 'zustand/shallow';
@@ -24,6 +26,7 @@ import shouldLayout from './util/should-layout';
 
 import type { WorkflowState } from '../workflow-editor/store';
 import type { Flow, Positions } from './types';
+import { getVisibleRect, isPointInRect } from './util/viewport';
 
 type WorkflowDiagramProps = {
   selection: string | null;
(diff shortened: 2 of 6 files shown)


taylordowns2000 (approved):
this looks great. i've tested everything i could and while there is a _little_ extra movement when removing placeholders, i think this is already more than good enough. green light from me.

christad92 (approved):

stuartc (approved):

PR #2236, opened 2024-06-27
Edges: Allow labels to exist outside of conditions

josephjclark:
The implementation team like to set Labels on their paths, regardless of whether or not there's a condition attached.

To do this today, they sort of have to hack it: set the Condition to Custom, then set `true` or something as the expression, and THEN set the label. Pretty tedious (and actually a little misleading)

This PR moves the edge label outside of the condition part of the form, so it's just a Label now, rather than a conditional label.

image

The only drawback I see to this is that if a label is set to a string, it might hide the actual behaviour. Like I can set a string "send email on error", but set the condition type to "on success". In fact as soon as I set the label, I lose visibility of the condition type entirely.

That's all basically true in production

## Label style updates

To resolve this issue, the label styles have been updated. Now the label shows an icon AND label at all times - so if the two are inconsistent, there is a visual clue in the UI

image

## Notes for the reviewer

I have NOT renamed `conditional_label` at the backend or  frontend. Maybe we should do that? Strictly speaking it should be `label`, but `condition_label` to be honest is close enough.

I just wanted a quick fix to a problem users are facing in the app today.

## Related issue

No issue that I can find (I'm sure I mentioned this, perhaps as a feature request)

## Review checklist


stuartc (commented):
Thanks dude, this seems very sensible... I'm kinda thinking renaming the field in the db as well.

But your points on having to update the label when going from an expression to a preset are valid.

@christad92 what do you think? Should we allow users to set labels on anything, or should be set the `condition_label` to nil when they change from an expression to a preset.

Btw, we currently set the `cron_expression` to nil _if_ you change from `cron` to `webhook`, so making it do that here would make sense in context of what we have right now.

My one concern is that I don't think users will know that if they leave the label blank it will default to the preset type?

josephjclark:
You know what @stuartc I think I'm just going to go ahead and do that...

josephjclark:
Took me while to finesse this, but it looks great IMO

image

josephjclark:
So now if you do something silly like this, the UI is giving you a pretty strong hint that something is up 
image

taylordowns2000 (changes requested):
this works but i'd love for the vertical positioning of the icon to be centered along with the text. the "X" is out of whack, while the others look good

josephjclark:
@taylordowns2000 ooh yeah it does look bad in that screenshot

I spent a long time on the alignment. I think you'll find it's technically correct, but the oversized X and the lowercase label are making it look off.

I'll take another look when I get a chance

josephjclark:
@taylordowns2000 Yeah I think the screenshot is misleading you:

Here's the label:

image

Looks a bit off to the eye maybe, but  if I set all caps:

image

I think zoom level affects the alignment - or at least perceived alignment. If I zoom out I get this, which looks awful

image

josephjclark:
@taylordowns2000 I might be able to do this, which might help distract the eye (and generally look better?)

image

josephjclark:
A cheaper and probably more robust solution - we colour code the icon red for on-fail, green for on-success, blue for anything else. Looks kinda neat

image

josephjclark:
Alright, we've settled for this:

image

The PR description is updated, as well as the changelog

stuartc (approved):

PR #2238, opened 2024-06-27
Improve help/docs for path expressions

### LINKED ISSUE #1903: Document edge conditions
The UI for edge conditions needs to include some basic information about their usage.

image

Some things that should be explained:

* must take an expression (no statement, no function wrapper)
* your state object is available
* return truthy to execute this edge
* use optional chaining (`?`) in state references

A set of common patterns would be useful

josephjclark:
A simple but considered little PR which adds some help text to the Path Expressions UI.

image

## Notes for the reviewer

**SUPER IMPORTANT** I have changed the top margin on the error message of the `input` component. I've had a click around and it looks OK, but there could be surprises as a result of this PR.

I've had to make the change  because otherwise the error text bleeds over the help.

Changes I've made:
* Add placeholder text to the expression
* Take the field out of the label, so that the text isn't all bolded (note that the error is no longer bold too)
* Added a collapsible help area with some immediate text and a link to docs

## QA Notes

As a general rule, the padding between error messages and the fields they relate to has been REDUCED. This looks great in the Workflow Diagram UIs, but I don't know if I've broken something else. Presumably the margin was set to a very high value for a reason.

## Related issue

Fixes #1903

## Review checklist

### DIFF
--- a/lib/lightning_web/components/new_inputs.ex
+++ b/lib/lightning_web/components/new_inputs.ex
@@ -412,7 +412,7 @@ defmodule LightningWeb.Components.NewInputs do
     ~H"""
     <p
       data-tag="error_message"
-      class="mt-3 inline-flex items-center gap-x-1.5 text-xs text-danger-600"
+      class="mt-1 inline-flex items-center gap-x-1.5 text-xs text-danger-600"
     >
       <.icon name="hero-exclamation-circle" class="h-4 w-4" />
       <%= render_slot(@inner_block) %>
--- a/lib/lightning_web/live/workflow_live/components.ex
+++ b/lib/lightning_web/live/workflow_live/components.ex
@@ -417,15 +417,17 @@ defmodule LightningWeb.WorkflowLive.Components do
             maxlength="255"
             placeholder="eg: !state.error"
           />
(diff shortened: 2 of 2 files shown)


lib/lightning_web/components/new_inputs.ex
```diff
@@ -412,7 +412,7 @@ defmodule LightningWeb.Components.NewInputs do
     ~H"""
     <p
       data-tag="error_message"
-      class="mt-3 inline-flex items-center gap-x-1.5 text-xs text-danger-600"
```
josephjclark, line 415: I want to make lots of noise about this change!!!!

I don't really know what it'll break.

christad92 (approved):

lib/lightning_web/components/new_inputs.ex
stuartc: It looks ok, it's the error description for a large set of inputs used in the app..

Before:
Screenshot 2024-07-02 at 14 22 15
After:
Screenshot 2024-07-02 at 14 22 00


Thanks for flagging!

stuartc (approved):

PR #2249, opened 2024-07-02
All models to use public schema

stuartc:
## Notes for the reviewer

Add `Lightning.Schema` macro so all models get the same defaults.

This explicitly sets the schema to `public`, making integration with other applications that use Ecto easier since the schema is known no matter what order the query is built.

## Review checklist


jyeshe (approved):
Nice cleanup!

PR #2256, opened 2024-07-04
Fix step selection issue/order of phx-click and links

### LINKED ISSUE #2253: "Switch to inspect" links broken after `v2.6.3`
These little icons (see image) are links to open the inspector for a different job in the workflow. Here's [how it works](https://www.loom.com/share/2422a4453eae4e93bd0f62ead9fdb949) in `v2.6.3`

image

When clicking the inspect icon next to the job title of the second step here ("another") we should show that job code in the middle panel (you'll see the whole inspector reload, exactly as desired) and then the link icon next to "another" should dissapear. (you can't navigate to where you already are.) you should see a "open the inspector for this job" link show up next to "New job" but you don't. [Here's how it now looks](https://www.loom.com/share/51709d7a7f8a4ec1afcc228afb334721) on `main` at `39a913da0f46edf2364614357a12c197bc93eda7`.

stuartc:
## Validation Steps

Open a Run from the History page, select different steps on the inspector, the job editor should reflect the correct code and the run viewer should update to not have an icon next to the selected item.

See issue for more details.

## Notes for the reviewer

This was a super weird one, there could be more than one thing at play. But it appeared that the `phx-click="select_step"` was no longer being called, or called too late because the link to the next step would trigger a navigation change.

The changes here use local pubsub to communicate the changes between the Edit and RunViewer liveviews; so the params go to the Edit LiveView and then the RunViewer receives just the changed job id.

The previous way was for the RunViewer to consume the click, handle the state change and then change the page, the responsibility is reversed now. 

## Related issue

Fixes #2253

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to
 
 ### Fixed
 
+- Fix issue when selecting different steps in RunViewer and the parent liveview
+  not being informed [#2253](https://github.com/OpenFn/lightning/issues/2253)
 - Workflow activation only considers new and changed workflows
   [#2237](https://github.com/OpenFn/lightning/pull/2237)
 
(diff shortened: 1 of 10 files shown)


taylordowns2000 (approved):
perfect! thank you

PR #2260, opened 2024-07-05
pretty print raw JSON in logs

taylordowns2000:
## Two questions:
1. Do you like this more than single-line JSON?
2. If yes, will it destroy our performance to `try/catch` JSON stringify on every log line?

### Pretty Print JSON in log viewer:
image

### Keep it as a single line (currently on production):
image

## Review checklist

### DIFF
--- a/assets/js/log-viewer/store.ts
+++ b/assets/js/log-viewer/store.ts
@@ -65,13 +65,26 @@ function coerceLogs(logs: LogLine[]): LogLine[] {
   }));
 }
 
+function isProbablyJSON(str: string) {
+  // Check if the string starts with '{' or '[' and ends with '}' or ']'
+  return (
+    (str.trim().startsWith('{') && str.trim().endsWith('}')) ||
+    (str.trim().startsWith('[') && str.trim().endsWith(']'))
+  );
+}
+
 function formatLogLine(log: LogLine) {
-  try {
-    const jsonObj = JSON.parse(log.message);
-    return `${log.source} ${JSON.stringify(jsonObj, null, 2)}`;
-  } catch {
-    return `${log.source} ${log.message}`;
+  const { source, message } = log;
+  if (isProbablyJSON(message)) {
+    try {
+      const jsonObj = JSON.parse(message);
+      return `${source} ${JSON.stringify(jsonObj, null, 2)}`;
+    } catch {
+      return `${source} ${message}`;
+    }
   }
+
+  return `${source} ${message}`;
 }
 
 export const createLogStore = () => {
--- a/assets/js/log-viewer/store.ts
+++ b/assets/js/log-viewer/store.ts
@@ -28,7 +28,9 @@ function findSelectedRanges(logs: LogLine[], stepId: string | undefined) {
   }>(
     ({ ranges, marker }, log) => {
       // Get the number of newlines in the message, used to determine the end index.
-      const newLineCount = [...log.message.matchAll(/\n/g)].length;
+      const newLineCount = [...possiblyPrettify(log.message).matchAll(/\n/g)]
+        .length;
+
       const nextMarker = marker + 1 + newLineCount;
 
       if (log.step_id !== stepId) {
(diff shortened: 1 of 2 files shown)


assets/js/log-viewer/store.ts
```diff
@@ -66,7 +66,12 @@ function coerceLogs(logs: LogLine[]): LogLine[] {
 }
 
 function formatLogLine(log: LogLine) {
-  return `${log.source} ${log.message}`;
+  try {
+    const jsonObj = JSON.parse(log.message);
```
josephjclark, line 70: This is untested code, but I think this approach will work a little better:
```
if (typeof log.message[0] === 'string') {
  return `${log.source} ${log.message}`;
}
return `${log.source} ${JSON.stringify(jsonObj, null, 2)}`;
```

josephjclark: That will only work if the first item is a string. So `console.log('result', state.data)` won't pretty print.

Message is always an array of strings or JSON (I think??).

I suppose either way we might need to iterate over the message array, pretty print any objects, and then log. Which is a little more complicated

taylordowns2000: the
```
if (typeof log.message[0] === 'string') {
  return `${log.source} ${log.message}`;
}
```
doesn't seem to work for me because we _do_ want to try it as JSON if it's a string. worker is always sending strings or arrays of strings, and it's up to us to give them the old college try and see if we can represent them as beautiful JSON

josephjclark: huh, are you sure the worker isn't sending objects?

I might have to go and look into that. My expectation is that a log event is a an object with `{ source, message }` etc keys, and message is always an array of strings (if actual strings) or objects (if actual objects). Like it shouldn't be double encoding them.

taylordowns2000: yeah peep this

image

seems to always be a string?

stuartc: Yes, we take whatever the worker says and turn it into a string: https://github.com/OpenFn/lightning/blob/main/lib/lightning/ecto_types.ex#L63

stuartc: I have always intended to come back to this and possible change the message to be either an object or a string... but for the sake of moving forward (when we first implemented this) we opted for storing everything as a string.

josephjclark: Oh interesting.

Alright, well for the time forcing a JSON.parse is the way to go.

I think in the long run JSON should be be send natively down to the client, which can render them as collapsible/formatted JSON in monaco. But that can all be far future stuff.

josephjclark (approved):

josephjclark:
Fine for me then, but changelogs etc should be added

taylordowns2000:
> Fine for me then, but changelogs etc should be added

Nice. Done! Also check out how nice it makes your new `worker@1.3` error messages look!
image

stuartc (commented):
This is fine in theory, but I do think a small optimisation should be done to test for the first character being one of `[{"` before attempting to parse it.

AishaHassen:
Q1: I like the pretty print. Much more readable, especially when we have a large json or an array json. I'm a little confused about the yellow line that is hanging on line 25 as opposed to the line spanning all the way down to line number 47/48. I will take back my comment if this is still a work in progress :) 
image

taylordowns2000:
> I'm a little confused about the yellow line that is hanging on line 25 as opposed to the line spanning all the way down to line number 47/48. I will take back my comment if this is still a work in progress :)

No this is a great comment! Thank you. If you _had_ to pick, would you prefer:

(a) pretty print now, yellow line later
(b) nothing now, both later

stuartc (approved):

AishaHassen (approved):

AishaHassen:
> it was worth fixing right away. so how about (c) both now?

Option c sounds great

PR #2264, opened 2024-07-05
Fix Bugy Snapshot Toggle Button

elias-ba:
## Validation Steps

1. Open a snapshot in the inspector
2. Toggle the latest version
3. Toggle back the snapshot
4. Keep repeating 2 & 3, be crazy and hit that toggle as fast as you can
5. See that the toggle is smooth enough and note that the changes indicator doesn't appear

## Related issue

Fixes #

## Review checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -1222,12 +1222,20 @@ defmodule LightningWeb.WorkflowLive.Edit do
   end
 
   def handle_event("push-change", %{"patches" => patches}, socket) do
+    %{workflow: workflow, snapshot: snapshot} = socket.assigns
     # Apply the incoming patches to the current workflow params producing a new
     # set of params.
     {:ok, params} =
       WorkflowParams.apply_patches(socket.assigns.workflow_params, patches)
 
-    socket = socket |> apply_params(params, :workflow)
+    version_type =
+      if snapshot == nil or workflow.lock_version == snapshot.lock_version do
+        :workflow
+      else
+        :snapshot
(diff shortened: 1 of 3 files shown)


lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1866,7 +1877,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     ~H"""
     <div class="relative">
       <div
-        :if={@changeset.changes |> Enum.any?()}
+        :if={@version == "latest" && @changeset.changes |> Enum.any?()}
```
taylordowns2000, line 1880: Isn't this a double fix? (Like, in a bad way? If you're making it so that the changeset doesn't inaccurately think there are changes, why must you also only show the red dot when version != latest?)

elias-ba: Removed it

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -638,27 +637,54 @@ defmodule LightningWeb.WorkflowLive.Edit do
     ~H"""
     <div
       id={"version-switcher-toggle-wrapper-#{@id}"}
-      class="flex items-center justify-between"
+      phx-click="switch-version"
+      phx-value-type="toggle"
```
stuartc, line 641: @elias-ba could you let me know why we need to repeat this click event?

elias-ba: @stuartc sorry I don't understand your question. What click event are we repeating?

PR #2266, opened 2024-07-08
UI small fixes

### LINKED ISSUE #2212: Standardize button styles and behaviour
We need to standardize buttons, styles, and behavior on Lightning and Thunderbolt. This helps to deliver consistency and eliminate bounces/flashes experienced when buttons are pressed on the app. See below for an example of how varying save buttons cause a bounce effect. 

https://www.loom.com/share/58feb3af170a42fdbe70a127843bb7cb

oliveirarleo:
Hi guys, sorry for the delay, I ended up going on a trip as I am in my vacation.

## Validation Steps

- Simply open the pages for workflow, history and settings page, the projects menu will be hidden when there is no project assigned yet.
- In the workflow job editor page, there is no flickering.
reference for the original bug https://www.loom.com/share/58feb3af170a42fdbe70a127843bb7cb
- The save button still blinks, but does not change size, the user does not have time to read still.
- Some texts are standardized with `...` for the `phx-disable-with` text.
- It's unclear if other buttons are flickering, I am new to this project, and did not find other pages with the same problems, but reading the code they should exist.

## Notes for the reviewer

At the `lib/lightning_web/live/components/menu.ex`, github fails to diff in a good way. You can notice that the previous code is the same as in the `else` block, this should simplify the review.

## Related issue

Fixes #2212

## Review checklist

### DIFF
--- a/lib/lightning_web/live/components/menu.ex
+++ b/lib/lightning_web/live/components/menu.ex
@@ -57,74 +57,82 @@ defmodule LightningWeb.Components.Menu do
 
   def projects_dropdown(assigns) do
     ~H"""
-    <div class="relative my-4 mx-2 px-2">
-      <button
-        type="button"
-        class="relative w-full cursor-default rounded-md bg-white py-1.5 pl-3
+    <%= if length(assigns[:projects]) == 0 do %>
+      <div class="p-2 mb-4 mt-4 text-center text-primary-300 bg-primary-800">
+        <span class="inline-block align-middle text-sm">
+          You don't have access to any projects
+        </span>
+      </div>
+    <% else %>
+      <div class="relative my-4 mx-2 px-2">
+        <button
+          type="button"
+          class="relative w-full cursor-default rounded-md bg-white py-1.5 pl-3
         pr-10 text-left text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300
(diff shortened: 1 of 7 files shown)


lib/lightning_web/live/components/menu.ex
```diff
@@ -57,74 +57,82 @@ defmodule LightningWeb.Components.Menu do
 
   def projects_dropdown(assigns) do
     ~H"""
-    <div class="relative my-4 mx-2 px-2">
-      <button
-        type="button"
-        class="relative w-full cursor-default rounded-md bg-white py-1.5 pl-3
+    <%= if length(assigns[:projects]) == 0 do %>
```
stuartc, line 60: When accessing assigns in Heex templates, please use `@projects` - that way the renderer can optimise for diffs/prevent rerenderers.

The example below, `assigns[:selected_project]` is an exception to the rule since in that specific case we don't know if there is an assign with that key.

But for any other case, we always use the `@name` style.

christad92 (changes requested):
With fixed width implemented for the save button, the text is small so we have too much space on both sides of the "Save" CTA.  See image below: 
image

My recommendation is to set the CTA to `Save edit`. This will reduce the padding and remove the _flashes_.

taylordowns2000:
@oliveirarleo , thanks for this! i'm closing as there are un-addressed changes and it's been a while. please don't hestitate to reopen if this pops back up for you. and thank you for contributing!

PR #2268, opened 2024-07-09
Do Not Require Lock Version In URL Parameters

### LINKED ISSUE #2267: FunctionClauseError: no function clause matching in LightningWeb.WorkflowLive.Edit.apply_action/3
The `v` query parameter should not be required when viewing the workflow canvas.

See: `apply_action(socket, :edit, %{"id" => workflow_id, "v" => version})` - basically, we should not pattern match on _any_ query params because query params should always be optional. This issue only covers removing the pattern matching here on the workflow canvas.

Sentry Issue: [LIGHTNING-E9](https://openfn.sentry.io/issues/5539533467/?referrer=github_integration)

```
FunctionClauseError: no function clause matching in LightningWeb.WorkflowLive.Edit.apply_action/3
  File "lib/lightning_web/live/workflow_live/edit.ex", line 909, in LightningWeb.WorkflowLive.Edit.apply_action/3
  File "lib/lightning_web/live/workflow_live/edit.ex", line 891, in LightningWeb.WorkflowLive.Edit.handle_params/3
  File "lib/phoenix_live_view/utils.ex", line 462, in anonymous fn/5 in Phoenix.LiveView.Utils.call_handle_params!/5
  File "/app/deps/telemetry/src/telemetry.erl", line 321, in :telemetry.span/3
  File "lib/phoenix_live_view/static.ex", line 282, in Phoenix.LiveView.Static.call_mount_and_handle_params!/5
...
```

elias-ba:
## Related issue

Fixes #2267 

## Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Do Not Require Lock Version In URL Parameters
+  [#2267](https://github.com/OpenFn/lightning/pull/2267)
+
 ## [v2.7.4] 2024-07-06
 
 ### Changed
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -920,11 +920,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     |> assign(page_title: "New Workflow")
   end
 
-  def apply_action(
-        socket,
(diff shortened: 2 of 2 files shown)


taylordowns2000 (approved):

christad92 (approved):
Works for me.. 

We need to make sure that if the local version is not found, we default to latest. Will create an issue for this.

Issue #1978, opened 2024-04-10
Add plumbing to fetch messages from a Kafka cluster


rorymckinley:

christad92:
- [x] Restart the consumer group if the config changes (e.g. username/password) 
- [x] Generate consumer group IDs and make them immutable
- [ ] Handle failures in the Pipeline by logging errors with payload and sending errors to Sentry
- [ ] Clean up the message log used for deduplication by configuring the purge period in the ENV
- [ ] Allow users to be able to configure time-outs via the trigger setup
- [ ] Look at Ayodele's bug

PR #2270, opened 2024-07-09
Experimental Kafka Trigger Implementation

rorymckinley:
## Validation Steps

**Note: To run the server locally the `KAFKA_TRIGGERS_ENABLED` ENV var must be set to `yes`. However, if you want
to run the tests, you need to set this to 'no' (it defaults to 'no', so you should only need to remove/comment out the ENV var).**

Either set up a [local Kafka cluster](https://github.com/OpenFn/lightning/compare/1978-kafka-review-1?expand=1#diff-dc7a76b8847f7134929ea2271954cb40787e1d7e3266256a3eeb5702da17e97bR204) or set up a free trial account on [Confluent Cloud](https://github.com/OpenFn/lightning/compare/1978-kafka-review-1?expand=1#diff-dc7a76b8847f7134929ea2271954cb40787e1d7e3266256a3eeb5702da17e97bR330).

Create a new workflow via the UI, selecting the Kafka trigger type. 

Publish a few messages to the Kafka cluster of your choice using [kcat/kafkacat](https://github.com/OpenFn/lightning/compare/1978-kafka-review-1?expand=1#diff-dc7a76b8847f7134929ea2271954cb40787e1d7e3266256a3eeb5702da17e97bR252) or via the Confluent Cloud UI (in both cases you will need to create a topic first).

After a few minutes, you should see the messages showing up in the run history.

Full testing information can be found in kafka_testing/KAFKA_ARCHITECTURE.md.

## Notes for the reviewer

I am creating this PR early due to the size, even though I still need to add some finishing touches:

Please refer to `kafka_testing/KAFKA_ARCHITECTURE.md` for a high-level overview of all the moving parts.

Excluding the UI changes, the Kafka functionality can be broken up into two functional areas:

- Getting Messages from the Kafka Cluster - in this area the primary modules in terms of fucntionality  are `KafkaTriggers`, `Workflows.Triggers.KafkaConfiguration`, `KafkaTriggers.Pipeline`, `KafkaTriggers.EventListener` and `KafkaTriggers.DuplicatetrackingCleanupWorker`.

- Handling received messages with the emphasis on preserving message sequence - the primary modules are 
`KafkaTriggers.MessageCandidateSetServer`, `KafkaTriggers.MessageCandidateSetWorker`, `KafkaTriggers.MessageHandling`.

### Caveats

- Adding some additional test coverage
- Implementing some crude feature flags so that we can disable the Kafka functionality at least until load testing is done.
- A possible bug fix for a Db connection issue that may be environmental.

I will add the above changes as new commits to make tracking easier.

I have also not implemented the Cachex solution as part of this PR for two reasons:

- I think what we have may be good enough for initial testing, and so I want to take my time with the Cachex solution rather than rushing it in so that it is ready for initial testing.
- The functionality that would rely on Cachex may end up being cut back considerably as there is (to my understanding) currently some debate around preserving message sequence and whether it should be the responsibility of the trigger.

I will update the Changelog when the R is ready to be merged in.

## Related issue

#1978 

## Review checklist


config/dev.exs
```diff
@@ -138,3 +138,5 @@ end
 config :lightning, :is_resettable_demo, true
 
 config :lightning, :apollo, endpoint: "http://localhost:3000"
+
+config :lightning, :kafka_triggers, run_supervisors: true
```
rorymckinley, line 142: @jyeshe I initially added this so that I could stop the Kafka trigger supervisors running in the test env. I was going to ask if there is a more idiomatic way to do this, but for now I am going to use this config to apply a ENV-driven feature flag so that the initial deploy to prod can have Kafka trigger functionality disabled.

lib/lightning/kafka_triggers/message_handling.ex
```diff
@@ -0,0 +1,151 @@
...
+    from t in TriggerKafkaMessage,
+      where: t.trigger_id == ^trigger_id and t.topic == ^topic and t.key == ^key,
+      order_by: t.offset,
+      limit: 1,
+      preload: [:work_order, trigger: [:workflow]]
+  end
+
+  # Stolen from the `webhooks_controller.ex` file and simplified until
+  # I understand the details (ask Roger).
+  defp check_skip_run_creation?(project_id) do
```
rorymckinley, line 139: @jyeshe As the comment says, I was unclear wrt existing cases if the above is an acceptable implementation?

lib/lightning/kafka_triggers/trigger_kafka_message.ex
```diff
@@ -0,0 +1,56 @@
...
+      :message_timestamp,
+      :metadata,
+      :offset,
+      :data
+    ]
+
+    message
+    |> cast(changes, cast_changes)
+    |> validate_required(required_changes)
+    # No test for this - test it
```
rorymckinley, line 53: @jyeshe I still need to add a test for the assoc_constraint

lib/lightning/runs.ex
```diff
@@ -133,14 +133,15 @@ defmodule Lightning.Runs do
...
+  # TODO Add test for kafka coverage
   @spec get_input(Run.t()) :: String.t() | nil
   def get_input(%Run{} = run) do
     from(d in Ecto.assoc(run, :dataclip),
       select:
         type(
           fragment(
             """
-            CASE WHEN type = 'http_request'
+            CASE WHEN type IN ('http_request', 'kafka')
```
rorymckinley, line 144: @jyeshe Need to add a test here - coming soon!

lib/lightning/vault.ex
```diff
@@ -85,4 +85,20 @@ end
 defmodule Lightning.Encrypted.Binary do
   @moduledoc false
   use Cloak.Ecto.Binary, vault: Lightning.Vault
+
+  # From https://github.com/danielberkompas/cloak/issues/84
+  # TODO How do I test this?
+  def embed_as(_format), do: :dump
```
rorymckinley, line 91: @jyeshe Still need to add a test here.

test/lightning/kafka_triggers_test.exs
```diff
@@ -0,0 +1,843 @@
...
+          enabled: true
+        )
+
+      expected_child_spec = child_spec(trigger: trigger, index: 1, sasl: false)
+      actual_child_spec = KafkaTriggers.generate_pipeline_child_spec(trigger)
+
+      assert actual_child_spec == expected_child_spec
+    end
+
+    # TODO merge with other configuration method
```
rorymckinley, line 470: @jyeshe I will still do this cleanup.

lib/lightning/kafka_triggers/pipeline_supervisor.ex
```diff
@@ -0,0 +1,16 @@
...
+  """
+  use Supervisor
+
+  def start_link(_opts) do
+    Supervisor.start_link(__MODULE__, [], name: :kafka_pipeline_supervisor)
+  end
+
+  @impl true
+  def init(_opts) do
+    Supervisor.init([], strategy: :one_for_one)
```
jyeshe, line 14: Considering this empty list, for children only started dynamically the `DynamicSupervisor` has the benefit of handling multiple pipelines more efficiently or, if there will be always a few number of kafka triggers, at least it handles shutdown more quickly (concurrently). Also using the `DynamicSupervisor` from the beginning would prevent the need to measure when it becomes more efficient on start up time and resource consumption.

lib/lightning/kafka_triggers/message_handling.ex
jyeshe: Nice finding Rory. Here the error `:too_many_runs` needs to be differentiated from `:runs_hard_limit`. With the former, Lightning still accepts the creation of workorders and does it without runs (the runs are blocked) while the with latter error no workorder is created.

lib/lightning/kafka_triggers/pipeline.ex
```diff
@@ -0,0 +1,212 @@
...
+
+    trigger_changeset =
+      Trigger
+      |> Repo.get(trigger_id |> Atom.to_string())
+      |> Trigger.kafka_partitions_changeset(partition, timestamp)
+
+    Multi.new()
+    |> Multi.insert(:record, record_changeset)
+    |> Multi.insert(:message, message_changeset)
+    |> Multi.update(:trigger, trigger_changeset)
```
jyeshe, line 88: I would prefer that the trigger object would be a result only from the UI, from user input, but I understand the straightforward approach to centralize the config in one place.

priv/repo/migrations/20240609092546_add_errors_to_trigger_kafka_messages.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Lightning.Repo.Migrations.AddProcessingDataToTriggerKafkaMessages do
+  use Ecto.Migration
+
+  def change do
+    alter table(:trigger_kafka_messages) do
+      add :processing_data, :map, default: %{}
+    end
+  end
+end
```
jyeshe, line 9: better to move the change to the `CreateTriggerKafkaMessages` to have a single file for `:trigger_kafka_messages`.

lib/lightning/kafka_triggers/pipeline.ex
```diff
@@ -0,0 +1,212 @@
...
+        default: [
+          concurrency: 10
+        ]
+      ],
+      batchers: []
+    )
+  end
+
+  @impl true
+  def handle_message(_processor, message, context) do
```
jyeshe, line 42: It might have been a good approach to start with, but for 100 messages with `handle_batch` we can reduce the amount of inserts and updates from 300 to 3. Or if there is no limit for the message size, we can decrease the batch size to 10 and leverage the `Multi.insert_all`.

jyeshe (commented):
Rory, great job. A part from the query we've discussed, these are some points that might be improved later or based on benchmarking results.

lib/lightning/kafka_triggers/message_handling.ex
rorymckinley: Thanks @jyeshe - I have updated as follows: https://github.com/OpenFn/lightning/pull/2270/files#diff-335ea25efd1cf677f88d6d071a096f13624ce535843b3190a6d537a6e56821d3R145.

lib/lightning/kafka_triggers/pipeline.ex
rorymckinley: @jyeshe Noted - the goal of this pass was to keep the footprint as small as possible, but there are already plans to  separate the cluster config from the consumer config and we can certainly extend that to have the tracking of partition timestamps be a new table as well. That way, we can isolate the operational updates.

priv/repo/migrations/20240609092546_add_errors_to_trigger_kafka_messages.exs
rorymckinley: @jyeshe Unfortunately, people other than you and I will already have run this migration, so I would prefer to just keep it in place so that they don't end up with an orphaned migration and no idea where it came from.

lib/lightning/kafka_triggers/supervisor.ex
```diff
@@ -0,0 +1,36 @@
...
+          {
+            Lightning.KafkaTriggers.MessageCandidateSetSupervisor,
+            type: :supervisor
+          },
+          {
+            Lightning.KafkaTriggers.PipelineSupervisor,
+            type: :supervisor
+          },
+          {Task, &Lightning.KafkaTriggers.start_triggers/0},
+          Lightning.KafkaTriggers.EventListener
```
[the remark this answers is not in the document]
rorymckinley: @jyeshe done.

lib/lightning/kafka_triggers/pipeline_supervisor.ex
jyeshe: Hi @rorymckinley, the child specs (for GenServer ids) are accepted also by the `DynamicSupervisor`s.

priv/repo/migrations/20240609092546_add_errors_to_trigger_kafka_messages.exs
jyeshe: As we are closing a version for staging this shouldn't be a concern but it's okay.

jyeshe:
@rorymckinley while I finish some tests could you see if there is a quick change for the input field ids on the Kafka trigger UI? The Chrome browser is considering as the login, suggesting the same values, and if you click on it the cursor doesn't appear:
Screenshot from 2024-07-15 14-29-53
It allows editing only if I start typing (without seeing the cursor in advance)

jyeshe (commented):
@rorymckinley this would be I think the main fix for now. Adding an error handling for non map messages (to discard them and unblock the processing of following messages):
```
[error] GenServer #PID<0.3195.0> terminating
** (MatchError) no match of right hand side value: {:error, #Ecto.Changeset<action: :insert, changes: %{type: :kafka, request: %{"headers" => [], "key" => "", "offset" => 6, "partition" => 1, "topic" => "some-jembi-topic", "ts" => 1721048733296}, body: "Hello World !", project_id: "4adf2644-ed4e-4f97-a24c-ab35b3cb1efa"}, errors: [body: {"must be a map", []}], data: #Lightning.Invocation.Dataclip<>, valid?: false>}   
```

About the scenario commented by @christad92 , I couldn't reproduce the failure on resuming the ingestion after a restart while adding few messages to the broker. On some load testing, this would be covered by the "rewind" to a certain offset that we've discussed and that is yet to be implemented.

rorymckinley:
> @rorymckinley this would be I think the main fix for now. Adding an error handling for non map messages (to discard them and unblock the processing of following messages):
> 
> ```
> [error] GenServer #PID<0.3195.0> terminating
> ** (MatchError) no match of right hand side value: {:error, #Ecto.Changeset<action: :insert, changes: %{type: :kafka, request: %{"headers" => [], "key" => "", "offset" => 6, "partition" => 1, "topic" => "some-jembi-topic", "ts" => 1721048733296}, body: "Hello World !", project_id: "4adf2644-ed4e-4f97-a24c-ab35b3cb1efa"}, errors: [body: {"must be a map", []}], data: #Lightning.Invocation.Dataclip<>, valid?: false>}   
> ```
> 
> About the scenario commented by @christad92 , I couldn't reproduce the failure on resuming the ingestion after a restart while adding few messages to the broker. On some load testing, this would be covered by the "rewind" to a certain offset that we've discussed and that is yet to be implemented.

@jyeshe Ah, a sneaky string encoded as JSON. This should be fixed - for now, I am treating it in the same way as I treat an error that results from a body that can't be deserialized: The TriggerKafkaMessage has errors appended to it, but it will block the sequence for the given key until this is corrected.

rorymckinley:
> @rorymckinley while I finish some tests could you see if there is a quick change for the input field ids on the Kafka trigger UI? The Chrome browser is considering as the login, suggesting the same values, and if you click on it the cursor doesn't appear: Screenshot from 2024-07-15 14-29-53 It allows editing only if I start typing (without seeing the cursor in advance)

@jyeshe Fixed in the blind as I could not reproduce on my Chrome (probably settings differences) - can you please take a look and see if this resolves things? I used a bit of a hack on the username, which is not great - another thing to potentially clean up.

jyeshe:
Both fixes working beautifully @rorymckinley !

jyeshe (approved):
Great to see this coming into play @rorymckinley !

The tech debts we've identified on this first experimental version are:
- Use `DynamicSupervisor` for improved teardown
- Cache the messages to avoid the full projection query (without indexation and `limit`)
- Handle Broadway processor failure by rewinding the offset
- Use `handle_batch` instead of `handle_message` for higher throughput and simpler offset rewinding.

christad92 (approved):

PR #2281, opened 2024-07-11
no need to map before uniq

taylordowns2000:


taylordowns2000:
ah my bad, i see @rorymckinley that you've got a new interface for this thing and it now expects only the IDs. let's just leave it as is 
image

PR #2293, opened 2024-07-12
Enhance OAuth Credential Validation via Changeset

taylordowns2000:
@elias-ba , for your consideration: Is there _any situation_ in which we would want to save a credential with an `oauth` schema to the DB if it does not have a `refresh_token` in the body?


elias-ba (approved):
@taylordowns2000 thanks a lot for this idea and the PR. I improved it by ensuring all required fields are present before saving to the database, specifically checking for the presence of access_token, refresh_token, and either expires_in or expires_at. Additionally, I implemented support for both string and atom keys for flexibility. I also added / edited tests to cover various scenarios, ensuring the validation works as expected. I believe this can go now to main as soon as you or @christad92 validates the UI.

midigofrank (approved):
Great job!

PR #2294, opened 2024-07-12
Allow Revocation Of OAuth Tokens

### LINKED ISSUE #2292: Make use of the revocation endpoint in the Oauth2 spec
Lots of Oauth2 servers provide a `revocation_endpoint`:

- https://login.salesforce.com/.well-known/openid-configuration
- https://accounts.google.com/.well-known/openid-configuration

We should:

- [x] Add this to the generic oauth2 client setup screen
- [x] Call this _before_ getting a new token when the user clicks "reauthorize"
- [x] Don't allow the user to reauthorize [when we've confirmed that everything works via `userinfo`](https://openfn.slack.com/archives/C05KZNPEJFN/p1720807588644119)
- [x] Allow the user to reauthorize [if `userinfo` isn't provided](https://openfn.slack.com/archives/C05KZNPEJFN/p1720807588644119) and we can't tell if things are set up properly
- [x] Call this when the user deletes the credential

elias-ba:
### Description

This PR adds support for the `revocation_endpoint` - a required part of the Oauth2 specification.

Closes #2292 

### Pre-submission checklist


lib/lightning_web/live/components/oauth.ex
```diff
@@ -98,7 +98,7 @@ defmodule LightningWeb.Components.Oauth do
   def success_message(assigns) do
     ~H"""
     <div class="text-sm">
-      <Heroicons.check_circle class="h-4 w-4 text-green-600 inline-block" />
+      <.icon name="hero-check-circle" class="h-4 w-4 text-green-600 inline-block" />
```
taylordowns2000, line 101: @elias-ba 🤣

taylordowns2000 (changes requested):
This is fantastic @elias-ba . Really great work. I have added the last bit (handling credential delete) and made some language changes. (Notably, we should do our best to _actually follow_ the Oauth2 specification. The field is called `revocation_endpoint`, not `revoke_endpoint` so I've made that adjustment.

Can you implement the tests and re-request review? Thanks!

christad92 (commented):
@elias-ba Amazing work. 

I have spent some time looking at OpenID configurations from different providers and I see that revocation URL is not adopted across board and might be confusing for the users. 

I have seen Microsoft and DHIS2 call it end_session_endpoint whilst some others refer to it as logout_endpoint. 

I'll suggest that we add a subtext to the title that says: "Also known as end_session_endpoint or logout_endpoint

See links below: 
[DHIS/OKTA](https://docs.dhis2.org/en/topics/tutorials/configure-oidc-with-okta.html#:~:text=oidc.provider.okta.end_session_endpoint%20%3D%20https%3A//%7Bclient_domain%7D/oauth2/v1/logout)

[Microsoft](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration#:~:text=%22end_session_endpoint%22%3A%20%22https%3A//login.microsoftonline.com/common/oauth2/v2.0/logout%22%2C)

There are also platforms like Github and Facebook that do not provide revocation URL but would have us make a DELETE request to a delete token URL. Do we make this field optional and have the fallback of telling the user to manually revoke permission if required?

cc: @taylordowns2000

taylordowns2000:
Hey @christad92 , I don't think we support OpenID. Do we?

These endpoints are all part of the required OAuth2 specification. (Not OpenID.)

Sorry if I'm misunderstanding something here.

christad92:
My point is some of these applications don't have this endpoint specified as "revocation endpoint" and it's called "end_session_endpoint" or "logout_endpoint".

Thats why I was suggesting that we have a subtext that communicates these options to users when revocation endpoint is not specified in the wellknown file.

taylordowns2000:
Hrmmm ok. Let's talk about it on Monday. Phrases like "end session" and "logout" make me think they do something different--is it definitely the same functionality? (We want to revoke a refresh token, not to revoke a session token or "log out" once logged in.)

Also see question above please. Can you find out if we support OpenID? Looks like the DHIS2 link is for OpenID. Is that a different protocol?

Excited to learn about this stuff next week.

lib/lightning_web/live/components/oauth.ex
elias-ba: 😆 thanks @taylordowns2000 !

midigofrank (approved):
Nicely done @elias-ba

christad92 (dismissed):

christad92 (commented):
I am still not certain about revocation URL. I have not seen it in any other specs yet.

christad92 (approved):

taylordowns2000:
@elias-ba please rebase (or resolve conflicts). this appears to be all green and good to go once that is done

christad92 (commented):
A few things have changed since the last review: 
- We should not show reauthorize if the user does not provide a revocation URL. We rather tell them to revoke existing access and reauthenticate from the third party app website/portal. 
- The re-authorize URL is set to /credential#
- When a user tries to add a new scope after an initial access, they are unable to remove the scope. So the user is forced to start authorization all over. They should be able to uncheck the box and retain existing auth. 

See video below: 
https://www.loom.com/share/4ddbefaf33364a75a5ab07001e9e3903?sid=68a8c360-4edd-415e-b195-599f3a739dab
image

taylordowns2000:
Hey @christad92 , shoot me some test data to use for an oauth client please (on slack dm) and I'll wrap this up on Monday!

.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "[elixir]": {
     "editor.trimAutoWhitespace": true,
-    "editor.defaultFormatter": "JakeBecker.elixir-ls"
+    "editor.defaultFormatter": "animus-coop.vscode-elixir-mix-formatter"
```
taylordowns2000, line 4: removing this!

PR #2305, opened 2024-07-18
Fix Work Order URL In Failure Alerts

elias-ba:
### Description

This PR fixes the broken work order URL that is sent in the failure alert emails.

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Fix work order URL in failure alerts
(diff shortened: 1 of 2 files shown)


taylordowns2000 (changes requested):
is there a reason that this implementation is different from the one on line `49`? (assuming yes, and it's _better_, please update line `49` to use the same style.)

if we should actively be supporting these different styles, please explain why in the comment and i'll withdraw my change request.

thank you @elias-ba !

taylordowns2000 (approved):

christad92 (approved):
Great. Validated that links work with History and Run views opened.

PR #2318, opened 2024-07-22
Restore Encrypted.Binary behaviour

### LINKED ISSUE #2316: New behaviour of Encrypted.Binary is causing issues with existing data
To support the embedded KafkaConfiguration,  I changed the [Encrypted.Binary](https://github.com/OpenFn/lightning/blob/main/lib/lightning/encrypted/binary.ex) behaviour to do Base64 encoding to allow embedding of encrypted data. This is now causing issues with existing data which has not (duh!) been base 64-encoded.

rorymckinley:
### Description

This PR fixes an issue caused when interacting with existing data encrypted using Ecto.Cloak

Closes #2316 

### Validation steps

drop your local DB.

```
git checkout 68958bbefdf102dccf6a6ebababe1f7711d8c7d9
mix ecto.migrate
```

Start Lightning locally.
Create a Webhook trigger with authentication.

From an IEx session confirm that `Lightning.Workflows.WebhookAuthMethod |> Lightning.Repo.all()` returns a result.

```
git checkout 1dd5450bfafacc746ad16d5ac0befce83c0a68b5
mix ecto.migrate
```

From an IEx session confirm that `Lightning.Workflows.WebhookAuthMethod |> Lightning.Repo.all()` produces  an '(ArgumentError) non-alphabet character found: "\x01" (byte 1)' error,

```
git checkout 2316-correct-encrypted-binary-behaviour
mix ecto.migrate
```

From an IEx session confirm that `Lightning.Workflows.WebhookAuthMethod |> Lightning.Repo.all()` returns a result and not an error.

I have confirmed that I the encrypted credentials for a Kafka trigger still work after this change, but please let me know if you would like to confirm this for yourself.

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to
 
 - Fix work order URL in failure alerts
   [#2305](https://github.com/OpenFn/lightning/pull/2305)
+- Fix error when handling existing encrypted credentials
+  [#2316](https://github.com/OpenFn/lightning/issues/2316)
 
 ## [v2.7.7] - 2024-07-18
 
--- a/lib/lightning/encrypted/binary.ex
+++ b/lib/lightning/encrypted/binary.ex
@@ -1,19 +1,4 @@
 defmodule Lightning.Encrypted.Binary do
   @moduledoc false
   use Cloak.Ecto.Binary, vault: Lightning.Vault
-
-  # From https://github.com/danielberkompas/cloak/issues/84
-  def embed_as(_format), do: :dump
-
-  def dump(nil), do: super(nil)
-
-  def dump(value) do
-    with {:ok, encrypted} <- super(value) do
-      {:ok, Base.encode64(encrypted)}
-    end
-  end
-
-  def load(nil), do: super(nil)
-
-  def load(value), do: super(Base.decode64!(value))
 end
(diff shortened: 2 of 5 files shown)


jyeshe (changes requested):
@rorymckinley for now I would apply the fix only to the kafka configuration because the `Lightning.Encrypted.Binary` is used somewhere else.

I don't see any difference on the `Lightning.Encrypted.EmbeddedBinary` (it's like the current `Lightning.Encrypted.Binary`). A commit missing to be pushed?

rorymckinley:
@jyeshe Apologies for not explaining properly,: 

`Encrypted.Binary` was originally in `Lightning.Vault` and looked like [this](https://github.com/OpenFn/lightning/commit/192eeba88e98b1d14dedc83d198ccb1cc305fc1f#diff-db9970fcdc03ec639fe2df14ecdb80a3f6227a6de60fdab8a52309726796a9d8R84).

I then added custom `load` and `dump` behaviour and moved `Encrypted.Binary` to the current location. The `load` and `dump` were needed to allow encryption of fields within an embedded schema.

This change is about restoring `Encrypted.Binary` to the way it was before I updated it for use by `KafkaConfiguration`. I removed the custom methods mentioned above so that `Encrypted.Binary` will, once again, use the default `load` and `dump` behaviour that is provided by Ecto.Cloak . You are correct thatother modules are using this and they should now be able to interact with existing data again.

 As I need to retain the custom `load` and `dump` behaviour for `KafkaConfiguration` I created a new module named `Encrypted.EmbeddedBinary` which (as you noted) has all the functionality that is currently in `Encrypted.Binary`. The only schema that currently makes use of this is `KafkaConfiguration` - I have left any other users of `Encrypted.Binary` unchanged.

jyeshe (approved):
Thanks for explaining about the revert @rorymckinley

PR #2319, opened 2024-07-22
Fix job editor switches to the snapshot version when body is changed

### LINKED ISSUE #2306: Editor switches to the snapshot/readonly mode without user toggling snapshot mode
When inspecting a job in snapshot mode I switched to the latest version to edit the job code. After typing a few characters, the editor switches to the snapshot version and read-only mode. When this happens, the snapshot version, toggle, and snapshot ID chip still indicate that the "latest" version is selected.

https://www.loom.com/share/22a7de9164354895b0b489f174080aa9

### LINKED ISSUE #2308: Misaligned "Retry from here" button on inspector page
There is a regression on the inspector page where there is a mismatch between the heights of the button and the dropdown as shown below.  
image

midigofrank:
### Description

This PR fixes the bug on the inspector that switches the job editor back to the snapshot version whenever the user types anything in. You can check the issue #2306 for a video on this

Closes #2306 
Closes #2308 

### Validation steps
In order to validate this, we'll need to have a workflow with at least one job in it. 
We'll create a manual work order using the job as it is, then later update the job body to have something else.
This way, we'll have a work order with an older version of the job body.

**These instructions assume you've done the the above**

1. Open up the inspector from the executed step of the work order in the history page
2. The inspector will open up in "snapshot mode". Now toggle to switch to the latest version
3. Notice that the job editor content gets updated to the latest job body.
4. If you try changing the job body, the content gets updated and the editor doesn't switch back to the snapshot version

### Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/editor_pane.ex
+++ b/lib/lightning_web/live/workflow_live/editor_pane.ex
@@ -78,6 +78,8 @@ defmodule LightningWeb.WorkflowLive.EditorPane do
     {:noreply, socket}
   end
 
+  # NOTE: This is dead code and should probably be removed. All events are
+  # being handled by 'push-change' in the parent liveview
   def handle_event("job_body_changed", %{"source" => source}, socket) do
     params =
       {socket.assigns.form[:body].name, source}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to
 
 - Fix work order URL in failure alerts
   [#2305](https://github.com/OpenFn/lightning/pull/2305)
+- Fix job editor switches to the snapshot version when body is changed
+  [#2306](https://github.com/OpenFn/lightning/issues/2306)
+- Fix misaligned "Retry from here" button on inspector page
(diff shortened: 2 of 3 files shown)


lib/lightning_web/live/workflow_live/editor_pane.ex
```diff
@@ -78,6 +78,8 @@ defmodule LightningWeb.WorkflowLive.EditorPane do
     {:noreply, socket}
   end
 
+  # NOTE: This is dead code and should probably be removed. All events are
```
elias-ba, line 81: Maybe we should remove it in this work, what do you think ?

elias-ba (approved):
@midigofrank this is looking good to me, great job 👏. I left a comment but up to you to handle it or not.

lib/lightning_web/live/workflow_live/editor_pane.ex
midigofrank: I don't feel confident enough to remove it at this time. I was thinking of removing it as a follow-up when everything is tested and works fine

christad92 (approved):
Loooooveeee eeeeeett!! 

Thanks mate!

Issue #1588, opened 2023-12-22
Don't render modals instead of hiding them


elias-ba:
Currently this is how we implement modals in the pages that list out resources (mind the `live_component` call in the loop)
```elixir
      <%= for credential <- @credentials do %>
        <.tr id={"credential-#{credential.id}"}>
          <.td class={if credential.scheduled_deletion, do: "line-through"}>
            <%= credential.name %>
          </.td>
          <.td>
            <%= credential.project_names %>
          </.td>
          <.td class={if credential.scheduled_deletion, do: "line-through"}>
            <%= credential.schema %>
          </.td>
          <.td>
            <%= if credential.production do %>
              <div class="flex">
                <Heroicons.exclamation_triangle class="w-5 h-5 text-secondary-500" />
                &nbsp;Production
              </div>
            <% end %>
          </.td>
          <.td>
            <%= if !credential.scheduled_deletion do %>
              <span>
                <.link phx-click={
                  show_modal("edit-credential-#{credential.id}-modal")
                }>
                  Edit
                </.link>
              </span>
              |
            <% end %>
            <.delete_action socket={@socket} credential={credential} />
            <.live_component
              id={"edit-credential-#{credential.id}-modal"}
              module={LightningWeb.CredentialLive.FormComponent}
              action={:edit}
              credential_type={@selected_credential_type}
              credential={credential}
              current_user={@current_user}
              projects={@projects}
              project={nil}
              show_project_credentials={true}
              can_create_project_credential={true}
              return_to={~p"/credentials"}
            />
          </.td>
        </.tr>
      <% end %>
```

This is not good at all and needs to be changed. Currently clicking on the cancel or close button of a modal, hides it using the `hidden` class of tailwind. That doesn't unmount it in the DOM. We should unmount it and not just hide it.

cc @stuartc

taylordowns2000:
Hey @elias-ba and @stuartc , I wonder if we should prioritize this for next week. The funniest behavior I'm seeing is that when a user goes to the credentials list view, it seems that `get_userinfo` is called for _all_ credentials on the screen, even though none of them have been opened. This is also related to https://github.com/OpenFn/lightning/issues/1578

taylordowns2000:
@stuartc notes that the `ModalPortal` might be a promising solution here

taylordowns2000:
@stuartc , @midigofrank - had a good call with Frank and confirmed that this is not yet done. I'm reopening, but noting that:

[...]

PR #2321, opened 2024-07-23
Apply the Bob Marley rule...

taylordowns2000:
🎵 _You can render some modals sometime, but you can't render all the modals all the time._ 🎶  

for #1588 - we'd need to do something like this all over the app, but this starts by taking on the place that's messing with us the most, the credentials list view. right now, when you go to /credentials code the checks and refreshes tokens is executed for all credentials simultaneously. (they're all rendered, even though they're not visible.)


taylordowns2000:
@elias-ba , @theroinaochieng , do we still have the "rendering all the modals all at the same time" problem?

I think it's been resolved, and that Elias used this draft PR as a concept, but that it should now be closed. Please feel free to re-open if I'm wrong!

PR #2326, opened 2024-07-24
2323 better nil key handling

### LINKED ISSUE #2323: Optimise handling of Kafka messages without a key
As Kafka only guarantees it will preserve the sequence of messages that have a key, we can introduce a more lightweight way of handling messages that do not have a key which should improve message handling performance and also prevent failures on unrelated keys from blocking message processing.

rorymckinley:
### Description

The test data that is being received as part of load testing the kafka cluster does not have message keys. This means that the existing message handling of grouping by trigger, topic and key is horribly inefficient.

This is a very crude band-aid that is being implemented to treat messages without keys differently so that load testing can continue. Once load-testing has completed, the nil-key message handling must be refactored to allow for distinct configuration in terms of processing delays as well as to remove duplication between the two code paths.

Closes #2323 

### Validation steps

Note: To run the server locally the KAFKA_TRIGGERS_ENABLED ENV var must be set to yes. However, if you want
to run the tests, you need to set this to 'no' (it defaults to 'no', so you should only need to remove/comment out the ENV var).

Setup a Kafka trigger (via the UI) pointing at either a [local](https://github.com/OpenFn/lightning/blob/main/kafka_testing/KAFKA_ARCHITECTURE.md#testing-locally-using-the-docker-kafka-cluster) or [remote](https://github.com/OpenFn/lightning/blob/main/kafka_testing/KAFKA_ARCHITECTURE.md#testing-using-a-confluent-cloud-instance) Kafka cluster. When setting up the trigger, set the 'Initial offset reset policy' to `:earliest`.

Publish a few messages using `kcat` aka `kafkacat`. You will want to try messages with and without keys.

With a key (using the local cluster, assuming you have a topic named `baz_topic`):

```
cat message.json | kcat -P -b 127.0.0.1:9094 -t baz_topic -k 202407211
```
Without a key:
```
cat other_message.json | kcat -P -b 127.0.0.1:9094 -t baz_topic
```

Confirm that the messages with keys and without keys were received (they may take a few minutes to show up).

Full testing information can be found in kafka_testing/KAFKA_ARCHITECTURE.md.

### Pre-submission checklist

### DIFF
--- a/lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
+++ b/lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
@@ -21,36 +21,24 @@ defmodule Lightning.KafkaTriggers.MessageCandidateSetSupervisor do
     number_of_workers = Keyword.get(opts, :number_of_workers, 1)
 
     mcs_children =
-      generate_child_specs(:message_candidate_set, number_of_workers)
+      generate_child_specs(MessageCandidateSetServer, number_of_workers)
 
     message_children =
-      generate_child_specs(:message, number_of_workers)
+      generate_child_specs(MessageServer, number_of_workers)
 
     Supervisor.init(mcs_children ++ message_children, strategy: :one_for_one)
   end
 
-  def generate_worker_specs(number_of_workers) do
-    no_set_delay =
-      Application.get_env(:lightning, :kafka_triggers)[
-        :no_message_candidate_set_delay_milliseconds
-      ]
+  def generate_child_specs(server, number_of_workers) do
+    {worker, id_prefix} =
+      case server do
+        MessageCandidateSetServer ->
+          {MessageCandidateSetWorker, "mcs_worker"}
 
-    next_set_delay =
-      Application.get_env(:lightning, :kafka_triggers)[
-        :next_message_candidate_set_delay_milliseconds
-      ]
+        MessageServer ->
+          {MessageWorker, "message_worker"}
+      end
 
-    0..(number_of_workers - 1)
-    |> Enum.map(fn index ->
-      {
-        MessageCandidateSetWorker,
-        [no_set_delay: no_set_delay, next_set_delay: next_set_delay]
-      }
-      |> Supervisor.child_spec(id: "mcs_worker_#{index}")
-    end)
-  end
-
-  def generate_child_specs(:message_candidate_set, number_of_workers) do
     no_set_delay =
       Application.get_env(:lightning, :kafka_triggers)[
         :no_message_candidate_set_delay_milliseconds
@@ -65,36 +53,12 @@ defmodule Lightning.KafkaTriggers.MessageCandidateSetSupervisor do
       0..(number_of_workers - 1)
       |> Enum.map(fn index ->
         {
-          MessageCandidateSetWorker,
+          worker,
           [no_set_delay: no_set_delay, next_set_delay: next_set_delay]
         }
-        |> Supervisor.child_spec(id: "mcs_worker_#{index}")
+        |> Supervisor.child_spec(id: "#{id_prefix}_#{index}")
       end)
 
-    [MessageCandidateSetServer] ++ workers
-  end
-
-  def generate_child_specs(:message, number_of_workers) do
-    no_set_delay =
-      Application.get_env(:lightning, :kafka_triggers)[
-        :no_message_candidate_set_delay_milliseconds
-      ]
-
-    next_set_delay =
-      Application.get_env(:lightning, :kafka_triggers)[
-        :next_message_candidate_set_delay_milliseconds
-      ]
-
-    workers =
-      0..(number_of_workers - 1)
-      |> Enum.map(fn index ->
-        {
-          MessageWorker,
-          [no_set_delay: no_set_delay, next_set_delay: next_set_delay]
-        }
-        |> Supervisor.child_spec(id: "message_worker_#{index}")
-      end)
-
-    [MessageServer] ++ workers
+    [server | workers]
   end
 end
--- a/lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
+++ b/lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
@@ -1,14 +1,16 @@
 defmodule Lightning.KafkaTriggers.MessageCandidateSetSupervisor do
   @moduledoc """
   Starts the server and worker processes responsible for converting messages
-  received from Kafka clusters. The sole purpose of this is to ensure that
-  messages with the same key (for a given cluster/topic configuration) are 
-  processed in the same order they were received.
+  received from Kafka clusters. There are two sets of workers and servers. This
+  is to accommodate messages that have keys (grouped into MessageCandidateSets) 
+  ans those that do not, which are processed individually.
   """
   use Supervisor
 
   alias Lightning.KafkaTriggers.MessageCandidateSetServer
   alias Lightning.KafkaTriggers.MessageCandidateSetWorker
+  alias Lightning.KafkaTriggers.MessageServer
+  alias Lightning.KafkaTriggers.MessageWorker
 
   def start_link(opts) do
     Supervisor.start_link(__MODULE__, opts, name: __MODULE__)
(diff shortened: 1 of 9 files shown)


lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
```diff
@@ -44,4 +49,52 @@ defmodule Lightning.KafkaTriggers.MessageCandidateSetSupervisor do
...
+      0..(number_of_workers - 1)
+      |> Enum.map(fn index ->
+        {
+          MessageWorker,
+          [no_set_delay: no_set_delay, next_set_delay: next_set_delay]
+        }
+        |> Supervisor.child_spec(id: "message_worker_#{index}")
+      end)
+
+    [MessageServer] ++ workers
```
jyeshe, line 98: It is okay but `[MessageServer | workers]` is more idiomatic for head and tail

lib/lightning/kafka_triggers/message_server.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  @impl true
+  def handle_call(:next_message, _from, current_messages) do
+    {message, remaining_messages} = pop_message(current_messages)
+
+    {:reply, message, remaining_messages}
+  end
+
+  defp pop_message([]) do
+    case MessageHandling.find_nil_key_message_ids() do
```
jyeshe, line 31: Can we have this function defined in this module? It is only used here and makes the code much easier to review. Also I appreciate your concern with making functions more testable but here even if it's private one can test it by calling `MessageServer.next_message()` or even directly 
`MessageServer.handle_call(:next_message, pid(), [])` if the module state is used for other test.

lib/lightning/kafka_triggers/message_server.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Lightning.KafkaTriggers.MessageServer do
```
jyeshe, line 1: Simpler name, thank you!

lib/lightning/kafka_triggers/message_worker.ex
```diff
@@ -0,0 +1,38 @@
...
+  def handle_info(:request_message, state) do
+    next_message_delay = Keyword.fetch!(state, :next_set_delay)
+    no_message_delay = Keyword.fetch!(state, :no_set_delay)
+
+    case MessageServer.next_message() do
+      nil ->
+        Process.send_after(self(), :request_message, no_message_delay)
+
+      message_id ->
+        MessageHandling.process_message_for(message_id)
```
jyeshe, line 32: Same here, there is a real setup where `process_message_for` is called and can be tested.

jyeshe (commented):
@rorymckinley what about using the offsets when the client decides to not use key on a topic?

lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
```diff
@@ -44,4 +49,52 @@ defmodule Lightning.KafkaTriggers.MessageCandidateSetSupervisor do
       |> Supervisor.child_spec(id: "mcs_worker_#{index}")
     end)
   end
+
+  def generate_child_specs(:message_candidate_set, number_of_workers) do
```
jyeshe, line 53: If the first parameter is the server module, the worker can be inferred and the function could be reused.

lib/lightning/kafka_triggers/message_handling.ex
```diff
@@ -33,6 +34,12 @@ defmodule Lightning.KafkaTriggers.MessageHandling do
     end)
   end
 
+  def find_nil_key_message_ids do
+    query = from t in TriggerKafkaMessage, where: is_nil(t.key), select: t.id
```
jyeshe, line 38: The offset would be used here to sort the messages.

rorymckinley:
> @rorymckinley what about using the offsets when the client decides to not use key on a topic?

@jyeshe The offset is defined per partition, not per topic unfortunately. Kafka guarantees sequence per key by making sure that all messages with the same key go to the same partition so under those circumstances sorting by offset is meaningful.

I decided not to incur the cost of sorting for 2 reasons: 

(1) I don't know enough about how Kafka distributes messages amongst partitions to be sure it will provide a meaningful benefit.
(2) If the publisher is not providing a key, they are essentially signalling that they do not care what order the messages are consumed in and so they won't mind what order we generate work orders in.

lib/lightning/kafka_triggers/message_server.ex
rorymckinley: @jyeshe Apologies for making review harder :(. 

I kept it separate to make the tests easier to follow. At the moment, given the 'refresh' mechanism contained within the server I would need to cycle through the messages, have it refresh and then see that we have definitely not gotten a message with a nil key. 

I could have it as a public method on the server and retain the simplicity of testing that it currently enjoys on MessageHandling, but the non-nil key counterpart currently lives on MessageHandling, so it felt more coherent to keep the method there.

At the moment, I am not sure if this code will survive load-testing :). Even if it does, the plan to move to Cachex would mean that much of this code will be changed anyway. So, with all those caveats, could we keep it here for now, please? 

I do not see this as the final implementation of nil-key handling, but rather a temporary implementation to deal with the fact that the load-testing data is sans key. Once I get a gap I would like to implement a better-engineered solution so this PR is just about adding enough wallpaper over the crack to allow us to continue with the load testing.

lib/lightning/kafka_triggers/message_worker.ex
rorymckinley: @jyeshe  This method shares some logic with `process_candidate_for` (`MessageHandling.create_work_order/1` and `MessageHandling.maybe_delete_candidate`), so it felt more coherent to keep this method there given that the primary goal is to first see if it works when drinking from the waterfall :).

I could certainly see an outcome where each server type passes in a finder query to a generalised `MessageHandling.process_message_for` method - that would be cleaner. 

However, I would like to ask for some leeway to do that - I believe that an important part of the load testing is to test the end-to-end solution, and we cannot see that until the Kafka message processing is stable.  So, I would like to confirm that this change (along with another PR yet to be crafted) delivers that and then I am happy to improve the engineering quality.

lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
rorymckinley: @jyeshe I have corrected this in a new PR.

lib/lightning/kafka_triggers/message_candidate_set_supervisor.ex
rorymckinley: @jyeshe I have corrected this in a new PR.

lib/lightning/kafka_triggers/message_server.ex
jyeshe: Okay @rorymckinley I appreciate the justification. It's not an issue after reviewing it again and refreshing the pipeline on our mind.

lib/lightning/kafka_triggers/message_worker.ex
jyeshe: Even though it's closely related to the `MessageWorker` I understand the reason you've put it on the `MessageHandling`. It's just part of the effort to make it simpler once the requirements already brings complexity to the code (from the need to have separate processing lanes for each key and now without keys).

jyeshe:
> > @rorymckinley what about using the offsets when the client decides to not use key on a topic?
> 
> @jyeshe The offset is defined per partition, not per topic unfortunately. Kafka guarantees sequence per key by making sure that all messages with the same key go to the same partition so under those circumstances sorting by offset is meaningful.
> 
> I decided not to incur the cost of sorting for 2 reasons:
> 
> (1) I don't know enough about how Kafka distributes messages amongst partitions to be sure it will provide a meaningful benefit. (2) If the publisher is not providing a key, they are essentially signalling that they do not care what order the messages are consumed in and so they won't mind what order we generate work orders in.

You're right @rorymckinley. Using the offset with nil keys would only work with the default partition strategy.

jyeshe (dismissed):

jyeshe (commented):
@rorymckinley could you confirm there is no sorting requirement at all? For messages close to one another, the `offset` or `inserted_at` would not be precise as two consumers might swap the order compared o the published one. However considering messages that are published when a patient or client goes to a service, it could help keeping the order (once the second message would be always be ingested/inserted after the first one using at least the `inserted_at`).

rorymckinley:
> @rorymckinley could you confirm there is no sorting requirement at all? For messages close to one another, the `offset` or `inserted_at` would not be precise as two consumers might swap the order compared o the published one. However considering messages that are published when a patient or client goes to a service, it could help keeping the order (once the second message would be always be ingested/inserted after the first one using at least the `inserted_at`).

@jyeshe The Kafka docs that I have read are quite explicit regarding the fact that message sequence can not be guaranteed without the key. I will query it with those who know this particular domain best.

If we wanted to sort it, we would need to sort by the timestamp that Kafka assigns to the message, and even then, when testing with my local cluster I was able to ingest messages that had the same timestamp (i.e. the speed with which Kafka ingested the message was within the millisecon-resolution that it reports timestamps with).

rorymckinley:
@jyeshe feedback from those more familiar with the domain is that we can proceed without sorting for now, and perhaps circle back if things change.

jyeshe (approved):

PR #2336, opened 2024-07-26
Limit Days To 31 Max for Monthdays In Cron Component

### LINKED ISSUE #2331: Limit number of days to 31 in Cron Scheduler
When configuring the Cron scheduler as a trigger type and every month is selected as the frequency, the day dropdown shows up[ to 60 days. We should fix the dropdown to ONLY show 1-30/1 days which is possible within the month.  See image below for reference.

image

elias-ba:
### Description

Closes #2331 

### Validation steps

1. Select a cron trigger in a workflow
2. Make the frequency to `Every Month`
3. See that the number of days is now max 31 instead of 60

### Pre-submission checklist


christad92 (commented):
Thanks LGTM

PR #2338, opened 2024-07-29
Moving runtime.exs into a module

stuartc:
### Description

Experiment to move `runtime.exs` into a module so it can be tested and called from other applications.

Instead of having one massive `runtime.exs` file, I've moved the logic into a module `Lighting.Config.Boostrap`.

This allows for easy calling from other applications, without the need for release overlays and as a bonus the config is more easily testable.

I've written a small test that exercises the `:prod` mode.

### Pre-submission checklist


stuartc:
@jyeshe I took inspiration from your config refactor the other day; which was great 🙌  - and after some playing around over the weekend I came up with this. Let me know if this raises any red flags for you?

jyeshe (approved):
@stuartc It seems it was a good insight. 
I would just avoid splitting the functionality into two steps due to this [condition](https://github.com/OpenFn/lightning/blob/22c5cea4bd20d6ab508ce22ebc975f2806340fe4/lib/lightning/config/bootstrap.ex#L18) and would call it `dump/0` or something similar instead of `setup/0` once it doesn't set the env vars by itself. It does a dump of config where you call it and depending on where it's called it gets the setup done.

stuartc:
> @stuartc It seems it was a good insight. I would just avoid splitting the functionality into two steps due to this [condition](https://github.com/OpenFn/lightning/blob/22c5cea4bd20d6ab508ce22ebc975f2806340fe4/lib/lightning/config/bootstrap.ex#L18) and would call it `dump/0` or something similar instead of `setup/0` once it doesn't set the env vars by itself. It does a dump of config where you call it and depending on where it's called it gets the setup done.

Interesting point, `dump` made me think it would return something but thinking further it does make sense like how database clients use the word.
Since it’s setting process dictionary keys, how about `configure/0`? That tells me it has side effects and doesn’t imply it does everything like `setup` might do?

stuartc:
@jyeshe I remember why I split out the env sourcing, its because we can't test the behaviour without splitting it out. The `source/1` function overrides any previously discovered env files/env vars in System.

I'll add a check to the function to throw an error if the env hasn't been sourced.

stuartc:
@jyeshe about to merge this, thanks for your input, I've dropped `setup` in favour of `configure`. In my opinion, the term configure better implies setting up or adjusting settings, which aligns with modifying configurations or state. On the other hand, dump suggests outputting or serialising data, which might not accurately describe the action of setting configuration options.

But thank you for bringing this up, I was never happy with the setup name in the first place! ✋

PR #2340, opened 2024-07-29
concept to fix #2339

### LINKED ISSUE #2339: 40mb dataclips can crash the web app
In version `v2.7.11`, when large dataclips (~38MB) are sent back to Lightning from the `ws-worker` between steps, the app can crash.

Still investigating this, but below are a few timestamps where it seems that a crash was caused by a large dataclip:

### Log Timestamps around app crashes
- 13:22 to 13:23 UTC
- 13:59 to 14:02 UTC

```js
fn(state => {
  const targetSizeMB = 40;

  // Calculate the number of characters needed (1 char = 2 bytes in UTF-16)
  const targetSizeChars = (targetSizeMB * 1024 * 1024) / 2;

  // Create a base string to use
  const baseString = "A".repeat(100); // 100 characters long

  // Calculate the number of times the base string needs to be repeated
  const repeatCount = Math.ceil(targetSizeChars / baseString.length);

  // Create the large array
  const largeArray = [];
  for (let i = 0; i < repeatCount; i++) {
    largeArray.push(baseString);
  }

  // Create the final JSON object
  const largeJsonObject = {
    data: largeArray
  };

  return { largeJsonObject };
})
```

taylordowns2000:
### Description

This PR drops dataclips when they're too big to fix #2339 

### Validation steps

1. Set up a workflow that creates an 11MB dataclip at some point in the workflow
2. Run it and verify that the dataclips are NOT saved for each step that's over 10MB

### Pre-submission checklist

### DIFF
--- a/lib/lightning_web/channels/run_channel.ex
+++ b/lib/lightning_web/channels/run_channel.ex
@@ -179,11 +179,31 @@ defmodule LightningWeb.RunChannel do
   end
 
   def handle_in("step:complete", payload, socket) do
+    max_size = Application.get_env(:lightning, :max_dataclip_size_bytes)
+
+    payload_size =
+      Map.get(payload, "output_dataclip")
+      |> :erlang.term_to_binary()
+      |> byte_size()
+
+    safe_payload =
+      if(payload_size > max_size) do
+        Logger.warning(fn ->
+          "Dataclip for run #{socket.assigns.run.id} too big. Dropping it."
+        end)
+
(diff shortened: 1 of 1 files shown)


taylordowns2000:
closing as this is likely too _late_ to check byte_size. https://github.com/OpenFn/kit/pull/740 will prevent us from sending really big payloads back to lightning, but we should still apply some more defensive stuff on this side

PR #2364, opened 2024-08-08
Ensure Users Confirm Their Accounts After Sign Up

### LINKED ISSUE #2389: Ensure users have real, valid emails
Appetite: 3 days

Context:
After a user signs up, we send them an email to verify their account by clicking a link. Clicking this link allows us to verify users and ensure that all users on the platform have signed up with their emails.  At the moment, don't check to ensure that every user has confirmed their email, and when they haven't, we can't enforce the user. 

Solution:
The goal of this project is to ensure that users confirm their account within 48 hours of signing up to their account. To validate if a user has confirmed their account, we will check the `confirmed_at` column in the `users` table to see if it is not null. The default state of the `users.confirmed_at` is `null` otherwise, the field will have the timestamp of when the user confirmed their account. 

When the user has not confirmed their email within 48 hours of creating an account, we should show them a danger banner at the top of the page. (For the banner, please see [OpenFn Spec](https://github.com/OpenFn/lightning/blob/c220c3ceafd0d5a29349e0447005cf6d94c8818c/lib/lightning_web/live/components/common.ex#L24)).

After 48 hours, we will block access to their projects and the user can request that we resend the confirmation email. Here is [how we currently send the confirmation email](https://github.com/OpenFn/lightning/blob/c220c3ceafd0d5a29349e0447005cf6d94c8818c/lib/lightning/accounts.ex#L806) to users. We don't need to change how `deliver_user_confirmation_instructions` works at the moment, we should ensure that we can call this function when a user requests that we resend the confirmation email. 

- The behaviour needs to exist across _all_ routes, note that you're not adding a `common.alert` to the top of each component
- Note that the banner uses the common component and the action is to request a resend of the confirmation email.
- You are using the same pattern as we do to display the "past your run limit" alert banner which _possibly_ lives in the route template for liveview... it's **outside** the component. (Might be slot?)

When the user is blocked: 
Unless they are viewing the `/profile` route, the user cannot see anything but the "your account needs verifying" modal. They should ONLY be able to ask that we resend the user confirmation email or navigate to `/profile` so that they can change their email. As soon as the user is confirmed, they can do as much as they'd love to. When this happens, we need to protect the paths and routes so that the users don't bypass the block screen.

- [ ] Confirm that all signup flows have email confirmation set up 
- [ ] Display an info alert telling the user that confirmation is pending on their account
- [ ] Resend the confirmation email when the user clicks the resend button
- [ ] Block user's actions if the user has not confirmed their emails after 48hours of signing up.
- [ ] Ensure that the user can _still_ navigate to `/profile` and change their email address.

https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App-(new)?node-id=4692-1206&t=wyW6Zg0T2S9ClNT1-0

### Things to be careful of
- The IDE doesn't currently accommodate various screen sizes. There is a risk that when the banner pushes everything down, it won't fill the remaining screen height, instead it will get cut off. Please be careful about this and come up with some not terrible solution.

elias-ba:
### Description

This PR introduces an alert banner and modal to remind users to confirm their accounts. When a user creates an account in Lightning, we send them an account confirmation email. Until the user confirms their account, an alert banner is displayed at the top of the app to remind them about the confirmation. If the user does not confirm their account within 48 hours, we block their access to all app pages except the profile page by displaying a modal that requires them to confirm their account. This modal allows them to resend the confirmation email and visit their profile page to modify their email.

Closes #2389 

### Validation steps

1. Account Creation
    - Create a new user account in Lightning.
    - Ensure that an account confirmation email is sent to the user's registered email address.

2. Alert Banner Display
    - Before the user confirms their account, log in to the app.
    - Verify that an alert banner is displayed at the top of the app, reminding the user to confirm their account.

3. Email Confirmation
    - Confirm the account using the link in the confirmation email.
    - Log in again and verify that the alert banner is no longer displayed.

4. 48-Hour Time Lapse
    - Create another test account without confirming it immediately.
    - Wait for 48 hours to pass (or simulate this time lapse if possible).

5. Modal Display
    - After 48 hours, attempt to log in with the unconfirmed account.
    - Verify that a modal appears, blocking access to all app pages except the profile page.
    - Ensure the modal clearly instructs the user to confirm their account.

6. Resend Confirmation Email
    - Use the modal to resend the confirmation email.
    - Check that a new confirmation email is received in the user's email inbox.
    - Confirm the account using the link in the email.
    - Log in and verify that the modal no longer appears and full app access is restored.

7. Profile Page Access
    - From the modal, navigate to the profile page.
    - Verify that the user can modify their email address on the profile page.

### Pre-submission checklist


christad92:
Thanks for this man. 

Please see my comments here: 
- For the message in the blocking modal, please let's remove "accounts" from this line : `... we have blocked access to your accounts, projects and workflows...`
- We need to rethink the placement of the alert message when the email is sent. I'd love something inexpensive, I'd have loved the toast but I get your concern on time and how expensive it can be. 
image

christad92:
Also, rather than sending the initial email for account verification, can we use this new [copy](https://docs.google.com/document/d/1FGO1lotlR0yrRUo8vuLV9fjIvelx5LqKewRGrNFTaco/edit#heading=h.9mol8dwiivlc)? The default version has the "Welcome to OpenFn" text which should not be sent to a user who have been on the platform for longer than 48 hours.  

@elias-ba

christad92:
Thanks. I have just one last comment: 

When I click on resend confirmation email from the banner, please can we show a flash message that says "Confirmation email sent to your email address.

image


All is good to me

taylordowns2000:
added css, will remove or fix tests tomorrow

taylordowns2000 (approved):

christad92 (approved):
Looks good to me! 

Thanks

midigofrank (changes requested):
Great job @elias-ba 🙌 . I've left a2 comments, but mainly:
 we-dont-talk-about-thunderbolt

midigofrank (approved):

Issue #2367, opened 2024-08-09
Allow deploy to manage credential associations for jobs in lightning


christad92:
Allow `cli deploy` to control the credentials that are associated with a given job in a workflow in a project.

- How should we uniquely identify credentials?
- First assume that all credentials are already available within the project (`project_credential` records exist) - can you associate jobs with those credentials?
- Then assume that the credentials belong to the deploying user but are _not yet_ available within the project (`project_credential` records don't exist) - can you prompt the user to see if they want to add `project_credential` associations to that project.

midigofrank:
@taylordowns2000 we should definitely extend the credential uniqueness to projects.

midigofrank:
Or maybe we can add the user email to the credential name to make it unique. Something like:

`"user@email.com:DHIS2 Prod"`

taylordowns2000:
I think user (not project) is the way to go. In v1 we show the user email alongside the cred for selection—that's a better UX anyway—so the concern is just deployment.

PR #2371, opened 2024-08-10
make Credentials unique by user by name

taylordowns2000:
### Description

This PR makes credentials unique by `user_id` and `name`. Right now, we allow duplicate credential names. This creates 2 problems:
1. it's hard to know which credential to select from a picklist (only partially solved now... we may consider extending the uniqueness to `project_id` via `project_credentials` if we want more collision avoidance, but that's doesn't seem necessary at the moment—will speak with @midigofrank on Monday to decide)
2. it's hard to generate credential slugs and associate jobs with credentials via `cli deploy` unless we have some sort of uniqueness here.

References #2367 

### Validation steps

1. Create several credentials with the same name for the same user
2. Run the migrations
3. Note that the credential names have been updated with `duplicatename-2`, `duplicatename-3`, etc.
4. Try to create a new credential with a duplicate name for the same user
5. Note the validation error

Only the note above on whether we should extend this to include unqiueness by `project_id` via the `project_credentials` relationship. 

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ and this project adheres to
 
 ### Changed
 
+- Enforced uniqueness on credential names _by user_.
+  [#2371](https://github.com/OpenFn/lightning/pull/2371)
+
 ### Fixed
 
 ## [v2.7.16] - 2024-08-07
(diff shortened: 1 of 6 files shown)


midigofrank (approved):
🙌

PR #2374, opened 2024-08-12
Use Swoosh to format User models into recipients

stuartc:
### Description

This PR makes changes to how we set the recipients on emails, previously we were just setting the email address.

This changes that to send the email with their first and last name as the recipient name and uses their email address.

Leveraging the Recipient protocol on Swoosh.

### Validation steps

Emails sent from Lightning should have `User Name (email@address.com)` instead of `email@address.com` as the `to:` field.

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ and this project adheres to
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


midigofrank (approved):
Nicely done 🙌

PR #2376, opened 2024-08-12
Test Harness for Layouts

josephjclark:
### Description

This PR creates a test harness for workflow layouts. It'll run a bunch of workflows and save screenshots of them to disk.

The idea is that I can make changes to the layout and run it against many workflows at once, so we can easily compare changes.

* Restore the old dev-server (which makes it easier to dev this stuff)
* Add repro cases to the dev server
* Set up a puppeteer script which will capture a screenshot for every chart.

Works great!

I've also added a bunch of little fixes to the dev server and some util functions to get it all working again. There are a couple of diffs against production code, but only very very minor

Here's a grid of screenshots I took.

image

## How to run

```
npm run test:workflow-snapshots
```

To get this to work I had to install chromium specially and pass the path to the bin. This should be looked at more but it's likely to only be me running this stuff, so I'm not too worried.

### Pre-submission checklist

- [ ] I have performed a **self-review** of my code.
- [ ] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [ ] I have updated the **changelog**.

Issue #2366, opened 2024-08-09
Edges/paths tend to overlap on the canvas 


christad92:
As the workflow in a canvas grows with more edges and paths, we have situations where edges and paths overlap as shown in the video below. 

https://www.loom.com/share/cbd61186553a482e86b8cc4e6dd59cf7?sid=216dfec4-4065-4c8a-9499-1c18a976c869

There are two potential ways we can do this: 
1. Have a collision detection for paths and edges and use that to ensure they don't overlap. 
2. Allow users to manually arrange edges on the canvas. This way, we can store the x/y positions of each edge.

christad92:
@josephjclark you might find this to be interesting. https://github.com/tisoap/react-flow-smart-edge

josephjclark:
On the left, our production layout with forked dagre. On the right, the dagre fork removed:

![image](https://github.com/user-attachments/assets/aa8a0048-4c9b-4668-8fa0-860674e475c6)

Uh oh.

What our dagre fork does is disable the "optimal order heuristic". Which, clearly, is designed to minimize edge crossings.

We switched it off because it makes the chart significantly less stable when adding and removing nodes.

josephjclark:
Unfortunately I can't reproduce Aisha's exact chart. I can produce the same problem on smaller charts, but on her chart it's working fine for me.

I have noticed that inconsistent edge (edges which don't connect to anything) cause weird layouts in my local testing - I wonder if Aisha has a ghost edge in her workflow?

Otherwise I suspect there's something to do order in which nodes are added (or maybe removed), or in the node id patterns maybe, which is triggering the problem

josephjclark:
@AishaHassen I am struggling to reproduce your issue precisely. I suspect an integrity issue in the workflow itself.

Is it possible to give me access to the project so I can inspect it? Or are you able to download the project.yaml and send that to me?

AishaHassen:
Hey @josephjclark, seems you already have access to the project. Here is a [link](https://app.openfn.org/projects/2fedca61-bb42-4e65-a908-7a60f1b25c1e/w/3350e498-75dd-4d9b-8ca7-a83c00ae21e7?v=503) to the workflow

josephjclark:
[...]

PR #2381, opened 2024-08-13
Workflow Diagram: Reduce edge crossings

josephjclark:
I think this is a reasonably small fix to #2366 (which is a very big problem).

When we forked dagre.js, we _disabled_ the layout algorithm's edge crossing optimization, because it made the chart very unstable when adding nodes.

The consequence of this is that the layout is not smart in cases where edges cross each other and behind other nodes, resulting in the problem in #2366

Looking more closely at dagre, I've made a more subtle fix. Now we're keeping edge crossing optimization, but we only adjust the node order if we find an edge crossing (where previously it always tampered with the ordering, even if there was no change)

So as a result of this PR:
* The layout should do better at preventing edges going behind nodes
* But it still isn't perfect, we will continue to encounter problems
* When adding new nodes, the chart SHOULD remain stable, but if there is an edge crossing in the chart, the nodes may re-order themselves "randomly".

## Issues

#2366

## QA

I'd love to test this against Aisha's original workflow. But I failed to reproduce the problem locally, so maybe it's not a clear-cut test.

- Check that the chart is reasonably stable when adding new nodes
- Check this in workflows which have crossing/overlapping edges
- Run the image generator against #2376  (`npm --prefix assets run test:canvas`)
- Run the same test here and compare results.


josephjclark:
Aaaand it's not working :(

josephjclark:
Ok, I've been digging in a bit. I think the ordering heuristic is working, perhaps by fluke, in some cases.

It only compares adjacent layers for overlaps - it doesn't even look at cross-generational overlaps, which is the case that Aisha's seeing.

I'm also becoming increasingly sure that Aisha's is a different bug. I think she has a disconnected edge in the chart and it's affecting the calculation.

PR #2403, opened 2024-08-19
Use Anthropic (claude)

josephjclark:
This PR renames `OPENAI_API_KEY` to `AI_ASSISTANT_API_KEY`, to reflect an upcoming change on the apollo backend.

⚠️ **If you are running a lightning deployment an use the AI assistant, make sure to update your secret!**

Note that this isn't technically needed to start using the new model. Setting the anthropic key values into the open AI key will work just fine.

### DIFF
--- a/lib/lightning/config/bootstrap.ex
+++ b/lib/lightning/config/bootstrap.ex
@@ -99,7 +99,7 @@ defmodule Lightning.Config.Bootstrap do
           :string,
           Utils.get_env([:lightning, :apollo, :endpoint])
         ),
-      openai_api_key: env!("AI_ASSISTANT_API_KEY", :string, nil)
+      ai_assistant_api_key: env!("AI_ASSISTANT_API_KEY", :string, nil)
 
     config :lightning, Lightning.Runtime.RuntimeManager,
       start:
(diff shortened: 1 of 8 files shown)


lib/lightning/config/bootstrap.ex
```diff
@@ -99,7 +99,7 @@ defmodule Lightning.Config.Bootstrap do
           :string,
           Utils.get_env([:lightning, :apollo, :endpoint])
         ),
-      openai_api_key: env!("OPENAI_API_KEY", :string, nil)
+      openai_api_key: env!("ANTHROPIC_API_KEY", :string, nil)
```
taylordowns2000, line 102: should the key name here be more generic? maybe change `openai_api_key` to `apollo_3rd_party_key`? or am i misunderstanding this change?

taylordowns2000 (changes requested):
hey i'd love resolution on the key name on line 102 of bootstrap. once that's clear happy to merge

lib/lightning/config/bootstrap.ex
josephjclark: That's a good question.

It certainly isn't a generic apollo API key. Different apollo services require different API keys.

It might make sense to rename to `AI_ASSISTANT_API_KEY` though. That way we're binding it to the service, rather than the implementation. Of course it's also less declarative from a devops point of view.

josephjclark:
Hi @taylordowns2000 and @stuartc where are we sitting on this? What do we want to do with the env var name?

taylordowns2000 (approved):

PR #2404, opened 2024-08-20
Correctly persist Kafka messages with headers

### LINKED ISSUE #2402: Handling Kafka Message Headers
When BroadwayKafka parses an incoming message that has headers, those headers are converted into a list of `{header, value}` tuples. Jason (understandably) does not support the serialisation of tuples. Therefore, these headers need to be converted into a list of `[header, value]` lists before persisting the message.

rorymckinley:
### Description

Kafka message headers are represented as a list of `{header, value}` tuples. This causes JSON serialisation issues, so this change converts the list of tuples to a list of `[header, value]` lists.

Closes #2402 

### Validation steps

- Setup a local Kafka cluster based on these [instructions](https://github.com/OpenFn/lightning/blob/main/kafka_testing/KAFKA_ARCHITECTURE.md#testing-locally-using-the-docker-kafka-cluster).
- Ensure that `KAFKA_TRIGGERS_ENABLED` is set to 'yes'
- Set up a Kafka trigger similar to the screenshot below (assuming you have  a topic named `baz_topic`) - note that the `Initial offset reset policy` is set to `latest` :

Screenshot from 2024-08-19 12-23-47

- Publish a message _without headers_ to the Kafka cluster, using kcat: `cat message.json | kcat -P -b 127.0.0.1:9094 -t baz_topic`.
- Look at the dataclip that is created for the message - the headers should be an empty list.
- Publish  a message _with headers_ to the Kafka cluster, using kcat: `cat message.json | kcat -P -b 127.0.0.1:9094 -t baz_topic -H headerone=headeronevalue -H headertwo=headertwovalue -H headerone=headerthreevalue`.
- The  complete headers should be represented in the dataclip created for this new message:

image

### Pre-submission checklist


jyeshe (commented):
Hey @rorymckinley is there a particular reason to not use a map instead of list of lists?

rorymckinley:
> Hey @rorymckinley is there a particular reason to not use a map instead of list of lists?

@jyeshe It is possible to specify the same header more than once and BroadwayKafka will retain each value for the header rather than overwriting it, so I picked a list of lists to preserve that behaviour (Note: I don't know enough about Kafka usage to understand *why* you would want to do that, but I did not want to introduce information loss).

lib/lightning/kafka_triggers/message_handling.ex
```diff
@@ -72,4 +74,14 @@ defmodule Lightning.KafkaTriggers.MessageHandling do
...
     end
   end
+
+  defp convert_for_serialisation(%{headers: headers} = metadata) do
+    metadata |> Map.merge(%{headers: list_of_lists(headers)})
+  end
+
+  defp list_of_lists(list) do
+    list
+    |> Enum.reduce([], fn tuple, acc -> [tuple |> Tuple.to_list() | acc] end)
```
jyeshe, line 84: I am not familiar with the input but id it needs to transform every tuple on a list into a list it can be done with a Enum.map. it would avoid the reverse.

jyeshe (approved):
`Enum` suggestion otherwise looks good to me.

stuartc (approved):

lib/lightning/kafka_triggers/message_handling.ex
stuartc: @rorymckinley I first changed it to:

```elixir
metadata |> Map.merge(%{headers: Enum.into(headers, [], &Tuple.to_list/1)})
```

Then I thought, thats a bit obnoxious; I've changed it to this:

```elixir
Map.put(metadata, :headers, Enum.map(headers, &Tuple.to_list/1))
```

👌

Issue #2416, opened 2024-08-22
Run shows as `failed` while step shows as `running`, in truth already failed


ritazagoni:
On both the [history page](https://app.openfn.org/projects/2fedca61-bb42-4e65-a908-7a60f1b25c1e/history?filters[workorder_id]=db766f95-1d55-40b6-bd55-47d743226422) and when [inspecting](https://app.openfn.org/projects/2fedca61-bb42-4e65-a908-7a60f1b25c1e/w/3350e498-75dd-4d9b-8ca7-a83c00ae21e7?v=655&a=a1ed7bf9-3830-4113-8e59-69ddf5dd9222&m=expand&s=879f3456-a196-44b3-9b48-59a8cc1d0f8d#log), a run is showing as `failed` while the step is showing as `running`. 

Meanwhile the step has already failed with `RuntimeError: TypeError: Cannot read properties of undefined (reading '0')`

This is a workflow that was triggered by another workflow.

v2.7.19

I have reproduced this locally on main:

- [ ] Yes
- [ x] No

![Screenshot 2024-08-22 150850](https://github.com/user-attachments/assets/cc765681-56eb-4e87-9bf2-14977bff4ea3)
![Screenshot 2024-08-22 150812](https://github.com/user-attachments/assets/75a0130b-01f2-402b-bccc-8d50defb5328)

taylordowns2000:
Could add some sort of trigger that sets steps to `failed|lost|crashed|skipped` (whatever the right decision is here!) when the run has finished and the channel is closed

stuartc:
@taylordowns2000 yeah we've been thinking the same; but we need to be really careful about how we do it - we're essentially changing data to something we don't know for certain. So some considerations about what the worker _might_ have been doing or how it might have failed and only apply this policy after a period of time where it would be theoretically impossible for it to change.

taylordowns2000:
Yeah I think we're on the same page. The ideal is "100% honest and as helpful as possible" - where honesty is more important than helpfulness. So "lost" would be just fine for all these as that's our shorthand for "I don't know what happened."

As in: "that step is `:lost` because I never got information about whether it succeeded or failed."

taylordowns2000:
@christad92 , why is this something we wont fix?

christad92:
[...]

PR #2418, opened 2024-08-22
Pull timeout

taylordowns2000:
this is a concept. @stuartc , we're trying to:
1. raise the timeout (make it ENV configurable) so that 11s query doesnt lead to `lost` (@josephjclark on it.)
2. make that query more efficient (and maybe look for _other things_ that might be slowing down this part of the pipeline under heavy load - you and @rorymckinley on it)
3. come up with some sort of failsafe that sweeps up false-claims (you sir must _FORFEIT YOUR CLAIM ON THAT RUN!_) and releases them for someone else to claim... but as you can see in the test placeholder, i really don't like this.

### DIFF
--- a/test/lightning/janitor_test.exs
+++ b/test/lightning/janitor_test.exs
@@ -6,7 +6,14 @@ defmodule Lightning.JanitorTest do
   alias Lightning.Repo
   alias Lightning.Run
 
-  describe "release_claimed_but_not_started/0" do
+  describe "forfeit_expired_claims/0" do
+    test "releases runs for reclaim if they have not been started after the pull timeout plus grace" do
+      dbg("eish, i don't love this. what if there was some other reason for them getting lost?")
+      dbg("like, what if they did start, and did the work, and all that, but we never heard back from them because of network issues?")
+      dbg("i wouldn't want to re-do the work.")
+      dbg("i wish there was some way to only mark them as claimed once we know that the worker actually got the run.")
+      dbg("can we check that our reply in the websocket channel was actually received by the ws-worker?")
+    end
   end
 
   describe "find_and_update_lost/0" do
--- a/test/lightning/janitor_test.exs
+++ b/test/lightning/janitor_test.exs
@@ -6,6 +6,9 @@ defmodule Lightning.JanitorTest do
   alias Lightning.Repo
   alias Lightning.Run
 
+  describe "release_claimed_but_not_started/0" do
+  end
+
   describe "find_and_update_lost/0" do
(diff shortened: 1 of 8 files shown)


test/lightning/janitor_test.exs
```diff
@@ -6,6 +6,16 @@ defmodule Lightning.JanitorTest do
   alias Lightning.Repo
   alias Lightning.Run
 
+  describe "forfeit_expired_claims/0" do
+    test "releases runs for reclaim if they have not been started after the pull timeout plus grace" do
+      dbg("eish, i don't love this. what if there was some other reason for them getting lost?")
+      dbg("like, what if they did start, and did the work, and all that, but we never heard back from them because of network issues?")
+      dbg("i wouldn't want to re-do the work.")
+      dbg("i wish there was some way to only mark them as claimed once we know that the worker actually got the run.")
```
rorymckinley, line 14: @taylordowns2000 If I may chip in (uninvited):

As you mentioned yesterday the janitor approach is a band-aid until a better fix is available. If I was applying the bandaid, I would not tie it to the timeout value - I would make it a large multiple of the timeout. E.g. if your timeout is 20 sec, maybe wait 10 min (after all you wait 2 South Effrican hours to determine if something is lost, right?). And, notify Sentry whenever this happens, so that we have an opportunity to take a look at what is going on and tune things. Wrt the concern of potentially redoing work, you could make this something that is user-controllable (i.e. if a Run is 'LostAfterClaimed' would you like us to automagically resubmit it - or would you rather do that manually?).

A longer-term fix (which I can only assume will mean a fair amount of work) would be to allow a run to be claimed by multiple worker processes but only started by one (whoever gets there first). This obviously creates inefficiency elsewhere. Personally, I would do the bandaid if it helps us satisfy the immediate requirements and then identify the issue - e.g. maybe it is a simple as scaling up web pods alongside worker pods?

taylordowns2000:
@rorymckinley , @stuartc , do you know if this code is still needed? would love to shut down this PR if it's not helpful (or has been implemented elsewhere in response to #2465 or #2416 )

rorymckinley:
> @rorymckinley , @stuartc , do you know if this code is still needed? would love to shut down this PR if it's not helpful (or has been implemented elsewhere in response to #2465 or #2416 )

@taylordowns2000 I have not implemented anything that addresses this  - did the changing of the timeout values make the original problem go away?

taylordowns2000:
lol @stuartc and @rorymckinley - look at this oldy but goody from August 2024 🤣

Issue #2432, opened 2024-08-26
Job editor: code assist doesn't appear for namespaces


josephjclark:
Code assist in Lighting works great for an adaptors operations, like `fn`:

![image](https://github.com/user-attachments/assets/de057af3-6750-4d80-b535-bd1938689756)

Some adaptors export non-operation functions as a "namespace". Common, for example, exports `http` with a bunch of helper functions.

But typings do not work for those namespaces :scream_cat: 

![image](https://github.com/user-attachments/assets/79260ce7-ea17-4b78-bd2a-d184ac326f33)

This is wrong and bad and ought to be fixed.

PR #2433, opened 2024-08-26
editor: fix namespaces in code assist

josephjclark:
This PR fixes (kinda) an issue in the editor where code assist is not provided for namespaces (like `http.get`). See #2432

This how it looks on this branch:

image

Playing with this stuff in Monaco is like pulling teeth. Some things work, some don't, the results always seem a bit random. FWIW the best example I've found for how to load typescript libraries is [this playground demo](). But that only deals with a single file.

For some reason, when we load index.d.ts, the namespaces and stuff exported from there don't get properly loaded into the environment. There's some strange voodoo needed to get the TS environment to load the right context.

In VSC this stuff just works, so I know that the actual typescript is correct. It's just our Monaco setup.

Anyway: this PR changers the way we load stuff so that namespaces kinda work.

I don't like it, but it's an improvement. It also takes a bunch of stuff out the environment that shouldn't be there.

One issue on this branch is that the options are not properly tracked `http.get` . Strange because the interface _IS_ is inside the namespace. Monaco just doesn't  find it.

It does track on other adaptors though -  I've got good namespaced trackign working on this experimental adaptor

image

### DIFF
--- a/assets/js/adaptor-docs/hooks/useDocs.tsx
+++ b/assets/js/adaptor-docs/hooks/useDocs.tsx
@@ -3,7 +3,7 @@ import { describePackage, PackageDescription } from '@openfn/describe-package';
 
 // Describe package is slow right now even if data is available
 // This in-mamory cache will help when switching tabs etc
-const cache: Record<string, PackageDescription | null | false> = {}
+const cache: Record<string, PackageDescription | null | false> = {};
 
 const useDocs = (specifier: string) => {
   const [docs, setDocs] = useState<PackageDescription | null | false>(null);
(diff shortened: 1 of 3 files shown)


assets/js/adaptor-docs/hooks/useDocs.tsx
```diff
@@ -3,7 +3,7 @@ import { describePackage, PackageDescription } from '@openfn/describe-package';
 
 // Describe package is slow right now even if data is available
 // This in-mamory cache will help when switching tabs etc
-const cache: Record<string, PackageDescription | null | false> = {}
+const cache: Record<string, PackageDescription | null | false> = {};
```
josephjclark, line 6: Uh this file is just a formatting diff

assets/js/editor/Editor.tsx
```diff
@@ -128,21 +128,27 @@ async function loadDTS(
...
+      // I don't understand why index.d.ts doesn't just do this though.
+      const fileName = filePath.split('/').at(-1).replace('.d.ts', '');
+
+      if (fileName === 'index' || fileName === 'Adaptor') {
+        results.push({
+          content: `declare ${type} { ${content} }`,
+        });
+      } else {
+        results.push({
+          content: `declare ${type} ${fileName} { ${content} }`,
```
josephjclark, line 146: This is the real fix - explicitly loading each file into its own namespace. It kinda works.

josephjclark:
Hey @stuartc I'd really like to bump this one into the next release - can you take a look at  fairly high priority?

It's an imperfect improvement, but an improvement nonetheless.

stuartc (approved):

PR #2437, opened 2024-08-27
Add Full Width Option for Select Inputs

elias-ba:
### Description

This PR introduces a new :full_width attribute to the input component in `LightningWeb.Components.NewInputs`. When set to true, this attribute allows the select input to expand to the full width of its container. This provides more flexibility in layout management and ensures that the select input can be displayed consistently alongside other form elements like text inputs. The default behavior remains unchanged, with the select input not expanding unless explicitly specified.

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to
 - Implement a combo-box to make navigating between projects easier
   [#241](https://github.com/OpenFn/lightning/pull/2424)
 
+- Fix select inputs not expanding to full width. Added a :full_width option to
+  control whether the input should expand to the full width or not
+  [#2437](https://github.com/OpenFn/lightning/pull/2437)
+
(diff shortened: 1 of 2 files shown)


stuartc (commented):
@elias-ba  I'm of the opinion that the component should stretch out all the time, what is the effect (generally as you move around the app) if the `flex-grow` class was always there? Does the app look weird, or certain forms look broken or too big?

elias-ba:
@stuartc yes the Add Collaborators forms would be broken. Because those forms are layed out horizontally. I think we should still handle the problem in the forms (grid layout or some other layout) but this PR was a tiny fix that would help me fix my problem the cheapest way

stuartc:
@elias-ba I'm going to close this, as discussed the other day; I believe the best approach is not extend our component but rather look into how the containing components/elements affect the size of this.

PR #2444, opened 2024-08-30
Configurable Cowboy idle timeout

### LINKED ISSUE #2443: Make Cowboy timeout configurable
Allow the Cowboy idle timeout to be configurable so that it can be aligned with conditions that are suited to the hosting environment e.g. when it is behind a load balancer that has a significantly longer timeout.

rorymckinley:
### Description

This PR adds a configurable timeout for Cowboy connections to address use cases where the Lightning idle timeout differs greatly from the timeout set on an upstream load balancer. In this situation, Lightning can disconnect before the load balancer disconnects resulting in 502s. By making the timeout configurable, Lightning's behaviour can be aligned with that of the load balancer.

Closes #2443 

### Validation steps

This is tricky :) - as the change is restricted to the production environment. I found that the easiest way to test was to copy a lightweight version of the changes to [here](https://github.com/OpenFn/lightning/blob/main/lib/lightning/config/bootstrap.ex#L407).

A 'lightweight' version in this case can be set up by applying the attached patch (GH would not let me upload a file with the patch extension). You can apply it by running `patch -p1 < test.txt` from Lightning root.
[test.txt](https://github.com/user-attachments/files/16811710/test.txt)

Once you have applied the patch, you can test it by running a test IEx session: 

```
MIX_ENV=test iex -S mix phx.server
``` 

From within the IEx session, you can check the value set:

```
LightningWeb.Endpoint.config(:http)
```

It will default to 60000 and can be overridden by the `IDLE_TIMEOUT` ENV var.

### Pre-submission checklist


taylordowns2000 (changes requested):
i think we should target `max_keepalive` based on this guy here: https://github.com/ninenines/cowboy/issues/1286

can you make that configurable as well?

PR #2273, opened 2024-07-09
use snapshots when initiating github sync

midigofrank:
## Validation Steps
You'll need these envs:
- `GITHUB_APP_ID`
- `GITHUB_APP_NAME`
- `GITHUB_APP_CLIENT_ID`
- `GITHUB_APP_CLIENT_SECRET`
- `GITHUB_CERT`

You also need a github repo. I would advise that you create an empty repo in your account with just the `README`. I have one just like that for testing these github work: https://github.com/midigofrank/openfn

If you're testing this locally, you'll need a tunnelling tool. I always use https://ngrok.com/docs/getting-started/ .
Assuming you're using `ngrok`, run the command below and copy the URL generated.
Replace call to `LightningWeb.Endpoint.url()` using that URL in this line https://github.com/OpenFn/lightning/blob/main/lib/lightning/version_control/version_control.ex#L548
Please remember to undo this change once you're done 

```bash
ngrok http 4000
```

**How to validate**
The aim is to initiate a sync and modify our workflow before the github action runs. We can achieve this by cancelling/stopping the github action and then later retry it.

1. Open the `Sync to Github` tab under project settings page
2. Click to connect your github account if you haven't already
3. Have your github repo open in another tab
4. Start the process, select your repo, branch, and chose the option to **Sync to Github** (OpenFn -> Github)
5. Click save and rush to your github repo, under the actions tab, wait for the action `.github/workflows/openfn-pull.yml` and cancel it. 
6. Now go back to lightning and update your workflow name, or any job name and save
7. Go back to your github repo actions and retry the action you cancelled in step 5
8. Once the `action` completes, we expect that the `yaml` committed in your github repo **NOT** to contain the changes you did in Step 6. It should have the original workflow and job names 

## Notes for the reviewer

- [x] Send workflow snapshot ids to github when initiating github sync.
- [x] Optimize creating snapshots for workflows without one. I need some eyes here
- [x] Update `priv/github/pull.yaml` to handle the `snapshot input` sent
- [x] Handle pulling of `state.json` and `project.yaml` with snapshots. It expects the user to hit the endpoint with `?snapshots[]=uuid1&snapshots[]=uuid2&snapshots[]=uuid3`.
The logic for getting this data in the contexts has been modified to accept `nil` or `list`
- [x] Modify the `cli-pull-action` to use the snapshots.  Here is the PR: https://github.com/OpenFn/cli-pull-action/pull/2.
Notice that it adds the `--snapshots` flag only if the `snapshots` are provided.
[...]

PR #2471, opened 2024-09-06
Provisioner only saves modified edges/jobs/triggers in snapshots

### LINKED ISSUE #2460: Workflow crashes with ValidationError: Cannot find job
When running a workflow, we get  ValidationError Cannot find a job: {ID}. We have checked the project spec and the job exists on the workflow, but the step is not discovered when the run is executed. 

This is currently happening on projects/workflows that are deployed from GitHub.

We've noticed that the Jobs that have raised this error eventually have successful runs later without any fixes but it is worth looking into the cause of these errors. 

image

### LINKED ISSUE #2461: Getting Cannot read properties of undefined (reading name)
This happens when a user tries to initiate a GitHub sync for a project. 

Image

midigofrank:
### Description
This PR fixes an issue where only changed jobs, edges and triggers were getting saved in the db.

In the provisioner, there existed a logic in the changesets for `edges`, `jobs` and `triggers` which was ignoring unchanged data. The logic is useful because it gets rid of empty maps when there is an error in one workflow but not the others

What this PR does is it forces the preload of all the assocs before creating snapshots

Closes #2461 
Closes #2460 

### Validation steps
You can validate this using the Github Sync feature in the UI or just the plain old cli. If you're testing locally and want to use Github, head to #2273, there's a nice description on how to do that easily.
 
Here's the gist of it:

1. In your project, open a workflow in the UI. It will redirect you to the "latest" version of that workflow. In the uri, note the version number, it's something like `?v=1`. For this purpose, let's use `1` as the lates version. PS: make sure your workflow has multiple jobs
2. Pull these changes. If using Github, head to project settings and click `initiate sync to github`. If using cli, `openfn pull project_id`
3.  Identify one job in your `spec.yaml`, change the name and commit. Or if using cli, `openfn deploy`. This creates snapshot version 2 of that particular workflow
4. Now in the workflow that you opened in step 1, if you type `?v=2` or your equivalent latest snapshot version, all the jobs will exist and the page would just fine. Previously, you would have found that only the job whose name you updated in step 3 exists.

Anyone who attempted to use `cli deploy` after snapshots were added to the Provisioner is probably affected by this.
Once deployed, we should run a script to correct the snapshots using the state in the `job`, `edges` and `triggers` tables. 

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ and this project adheres to
 
 - Fix `workflow_id` presence in state.json during Github sync
   [#2445](https://github.com/OpenFn/lightning/issues/2445)
+- Provioner creates invalid snapshots when doing CLI deploy
+  [#2461](https://github.com/OpenFn/lightning/issues/2461)
+  [#2460](https://github.com/OpenFn/lightning/issues/2460)
 
 ## [v2.8.2] - 2024-09-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to
 
 ### Fixed
 
+- Provisioner creates invalid snapshots when doing CLI deploy
+  [#2461](https://github.com/OpenFn/lightning/issues/2461)
+  [#2460](https://github.com/OpenFn/lightning/issues/2460)
+
 ## [v2.9.0] - 2024-09-06
 
 ### Added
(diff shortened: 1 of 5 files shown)


elias-ba (approved):
This is looking good to me. Bravo @midigofrank 👏🏽

stuartc (approved):

PR #2478, opened 2024-09-09
AI assistant:  Responsible AI Review

josephjclark:
Surgical strike to tidy up and apply responsible AI practices to the AI chat assistant.

image
image
image

# Changes

* Re-write of the "disclaimer". It's help text now.
* Fix "PPI" typo (this is probably my fault anyway)
* Add a fixed disclaimer "this message was generated by an algorithm". I'm open to better styling on this - I'd like a different font but I can't find a good one to hand :thinking:  This is designed to wedge a tiny spark of doubt in the user's mind and encourage critical thinking about the response
*  Added a tooltip to the user icon and disabled text selection
image
* Clean up the  styling of the "you haven't set up the assistant" message
* Made a footer along the bottom with:
  *  responsible AI policy link 
  *  Link to the help content 
* ~~Verify that input and logs are not send to assistant~~ Looks good
* review changelog

Closes #2411 https://github.com/OpenFn/lightning/issues/2476

Note that the policy link is dead until the policy is live on the marketing site (won't be long and the URL is correct)

PR #2486, opened 2024-09-10
Disable AI assistant submit button incase the job isn't saved yet

### LINKED ISSUE #2479: AI Assitant crashes on a job that is not saved yet
After creating a new job on the workflow and before saving it, the AI Assistant crashes when the user tries to send an AI query.

**Version number** 
v2.9.0

- [x] Yes
- [ ] No

scenario A)

Create a new project
Create a new workflow but don't save it
Click on the job and on AI Assistant tab
It blinks and shows logs an exception when sending an AI question/query
scenario B)

On an existing workflow
Create a job but don't save it
Click on the job and on AI Assistant tab
It blinks and shows logs an exception when sending an AI question/query

**Expected behavior**

Depends on the requirement. A suggestion is to disable the AI Assistant tab when the job is not yet saved.

midigofrank:
### Description

This PR disables AI assistant submit button incase the job isn't saved yet

Closes #2479 

### Validation steps
You need to have `OPENAI_API_KEY` and `APOLLO_ENDPOINT` envs in order to validate this

1. Open up an existing workflow and add a new job. Alternatively, you can create a new workflow and open up the inspector
2. In the AI assistant tab, you'll see that the button to send the message is disabled. When you hover the button you'll get a tooltip saying `Save the job first in order to use the AI Assistant`

### Pre-submission checklist


josephjclark:
Just need to flag that this may conflict with #2478

test/support/workflow_live_helpers.ex
```diff
@@ -201,10 +201,10 @@ defmodule Lightning.WorkflowLive.Helpers do
     |> render_hook("push-change", %{patches: patches})
   end
 
-  def add_job_patch(name \\ "") do
+  def add_job_patch(name \\ "", id \\ Ecto.UUID.generate()) do
```
stuartc, line 204: TIL that putting a function call inside a default _doesn't_ evaluate it on compilation. 

I totally expect that function to be called and all default values would be the same.

stuartc (approved):

PR #2496, opened 2024-09-13
Unlock Inspector Page for Multi-User Step Retries and Work Orders

### LINKED ISSUE #2282: Preventing simultaneous editing should not stop users from running workflows
This isn't quite a bug, but a pretty high priority issue. In #2241 we were attempting to prevent a possible data loss situation where two edits would overwrite one another.

That feature was all about preventing simultaneous _editing_ of a workflow. As it was delivered, however, it also included code that prevents user B from _running_ a workflow while user A is looking at the same workflow.

_We don't want this. It's OK for user B to run a workflow while user A is looking at that workflow._

What's more, with the current implementation it's still possible for user B to run the workflow while user A is looking at it, but they're forced to go on a bit of a journey... they're only blocked from running the workflow from the Inspector interface and they're forced to navigate to the History page before clicking "retry" or to CURL the webhook URL to create a new workorder.

We may find that the fix is as simple as removing the `|| !@has_presence_edit_priority` from this code:
```ex
disabled={
  @save_and_run_disabled ||
  @snapshot_version_tag != "latest" ||
  !@has_presence_edit_priority
}
```
But careful testing should be done.

Note that the run button on the Inspector may be built in such a way that it always "saves and then runs". It will probably Just Work :tm: once we remove the `has_presence_edit_priority` bit since we’re already blocking edits to the editor—so they won't have updated anything even if they "save and run", but I’d rather take some more time to guarantee this, maybe through tests?

elias-ba:
### Description

This PR allows users with read-only access to the workflow canvas / inspector to be able to rerun the steps or create new work orders without having edit access.

Closes #2282 

### Validation steps

1. Make sure you have two users who have access to the same workflow and some runs of that workflow in the history page.
2. Log in with the first user and open the workflow canvas.
3. Log in with the second user (maybe in incognito window?).
4. With the second user, visit the history page and inspect a step of a run of the workflow.
5. Normally you now have access to the inspector page with a step loaded in it but with read-only access.
6. Click on the 'Retry run from here' button and note that it is not disabled and the step is retried successfully.
7. Click on the 'Create new work order' button and note that it is not disabled and that a new work order is created successfully.

### Pre-submission checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,8 +12,23 @@ Closes #__
 
 1. *(Is there anything else the reviewer should know or look out for?)*
 
+## AI Usage
+
+Please disclose how you've used AI in this work (it's cool, we just want to know!):
+
+- [ ] Code generation (copilot but not intellisense)
+- [ ] Learning or fact checking
+- [ ] Strategy / design
+- [ ] Optimisation / refactoring
+- [ ] Translation / spellchecking / doc gen
+- [ ] Other
+- [ ] I have not used AI
+
+You can read more details in our [Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)
+
 ### Pre-submission checklist
 
 - [ ] I have performed a **self-review** of my code.
 - [ ] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
 - [ ] I have updated the **changelog**.
+- [ ] I have ticked a box in "AI usage" in this PR
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,12 +24,16 @@ and this project adheres to
 
 ### Changed
 
+- Notify other present users when the promoted user saves the workflow
+  [#2282](https://github.com/OpenFn/lightning/issues/2282)
+
 ### Fixed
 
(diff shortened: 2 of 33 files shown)


christad92 (dismissed):
Tested with Elias and looks good!

Thanks

midigofrank (approved):

christad92 (changes requested):
ONLY the editor panel should be disabled. users should be able to provide inputs or select input for a new workorder

elias-ba:
@christad92 thanks for the review, I have enabled the dataclip form in my last commit. Do you wanna pull and do another round of testing please ? Thanks and let me know !

taylordowns2000:
@christad92 @elias-ba can you close or mark this PR as draft until it's ready?

elias-ba:
@christad92 for the AM, please can you help QA this. I have implemented the option 1 described [in this comment here](https://github.com/OpenFn/lightning/issues/2282#issuecomment-2371128630).

christad92 (changes requested):
This looks good so far. 

When a change is published, the message should be _"This workflow has been updated. You're no longer on the latest version"_.

For a weird moment, 2 simultaneous editors could edit a workflow at the same time.
image

elias-ba:
@christad92 thanks for the correct flash message. I updated it.

christad92 (approved):
Thanks

PR #2502, opened 2024-09-16
Fix Buggy Change Email Flow

### LINKED ISSUE #2365: Buggy change email flow
Looks like two bugs here:
1. "Passwords don't match" text is being displayed. This is neither proper english (there are not two passwords, there is just one password) nor good UX! We don't want to test the password against the crypted pass in the DB on each keystroke. Think about a "login" form... you only get feedback on the password when you hit submit.
2. Both the "warning: your email was updated" and the "please confirm this new email" go to the old email address

https://www.loom.com/share/0ec5c6a587f84087a1507f2afc650807

elias-ba:
### Description

Closes #2365 

### Validation steps

### Pre-submission checklist

### DIFF
--- a/lib/lightning/accounts.ex
+++ b/lib/lightning/accounts.ex
@@ -513,7 +513,11 @@ defmodule Lightning.Accounts do
     )
 
     UserNotifier.deliver_update_email_instructions(
-      user,
+      %User{
+        email: current_email,
+        first_name: user.first_name,
+        last_name: user.last_name
+      },
       update_email_url_fun.(encoded_token)
     )
   end
--- a/test/lightning/accounts_test.exs
+++ b/test/lightning/accounts_test.exs
@@ -16,6 +16,7 @@ defmodule Lightning.AccountsTest do
 
   import Lightning.AccountsFixtures
   import Lightning.Factories
+  import Swoosh.TestAssertions
 
   test "confirmation_required?/1 returns false for users who are already confirmed" do
     user = insert(:user, confirmed_at: DateTime.utc_now())
@@ -779,11 +780,13 @@ defmodule Lightning.AccountsTest do
     end
 
     test "sends token through notification", %{user: user} do
+      new_email = "current@example.com"
+
       token =
         extract_user_token(fn url ->
           Accounts.deliver_update_email_instructions(
             user,
-            "current@example.com",
+            new_email,
             url
           )
         end)
@@ -794,8 +797,23 @@ defmodule Lightning.AccountsTest do
                Repo.get_by(UserToken, token: :crypto.hash(:sha256, token))
 
       assert user_token.user_id == user.id
-      assert user_token.sent_to == "current@example.com"
+      assert user_token.sent_to == new_email
       assert user_token.context == "change:#{user.email}"
+
+      assert_email_sent(
+        subject: "Your OpenFn email was changed",
+        to: Swoosh.Email.Recipient.format(user)
+      )
+
+      assert_email_sent(
+        subject: "Please confirm your new email",
+        to:
+          Swoosh.Email.Recipient.format(%User{
+            email: new_email,
+            first_name: user.first_name,
+            last_name: user.last_name
+          })
+      )
     end
   end
 


midigofrank (changes requested):
Nicely done @elias-ba , I've left a few comments

lib/lightning/accounts.ex
```diff
@@ -513,18 +513,22 @@ defmodule Lightning.Accounts do
     )
 
     UserNotifier.deliver_update_email_instructions(
-      user,
+      %User{
+        email: current_email,
```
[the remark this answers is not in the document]
elias-ba: Done

christad92:
@elias-ba I will review this today.

lib/lightning/accounts.ex
```diff
@@ -513,18 +513,22 @@ defmodule Lightning.Accounts do
...
+        first_name: user.first_name,
+        last_name: user.last_name
+      },
       update_email_url_fun.(encoded_token)
     )
   end
 
-  def validate_change_user_email(user, params \\ %{}) do
+  def validate_change_user_email(user, params \\ %{}, validate_password \\ false) do
```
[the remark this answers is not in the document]
elias-ba: Hey Frank, thanks for this. I am also not a huge fan of adding boolean params. But your suggestion here won't fit the business needs.

What we need is to validate the password in specific contexts (like form submission) regardless of whether the password field has changed. Your approach would trigger validation whenever there's a change to the password field, which would be same as the bug we are fixing here (we are avoiding too many database calls).

But I have reduced the ugliness of the function taking extra boolean param, by just using keyword arguments instead. This way, we can maintain the necessary control over when password validation occurs while improving the function's readability. Here's how it looks now:

```elixir
validate_change_user_email(user, params, opts \\ []) do
  # ... existing code ...
  |> maybe_validate_password(user, opts[:validate_password] || false)
end
```

This approach allows us to meet our specific business requirements while addressing some of the concerns about function signature clarity. What do you think about this compromise?

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Fix Buggy Change Email Flow
```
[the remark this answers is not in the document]
elias-ba: Done

elias-ba:
> @elias-ba I will review this today.

@christad92 hit me up if you need any help.

elias-ba:
@midigofrank wanna give me another round of review? Thanks man

midigofrank (approved):

lib/lightning/accounts.ex
```diff
@@ -492,44 +494,55 @@ defmodule Lightning.Accounts do
...
+        first_name: user.first_name,
+        last_name: user.last_name
+      },
       update_email_url_fun.(encoded_token)
     )
   end
 
-  def validate_change_user_email(user, params \\ %{}) do
+  def validate_change_user_email(user, params \\ %{}, opts \\ []) do
+    validate_password? = Keyword.get(opts, :validate_password, false)
```
stuartc, line 528: Why do we conditionally validate the password?

lib/lightning/accounts.ex
```diff
@@ -492,44 +494,55 @@ defmodule Lightning.Accounts do
...
-    |> Ecto.Changeset.validate_format(:email, ~r/^[^\s]+@[^\s]+$/,
+    |> Changeset.cast(params, Map.keys(types))
+    |> Changeset.validate_required([:email])
+    |> Changeset.validate_format(:email, ~r/^[^\s]+@[^\s]+$/,
       message: "must have the @ sign and no spaces"
     )
-    |> Ecto.Changeset.validate_length(:email, max: 160)
-    |> Ecto.Changeset.validate_change(:email, fn :email, email ->
+    |> Changeset.validate_length(:email, max: 160)
+    |> validate_email_change(user)
```
stuartc, line 540: We already have functions in User that should validate the email so it looks like parts of this function and `validate_email_change/2` are redundant. It's been this way for a while, could you use a function inside `User` instead and ensure the behaviour is the same.

lib/lightning/accounts.ex
```diff
@@ -492,44 +494,55 @@ defmodule Lightning.Accounts do
...
 
-      iex> deliver_update_email_instructions(user, current_email, &Routes.user_update_email_url(conn, :edit, &1))
+      iex> deliver_update_email_instructions(user, new_email, &Routes.user_update_email_url(conn, :edit, &1))
       {:ok, %{to: ..., body: ...}}
 
   """
   def deliver_update_email_instructions(
         %User{} = user,
-        current_email,
+        new_email,
```
stuartc, line 503: Instead of passing in just the email, could we pass in an unpersisted User struct. That way the user will get the email with their name on it as well instead of just `their@email.com`.

stuartc (changes requested):

christad92 (approved):
Thank! tested and it works.

lib/lightning/accounts.ex
elias-ba: @stuartc the only reason why we are validating conditionally is to avoid validating the password after every keystroke. So I introduced the `validate_password` flag to be able to use the same function during form validation function and during form submission.

lib/lightning/accounts.ex
elias-ba: This is done, thanks

lib/lightning/accounts.ex
elias-ba: @stuartc I don't quite understand relevancy of this comment (sorry I might be missing something). This function is a wrapper around the `UserNotification.deliver_update_email_instructions/2` which takes a User struct and a URL. The `Accounts.deliver_update_email_instructions/3` function takes a User struct, an email string, and a URL string, build the token, and assigns the new email to the User struct before calling the `UserNotification.deliver_update_email_instructions/2`. So I don't understand why we need to have another User struct.

```
  def deliver_update_email_instructions(
        %User{} = user,
        new_email,
        update_email_url_fun
      )
      when is_function(update_email_url_fun, 1) do
    {encoded_token, user_token} =
      UserToken.build_email_token(user, "change:#{user.email}", new_email)

    Repo.insert!(user_token)

    UserNotifier.deliver_update_email_warning(
      user,
      new_email
    )

    UserNotifier.deliver_update_email_instructions(
      %User{
        email: new_email,
        first_name: user.first_name,
        last_name: user.last_name
      },
      update_email_url_fun.(encoded_token)
    )
  end
```

PR #2510, opened 2024-09-17
Add quotes when special YAML characters are present in the exported project

### LINKED ISSUE #2446: Don't generate invalid yaml
### Projects having workflows with special YAML characters like `:` fails to pull

Yaml does not do well with special characters and that means cli pull and deploy operations will always run into an error when a project has workflows with those characters in names. 

There are two ways to do this, enforce some data validation on lightning or do data manipulation. I'd default validation because data manipulation might be expensive with multiple edge cases to deal with/test.

midigofrank:
### Description

This PR adds quotes to special YAML characters when exporting the project into YAML.

It uses a regex to do the validation.

A valid YAML key:
- starts with alphanumeric
- followed by alphanumeric or hyphen or underscore or @ or .
- ends with alphanumeric

A valid YAML value:
- starts with alphanumeric
- followed by alphanumeric or hyphen or underscore or @  or . or space
- ends with alphanumeric

This means if a job name / workflow name begins with a space, it will get quoted in the exported YAML

Closes #2446

### Validation steps

1. In your project, try renaming the project/workflow/job/edge condition names  to contain a character like `:`. Or maybe you append/prepend a space to the name and save
2. Head to the project settings and click `export project`.
3. You'll see that names with special characters have been quoted.
4. Try using the `cli` to pull that particular project, i.e. `openfn pull <project id>`
5. You should get the same result as step 3

Initially, `cli pull` would fail with this error:
```sh
[CLI] ♦ Downloading the project spec (as YAML) from the server.

[CLI] ✘ Command failed!
[CLI] ✘ Error: Document with errors cannot be stringified
    at Document.toString (/opt/hostedtoolcache/node/18.20.4/x64/lib/node_modules/@openfn/cli/node_modules/yaml/dist/doc/Document.js:321:19)
    at extractJobsToDisk (file:///opt/hostedtoolcache/node/18.20.4/x64/lib/node_modules/@openfn/cli/node_modules/@openfn/deploy/dist/index.js:626:14)
    at async pullHandler (file:///opt/hostedtoolcache/node/18.20.4/x64/lib/node_modules/@openfn/cli/dist/process/runner.js:1587:25)
    at async parse (file:///opt/hostedtoolcache/node/18.20.4/x64/lib/node_modules/@openfn/cli/dist/process/runner.js:1764:12)
```
  

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Add quotes when special YAML characters are present in the exported project
+  [#2446](https://github.com/OpenFn/lightning/issues/2446)
+
 ## [v2.9.4] - 2024-09-16
 
 ### Changed
--- a/lib/lightning/export_utils.ex
+++ b/lib/lightning/export_utils.ex
@@ -163,9 +163,9 @@ defmodule Lightning.ExportUtils do
 
   defp yaml_safe_string(value) do
     # starts with alphanumeric
-    # followed by alphanumeric or hyphen or underscore or @ or . or space
+    # followed by alphanumeric or hyphen or underscore or @ or . or > or space
     # ends with alphanumeric
-    if Regex.match?(~r/^[a-zA-Z0-9][a-zA-Z0-9_\-@\. ]*[a-zA-Z0-9]$/, value) do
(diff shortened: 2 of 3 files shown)


stuartc (approved):

christad92:
@midigofrank I have tested this we need to add the condition expressions in quotes ("").  If you put a condition expression that has && in it and try a pull, you'll see the order is distorted and when you try a deploy, it removes half of the expression.

PR #2516, opened 2024-09-19
When the selected job was not run, do not say the dataclip is wiped

### LINKED ISSUE #2303: Update "No Input Data Here" message when an inspected job was not executed in a run.
When trying to view a job that is not included in a run, the input and output tabs (on the inspector) expectedly do not show any data as there is nothing to show. But the message we display to the user tells them they have enabled zero persistence when that is not the case. 

We need to fix the message to show the user that "This job was not included in this Run. Select a step from this run in the Run tab or create a new work order by clicking here to select/build an input" 

Image

midigofrank:
### Description

This PR fixes an issue on the dataclip selector viewer whenever a run is followed and the selected job isn't part of the run.

Initially, all it was saying is the dataclip is wiped, but now, it checks to see if the `step` exists or not before making the conclusion on whether it was wiped or not

Closes #2303 

### Validation steps

1. Ensure you have a workflow with multiple steps. The `openhie-project` can do
2. Make the workflow crash in the first job. A simple typo in the job code would do
3. Now get the `run id` and open any of the subsequent jobs in that workflow and add `a={run_id_here}` to the query params.
4. You'll see that on the dataclip viewer, the message says that the job wasn't run. Attached is a screenshot of the same
Screenshot 2024-09-19 at 15 18 55

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to
 
 ### Fixed
 
+- Dataclip selector always shows that the dataclip is wiped even when the job wasn't run
+  [#2303](https://github.com/OpenFn/lightning/issues/2303)
+
 ## [v2.9.5] - 2024-09-18
 
 ### Changed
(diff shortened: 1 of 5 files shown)


stuartc (approved):

christad92:
@midigofrank please fix the `was not/is not` in the sentence here. can we make it `was not` since all runs are in the immediate past? 
image

midigofrank:
So the problem is that the run could still be in progress

christad92:
So far the run is in progress... it is in the "immediate" past :-)

PR #2532, opened 2024-09-26
2531 do not track timestamps

### LINKED ISSUE #2531: Remove locking of trigger to update timestamps
We lock the trigger so that we can track timestamps received for partitions and avoid race conditions. Unfortunately, this may be the cause of intermittent deadlocks. Remove the tracking of partition timestamps to see if it resolves the issue.

rorymckinley:
### Description

Existing KafkaTrigger code tracks the last timestamp received per partition to cater for cases where a Kafka Trigger has been disabled long enough to exceed the cluster's offset retention period. If the trigger is enabled after this period, the cluster will revert to whatever the originally configured reset was (e.g. `earliest') and this may result in a duplicate messages or data loss.

By persisting the timestamp the trigger can communicate a minimum start point that reflects the actual data received thereby minimising duplication in the above scenario. Unfortunately, to track this without race conditions the trigger record must be locked for the update and intermittent DB deadlocks were observed when testing in a multi-node environment.

The decision was made to remove this functionality and, for now, include documentation describing the risks of the trigger being disabled for an extended period. This is not a great solution by any means, but it is deemed acceptable while the Kafka trigger remains an experimental feature. As the Kafka triggers mature to the point where more general availability is possible, it will be necessary to add some form of alert in the UI that will be displayed if a user disables a trigger that was previously enabled.

### Testing

- Setup a local Kafka cluster based on these [instructions](https://github.com/OpenFn/lightning/blob/main/kafka_testing/KAFKA_ARCHITECTURE.md#testing-locally-using-the-docker-kafka-cluster) but use a **different docker-compose file**: `kafka_testing/docker-compose.kafka-testing-forgetful-cluster.yml`. This docker compose file will setup a cluster configured to 'forget' consumer offsets after 5 minutes.
- Setup a Kafka trigger via the UI as follows (note that to make this test easier, we set the 'Initial offset reset policy' to `earliest`:

image

- Publish a small number of messages to the cluster : `kafka_testing/lots_of_messages_sans_headers.sh 5`
- Wait until these messages reflect in the Workflow history and then, open an IEx session.
- Start by pulling some metadata re: the messages

```
alias Lightning.Repo
alias Lightning.KafkaTriggers.TriggerKafkaMessageRecord
before_tpos =  TriggerKafkaMessageRecord |> Repo.all() |> Enum.map(&(&1.topic_partition_offset)) |> Enum.sort()
```

- Disable the trigger via the UI.
- Clear any existing TriggerKafkaMessageRecord instances:

```
TriggerKafkaMessageRecord |> Repo.delete_all()
```

- Let at least 5 minutes pass..... and then re-enable the trigger via the UI.
- Wait until you see that more WorkOrders have been enqueued and then run the following from your IEx session:

```
alias Lightning.Repo
alias Lightning.KafkaTriggers.TriggerKafkaMessageRecord
after_tpos =  TriggerKafkaMessageRecord |> Repo.all() |> Enum.map(&(&1.topic_partition_offset)) |> Enum.sort()
after_tpos == before_tpos # should return true
```

Closes #2531 

### Pre-submission checklist


jyeshe (approved):
For the experimental purpose looks good. Can you provide another PR persisting the message with some metadata to identify the latest successful @rorymckinley (provided it halts on failure)? It needs to be able to reset to a certain offset, if not the oldest offset failure that was not processed yet.

rorymckinley:
> For the experimental purpose looks good. Can you provide another PR persisting the message with some metadata to identify the latest successful @rorymckinley? As long as it halts on failure, it needs to be able to reset to a certain offset, if not the oldest offset failure that was not processed yet.

Thanks @jyeshe - the 'handling persistence failure' work is still in progress - it will most likely need to change as a result of this change.

stuartc (approved):

Issue #2386, opened 2024-08-14
Allow Kafka pipeline to rollback offset on persistence failure


rorymckinley:
The design of the BroadwayKafka producer means that it always acknowledges the offset even if the pipeline did not produce a successful result. 

In the case where we have had a transient database failure, we may need to request a message again to try and reprocessing.

rorymckinley:
No further work is required on this.

PR #2568, opened 2024-10-08
Notification for Kafka persistence failures.

rorymckinley:
### Description

This PR adds functionality to send certain users associated with a project a notification that a Kafka message could not be persisted to the DB. 

When a message fails to persist, the code will send the associated users an email. It will also keep track of when the email has been sent. By tracking the time that the email was sent, we can ensure that the recipients do not get spammed with multiple emails if the problem persists. There is a configurable embargo window that has to pass before another email is sent.

The timestamp is also published as an event to ensure that other nodes in the cluster use this timestamp when calculating if they should be sending another email. 

Should other nodes fail to receive the  timestamp, due to some form of cluster event, the worst that will happen is that the recipients may receive a small number of duplicate notifications.

The changes to `KafkaTriggers.Pipeline` and `KafkaTriggers` are probably a good place to start.

NOTE: This PR does not address the persistence of failed messages to the local file system, that will be addressed in a subsequent PR.

 #2386 

### Validation steps

- Setup a local Kafka cluster based on these [instructions](https://github.com/OpenFn/lightning/blob/main/kafka_testing/KAFKA_ARCHITECTURE.md#testing-locally-using-the-docker-kafka-cluster)
- Setup an enabled Kafka trigger as follows:

image

In an IEx session:

```
import Ecto.Query
alias Lightning.Repo
alias Lightning.KafkaTriggers.TriggerKafkaMessageRecord

TriggerKafkaMessageRecord |> Repo.delete_all()
%{id: trigger_id} = (from t in Trigger, where: t.enabled == true and t.type == :kafka, order_by: [desc: t.inserted_at]) |> Repo.one()
:persistent_term.put({:kafka_trigger_test_failure, trigger_id}, true)
```

In the terminal, add 5 messages to the cluster:

```
kafka_testing/lots_of_messages_sans_headers.sh 5
```

Give the trigger a couple of seconds to process the messages, and then navigate to http://localhost:4000/dev/mailbox.

You should see the mails in the mailbox, and the URL in the mail body should take you to the workflow with the affected Kafka trigger.

## AI Usage

- [ ] Strategy / design
- [x] Other
- [ ] I have not used AI

### Pre-submission checklist


lib/lightning/kafka_triggers.ex
```diff
@@ -191,4 +194,57 @@ defmodule Lightning.KafkaTriggers do
...
+
+      track_notification_sent(trigger_id, sending_at)
+
+      notify_any_other_nodes(trigger_id, sending_at)
+    end
+  end
+
+  defp last_notification_sent_at(trigger_id) do
+    :persistent_term.get(failure_notification_tracking_key(trigger_id), nil)
+  end
```
stuartc, line 212: Persistent terms are localised to a single node, how are we guaranteeing that the calls to this function originate from the same process?

rorymckinley: @stuartc We are not. This means that there is a chance that the Pipeline on another node also wants to send a notification before it has received the [event](https://github.com/OpenFn/lightning/pull/2568/files#diff-5277a74a56ffb8714c3f4d6ae7afed0781ff27eef37ada7cdd8e606d71392363R34) notifying it that a notification has been sent.

In the worst case, this would mean that there would be a duplicate message for each node in the cluster. However, this risk exists already - at a high enough message rate, it is conceivable that each node experiences such an error simultaneously and they all send the notifications. 

However, once the event is received by all nodes, the window for such duplicated notifications will close. If a node never receives the event, then it will be vulnerable to sending a duplicate notification, but then it will also update the local :persistent_term and will not send any further.

Given that the probability of a persistence error is currently hard to quantify, I elected to bias towards a simpler process that has the risk of sending a small number of duplicate messages rather than sending none at all. In fact, this can be considered the principle for all the code I am going to want to merge in for persistence error handling: 'Messy, but hopefully robust' :).

Note: I have marked this PR as draft for now as I am awaiting feedback on the email contents, but please feel free to add further comments.

stuartc: Ok cool, yeah I agree generally with your approach - we _do_ have the rate limiter used to debounce emails. So that is also an option - but since this is an edge case and has a limited impact I'm comfortable with this.

rorymckinley: @stuartc Ah, I was not aware that there was existing functionality. Do you mean [this](https://github.com/OpenFn/lightning/blob/main/lib/lightning/pipeline/failure_alerter.ex#L36)?

jyeshe (approved):
I would rather go with an automated offset reset but there might be some business requirement I am unaware of so for it's purpose this looks good to me.

lib/lightning/kafka_triggers.ex
stuartc: Yes thats it, well more specifically wrapping calls in `Hammer.check_rate` - under the hood it uses mnesia to keep counts in buckets.

rorymckinley: Ah, cool. I did originally consider Hammer but it seemed overkill for what I needed. Given that we already have the plumbing in place, I can take another look.

Happy for me to do so in a follow-up PR?

lib/lightning/accounts/user_notifier.ex
```diff
@@ -107,14 +107,15 @@ defmodule Lightning.Accounts.UserNotifier do
   """
   def deliver_project_addition_notification(user, project) do
     role = Projects.get_project_user_role(user, project) |> Atom.to_string()
-    url = LightningWeb.RouteHelpers.project_dashboard_url(project.id)
```
stuartc, line 110: @rorymckinley just noting here that I changed the url functions here across this file to use verified routes, so we don't need to:

- use the `RouteHelpers` module anymore
- provide an Endpoint to the `url` function, since the `~p` ast carries the endpoint on it.

test/lightning_web/live/workflow_live/edit_test.exs
```diff
@@ -2110,22 +2108,22 @@ defmodule LightningWeb.WorkflowLive.EditTest do
       assert high_priority_view
              |> has_element?("#inspector-workflow-version", "latest")
 
-      dynamically_absorb_delay(
-        fn ->
-          !(low_priority_view
-            |> has_element?("#inspector-workflow-version", "latest"))
-        end,
-        sleep: 2
-      )
```
stuartc, line 2119: @rorymckinley I'm trying something new here, you're not the first and won't be the last to deal with tricky timing issues on tests, the `retry` library has been in the back of my mind for a while, thought it would be worth taking it for a spin.

PR #2581, opened 2024-10-15
Fix work order retry sorting and avoids loading dataclips

### LINKED ISSUE #2582: Retrying work orders doesn't respect the order of existing/base ones
**Describe the bug** 

This is a flaky that gladly happened more than once:
```
  1) test retry_many/2 for WorkOrders retrying multiple workorders preserves the order in which the workorders were created (Lightning.WorkOrdersTest)
     test/lightning/work_orders_test.exs:1558
     Expected truthy, got false
     code: assert retry_run_1.inserted_at |> DateTime.before?(retry_run_2.inserted_at)
     stacktrace:
       test/lightning/work_orders_test.exs:1628: (test)
```

The work orders should be retried at the same order of original ones. It's not sorting by `inserted_at`.

2.9.9

- [x] Yes
- [ ] No

It's hard to reproduce but this test failed multiple times on the CI and the query doesn't have an order by.

jyeshe:
### Description

This PR fixes the sorting of retried work orders. It also prevents from loading dataclips to memory once each of them can have MB magnitude and a retry might enqueue thousands of work orders.

Closes #2582 

### Validation steps

Run the test case from the error multiple times and check that the code now has a `order_by`.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### Pre-submission checklist


stuartc (approved):
Nice one 🙌 Fyi, you don't need to avoid preloading Dataclips - by default the body is not returned in any preload for data clips, so the size of the records that come down is no bigger than any other table. But yeah this is a nice change, easier to understand too. 👍

jyeshe:
> Nice one 🙌 Fyi, you don't need to avoid preloading Dataclips - by default the body is not returned in any preload for data clips, so the size of the records that come down is no bigger than any other table. But yeah this is a nice change, easier to understand too. 👍

`field :body, :map, load_in_query: false` :+1:

PR #2646, opened 2024-11-07
Replace DB transaction with on demand stream

### LINKED ISSUE #2645: Make Lightning resilient to long collection fetch
### User story

By the original design, the data should be streamed from the database so that it can be streamed to the client/worker.

The current implementation was made available for the worker tests however it exposes Lightning to high user limits blocking the database connection (a limited resource, currently set to 20) for more than 15 seconds on AWS Lagos setup.

In order to avoid any concern about the progress of the consumption of the streamed data by the worker or any other client, in case of heavy workload or high custom limit, the backend should release database connection as possible instead of holding it for the whole duration of the streaming. This is particularly important as well for deployments involving databases with low throughput. 

### Details

- Database streaming can be switched to regular query
- The abstraction is applicable to both streaming all and matches.

### User acceptance criteria

The cursor and limits continue working the same, i.e the user limit on the query param is respected.

jyeshe:
### Description

This PR changes the way that the items are streamed to the reducer. Instead of relying on the database streaming, which blocks a transaction or a connection, now that it accepts a custom limit it's safer to do iterations that might block only on the final part of response processing. 

The iterations are now composed by:
1. Fetching some database data (up to a max limit of 5k items)
2. Right after the fetching the database connection is released back to the pool
3. Stream to network in chunks up to user limit
4. If the user limit is higher than 5k, it continues from the step 1, otherwise it finishes the chunking.

Closes #2645

### Validation steps

- Repeat same requests with cursor and limits as before.

### Additional notes

The database streaming is replaced by regular fetching once the data is streamed on chunks.
 
## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to
 
 ### Changed
 
-- Make Lightning resilient to long collection fetch
+- Reduce transaction time when fetching collection items by fetching upfront
   [#2645](https://github.com/OpenFn/lightning/issues/2645)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to
 
 ### Changed
 
+- Make Lightning resilient to long collection fetch
+  [#2645](https://github.com/OpenFn/lightning/issues/2645)
+
 ### Fixed
 
 ## [v2.10.0-rc.0] - 2024-11-07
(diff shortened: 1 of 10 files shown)


jyeshe:
Flaky on
  1) test complete_step/2 wipes the dataclip if erase_all retention policy is specified at the project level when the run is created (Lightning.RunsTest)
     test/lightning/runs_test.exs:290
     Assertion with == failed
     code:  assert step.output_dataclip.wiped_at == DateTime.utc_now() |> DateTime.truncate(:second)
     left:  ~U[2024-11-07 15:40:32Z]
     right: ~U[2024-11-07 15:40:33Z]
     stacktrace:
       test/lightning/runs_test.exs:331: (test)

stuartc (approved):

PR #2650, opened 2024-11-09
Stay on-palette?

taylordowns2000:
@stuartc , what do you think about this? I have to admit I found the blue-to-purple a little jarring—not because it's not a fine transition in another context, but because neither is part of our palette—AFAIK.

I'm all for introducing a new color (or even two!) but if we do we should make a proper design rule for it and use it for a certain type of thing. As far as I can tell these are the only things that use blue and purple backgrounds.

While the below is certainly boring, at least it feels like we're presenting a consistent design:

image


josephjclark:
Ship it!

Please!!

stuartc (approved):

Issue #2627, opened 2024-11-04
Inspector: improve code assist UX


josephjclark:
### User story

The UX for code assist in Monaco in the inspector is bad, actually. This issue is a little wishlist of features/improvements that we might get to one day.

![Image](https://github.com/user-attachments/assets/dba0ee89-ee1d-422a-a018-b0ad448b9ecf)

* Typescript Signatures appear at the top of the prompts, but they are not particularly helpful for our users 
* We force ALL common exports into the global namespace so that they appear - but most adaptors do not export all common functions
* We show JSdoc comments, badly formatted, and with lots of annotations that don't help the user.
* We should show the first line of description to the right of the suggestion (sometimes it'll show a ts signature there - not useful)
* It can be kinda hard and fiddly to get docs for the parameters
* When a function is inserted, we should also insert the opening bracket, so that the code assist is instantly in the right place
* You can open a small suggestion prompt with ctrl-space. To get the details panel on the right, you have to ctrl + space again. This is annoying and intuitive. I'd like to work out a better way to show some help for this

### Notes

Editing the appearance of the code assist is hard. We might have to switch off all default language features and develop our own custom suggestions (I'm worried about what this might mean for stuff like object properties).

The ideal solution would be to take the default suggestions and customise them - removing stuff that isn't `@public` for example, or updating descriptions.

This might be useful hack to this end:
```
const x = editor.getContribution('editor.contrib.suggestController')
```
This gets the suggestion controller component right outta Monaco. You can then inspect it in the debugger and trace through all sorts of interesting things, including a couple of callbacks.

So far I haven't found the exact callbacks I think we need, but I wanted to record the find.

josephjclark:
I have had to restore the JSDoc attributes  in #2669

Maybe for now we have to live with jsdoc attributes appearing in the suggestions, no matter how untidy they are. 

This speaks to the idea that we need to generate our own suggestions from scratch and not use the default service.

PR #2662, opened 2024-11-12
Fix code assist modules

### LINKED ISSUE #2641: Intellisense seems to be broken (see magic function for DHIS2)
When using the DHIS2 adaptor with the credentials well configured and metadata loaded, the IntelliSense did not load. Even when you try to manually trigger it by entering ctrl/cmd + space, it doesn't load.

josephjclark:
### Description

This PR fixes an issue where code assist is not available for the second argument of dhis2.create.

Closes #2641

Improves #2627

### Implementation nodes

Global types declared in an adaptor's `types.d.ts` are not made visible to each adaptor in Monaco, as a result of changes made in #2433 where each file is loaded into its own module.

We've also got problems where functions which are not exported from an adaptor are exposed as code-complete suggestions. 

I've had to do a bunch of things:

* Everything from common is loaded into a single uber-module called `@openfn/language-common`. None of these definitions are exposed globally to the script.
* BUT imports from `@openfn/common` now work organically inside our adaptor's typings, so this is good
* The main adaptor loads each file into its own module, like `@openfn/language-dhis2/Util.d.ts`. These are also scoped and not available globally
* EXCEPT for index.d.ts and adaptor.d.ts. These are loaded into the global namespace, which means all main adaptor exports and namespaces declared in index.d.ts are loaded globally and available to the user
* To make this work in Monaco, we have to do some nasty regex refactoring of the code (see commends in `Editor.tsx:154`-ish)
* This gives us a much more realistic projection of the adaptor's types into Monaco
* BUT global types still won't load. In Adaptor.d.ts we refer to `Dhis2Data`, a "global" type which, because of how we load modules, is loaded into its own private scope. So after all this work, `dhis2` still doesn't work.
* As a workaround, we force the contents of global types in `types.d.ts` into each module scope.

And this works! :tada: 

Note that global types in other d.ts files still won't work. I've hard-coded support for `types.d.ts`.

### QA

I really need help with QA on this!

First, please test dhis2.

Load up a dhis2 job and type `create('trackedEntityInstance', {})`. You should get code assist on the object. You do NOT need metadata or a valid credential (but if you have one, you should get smart metadata too).

Then please test around some other adaptors.

- Pick an adaptor
- Open its adaptor docs in a new window
- Check that functions and namespaces exported by that adaptor are available in code assist
- You might also check whether you get code assist on the arguments to each function (many adaptors will not)
- Check that functions from common are not exported (unless the adaptor explicitly exports them)

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


josephjclark:
For DHIS2 I get a decent list of suggested functions:

image

And I get code assist on the object of `create`:

image

josephjclark:
For `fhir-ndr-et` - which has some hairy code assist stuff, I get reasonable top level suggestions:

image

I get code assist on profile names:

image

And I get code assist on Patient properties:

image

I do NOT get code assist on resources with multiple profiles, but  this is a known issue

josephjclark:
In common I get code assist for http.get:

image

Although I do not see `http` listed at the top level :thinking: 

image

I do not get code assist on http.options  - but I don't in production

image

That's a bit wierd because there IS a typing in the d.ts file

josephjclark:
Ah, I think the problem in common is that I'm loading common/util/http.d.ts (which is not well typed) and overwriting common/http.d.ts, which IS.  There's sort of a naming conflict. I might try ti fix this here tomorrow.

josephjclark:
We also don't get typings on http's params argument.

but that's because the d.ts file is basically broken:
```
export type RequestOptions = any;
```
This is a bit of a problem - we should be able to provide good code assist on this stuff.

Sigh.

stuartc (approved):
Works locally for me, this seems pretty reasonable for now and the logic is nicely commented and isolated 👌

PR #2667, opened 2024-11-13
2636 reduce usage tracker memory usage

### LINKED ISSUE #2636: Usage tracker can cause OOM restarts
This is a placeholder issue until we have more data. For now, it seems that we can run into memory problems when the Lightning.UsageTracker.ReportWorker runs.

In Oban, the jobs appear to be "executing" when in reality they are all orphaned after they (most likely) are the cause of an OOM restart.

rorymckinley:
### Description

This is a blind attempt to reduce memory usage when generating UsageTracking reports. I have taken an educated guess at a possible source of the issue but the proof of the pudding would be to run the reports on the affected system and see if memory issues are addressed.

At the very least there is now an ENV var that can be used to control how many Run instances get streamed to the report at any given time, which will hopefully allow memory usage to be reduced to a manageable level.

There is a fair bit going on, so I grouped the code changes into three commits:

- Introduce the streaming code (as well as some additional tests)
- Remove redundant tests
- Remove redundant code

Closes #2636 

### Validation steps

- Start a local copy of ImpactTracker on port 4001.
- Start with an empty DB for Lightning
- Ensure that `USAGE_TRACKING_ENABLED` is either commented out or set to true.
- Set `USAGE_TRACKER_HOST` to 'http://localhost:4001'
- Start Lightning

```
Lightning.SetupUtils.setup_demo(create_super: true)
date = Date.utc_today()
Lightning.UsageTracking.enable_daily_report(DateTime.add(DateTime.utc_now(), -1, :day))
Oban.insert(Lightning.Oban, Lightning.UsageTracking.ReportWorker.new(%{date: date}))
```

After a few seconds, the ImpactTracker logs should show activity as Lightning submits the report data.

## AI Usage

- [ ] Strategy / design
- [x] Other
- [ ] I have not used AI

### Pre-submission checklist


lib/lightning/usage_tracking/workflow_metrics_query.ex
```diff
@@ -0,0 +1,31 @@
...
+
+    start_of_next_day =
+      date
+      |> Date.add(1)
+      |> DateTime.new!(~T[00:00:00])
+
+    from r in query,
+      where: not is_nil(r.finished_at),
+      where: r.finished_at >= ^start_of_day,
+      where: r.finished_at < ^start_of_next_day
```
stuartc, line 29: @rorymckinley I just realised we don't have an index on `finished_at`, while this wouldn't be a cause of the memory issue, it probably is a performance one right?

rorymckinley: @stuartc If you view this portion of the query in isolation, then yes it could have an impact. However, the overall reporting query also filters runs by workflow so there is a chance Postgres would use the those indexes when retrieving the records.


Of course, this also depends on the nature of the data in the Db at the time and whether the query planner has a case of the Mondays :D. It is also not to say that Postres would not prefer to use the `finished_at`  if it had an index.

Happy to run an `EXPLAIN ANALYZE` on prod if you like?

rorymckinley: @stuartc And of course the counterargument is: Do you want to potentially slow down run insertions to allow the reports to complete quicker?

stuartc: Yeah please do, find a project with lots of runs and do a check - we're also looking at adding indexes for some tables that get hammered when retention jobs come along and trigger cascade deletes - so it's been on my mind.

I'll also take a look around for things that also use `finished_at` and co for things like ordering. So I'm on your side here, we don't want to add indexes for the sake of it - which ironically is the same practice that has created some of the performance issues we've observed.

lib/lightning/usage_tracking/workflow_metrics_service.ex
```diff
@@ -4,25 +4,79 @@ defmodule Lightning.UsageTracking.WorkflowMetricsService do
...
+      active_jobs = RunService.unique_job_ids(steps, date)
+
+      Map.merge(
+        acc,
+        %{
+          no_of_runs: no_of_runs + 1,
+          no_of_steps: no_of_steps + Enum.count(steps),
+          job_ids: Enum.concat(job_ids, active_jobs)
+        }
+      )
```
stuartc, line 71: While this is similar to the original implementation I'm curious if all of this could be rolled into a query as well. That way we could skip the cost of hydrating steps as well. But lets roll with this for now, the preload is much much smaller and with the chunking we've probably cut this down significantly.

lib/lightning/usage_tracking/workflow_metrics_query.ex
rorymckinley: I think holding off on indexing until you see performance issues is the accepted approach, I just tend to see them too late :).

lib/lightning/usage_tracking/workflow_metrics_service.ex
rorymckinley: @stuartc I did consider this as a follow-up PR, but I stopped because I think a single query rolling up all the things is much harder to debug. For example, if there is an issue with some of the data going through, I would find it easier to pull this approach apart in an IEx session than I would a single query.

So, I held off on the suggestion until I have more data that indicates that this optimisation (i.e. a smarter query) is needed.

test/lightning/usage_tracking/workflow_metrics_service_test.exs
```diff
@@ -1,5 +1,6 @@
 defmodule Lightning.UsageTracking.WorkflowMetricsServiceTest do
-  use Lightning.DataCase, async: true
+  # Async false becuase Mox returns nil whenit should return 100 for seed 976623
+  use Lightning.DataCase, async: false
```
stuartc, line 3: I tracked down the cause of this, it's because of the use of:

```
put_temporary_env(:lightning, :usage_tracking,
  cleartext_uuids_enabled: cleartext_uuids_enabled,
  enabled: true,
  host: @host
)
```

in `test/lightning/usage_tracking/report_worker_test.exs`.

Could you update those tests to use Mox (at least in the cases of `put_temporary_env`) instead.

stuartc (changes requested):

stuartc (approved):

stuartc:
Thanks @rorymckinley, I merged and rearranged the `WorkflowMetricsService` tests - the cost of creating all those fixtures was almost a second for each test; I cut the test down into two tests, one with the fixtures and the other without (since it didn't actually use them).

I do appreciate the granularity of the tests you put there, but in this case I think saving 5-7% of the test suite execution for a denser test is better? 🤷‍♂️

test/lightning/usage_tracking/workflow_metrics_service_test.exs
stuartc: I think we're good here for now, but it is curious why codecov thinks those functions aren't being called though, because they are - or maybe it's because we are stubbing them?

rorymckinley: @stuartc Yeah, maybe it is because they are stubbed in all the other tests, whereas other methods have at least one test where they actually hit? That being said, I tried breaking some of the methods and no test breakage was observed, so CodeCov moves in mysterious ways...

rorymckinley:
> I do appreciate the granularity of the tests you put there, but in this case I think saving 5-7% of the test suite execution for a denser test is better? 🤷‍♂️

Yeah, I think it is good call @stuartc

PR #2702, opened 2024-11-21
2701 extend diff

### LINKED ISSUE #2701: Extend LightningWeb.AuditLive.Index.diff/1 behaviour
The `diff` function is responsible for representing the before and after changes of an audit entry in the UI.

As part of auditing the GH connection removal,  I need to make some changes to `diff` as it does not appear to support a use-case where there is `before` data but no `after` data. 

While looking at the implementation, I noticed that there were some edge cases which I ran into that I thought might affect future work around audits.

The work for this issue will try to smooth out some of the wrinkles while also adding in some new functionality to support auditing the deletion of GH connections.

rorymckinley:
### Description

This extends the display of audit events to cover a use case where an audit entry has `before` changes but no after `changes`.  This is required to address the display of audit entries relating to the removal of a Github connection.

While setting up covering tests for the functional component, I discovered some inconsistent behaviour when using maps with string keys vs atom keys as well as if the before and after maps have dissimilar keys. As both of these are errors that a particular kind of idiot (i.e. me) will make, I decide to make the behaviour more predictable.

(I think the inconsistent behaviour between string keys and atom keys probably occurred as a result of changes made last year to [OTP 26](https://fly.io/phoenix-files/taking-control-of-map-sort-order-in-elixir/).

Closes #2701 

### Validation steps

The Github auditing changes are not in main yet, so the only test that is cheap to do would be to ensure that the display of audit events ('Admin Menu -> Audit') should look exactly the same as it does when running on `main`.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/audit_live/index.ex
+++ b/lib/lightning_web/live/audit_live/index.ex
@@ -46,38 +46,47 @@ defmodule LightningWeb.AuditLive.Index do
     |> assign(:page, Auditing.list_all(params))
   end
 
+  def diff(%{metadata: %{before: nil, after: nil}} = assigns) do
+    no_changes(assigns)
+  end
+
+  def diff(%{metadata: %{before: lhs, after: rhs}} = assigns)
+      when map_size(lhs) == 0 and map_size(rhs) == 0 do
+    no_changes(assigns)
+  end
+
   def diff(assigns) do
-    with lhs <- assigns.metadata.before || %{},
-         rhs <- assigns.metadata.after || %{},
-         true <- lhs != %{} || rhs != %{} do
-      changes =
-        lhs
-        |> Map.keys()
-        |> Enum.concat(Map.keys(rhs))
-        |> Enum.sort()
-        |> Enum.uniq()
-        |> Enum.map(fn key ->
-          {key, Map.get(lhs, key), Map.get(rhs, key)}
-        end)
+    lhs = assigns.metadata.before || %{}
+    rhs = assigns.metadata.after || %{}
 
-      assigns = assign(assigns, changes: changes)
+    changes =
+      lhs
+      |> Map.keys()
+      |> Enum.concat(Map.keys(rhs))
+      |> Enum.sort()
+      |> Enum.uniq()
+      |> Enum.map(fn key ->
+        {key, Map.get(lhs, key), Map.get(rhs, key)}
+      end)
 
-      ~H"""
-      <.td colspan="4" class="font-mono text-xs break-all">
-        <%= for {field, old, new} <- @changes do %>
-          <li><%= field %>&nbsp; <%= old %>
-            <.icon name="hero-arrow-right" class="h-5 w-5 inline-block mr-2" />
-            <%= new %></li>
-        <% end %>
-      </.td>
-      """
-    else
-      false ->
-        ~H"""
-        <.td colspan="4" class="font-mono text-xs">
-          No changes
-        </.td>
-        """
-    end
+    assigns = assign(assigns, changes: changes)
+
+    ~H"""
+    <.td colspan="4" class="font-mono text-xs break-all">
+      <%= for {field, old, new} <- @changes do %>
+        <li><%= field %>&nbsp; <%= old %>
+          <.icon name="hero-arrow-right" class="h-5 w-5 inline-block mr-2" />
+          <%= new %></li>
+      <% end %>
+    </.td>
+    """
+  end
+
+  defp no_changes(assigns) do
+    ~H"""
+    <.td colspan="4" class="font-mono text-xs">
+      No changes
+    </.td>
+    """
   end
 end
--- a/lib/lightning_web/live/audit_live/index.ex
+++ b/lib/lightning_web/live/audit_live/index.ex
@@ -47,34 +47,41 @@ defmodule LightningWeb.AuditLive.Index do
   end
 
   def diff(assigns) do
-    rhs = assigns.metadata.after
+    rhs = assigns.metadata.after || %{}
 
     lhs =
       assigns.metadata.before ||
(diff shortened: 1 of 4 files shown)


test/lightning_web/live/audit_live_test.exs
```diff
@@ -60,4 +61,296 @@ defmodule LightningWeb.AuditLiveTest do
...
+        }
+      }
+
+      html = render_component(&AuditLive.Index.diff/1, assigns)
+
+      assert html =~ ~r"<li>foo.+foo_before.+foo_after</li>"s
+      assert html =~ ~r"<li>bar.+bar_before.+bar_after</li>"s
+    end
+
+    test "correctly lists changes with both before and after (atom keys)" do
```
rorymckinley, line 86: Having a string variant and an atom variant of each test would usually be overkill. However, I needed them to support the initial changes (ensuring consistent behaviour for strings and atoms) and, as they are lightweight and already 'paid for' :) we may as well keep them.

lib/lightning_web/live/audit_live/index.ex
```diff
@@ -47,24 +47,37 @@ defmodule LightningWeb.AuditLive.Index do
...
-      <li><%= field %>&nbsp; <%= old %>
-        <Heroicons.arrow_right class="h-5 w-5 inline-block mr-2" />
-        <%= new %></li>
-    <% end %>
-    """
+      ~H"""
+      <.td colspan="4" class="font-mono text-xs break-all">
+        <%= for {field, old, new} <- @changes do %>
+          <li><%= field %>&nbsp; <%= old %>
+            <Heroicons.arrow_right class="h-5 w-5 inline-block mr-2" />
```
jyeshe, line 69: We are trying to concentrate on a single way of using the icons that is common to the latest Phoenix versions way using the `<.icon name="hero-arrow-right" ...`

lib/lightning_web/live/audit_live/index.html.heex
```diff
@@ -44,15 +44,7 @@
...
-          <%= if audit.changes.after do %>
-            <.td colspan="4" class="font-mono text-xs break-all">
-              <.diff metadata={audit.changes} />
-            </.td>
-          <% else %>
-            <.td colspan="4" class="font-mono text-xs">
-              No changes
-            </.td>
-          <% end %>
+          <.diff metadata={audit.changes} />
```
jyeshe, line 47: I like the naming, very intuitive

lib/lightning_web/live/audit_live/index.ex
```diff
@@ -47,24 +47,37 @@ defmodule LightningWeb.AuditLive.Index do
...
+    with lhs <- assigns.metadata.before || %{},
+         rhs <- assigns.metadata.after || %{},
+         true <- lhs != %{} || rhs != %{} do
+      changes =
+        lhs
+        |> Map.keys()
+        |> Enum.concat(Map.keys(rhs))
+        |> Enum.sort()
+        |> Enum.uniq()
+        |> Enum.map(fn key ->
```
jyeshe, line 59: It would make sense but not sure if that's the case, if the key for `after` only exists conditioned to an existing `before`, couldn't it just iterate the `lhs`? Would appreciate to understand this better.

lib/lightning_web/live/audit_live/index.ex
```diff
@@ -47,24 +47,37 @@ defmodule LightningWeb.AuditLive.Index do
   end
 
   def diff(assigns) do
-    rhs = assigns.metadata.after
+    with lhs <- assigns.metadata.before || %{},
+         rhs <- assigns.metadata.after || %{},
+         true <- lhs != %{} || rhs != %{} do
```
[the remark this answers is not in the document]
jyeshe: As a suggestion, I think pattern matching on the contract helps to understand what is the purpose or condition to run certain function definition.
As on `def diff(%{metadata: %{before: before, after: after}} = assigns) when map_size(before) > 0 and map_size(after) > 0 do`

jyeshe (commented):
The best test coverage as usual (: please find some questions and suggestions on readability/idioms.

lib/lightning_web/live/audit_live/index.ex
rorymckinley: @jyeshe I wanted to try and make the display as 'failsafe' as possible, given the importance of audit events.

With that in mind, I asked myself: "If I make a mistake and I have a key in `after` that is not present in `before` (for example) what is the safest thing that the code could do?" To me the answer would be that it would be better to show this 'weird' setting rather than not showing anything at all (which is what would happen if I just iterate LHS).

By showing something, the data is still presented in the UI and so the information is still presented to the user (in case it is useful) and it s more likely that the underlying cause (key mismatch)  will be identified and corrected.

jyeshe: Got it, first implementation to make sure that everything is fine 👍

lib/lightning_web/live/audit_live/index.html.heex
rorymckinley: I can't take credit for it, unfortunately :)

lib/lightning_web/live/audit_live/index.ex
rorymckinley: @jyeshe Thanks - fixed in a follow-up commit.

lib/lightning_web/live/audit_live/index.ex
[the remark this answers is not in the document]
rorymckinley: Thanks @jyeshe - fixed in a follow-up commit.

rorymckinley:
@jyeshe Thank you for your feedback. I have updated as per your suggestions.

jyeshe (approved):

stuartc (approved):

Issue #2410, opened 2024-08-20
AI Assistant: Bad rendering of line breaks


josephjclark:
When the response from the AI agent is formatted, line breaks aren't always handled cleanly.

The response to "how do I use salesforce?" renders really horribly:

![image](https://github.com/user-attachments/assets/48a791c5-5fc4-485b-9a63-64203d9fa0fe)

Here's the raw markdown:

````
To use Salesforce with OpenFn, you'll need to:

1. Set up your Salesforce credentials in your project configuration.

2. Use the Salesforce adaptor functions in your job. Here are some common operations:

- Query records:
```javascript
query("SELECT Id, Name FROM Account LIMIT 5");
```

- Create a record:
```javascript
create("Account", { Name: "New Account" });
```

- Update a record:
```javascript
update("Account", { 
  Id: "001XXXXXXXXXXXXXXX",
  Name: "Updated Account Name" 
});
```

- Upsert a record:
```javascript
upsert("Contact", "Email", {
  Email: "user@example.com",
  FirstName: "John",
  LastName: "Doe"
});
```

- Delete a record:
```javascript
destroy("Account", ["001XXXXXXXXXXXXXXX"]);
```

Remember to handle state and use data from previous operations as needed. For example:

```javascript
query("SELECT Id FROM Account LIMIT 1");

fn(state => {
  const accountId = state.references[0].records[0].Id;
  return { ...state, accountId };
});

update("Account", state => ({
  Id: state.accountId,
  Description: "Updated via OpenFn"
}));
```

````

josephjclark:
Is there anything we can do on the AI side to help rendering? Or is this just a case of plugging in a good markdown renderer?

josephjclark:
Related? `<ol>` list items don't render well either

![image](https://github.com/user-attachments/assets/3ed4ca63-9da4-4be5-a03f-b0c4332122c8)
![image](https://github.com/user-attachments/assets/78df2992-fbbd-47b1-a3c1-1e80bde63321)

The HTML is fine but tailwind styles show list-style none on `<ol>`

![image](https://github.com/user-attachments/assets/3ff507d9-1248-41f8-a30c-dbe88b2dc656)

Are classes missing on the returned HTML?

PR #2721, opened 2024-11-27
Style the AI Chat Messages

### LINKED ISSUE #2484: AI Assistant: Format and Open links in new tab
When the AI assistant returns links in the response, it would be great to have them formatted to distinguish it from other content. We would also like to open the link in a new tab instead of replacing the current window.

elias-ba:
### Description

This PR formats the links and makes them open in a new tab when the AI assistant sends a text with links. It actually introduces a function named `apply_styles/1` which parses the AI assistant message AST and injects some styles to format it.

Closes #2484 (and maybe #2410)

### Validation steps

1. Initiate a discussion with the AI assistant
2. Ask it to send you some links to learn more about OpenFn
3. See the links formatted in blue and note the hover effect on them
4. Click on the links to see them open in a new tab

I turned Dialyzer off for the function that formats the links to avoid a repetitive error due to the types of `Earmark`. In fact in the type specs, `Earmark.Parser.as_ast/1` returns a list while `Earmark.Transform.map_ast/3` expects a binary as first argument. Dialyzer was yelling at me for that with the following error.

```
lib/lightning_web/live/workflow_live/ai_assistant_component.ex:591:call
The function call will not succeed.

Earmark.Transform.map_ast(_ast :: binary(), (_ -> any()))

will never return since the 1st arguments differ
from the success typing arguments:

(
  [
    binary()
    | maybe_improper_list(
        binary() | maybe_improper_list(any(), binary() | [] | {_, _, _, _}) | {_, _, _, _},
        binary() | [] | {_, _, _, _}
      )
    | {_, _, _, _}
  ],
  any()
)

________________________________________________________________________________
lib/lightning_web/live/workflow_live/ai_assistant_component.ex:592:call
The function call will not succeed.

Earmark.Transform.transform([any()])

will never return since the 1st arguments differ
from the success typing arguments:

(
  binary()
  | {binary(), [{binary(), binary()}], [binary() | {binary(), [any()], [any()], _}], map()}
)
```

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### Pre-submission checklist


test/lightning_web/live/workflow_live/edit_test.exs
```diff
@@ -2357,6 +2357,311 @@ defmodule LightningWeb.WorkflowLive.EditTest do
...
+      assert docs_link != nil
+
+      assert Floki.attribute(docs_link, "class") == [
+               "text-primary-400 hover:text-primary-600"
+             ]
+
+      assert Floki.attribute(docs_link, "target") == ["_blank"]
+
+      assert Floki.find(html, "li")
+             |> Enum.any?(&(Floki.text(&1) == "\nPlain text  "))
```
stuartc, line 2451: @elias-ba I've been thinking about this and I think instead of testing the chat component (at least this deeply), we should be testing the `apply_styles` function, perhaps that function becomes a function component that renders out html.

```
<.formatted_markdown content={message.content} />
```

And we do a function component test with different sets of markdown and test against that output?

elias-ba: Thanks @stuartc, this is a great suggestion. I have implemented it.

lib/lightning_web/live/workflow_live/ai_assistant_component.ex
```diff
@@ -584,6 +584,59 @@ defmodule LightningWeb.WorkflowLive.AiAssistantComponent do
...
+        {:ok, ast, _} ->
+          process_ast(ast, merged_attributes) |> raw()
+
+        _ ->
+          assigns.content
+      end
+      |> then(&assign(assigns, :content, &1))
+
+    ~H"""
+    <article><%= @content %></article>
```
stuartc, line 618: @elias-ba I made a small change here, I moved the `raw` call to the result of `process_ast`, because in the rare chance the parser doesn't do anything to the output - we shouldn't assume it's safe and should render it out as `{:safe, str}` (which is what `raw` does).

stuartc (approved):

PR #2792, opened 2024-12-17
Add workflows API to create, get, update and list

### LINKED ISSUE #1887: API for Workflows
WCS has a [use case where they want to create and update workflows](https://openfn.github.io/ConSoSci/kobo-automation/)(with steps) from a step. This was available in V1, and it blocks migration to v2 for this edge case.

Image

A typical use case can be to build a new workflow every time a new form is created in CommCare.

We need to expose an API that developers can call from OpenFn or an external application to: 
- Get a list of workflows (incl. triggers, jobs, paths),
- Create workflows, and
- Update workflows (in their entirety… the whole thing at once.)

~All endpoints should be authenticated with the user’s personal access token~ (Whatever is best, technically.)

We also need to add documentation on how to use the Workflow JSON API.

No changes to the UI are required here.

Two important things to note:
1. In v1, we used [the openfn adaptor](https://github.com/OpenFn/adaptors/blob/main/packages/openfn/src/Adaptor.js) to do this. While we may update this adaptor to use a personal access token for v2, that's probably out of scope for now.
2. Our GovStack compliant JSON api for jobs (`lib/lightning_web/controllers/api/job_controller.ex`) and runs (`lib/lightning_web/controllers/api/run_controller.ex`) is now outdated. Please _consider_ how the new workflows REST api _may_ (??) replace the outdated `jobs_controller` and how it does or does not relate to the upcoming workorders/runs API for GovStack. #1656

jyeshe:
### Description

This PR adds an API that allows to create, update, get and list workflows. The authentication uses a Bearer token and the  user of the Personal Access Token must belong to the project related to the workflow. 

For POST and PUT the body should be a JSON encoded workflow having {name, project_id, edges, jobs, triggers}.
A PATCH can be also used to update a workflow passing only a difference of the fields that needs to be changed.

When certain workflow preconditions are not met, it returns 422 (Unprocessable Entity) with an error message describing the violation:
- "The triggers cannot be replaced, only edited or added." (cannot delete a webhook trigger for example) 
- "A workflow can have only one trigger enabled at a time."

A POST or PUT or PATCH might also fail if the outcome would be having more active workflows than the limit (when running on the billing app). For this case it also returns a 422 with the message: 

"Your plan has reached the limit of active workflows."

Some additional information on protocol syntax in regards to paths and responses:
- All paths follow a nested REST resource on projects (e.g PATCH /api/projects/:project_id/workflows/:id).
- Getting a single workflow returns a JSON with a `workflow` 
- Getting a list of workflows returns a JSON with a `workflows`
- All other paths return a JSON {"id": <workflow_uuid>, "error": <error_message>}

Closes #1887 

### Validation steps

Create a Personal Access Token (PAT) on Lightning user profile menu and make these curl requests with `authorization: Bearer <the_PAT>`, `accept: application/json`and `content-type: application/json`

Example of POST to create using $token and project_id 86a7888e-8a07-4673-9dbf-43ba016dc790
```
curl -s -X POST -H "content-type: application/json" -H "authorization: Bearer $token" localhost:4000/api/projects/86a7888e-8a07-4673-9dbf-43ba016dc790/workflows -d "{\"id\":null,\"name\":\"work1\",\"project_id\":\"86a7888e-8a07-4673-9dbf-43ba016dc790\",\"edges\":[{\"enabled\":true,\"id\":\"a8f136af-4527-4009-9122-f6e67fdf0a44\",\"source_trigger_id\":\"858f21e6-f301-4c56-8bcf-66b7378d1e54\",\"target_job_id\":\"66810fb2-e770-4d71-9c3c-44a0354f650d\",\"condition_type\":\"always\",\"source_job_id\":null}],\"jobs\":[{\"id\":\"66810fb2-e770-4d71-9c3c-44a0354f650d\",\"body\":\"fn(state => { return {...state, extra: \\\"data\\\"} })\",\"name\":\"job-9\",\"adaptor\":\"@openfn/language-common@latest\",\"inserted_at\":null,\"updated_at\":null}],\"triggers\":[{\"id\":\"858f21e6-f301-4c56-8bcf-66b7378d1e54\",\"comment\":null,\"custom_path\":null,\"cron_expression\":null,\"type\":\"webhook\",\"enabled\":true}],\"inserted_at\":null,\"updated_at\":null}"
{"error":null,"id":"2eed727b-99a1-4bf8-8312-1ab460f4fd47"}
```
it shall return a JSON in this format: {"error":null,"id":"2eed727b-99a1-4bf8-8312-1ab460f4fd47"}

After that execute curl commands for:
GET    /api/projects/:projectId/workflows
GET    /api/projects/:projectId/workflows/:workflowId
PUT    /api/projects/:projectId/workflows/:workflowId
PATCH  /api/projects/:projectId/workflows/:workflowId

Currently PUT and PATCH are handled by the same `update` function so can be used interchangeably until a final validation is added.
 
## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### Pre-submission checklist


josephjclark:
Hmm, possible problem here

The client is responsible for generating UUIDs for new steps, triggers and workflows.

This is likely to be an issue for clients using the API. First off, we don't report very good errors right now. Secondly, I can imagine cases where the user gets into all sorts of trouble if they're in control of ids. And thirdly, relying on clients to generate a UUID may not be realistic (tooling like adaptors can help but it's not a very friendly  API).

Here are some more detailed observations and ideas:

## ID Conflict Errors

When I post a new workflow with ids which already exist, I get a huge stack trace out of lightning:

```
# Ecto.ConstraintError at POST /api/projects/24da840c-011c-4a4b-bad7-6746b09886cc/workflows

Exception:

    ** (Ecto.ConstraintError) constraint error when attempting to insert struct:
    
        * "triggers_pkey" (unique_constraint)
    
    If you would like to stop this constraint violation from raising an
    exception and instead add it as an error to your changeset, please
    call `unique_constraint/3` on your changeset with the constraint
    `:name` as an option.
    
    The changeset defined the following constraints:
    
        * "triggers_workflow_id_fkey" (foreign_key_constraint)

// ...etc
```

It would be better to catch this error and report it in in the `error` key in the returned JSON.

## ID validation

Related to the above, perhaps its a good idea to validate incoming IDs before creating the workflow, and if any new id ids already exist, or references do not exist in the JSON, we report an error.

## Existing ID references

I should not be able to post a workflow where one step references the id of pre-existing step outside this workflow. That's a no-no. So we need to throw an error in this case.

Basically all the ids in the submitted workflow should be totally self-contained.

## ID Replacement

Would it make sense to treat all IDs in the incoming data as placeholders, and replace them with UUIDs of Lightning's choosing?

So I can POST a workflow like `{ triggers: [{ id: 't' }], jobs: [{ id: 'j' }], edges: [{ source: 't', target: 'j' }] }`, and all the ids will be swapped out for UUIDs?

This would have to work for PUT and POST too, where new ids must be detected and replaced. Ie if I add a new step I can call it "step1" locally, and lightning will replace it (and refereces to it) with a UUID.

If we did this stuff, we'd have to return the full modified workflow back to the client.

Note that neither the Workflow Diagram nor the Adaptor will need this - they'll create their own UUIDs.

If we DON'T do this, users will create workflows with step ids like "my-cool-step". I don't know if this is technically a problem, but it'll result in inconsistent looking ids. And that id can't ever be changed, presumably

josephjclark:
Unrelated but relevant: If I create a Workflow with the API, then remove it with the Lightning UI, then post it again, I get the following error:
```
"a workflow with this name already exists in this project."
```
Presumably we still have the workflow, and all steps and edges, in the database, so they'll all conflict when I try and create the same workflow again.

It's the kind of problem I expect to run into when testing a workflow out. I probably want to keep posting the same workflow to a project, testing it, maybe editing it, and removing it again.

josephjclark:
When I POST a new workflow, then edit a step name, and PUT the edited workflow, I get:
```
{"error":"Conflict"}
```
The POST and PUT structures are basically identical at this point.

Setting workflow or project ids in the payload makes no difference

josephjclark:
Also, on POST, if the project id in the body is different to the project in the URL I posted to, I do not get an error

jyeshe:
> Also, on POST, if the project id in the body is different to the project in the URL I posted to, I do not get an error

done

jyeshe:
> Unrelated but relevant: If I create a Workflow with the API, then remove it with the Lightning UI, then post it again, I get the following error:
> 
> ```
> "a workflow with this name already exists in this project."
> ```
> 
> Presumably we still have the workflow, and all steps and edges, in the database, so they'll all conflict when I try and create the same workflow again.
> 
> It's the kind of problem I expect to run into when testing a workflow out. I probably want to keep posting the same workflow to a project, testing it, maybe editing it, and removing it again.

It's a Lightning restriction. Probably the workflow is not completed deleted to give sometime for the user to recover it. @taylordowns2000 when a workflow is deleted, the record is still there and blocks the creation of new ones with same name. Should the API overwrite it?

josephjclark:
I would be in favour of the API overriding this restriction, but I'm probably biased my my testing requirements

Alternatively, maybe we should add support for DELETE on a workflow, which is an instant delete with no recovery.

So users using the UI get a grace period, but users using the API do not. Perhaps there's a flag for this, like a hard and soft delete.

This would give me what I want for testing and automation - I'm not blocked by the Lightning soft delete. But we also don't lose that behaviour entirely.

josephjclark:
Looking good - some test results with only a couple of minor grumbles

:heavy_check_mark:  allow me to create a workflow
:heavy_check_mark:  allow me to modify workflow
:heavy_check_mark:  error when the workflow is open and edits are blocked
:heavy_check_mark:  good error when referencing a step that is not in the payload
:no_entry:  Stack trace error (`"jobs_pkey" (unique_constraint)`) when I try to define a step that exists elsewhere (ie, I copy the trigger from another job and use that ID as the trigger in this job)
:no_entry: I don't get an error error when project id in the payload is different to the one in the URL (I expect an error here!)
:heavy_check_mark: : GET `projects/:id/workflows` looks great
:no_entry: But  GET `projects/:id/workflows/:workflowid` seems to give me a 403 and a login HTML page in the body. The same workflow ID words fine in lightining if I copy it into the URL, and I am definitely attaching my credential.

josephjclark:
Open question on documentation - where and how should we document this stuff? Even the collections rest API isn't really documented anywhere (I refer back to the original issue when things come up)

jyeshe:
@josephjclark, for this I have tested again on the PATCH and it works for the PATCH, PUT and GET:
 ⛔ I don't get an error error when project id in the payload is different to the one in the URL (I expect an error here!) 
image
The pending one for POST is available for testing now.

jyeshe:
> But GET `projects/:id/workflows/:workflowid` seems to give me a 403 and a login HTML page in the body. The same workflow ID words fine in lightining if I copy it into the URL, and I am definitely attaching my credential.
Hey Joe, this happens when the user doesn't have access to the project.
image

jyeshe:
> ⛔ If I do not pass a UUID, but use my own ids (ie `my-special-id-1092`) I get a stacktrace error body back with
Done

jyeshe:
> Hmm, possible problem here
> 
> The client is responsible for generating UUIDs for new steps, triggers and workflows.
> 
> This is likely to be an issue for clients using the API. First off, we don't report very good errors right now. 

The pre-validations and changesets were extended to handle IDs formats and uniqueness.

> Secondly, I can imagine cases where the user gets into all sorts of trouble if they're in control of ids. 

For the record, the users need to set the ids to build a workflow in order to connect the vertices of the edges. We could allow them using their own domain of IDS (as names or integers for example instead of UUIDs) and convert them to UUIDs but can you ask this question on the requirements issue if it's something that you think that would be helpful?

One caveat on flexibilizing the ID domain is that for the PUT we need to rely on the actual ids. 

> And thirdly, relying on clients to generate a UUID may not be realistic (tooling like adaptors can help but it's not a very friendly API).

As UUIDs are pretty common and the users of this API are technical ones would the generation of UUIDs be a concern? This falls back the second question whether they could use a different ID domain. I would gladly implement it for the POST if it becomes a requirement to help users to see more clearly the edges connections with integer ids for example. 

> Here are some more detailed observations and ideas:
> 
> ## ID Conflict Errors
> When I post a new workflow with ids which already exist, I get a huge stack trace out of lightning:

Uniqueness check ✅ 

> ## ID validation
> Related to the above, perhaps its a good idea to validate incoming IDs before creating the workflow, and if any new id ids already exist, or references do not exist in the JSON, we report an error.

Uniqueness check ✅ 
Consistency of ids on edges, jobs and triggers ✅ (symmetric diff, what exists on edges should be on jobs and triggers and vice-versa)  
 
> ## Existing ID references
> I should not be able to post a workflow where one step references the id of pre-existing step outside this workflow. That's a no-no. So we need to throw an error in this case.

WIP ...

> ## ID Replacement
> Would it make sense to treat all IDs in the incoming data as placeholders, and replace them with UUIDs of Lightning's choosing?

TBD on requirement

lib/lightning_web/router.ex
```diff
@@ -82,10 +82,13 @@ defmodule LightningWeb.Router do
 
     resources "/projects", API.ProjectController, only: [:index, :show] do
       resources "/jobs", API.JobController, only: [:index, :show]
-      resources "/workflows", API.WorkflowsController, except: [:delete]
       # resources "/runs", API.RunController, only: [:index, :show]
     end
 
+    resources "/v1/projects", API.ProjectController, only: [:index, :show] do
```
josephjclark, line 88: :exclamation: I am worried about this because Stu was keen on implementing header-based versioning for the Collections API

I strongly  recommend we either implement this as a header, or revert the change and wait discuss with Stu next week.

jyeshe: We should for him, thanks!

jyeshe: One factor to take into consideration is that the Collections is an internal API while the workflows will be available for other kinds of integrations. But if we are not sure how stable the API will be on the fields and structure the Headers versioning would be a better alternative (otherwise, considering the evolution of Triggers, Edges and Jobs fields the URL is the most explicit and clear one).

josephjclark:
> But GET projects/:id/workflows/:workflowid seems to give me a 403 and a login HTML page in the body. The same 
workflow ID words fine in lightining if I copy it into the URL, and I am definitely attaching my credential.

> Hey Joe, this happens when the user doesn't have access to the project.

hmm, this one is a bit tricky :thinking:  If I do  `projects/:id/workflows` for a project that doesn't exist, do we expect a 403 or a 404? No access or project doesn't exist?

I'd expect a 404 unless project ids are super secret.

Also, I notice that if I make a request where the workflow id or project id happens to not be a valid id, I get a 400 web page back with this error:
```
Exception:

    ** (Ecto.Query.CastError) deps/ecto/lib/ecto/repo/queryable.ex:477: value `"79efba60-072a-4d4f-8d6c-22dfd3852"` cannot be dumped to type :binary_id in query:
    
    from p0 in Lightning.Projects.Project,
      prefix: "public",
      where: p0.id == ^"79efba60-072a-4d4f-8d6c-22dfd3852",
      select: p0
    
        (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
        (elixir 1.16.2) lib/enum.ex:1826: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir 1.16.2) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
        (ecto 3.12.4) lib/ecto/repo/queryable.ex:214: Ecto.Repo.Queryable.execute/4
        (ecto 3.12.4) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
        (ecto 3.12.4) lib/ecto/repo/queryable.ex:154: Ecto.Repo.Queryable.one/3
        (lightning 2.10.8) lib/lightning_web/controllers/api/project_controller.ex:21: LightningWeb.API.ProjectController.show/2
```

I would prefer to just get a 404 back. I don't know if it's worth handling the case - although it's very easy to accidentally corrupt the ids when building cURL commands

lib/lightning_web/controllers/api/workflows_controller.ex
```diff
@@ -0,0 +1,461 @@
...
+          :jobs,
+          "The jobs #{inspect(job_ids)} should be present both in the jobs and on an edge."
+        )
+
+      {:error, :edges_misses_a_trigger} ->
+        reply_422(
+          conn,
+          workflow_id,
+          :edges,
+          "Missing edge with source_trigger_id."
```
josephjclark, line 403: This error occurs when an edge points to a source trigger that does not exist. But the error implies that it's the edge that is missing

Maybe better wording is `Trigger not found for source_trigger_id`

lib/lightning_web/controllers/api/workflows_controller.ex
```diff
@@ -0,0 +1,461 @@
...
+
+  defp maybe_handle_error(conn, result, workflow_id)
+       when is_tuple(result) do
+    case result do
+      {:error, :invalid_jobs_ids, job_ids} ->
+        reply_422(
+          conn,
+          workflow_id,
+          :jobs,
+          "The jobs #{inspect(job_ids)} should be present both in the jobs and on an edge."
```
josephjclark, line 395: As per our conversation this morning, it's quite possible to have an "orphaned" job with no edges. Either this wording needs to be tweaked (I take issue with the "and on an edge" bit in particular"), or this whole validation should be removed.

To be fair I think it's probably OK to remove the validation. The diagram should still render the graph perfectly well, and the runtime has its own validation rules and will report errors if it finds serious issues.

josephjclark:
Results from my latest round of testing:

✔️ get workflows for project
⛔ get workflows with invalid project id (expect 404 and error object, got 400 webpage)
⛔ get workflows for project that does not exist (expect 404, got 401)
✔️ get one valid workflow id (200)
✔️ get one valid workflow id that does not exist (404)
⛔ get one invalid workflow id (expect 404, got 400 + stacktrace)
✔️create a workflow
✔️ modify workflow with PUT
⛔ modify workflow with PATCH (see below)
✔️ error when the workflow is open and edits are blocked
✔️error when referencing a step that is not in the payload
⛔ I know this is still in discussion, but for the record ids in the body must still be valid UUIDs
✔️ Creating and editing steps with invalid job names throws an error
✔️ I can PUT new steps and edges to an existing workflow
✔️ Error if I try and use a non-UUID id (this is the correct behaviour for this commit but I'm gonna question it!)
⛔ Error if the workflow id in the body is different to the one in the URL

The first 3 issues are covered in my prior comment: is it a 404 or 403 for a project that doesn't exist; and handle invalid project/workflow UUIDs are regular 404s.

What is the design of PATCH right now? Is it just an alias for PUT? In the long term, PATCH should use JSONpatch to allow incremental/partial updates to a workflow. I don't think we need to implement this today, but until we support it, we should not support PATCH at all.

I've asked about ID replacement in the issue. Tbh I really think we should just do it (for new steps and edges, for POST, PUT and PATCH)

Finally, I'm still not getting an error for PUT or PATCH if the workflow id in the body differs to the one in the URL

lib/lightning_web/controllers/api/workflows_controller.ex
jyeshe: Yeah, this validation needs to be removed. This morning I was wondering about validating the dangling edges but as the UI permits dangling jobs it makes sense to loosen it up here too.

jyeshe:
> What is the design of PATCH right now?
A simple patch without passing the whole workflow should work. On the automated test it only appends a job and sends this patch:
```
patch =
        build(:job)
        |> then(fn job ->
          %{
            name: "work1.1",
            edges:
              edges ++
                [
                  build(:edge,
                    source_job_id: List.last(jobs).id,
                    target_job_id: job.id,
                    condition_type: :on_job_success
                  )
                ],
            jobs: jobs ++ [job]
          }
        end)
```

But I am getting some errors vai curl as well, ...

josephjclark:
@jyeshe Yeah I did try a partial update through curl and got errors. I still stand by the JSON patch thing though - the workflow diagram  is currently using JSON patch to submit diffs to backend. I don't know the intimate details of this. But One day soonish we'll be re-writing the diagram to use _this_ API - and if we support JSONPatch through PATCH then it'll be an easier transition.

lib/lightning/policies/workflows.ex
```diff
@@ -0,0 +1,28 @@
...
+  @moduledoc """
+  The Bodyguard Policy module for Workflows API.
+  """
+  @behaviour Bodyguard.Policy
+
+  alias Lightning.Accounts.User
+  alias Lightning.Projects.Project
+  alias Lightning.Run
+
+  @type actions :: :access_write | :access_read
```
midigofrank, line 11: The other bodyguard polices name the specific actions to be done, could we do the same here as well? 
Something like `:create_workflow`, `:update_workflow` ...

lib/lightning/work_orders.ex
```diff
@@ -394,13 +394,7 @@ defmodule Lightning.WorkOrders do
...
        }) do
     edges
-    |> Enum.reduce(Graph.new(), fn edge, graph ->
-      graph
-      |> Graph.add_edge(
-        edge.source_trigger_id || edge.source_job_id,
-        edge.target_job_id
-      )
-    end)
+    |> Enum.reduce(Graph.new(), &Graph.add_edge(&2, &1))
```
midigofrank, line 397: Could we maintain the long `fn` block here? 😅 the capture operator can  get confusing real quick. I think just using `edge` and `graph` makes it really easy to know what's going on.

lib/lightning/workflows/presence.ex
```diff
@@ -41,22 +43,23 @@ defmodule Lightning.Workflows.Presence do
...
 
   ## Examples
 
-      iex> Lightning.Workflows.Presence.track_user_presence(%User{id: 1}, "room:lobby", self())
+      iex> Lightning.Workflows.Presence.track_user_presence(%User{id: user_id}, %Workflow{id: workflow_id}, self())
       :ok
 
   """
-  def track_user_presence(user, topic, pid) do
+  def track_user_presence(user, %Workflow{id: workflow_id}, pid) do
```
midigofrank, line 60: Nicely done 🙌 . I was worried for a second then later noticed that it's already scoped for `workflows`

lib/lightning_web/controllers/api/workflows_controller.ex
```diff
@@ -0,0 +1,461 @@
...
+  require Logger
+
+  def index(conn, %{"project_id" => project_id}) do
+    with :ok <- authorize_read(conn, project_id) do
+      list =
+        Workflows.list_project_workflows(project_id,
+          include: [:edges, :jobs, :triggers]
+        )
+
+      json(conn, %{workflows: list, errors: []})
```
midigofrank, line 31: Why do we return the `errors` even when it's `200`? Does the `spec` require that it be present?

lib/lightning/policies/workflows.ex
```diff
@@ -0,0 +1,28 @@
...
+  @spec authorize(actions(), User.t() | Run.t(), Project.t()) ::
+          :ok | {:error, :unauthorized}
+  def authorize(access, %User{} = user, project)
```
midigofrank, line 14: I think this is a duplicate policy check. In liveview, we use the `ProjectUsers` module for this.

lib/lightning_web/controllers/api/workflows_controller.ex
```diff
@@ -0,0 +1,461 @@
...
+
+      :else ->
+        workflow
+        |> Workflows.change_workflow(params)
+        |> save_workflow(active_triggers_count > 0, workflow.project_id, user)
+    end
+  end
+
+  defp save_workflow(params_or_changeset, activate?, project_id, user) do
+    with :ok <- check_limit(activate?, project_id),
```
midigofrank, line 115: There is a `lib/lightning/workflows/workflow_usage_limiter.ex` module with these helpers already, did you consider using it? 
If given a `changeset`, it will determine the activated ones  and call the limiter.

midigofrank (changes requested):
Nicely done @jyeshe . I have left some comments, mostly cosmetic improvements.

jyeshe:
> @jyeshe Yeah I did try a partial update through curl and got errors. I still stand by the JSON patch thing though - the workflow diagram is currently using JSON patch to submit diffs to backend. I don't know the intimate details of this. But One day soonish we'll be re-writing the diagram to use _this_ API - and if we support JSONPatch through PATCH then it'll be an easier transition.

The v1 had got back on the way 🤕 Now it passes:
image

lib/lightning/policies/workflows.ex
jyeshe: From collections we've started using this pattern of access to make it more noticeable and different from the operation themselves (borrowed from ACL idea).  (https://github.com/OpenFn/lightning/blob/8e082b20e60e072d0bcd2cdb189e541a41d6461c/lib/lightning/policies/collections.ex#L16)

Another reason is that the `:access_write` is the same permission check for `:create_workflow` and `:updated_workflow`.

lib/lightning_web/controllers/api/workflows_controller.ex
jyeshe: On the collections API we where always returning the field, on the client side I think it helps to avoid an extra check if it's available and only then reading.

josephjclark:
@jyeshe I think this is academic? These PATCH requests are not JSON patch. They are not compatible with the workflow diagram. If we continue down this track, we will have to version the API later to make it compatible. Unless there's some way we can support multiple body formats in PATCH :thinking: 

I really think we should pause the PATCH work. Switch it all off until later. It is not needed for WCS.

jyeshe:
⛔ get workflows with invalid project id (expect 404 and error object, got 400 webpage) 
Done
⛔ get workflows for project that does not exist (expect 404, got 401) 
The logic here is that there is only one query for the `project_users` relationship. If there is none the user is not authorized for the operation. 
⛔ get one invalid workflow id (expect 404, got 400 + stacktrace) 
Done
⛔ modify workflow with PATCH (see below) 
Working but commented as requested
⛔ I know this is still in discussion, but for the record ids in the body must still be valid UUIDs 
Done
⛔ Error if the workflow id in the body is different to the one in the URL
Done

lib/lightning_web/controllers/api/workflows_controller.ex
jyeshe: During todays standup we've decided to keep this validation as it's helpful for the user to make sure that a fully connected/executable graph is on the body (for POSTs and PUTs).

Interventions on the UI that would lead to this state (of dandling nodes) would be validated by the API in a way that the user can either use a PATCH or use a PUT removing the jobs that are not on the edges.
cc @theroinaochieng

josephjclark:
Two spin-out issues:

https://github.com/OpenFn/lightning/issues/2802
https://github.com/OpenFn/lightning/issues/2803

#2803 actually worries me @jyeshe  - I know I said it was OK to be strict here if we had a good error message. But as soon as we port the Workflow diagram we're going to bump into this issue :thinking:

jyeshe:
> Two spin-out issues:
> 
> #2802 #2803
> 
> #2803 actually worries me @jyeshe - I know I said it was OK to be strict here if we had a good error message. But as soon as we port the Workflow diagram we're going to bump into this issue 🤔

I think for these cases when the UI allowed a dangling node, the users could use the PATCH to edit the workflow or easily make a change on the input based on the error informed by the API. Receiving the implemented error message 
"The jobs #{inspect(job_ids)} should be present both in the jobs and on an edge."
the user could quickly remove the job with disconnected id (not used on an edge) and alternatively make a PUT.

On the other hand, I understand the question lies on whether draft workflows (or worfklows with jobs for the future use) should be accepted or not (a product question). From engineering perspective I would prioritise helping as much as possible to inform the user if the workflow is fully executable because when using the API there is visual representation and it's harder classify on what is a mistake from the user or a decision from the user to save some job for future use. In doubt an error would be show taking a more strict/conservative path. To save draft jobs or jobs to be used in the future, the users would do it locally, use the GitHub sync feature or export the yaml ("I am going to save/sync this workflow to GH because I don't need this job C but I might need it in the future").

lib/lightning/work_orders.ex
midigofrank: Perfect! 🙌 
Generally, I only use the short form when it's one argument. The moment it's more than 1, it gets confusing for me, even for `reduce` 😅

lib/lightning_web/controllers/api/workflows_controller.ex
midigofrank: Sounds good

midigofrank:
@jyeshe any idea why the tests are timing out 🤔 ?

jyeshe:
> @jyeshe any idea why the tests are timing out 🤔 ?

The refactoring to reuse ProjectUsers.

josephjclark:
This is where I'm at now:

✔️ get workflows for project
⛔ get workflows with invalid project id (expect 404 and error object, got 400 webpage)
⛔ get workflows for project that does not exist (expect 404, got 401)
✔️ get one valid workflow id (200)
✔️get one valid workflow id that does not exist (404)
✔️ get one invalid workflow id (404)
✔️create a workflow
✔️create a workflow with arbitrary ids (but workflow not returned)
✔️ modify workflow with PUT
✔️ modify workflow with PUT if there's no diff
⛔ PUT a new job with arbitrary ids
✔️ PUT an edge without an id 
⛔ PUT a new orphaned job without an id
✔️PUT a new orphaned job with an id
✔️ error when the workflow is open and edits are blocked
✔️ error when referencing a step that is not in the payload
✔️ error when referencing a step defined in another workflow
✔️ Error if the workflow id in the body is different to the one in the URL
 
In summary:

* Basically if the project id is invalid, we get a 401 error not a  400 or 404.  This is not ideal but I'm happy to waive it at this stage.
* Arbitrary ids / id replacement works for POST but not for PUT. I was expecting it to be available for PUT :thinking: 
* And on that note I expect that if I post a new job, and don't include a reference to it, I don't need to include an id at all. But this is an edge case, not important. Happy to waive it.

Also - I've given id replacement a pass on POST but the mapped ids are not returned to the client. I think I'd prefer an id map to come back with the result (ie, `{ id, errors, replacedIds }` or something). But I'll also accept just returning the whole workflow with ids in situ. Strictly speaking I don't think either thing is needed for WCS (although I wouldn't guarantee it)

jyeshe:
> This is where I'm at now:
⛔ get workflows with invalid project id (expect 404 and error object, got 400 webpage) 
Sorry Joe, previously it was implemented only on the get workflow, now it's done for the index
⛔ get workflows for project that does not exist (expect 404, got 401) 
The reason it's returning 401 Unauthorized it's because the project id is used for ACL. The user could check the project id on 401 but if 404 is really helpful for the user we can add a query or subquery for all operations. In any case I've added the 401 on the spec with a description "User not allowed to access the project" for existing projects.
⛔ PUT a new job with arbitrary ids 
The arbitrary ids are supported only on the POST because the triggers cannot be replaced. If arbitrary ids are allowed only for jobs we could have a single rule for POST and PUT. However the PUT we need to be aware that all jobs would be replaced and I need to check what is behaviour of the edges on this jobs full update. 
⛔ PUT a new orphaned job without an id 
On PUT everything is expected to have an id for being an update. It's doable however.

jyeshe:
Take aways from our meeting:
- There is no particular rule for disconnected/orphan jobs in regards to ids. Client apps need to pass ids on all jobs.
- All operations will return the resulting workflow.
- Make sure that PUT is able to remove jobs and edges.

jyeshe:
Tests failing on unrelated flaky:
  1) test .generate_metrics/3 - cleartext disabled excludes the cleartext uuid (Lightning.UsageTracking.ProjectMetricsServiceTest)
     test/lightning/usage_tracking/project_metrics_service_test.exs:105
     ** (FunctionClauseError) no function clause matching in Stream.chunk_every/4

lib/lightning/policies/workflows.ex
jyeshe: to be continued here: https://github.com/OpenFn/lightning/issues/2822

jyeshe:
> @jyeshe any idea why the tests are timing out 🤔 ?

to be continued here: https://github.com/OpenFn/lightning/issues/2822

midigofrank (changes requested):
Hey @jyeshe , this looks really great.

There's a comment I've left regarding using `if..else` instead of `cond`. 
Also, there's a conflict on the `CHANGELOG`.

lib/lightning/workflows/edge.ex
```diff
@@ -30,6 +30,16 @@ defmodule Lightning.Workflows.Edge do
...
+
+  @derive {Jason.Encoder,
+           only: [
+             :id,
+             :condition_type,
+             :enabled,
+             :source_job_id,
+             :source_trigger_id,
+             :target_job_id
+           ]}
```
stuartc, line 42: I think we also need `condition_expression` and `condition_label` to allow for JS based conditions no?

priv/gettext/en/LC_MESSAGES/errors.po
```diff
@@ -2,7 +2,7 @@ msgid "can't be blank"
 msgstr "This field can't be blank."
 
 msgid "has already been taken"
-msgstr "This email address already exists."
+msgstr "This value should be unique."
```
stuartc, line 5: Is there not a way to make the distinction between emails (in this case) and other uniqueness violations?

stuartc (commented):
Great work @jyeshe, left some comments; I think at least we need to add the other `condition_...` fields to the Edge `@derive` property.

priv/gettext/en/LC_MESSAGES/errors.po
jyeshe: It's possible but we gonna need to call a different translation. Opened an issue for identifying and changing on all places of the app to have a specific message for email: https://github.com/OpenFn/lightning/issues/2824

lib/lightning/workflows/edge.ex
jyeshe: Done

midigofrank (approved):

stuartc (approved):

PR #2801, opened 2025-01-08
Monaco offline support

### LINKED ISSUE #2786: Get Monaco working fully offline
Investigate how we can get Monaco to work offline so that offline users can see Editor and Output/Logs panels.

midigofrank:
### Description

This PR enables monaco editor to work offline.  It configures the `loader` to look for the files in the `assets/` directory.
This PR also creates another `esbuild` profile responsible for copying over the required monaco files. I've found this to be easier compared to configuring the worker: https://github.com/suren-atoyan/monaco-react/blob/master/README.md#loader-config

Closes #2786 

### Validation steps
Ensure you don't have any cache. You can test this on a private window

 ~Run `mix esbuild.install` to get the newly configured version~ EDIT: No longer needed

1. Run `npm install --prefix assets` to get the updated monaco editor. `v0.43.0`
2. Turn off your internet
3. Start the server and try editing any  job's body 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### Pre-submission checklist


assets/js/monaco/index.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { useRef, useCallback } from 'react';
 import ResizeObserver from 'rc-resize-observer';
-import Editor, { Monaco } from '@monaco-editor/react';
+import Editor, { Monaco, loader } from '@monaco-editor/react';
 import type { editor } from 'monaco-editor';
 
+loader.config({ paths: { vs: '/assets/monaco-editor/vs' } });
```
midigofrank, line 6: What I'm doing here is just specifying where to get the monaco files. By default, it points to the `cdn`: https://github.com/suren-atoyan/monaco-loader/blob/master/src/config/index.js .

@josephjclark this is what I wanted your feedback on.

config/config.exs
```diff
@@ -101,6 +101,14 @@ config :esbuild,
          --outdir=../priv/static/assets --external:/fonts/* --external:/images/*),
     cd: Path.expand("../assets", __DIR__),
     env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
+  ],
+  monaco: [
```
midigofrank, line 105: Here we instruct `esbuild` to copy the `monaco files` to our assets directory

josephjclark:
@midigofrank when do we need to run `mix esbuild.install`? Do new users need to do this?

I ask because once this is merged to main, local editing will break for everyone, and they'll need to know to run that command

assets/js/monaco/index.tsx
josephjclark: This is perfect :100:

midigofrank:
@josephjclark `mix esbuild.install` needs to be run either way, even for new users. When you run `mix setup`, `mix esbuild.install` is run automagically for you.

Also, when you don't have the correct version, a warning is logged:
```
[warning] Outdated esbuild version. Expected 0.17.18, got 0.24.2. Please run `mix esbuild.install` or update the version in your config files.
```

config/config.exs
josephjclark: In dev, this adds about 50mb of _stufff_ to `static/assets`. I presume it's the same in prod. 

I don't _think_ this matters? Most of it will sit on the filesystem and never be used. The client is downloading exactly the same files, just from a different location

josephjclark:
@midigofrank when I checked out this branch my monaco was dead. I'm sure there was a warning but there's SO MUCH logging that I didn't see it.

`mix setup` isn't mentioned at all in the RUNNINGLOCAL.md file :thinking:  Is that handled by some other magic?

Is there any other way we can alert users about this change?

josephjclark:
Oh @midigofrank  I have an unchecked in mix.lock file after running the esbuild install - maybe that's the issue?

josephjclark:
@midigofrank  the other thing I noticed here is that the file sizes of the monaco build have bloated _significantly_

Here's local dev on main (prod looks about the same):

monaco-main

Here's the same files on this

monaco-branch
 branch:

Note that editor.main.js is up to 18mb and tsWrorker.js is 17mb.

I've had a quick look at the worker file and it's not obvious to me what the difference is. The new file is many many more lines, but when I skim through it seems to have the same formatting and contents. Weird. I note that you're loading the `min/vs` folder so you should be doing the right thing :thinking: 

I don't have time to investigate this any further right now - are you able to take a look?

midigofrank:
Whoa!! Good catch. Let me try and sift through

midigofrank:
It seems the problem was that they weren't getting minified. In production we add the flag `--minify`. Here is the result, not as good as the cdn though:  
Screenshot 2025-01-09 at 16 47 58

josephjclark:
Huge Improvement, thank you @midigofrank !

josephjclark:
Hmm, hang on 🤔

The unminified code in dev, on main and branch, is very different. Neither is minified in dev right? Doesn't need to be.

But there's a lot more content coming in the new version 

That'll be why the files still arent as small as the CDN build. It's not the minification, it's the content

Those prod sizes are acceptable I think but I'd still like to understand the difference

midigofrank:
Hmm, yeah it could be the difference in versions. In staging and prod, the cdn is pointing to `0.43.0`. Locally, I have `0.34.1`.

josephjclark:
@midigofrank ooh that's interesting. A very likely culprit. And scary! It's _critical_  that local dev and prod are running the same versions of software.

Can you work out how to lock down those versions to 0.43.0?

midigofrank:
Okay, `0.43.0` is actually bigger in size:
Screenshot 2025-01-09 at 17 46 43

midigofrank:
I have upgraded it to `0.43.0`. This came with a regression. I could no longer just run `esbuild --minify` on the lightning bundle, I have been forced to specify the platform `--platform=node`. 
Screenshot 2025-01-09 at 18 24 13

josephjclark:
Thanks @midigofrank 

I've just run through to check a bunch of things:

- Production right now uses monaco 0.43.0 from jsdeliver. The tsworker is 1.1mb on the wire. Source looks minified (although the typescript defs are not)
- [here it is!](https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/language/typescript/tsWorker.js)
- Dev right now uses exactly the  same file, 0.43.0
- Dev is about 4mb on the wire

Both tsworker files are about 4mb on disk.

So basically, the difference between the CDN file and the local file is better compression over the wire through some CDN magic.

josephjclark (approved):
:sparkles:

midigofrank:
Huh!! I have run `mix phx.digest` and this is the result. The prod `Dokerfile` doesn't run this command 🤷 . 
This PR shouldn't get merged until this is clarified
Screenshot 2025-01-10 at 08 46 38

josephjclark:
Hey @midigofrank I'll take a look later. Remember that prod is actually on 0.43 so it's not the version change, it's the build change generally

josephjclark:
Call summary:

- the actual problem here is the esbuild version bump, which breaks the standard esbuild profile
- I think this is  caused by [esbuild 22](https://github.com/evanw/esbuild/releases/tag/v0.22.0) (Click for release notes). One day this is going to bite us in the unmentionables
- But not today. We don't actually need esbuild to process the monaco files - we can just do a simple `cp` from `assets/node_modules/...` to `priv/assets` (or whatever) and serve the minified monaco files directly

theroinaochieng:
@josephjclark thanks for flagging this, I've noted your concerns about potential future issues related to esbuild 22, dropped a note to understand how the implications of this release so we can figure out scope and  prioritization

@midigofrank I can't seem to see any notes for QA on this issue description, do you mind adding them? Also reached out about who our second reviewer on this should be. Major claps for finishing this!

midigofrank:
@josephjclark after several deliberations I have found a way through with just `esbuild`. Given that `esbuild 0.17` doesn't support `glob` matching, all I have done is expanded the glob from elixir's end.

Using `cp`  is indeed simple but in the grand scheme of things, it means we have 2 separate ways of building the asset pipeline and I was heavily trying to avoid that.

taylordowns2000 (changes requested):
Hey @midigofrank , I got a review request here. (CC @theroinaochieng .)

I've just attempted the validation steps but my Monaco is dead:
image

```ex
[info] GET /assets/monaco-editor/vs/loader.js
[debug] ** (Phoenix.Router.NoRouteError) no route found for GET /assets/monaco-editor/vs/loader.js (LightningWeb.Router)
    (lightning 2.10.9) deps/phoenix/lib/phoenix/router.ex:541: LightningWeb.Router.call/2
    (lightning 2.10.9) lib/lightning_web/endpoint.ex:1: LightningWeb.Endpoint.plug_builder_call/2
    (lightning 2.10.9) lib/lightning_web/endpoint.ex:1: LightningWeb.Endpoint."call (overridable 3)"/2
    (lightning 2.10.9) deps/plug/lib/plug/debugger.ex:136: LightningWeb.Endpoint."call (overridable 4)"/2
    (lightning 2.10.9) lib/lightning_web/endpoint.ex:1: LightningWeb.Endpoint.call/2
    (phoenix 1.7.18) lib/phoenix/endpoint/sync_code_reload_plug.ex:22: Phoenix.Endpoint.SyncCodeReloadPlug.do_call/4
    (plug_cowboy 2.7.2) lib/plug/cowboy/handler.ex:11: Plug.Cowboy.Handler.init/2
    (cowboy 2.12.0) /Users/taylor/build/lightning/deps/cowboy/src/cowboy_handler.erl:37: :cowboy_handler.execute/2
    (cowboy 2.12.0) /Users/taylor/build/lightning/deps/cowboy/src/cowboy_stream_h.erl:306: :cowboy_stream_h.execute/3
    (cowboy 2.12.0) /Users/taylor/build/lightning/deps/cowboy/src/cowboy_stream_h.erl:295: :cowboy_stream_h.request_process/3
    (stdlib 5.2.3) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

I also see an approval from Joe, but what sounds like an open conversation. Can you only add me back as a reviewer once this thing is ready to test out?

Or, if it is ready to test, please drop in some validation instructions that replace this current set of instructions:

> Ensure you don't have any cache. You can test this on a private window
> 
> 1. Run mix esbuild.install to get the newly configured version
> 2. Turn off your internet
> 3. Start the server and try editing any job's body

As a separate bit... I'm worried that there's now a `RUNNINGLOCAL.md` file. I see this was created 4 months back, and I'm sorry for not noticing and speaking up then, but why does this exist and duplicate so much of the official https://openfn.github.io/lightning/readme.html#dev-on-lightning-locally instruction?

## Conclusion
Two things to do here:
1. Let me know if/when this is ready for review. (If now, provide updated review instructions.)
2. Delete `RUNNINGLOCAL.md` or... justify its existence, remove duplication from the README.md, and link to it from the README.md

Thank you! This is awesome work and I know MSF will be delighted to see this in action.

theroinaochieng:
@taylordowns2000 thank you for this feedback. I'll circle back with Frank and re-request a review when the provided actions are completed.

midigofrank:
Hey @taylordowns2000 I have just pushed the fix for the issue you were experiencing, sorry my bad, I've also updated the validation instructions accordingly.

As for `RUNNINGLOCAL.md`, it is separate because there was need for a guide that can help a dev setup and run lightning locally. Yes, most of it is duplicated from the readme but the whole idea was to declutter the README

taylordowns2000 (changes requested):
Thanks for the update. Looking better. 2 things:

1. See image below. The editor sorta seems to work, but I'm getting this perpetual "Loading types" spinner. Is that simply because I don't _also_ have offline language packages, or is this a separate issues? If it's meant to work without the spinner, can you give me some instructions to validate this the way that MSF will see it? (Maybe some adaptor install instructions?) Basically, for validation, I want to know what _YOU DO_ when you validate that this all works before pushing so that I can follow those same steps and see the same manual checks you're performing via the UI all pass on my side too.
2. If the idea is to declutter the README, please go ahead and dry it up. Don't duplicate the code elsewhere (that means there are two separate sets of instructions that need to be maintained) but instead delete everything in the README and in the section where we used to explain how to run locally, provide a link to where the new instructions live. Also please change the title of the document from `RUNNINGLOCAL.md` to `DEVELOPMENT.md`. The official way to run lightning locally is the first section under "Getting Started" and requires docker: https://openfn.github.io/lightning/readme.html#run-via-docker . What you're describing in this new file, it seems, is how to build stuff using Erland and Elixir from scratch in case you want to do local development. This is the [Dev on Lightning Locally](https://openfn.github.io/lightning/readme.html#dev-on-lightning-locally) section.

image

Thanks @midigofrank , my sense is that you've probably got this all working perfectly—congrats—just need to tidy up and we're good to go!

josephjclark:
@taylordowns2000 the RUNNINGLOCAL stuff is unrelated to this PR. It's not a new file or anything, nor is it significantly changed here

I take your points about the structure, and actually this stuff is none of my business. But maybe those comments need to be spun out into a new issue? Might needs Stu's eyes too - I don't know who created RUNNINGLOCAL or when or why 🤔

josephjclark:
That's also a really good point about types, metadata, docs. They might have hidden internet dependencies. I'll have a look later.

We should definitely cut cloth there and make sure those services fail gracefully. That might indeed be a new issue. I'll get back to you

midigofrank:
Hey @taylordowns2000 @josephjclark ,
RE:

1.) Regarding the loading spinner, yes it is related to loading of types and adaptor metadata from npm. I agree with @josephjclark here, we should create a separate issue for this. While trying to debug this, I found and fixed another unrelated bug. Esbuild was rewriting the `.ttf` file to `.js`. I have tested this over and over again with different jobs and it seems to be working just fine.

2.) About `RUNNINGLOCAL.md`, I have just created a separate issue to handle it: https://github.com/OpenFn/lightning/issues/2812. One thing we need to be careful of though is that we have links(solutions in community) that point to the readme file

josephjclark:
Looking good to me @midigofrank 

The Loading Types thing is very annoying. I'll open an issue for that now. I'll if I can raise a quick fix but if it doesn't happen before EOD Tuesday, it's not going to happen from me

midigofrank:
Hey @josephjclark , I'm waiting for your 2nd approval on this before merging

josephjclark (approved):

PR #2831, opened 2025-01-16
Component Injection for AI Responses Feedback

### LINKED ISSUE #2495: AI Assistant: Get feedback on AI responses
Here are a few things we want to do: 

- Add a flag that lets user report an unsatisfactory response with comment. 
- A posthog survey to collect experience.

elias-ba:
### Description

This PR allows to inject custom components for collecting user feedback on the AI assistant responses.

Closes #2495 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/ai_assistant_component.ex
+++ b/lib/lightning_web/live/workflow_live/ai_assistant_component.ex
@@ -817,7 +817,11 @@ defmodule LightningWeb.WorkflowLive.AiAssistantComponent do
               <div :if={@ai_feedback} class="flex justify-start mt-4">
                 <%= Phoenix.LiveView.TagEngine.component(
                   @ai_feedback.component,
-                  %{session_id: @session.id, message_id: message.id},
+                  %{
+                    session_id: @session.id,
+                    message_id: message.id,
+                    myself: @target
+                  },
                   {__ENV__.module, __ENV__.function, __ENV__.file, __ENV__.line}
                 ) %>
               </div>
--- a/lib/lightning_web/live/workflow_live/ai_assistant_component.ex
+++ b/lib/lightning_web/live/workflow_live/ai_assistant_component.ex
@@ -745,10 +745,7 @@ defmodule LightningWeb.WorkflowLive.AiAssistantComponent do
   attr :target, :any, required: true
 
   defp render_individual_session(assigns) do
-    assigns =
-      assign(assigns,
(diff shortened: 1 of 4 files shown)


jyeshe (commented):
Hey @elias-ba this is great. I was just wondering if we couldn't implement this component in Lightning once the Apollo client is already here. The billing app would just reuse Lightning ai_feedback component.

I would do that for 2 reasons: i) for a small aesthetics detail on rendering the component once using the `Phoenix.LiveView.TagEngine` it becomes more verbose; ii) specially to make configuration simpler (decreasing the surface area of external API).

lib/lightning_web/live/workflow_live/ai_assistant_component.ex
```diff
@@ -962,4 +971,8 @@ defmodule LightningWeb.WorkflowLive.AiAssistantComponent do
     user_messages = Enum.filter(session.messages, &(&1.role == :user))
     length(user_messages) > 1
   end
+
+  defp ai_feedback do
+    Application.get_env(:lightning, :ai_feedback)
```
midigofrank, line 976: Hey @elias-ba, is this a new config? Is it a flag? Maybe we should add it to `Lightning.Config` instead

midigofrank (changes requested):

jyeshe (approved):
Thanks for the clarification @elias-ba . I am waiving my request on his PR because even though it would simplify on the component side, it follows the current structure for google tag (`gtag`) processing on the billing app. As we've agreed, let's open an Engineering issue to move this code from the root to one or more Hooks. I will complement on it that there is a building complexity to achieve it without copying the file on build time. It will require further investigation.

lib/lightning_web/live/workflow_live/ai_assistant_component.ex
elias-ba: The reason why I did it this way was to avoid introducing a new config variable in Lightning. We are trying to do the least possible to be able to inject the feedback component without defining how we inject it in Lightning.

elias-ba: hey @midigofrank, thanks a lot for the clarifications. I have added the `:ai_feedback` config into the `config/config.exs`

midigofrank (approved):

Issue #2617, opened 2024-10-29
Editor loading very slowly


christad92:
Today during a demo I noticed that the editor was very slow when loading and had to wait for about one minute for it to load. The input and output panels were loaded but the editor panel was blank.  

Workflow:

stuartc:
I can't reproduce this, has it happened subsequently? With my cache disabled I timed this as fully usable in 5 seconds.

josephjclark:
I suspect this is related to cache and slow internet. There's like 6mb of editor stuff that needs downloading and if its not cached then it will take some time.

josephjclark:
Huh - I just saw this!

I loaded [this workflow](https://app.openfn.org/projects/2fedca61-bb42-4e65-a908-7a60f1b25c1e/w/ce0cd63c-4aa1-4e5b-88be-813d125ebb8a?s=6637fdf1-62f4-446d-8f59-98b7b37dc3e3&a=2a30533e-0b31-4d5c-ba62-db49de5235ad&v=72&m=expand#log) on app.openfn.org, using a good fast stable connection and _surely _ with the actual editor source cached.

It took like 30 seconds for the job code to load.

My dev console shows this:

```
Job not found 6637fdf1-62f4-446d-8f59-98b7b37dc3e3
```
![Image](https://github.com/user-attachments/assets/be394e89-e618-4f0b-a8d8-da8705bd2034)

That is the job I'm looking at, which _did_ subsequently load.

I think this issue might actually: _sometimes there's an error loading job expressions into the inspector and that stops the editor loading_

josephjclark:
Yes, the editor component will not mount until the workflow has been loaded into the zustand store.

The error in my console was thrown in the editor's mount code:
```js
  render() {
    const { adaptor, disabled, disabledMessage, jobId } = this.el.dataset;

    checkAdaptorVersion(adaptor);

    this.findWorkflowEditorStore().then(workflowStore => {
      const job = workflowStore.getState().getById<Lightning.Job>(jobId);

      if (!job) {
        console.error('Job not found', jobId);
        return;
      }

      if (JobEditorComponent) {
        this.componentRoot?.render(
          <JobEditorComponent
            adaptor={adaptor}
            source={job.body}
            metadata={this.metadata}
            disabled={disabled === 'true'}
            disabledMessage={disabledMessage}
            onSourceChanged={src => this.handleContentChange(src)}
          />
        );
      }
    });
  },
```

josephjclark:
[...]

PR #2851, opened 2025-01-23
2617 console instrumentation

rorymckinley:
### Description

At the moment, the exact cause of the job editor slow loading is unknown - so this is the first of 2 changes to attempt to at least instrument the slow behaviour. 

As I am unsure how this data will be used - I have included both start/end times (useful to identify possible overlaps) as well as span duration (useful to avoid a bunch of subtraction :)).

#2617 

### Validation steps

Confirm that both the workflow editor and the job editor still load correctly. In the browser console you should see logging similar to the below:

image

As far as I can tell, there are no tests that cover this code, other than the LiveView tests?

M & Ms are named after the two developers of the sweet - Mars and Murrie. Forrest Mars (of Mars Bar fame) is suspected of basing them off a British sweet, Smarties, during WW2. Both M&Ms and Smarties were developed to allow for more chocolate sales in summer as the coating meant that both sellers and consumers had less issues as result of melted chocolate.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
 
 - Audit the provisioning of projects via the API
   [#2718](https://github.com/OpenFn/lightning/issues/2718)
+- Temporary instrumentation for JobEditor to help identify performance issues.
+  [#2617](https://github.com/OpenFn/lightning/issues/2617)
 
 ### Changed
 
--- a/assets/js/job-editor/mount.tsx
+++ b/assets/js/job-editor/mount.tsx
@@ -81,13 +81,7 @@ export default {
 
       instrumentFinish('editor load');
 
-      this.requestMetadata().then(() => {
-        instrumentStart('Render - metadata received');
-
-        this.render()
-
-        instrumentFinish('Render - metadata received');
-      });
+      this.requestMetadata().then(() => this.render());
     });
   },
   handleContentChange(content: string) {
@@ -136,13 +130,7 @@ export default {
   },
   requestMetadata() {
     this.metadata = true; // indicate we're loading
-
-    instrumentStart('Render - prior to requesting metadata');
-
     this.render();
-
-    instrumentFinish('Render - prior to requesting metadata');
-
     return new Promise(resolve => {
       const callbackRef = this.handleEvent('metadata_ready', data => {
         this.removeHandleEvent(callbackRef);
--- a/assets/js/job-editor/mount.tsx
+++ b/assets/js/job-editor/mount.tsx
@@ -60,6 +60,8 @@ export default {
     // this.handleContentChange(this.currentContent);
   },
   mounted(this: JobEditorEntrypoint) {
+    instrumentStart('Mount to first render');
+
     window.jobEditor = this;
 
     this._debouncedPushChange = pDebounce(this.pushChange, EDITOR_DEBOUNCE_MS);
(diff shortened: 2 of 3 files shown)


josephjclark:
hmm - it's very noisy, and I find it hard to know which noise I'm supposed to be looking at.

I'd suggest:
- we only log durations, not timestamps
- Remove the metadata logging entirely
- We log something like `editor load` to capture the load from mount until Monaco gets first rendered. This, I think, is actually the ` this.componentRoot?.render(` line in `job-editor/mount.js`. I'm pretty sure that this only runs after the workflow_params are ready.
 - We log something like `workflow-params load` in place of `get-initial-state -> current-workflow-params`. I'd change the label but leave the logic as the same.

So that would just be the two events? And I'd expect the `workflow-params load` to be like 90% of the `editor load` duration (especially  in these load cases). 

Then in metrics, what we're looking for is any place where `workflow-params load` is greater than a couple of seconds.

rorymckinley:
@josephjclark Thank you for the detailed feedback - I have replied inline below:

> hmm - it's very noisy, and I find it hard to know which noise I'm supposed to be looking at.
> 

Yes, it is noisy. Given that it will take a non-trivial amount of time between deciding that we need extra instrumentation and that instrumentation becoming available in prod, I deliberately erred on the side of noisiness based on the fact that we are not really sure where the error is occurring and we will hopefully only be looking at a small handful of these cases.

> I'd suggest:
> 
> * we only log durations, not timestamps

The one time that I have been able to produce something that seems similar (on local dev) involved overlapping activities - current-workflow-params had not arrived but the code was querying the store for job details. While I am not sure if this is related, I thought it might be useful to be able to identify if an overlap is occurring. For this, timestamps provide a more concrete answer than durations.

Of course, if we have timestamps, we can calculate durations from these, but I thought I would leave it in as a convenience. I could just go with timestamps as they provide 'richer' data?

> * Remove the metadata logging entirely
> * We log something like `editor load` to capture the load from mount until Monaco gets first rendered. This, I think, is actually the ` this.componentRoot?.render(` line in `job-editor/mount.js`. I'm pretty sure that this only runs after the workflow_params are ready.

When I edit a job, the `render()` method is called 3 times (after mounting, before requesting metadata, after metadata has arrived). Each time `render` is called, the [code](https://github.com/OpenFn/lightning/blob/main/assets/js/job-editor/mount.tsx#L113) within `if (JobEditorComponent) { .. }` is executed (based on observations when running locally).

This means that, if I end the timer in this block of code, it will succeed on the first call and fail on the two subsequent calls. This generates noise in the console that is (arguably) less useful than the noise I chose to add to the console :).

My understanding of the React API is that `componentRoot?.render` does not return an indication that has changed something or not so I cannot cheaply identify if anything has changed.

Once again, taking into account the level of uncertainty that we currently have, I erred towards being noisier (by logging the additional renders) in the hopes that it would reduce the number of 'Observe-Investigate-Instrument` cycles we would need to perform before being able to resolve the issues.

> * We log something like `workflow-params load` in place of `get-initial-state -> current-workflow-params`. I'd change the label but leave the logic as the same.

I will change this label as well as the label to identify `editor load`.

josephjclark:
Okay, I hear you about the "noise". Let's leave the timestamps and durations in place. 

My concern about this stuff is: if we have to ask another user to open their dev tools and read the timings back to us, that process is a lot harder when there's more noise. But if we can track this stuff from the backend or via sentry than that's less relevant.

> When I edit a job, the render() method is called 3 times (after mounting, before requesting metadata, after metadata has arrived). Each time render is called, the [code](https://github.com/OpenFn/lightning/blob/main/assets/js/job-editor/mount.tsx#L113) within if (JobEditorComponent) { .. } is executed (based on observations when running locally).

Sure, but the metadata renders really aren't  important. What we care about is the time until the user sees the code. If you disable the metadata renders completely, you'll see that the editor still appears with the user's code.

So I really think it's just that very first render we need to track, and we can ignore the others.

rorymckinley:
@josephjclark I have removed the metadata instrumentation.

josephjclark (approved):
Great! I'm happy

PR #2866, opened 2025-01-24
fix: remove 'Gmail' from list for #2854"

### LINKED ISSUE #2854: Strange behaviour for credential types
On `v2.10.12 (1.4.12)` I'm seeing two strange things related to credentials:
1. the `hubtel` credential type doesn't show up, despite a seemingly valid configuration schema existing [here](https://github.com/OpenFn/adaptors/blob/main/packages/hubtel/configuration-schema.json) - this was generated directly by the `pnpm generate <adaptor>` command.
2. there is a `gmail` credential type, despite gmail using Oauth. It asks the user to enter an access token manually, but access tokens are automatically generated via the Oauth flow.

- [x] Yes
- [ ] No

1. Go to https://app.openfn.org/credentials
2. Click on "Add new"
3. Check out the "Gmail" type and the _missing_ "Hubtel" type

taylordowns2000:
This fixes #2854 (the other part of that issue—Hubtel—will be resolved by the next build) and closes the bug.

The open question is how to handle this TODO for the future @elias-ba , @stuartc . (Don't think this is high priority, but will come up again next time we add another OAuth based adaptor, @theroinaochieng .)

See [here](https://github.com/OpenFn/lightning/blob/main/lib/lightning_web/live/credential_live/credential_form_component.ex#L604-L605).
```
    schemas_options
    |> Enum.concat([{"Raw JSON", "raw", nil, nil}])
    # TODO: Gracefully determine which adaptors define custom credential types
    # and which adaptors use OAuth2 credential types.
    |> List.delete({"Googlesheets", "googlesheets", nil, nil})
    |> List.delete({"Gmail", "gmail", nil, nil})
    |> Enum.sort_by(& &1, :asc)
```

Please feel free to open an issue and prioritize as you see fit!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to
 - Bump PostCSS [#2863](https://github.com/OpenFn/lightning/pull/2863)
 - Replaced HTTPoison with Tesla in the AdaptorRegistry.
   [#2861](https://github.com/OpenFn/lightning/pull/2861)
+- Remove googlesheets, gmail and collections from credential schemas list
+  [#2854](https://github.com/OpenFn/lightning/issues/2854)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


elias-ba (approved):
Nice one, Flash !

download

stuartc (approved):

PR #2939, opened 2025-02-17
Enable project level concurrency limit

### LINKED ISSUE #2906: Control run concurrency at a project level

Some projects which have more than one workflow calling to the same destination system are prone to having conflicts in concurrent situations.

For example, when upserting to the same Salesforce object with two workflows at the same time results in an error from Salesforce.

Setting the concurrency for each respective workflow limits the potential for this happening, but does not outright prevent it.

**Tell us about your idea - describe the solution you'd like.**

Being able to restrict _all_ runs into a maximum of one at a time would resolve this issue.

In the project settings we would need a place to specify the concurrency, along with some short but helpful copy explaining the impact of setting the number to 1.

When looking at the workflow settings, _if_ the project concurrency is lower than the value in the workflow, there should be some information explaining that it is being overridden.

**Additional context**

While there are other techniques that can be used to avoid this issue, like having both conflicting jobs have some retry logic (backing off for a few seconds) would resolve a specific class of issues. However this only applies to idempotent operations like upserts.

It should be noted that this feature is not something we want users to default to. It would severely limit the total possible performance for the project.

Extending off the queue claim SQL query, we would have the concurrency value be the equivalent of `min(workflow_concurrency, project_concurrency)`. Which would override any concurrency settings on the workflow.

**Concerns**

There is concern here, in environments that employ autoscaling based on the number of runs waiting to be claimed - this number would stay higher for longer, and any scaling decisions that are made downstream would consider this high number as a signal to provision more workers, but only one run would be claimed at the same time.

jyeshe:
## Description

This PR enables the RunQueue (the scheduler that processes run claims) to limit the concurrency on the project level. This concurrency control allows disabling the execution of parallel runs as shown on the option given in the UI:

image

Closes #2906 

## Validation steps

1. Go to settings and disable the parallel execution of runs.
2. On 2 workflows of the same project, add a job after the trigger that calls 
```
fn(state => {
  return new Promise((resolve) => {
    setTimeout(() => {
      resolve(state);
    }, 10000); // 10 second delay
  });
});
```
4. Send a POST request to each webhook of these workflows
5. While one is executing the other one is not.

Using default configuration:
image
After disabling concurrency on project settings:
image

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/runs/query.ex
+++ b/lib/lightning/runs/query.ex
@@ -67,7 +67,7 @@ defmodule Lightning.Runs.Query do
   - `id`, the id of the run
   - `state`, the state of the run
   - `row_number`, the number of the row in the window, per workflow
-  - `limit`, the maximum number of runs that can be claimed for the workflow
+  - `concurrency`, the maximum number of runs that can be claimed for the workflow
   """
   @spec in_progress_window() :: Ecto.Queryable.t()
   def in_progress_window do
@@ -82,19 +82,11 @@ defmodule Lightning.Runs.Query do
         partition_by:
           fragment(
             """
-              CASE
-                WHEN ? IS NULL AND ? IS NULL THEN ?
-                WHEN ? IS NULL THEN ?
-                WHEN ? < ? THEN ?
-                ELSE ?
-              END
+            CASE
+              WHEN ? IS NULL THEN ?
+              ELSE ?
+            END
             """,
-            w.concurrency,
-            p.concurrency,
-            w.id,
-            w.concurrency,
-            p.id,
-            w.concurrency,
             p.concurrency,
             w.id,
             p.id
@@ -110,25 +102,7 @@ defmodule Lightning.Runs.Query do
       # calculated here?
       row_number: row_number() |> over(:row_number),
       project_id: w.project_id,
-      concurrency: fragment(
-        """
-          CASE
-            WHEN ? IS NULL AND ? IS NULL THEN ?
-            WHEN ? IS NULL THEN ?
-            WHEN ? < ? THEN ?
-            ELSE ?
-          END
-        """,
-        p.concurrency,
-        w.concurrency,
-        nil,
-        p.concurrency,
-        w.concurrency,
-        w.concurrency,
-        p.concurrency,
-        w.concurrency,
-        p.concurrency
-      ),
+      concurrency: coalesce(p.concurrency, w.concurrency),
       inserted_at: r.inserted_at
     })
   end
--- a/lib/lightning/runs/query.ex
+++ b/lib/lightning/runs/query.ex
@@ -133,7 +133,7 @@ defmodule Lightning.Runs.Query do
     |> where(
       [r, ipw],
       r.state == :available and
-      (is_nil(ipw.concurrency) or ipw.row_number <= ipw.concurrency)
+        (is_nil(ipw.concurrency) or ipw.row_number <= ipw.concurrency)
     )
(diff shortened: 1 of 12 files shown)


lib/lightning_web/live/workflow_live/components.ex
```diff
@@ -145,16 +146,27 @@ defmodule LightningWeb.WorkflowLive.Components do
...
-            :if={Enum.empty?(@form[:concurrency].errors)}
+            :if={
+              not @project_concurrency_disabled and
+                Enum.empty?(@form[:concurrency].errors)
+            }
             class="text-xs text-slate-500 italic"
           >
-            <%= case @form[:concurrency].value do
-              nil -> "Unlimited (up to max available)"
+            <%= case @form[:concurrency].value || "" do
```
stuartc, line 155: I never noticed this in the original implementation, should we try and enforce either `nil` or "positive number" here. I'm wondering why I ever needed to check for a string 🤔.

jyeshe: Just simplified a bit the pre-existing code that was chexking both for the string and nil with the same message duplicated. Maybe there was a string due to the form params but as far as I remember the Form module would cover this and set as nil.

lib/lightning/runs/query.ex
```diff
@@ -115,14 +126,14 @@ defmodule Lightning.Runs.Query do
...
+      on: r.id == ipw.id,
       as: :in_progress_window
     )
     |> where(
-      [r, w],
-      (is_nil(w.concurrency) or w.row_number <= w.concurrency) and
-        w.state == "available"
+      [r, ipw],
+      r.state == :available and
+        (is_nil(ipw.concurrency) or ipw.row_number <= ipw.concurrency)
```
stuartc, line 136: Super interesting, does this give us a small lift in query performance?

stuartc (commented):
This is great @jyeshe, really great! I left few comments, but one that made me pause (not a dealbreaker) was the old behaviour around the form checking for an empty string on `concurrency`; was wondering if we could do something about in this PR.

I'll wait for language comments before considering this done done. 👌

lib/lightning/runs/query.ex
jyeshe: A small one but repeatedly might save us some cycles by discarding right away the 'claimed' and 'started' from ipw, that were inserted before the 'available' ones.

lib/lightning_web/live/workflow_live/components.ex
jyeshe: For future reference the `Form`module emulates the params converting nil values to empty strings.

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -114,6 +116,31 @@
...
+                    disabled={!@can_edit_project}
+                    label=""
+                    checked_value="1"
+                    unchecked_value={nil}
+                  />
+                </div>
+              </div>
+              <div>
+                <small class="block my-1 text-xs text-gray-600">
+                  Additionally you can setup all runs from this project, accross multiple workflows, to not execute in parallel.
```
elias-ba, line 140: Could we replace the word "Additionally" by "Note that", for more clarity on the note ?

elias-ba (approved):
@jyeshe this is looking good to me. I have one tiny comment on the word "Additionally" but it's not blocking at all. Great work

stuartc (approved):

PR #2946, opened 2025-02-19
Transfer credentials

elias-ba:
## Description

Closes #__

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+      nil ->
+        changeset
+
+      email ->
+        case Lightning.Repo.exists?(User |> where(email: ^email)) do
+          false ->
+            changeset
+
+          true ->
+            user = Lightning.Accounts.get_user_by_email(email)
```
jyeshe, line 182: Can be done in a single query and check for `nil`

lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+
+            projects =
+              credential
+              |> Lightning.Credentials.diff_project_credentials_and_project_users(
```
jyeshe, line 186: What you think about naming this one by the purpose (goal of the function)? I was trying to understand this function because project credentials and project users are different entities so I was trying to grasp what would be the meaning to diff them? By reading your docs I could see that it's about detecting invalid projects.

lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+                user
+              )
+              |> Lightning.Projects.get_projects_by_ids()
```
jyeshe, line 189: I would go with a `Enum.map(&Projects.get_project/1)` instead of creating a function to get for a list

lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+
+            if Enum.empty?(projects) do
+              changeset
+            else
+              projects_names = Enum.map_join(projects, ", ", & &1.name)
```
jyeshe, line 194: neat! 👏

lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+                :email,
+                "User doesn't have access to these projects: #{projects_names}"
+              )
+            end
+        end
+    end
+  end
+
+  def validate_not_same_user(changeset, field, current_user, opts \\ []) do
+    message = opts[:message] || "Cannot be the same as current user"
```
jyeshe, line 207: maybe reword `current user` to `the user logged in`

lib/lightning/accounts/user.ex
```diff
@@ -164,14 +164,74 @@ defmodule Lightning.Accounts.User do
...
+    cond do
+      is_nil(value) ->
+        changeset
+
+      value == current_value ->
+        add_error(changeset, field, message)
+
+      true ->
+        changeset
+    end
```
jyeshe, line 220: the `current_user` will always have an email so two conditions for `value == current_value` true or false would be enough. As this function is used to check equality based on the email, would you mind starting with a specific one (without the `field` parameter)? If need in the future in could be generalised unless you know that it will be or that it is very likely to be used with another field. Abstract functions are more flexible but specific ones have a more clear purpose which makes easier to maintain.

lib/lightning/accounts/user_notifier.ex
```diff
@@ -415,4 +415,54 @@ defmodule Lightning.Accounts.UserNotifier do
...
+        token
+      ) do
+    validity_days = Lightning.Config.credential_transfer_token_validity_in_days()
+    validity_text = format_validity_period(validity_days)
+
+    confirmation_url =
+      url(~p"/credentials/transfer/#{credential.id}/#{receiver.id}/#{token}")
+
+    deliver(owner, "Confirm your credential transfer", """
+    Hi #{owner.first_name},
```
jyeshe, line 435: Nice way to enforce the real person in case the password/login is shared among multiple people.

lib/lightning_web/controllers/credential_transfer_controller.ex
```diff
@@ -0,0 +1,35 @@
...
+  use LightningWeb, :controller
+
+  alias Lightning.Credentials
+
+  def confirm(conn, %{
+        "token" => token,
+        "receiver_id" => receiver_id,
+        "credential_id" => credential_id
+      }) do
+    %{assigns: %{current_user: %{id: owner_id}}} = conn
```
jyeshe, line 11: For transfering a credential seems to be the most responsible and safe way by requiring the user to be authenticaed.

lib/lightning_web/live/credential_live/transfer_credential_modal.ex
```diff
@@ -0,0 +1,317 @@
...
+    <% else %>
+      Transfer Credential Ownership
+    <% end %>
+    """
+  end
+
+  defp modal_subtitle(assigns) do
+    ~H"""
+    <%= if @revoke_transfer do %>
+      A transfer of this credential is pending. Click the revoke button to cancel it.
```
jyeshe, line 311: 💯

jyeshe (changes requested):
Congrats for thinking about this thoroughly including a way to unblock pending transfers and being cautious on the transfer authorisation. Some comments mainly on maintainability ...

lib/lightning/accounts/user.ex
elias-ba: This is fixed, thanks for the catch

lib/lightning/accounts/user.ex
elias-ba: This is also done, thanks a lot

lib/lightning/accounts/user.ex
elias-ba: Mhmmm I disagree on this one. Having one function that fetches all the projects once seems better to me than having to hit the database n times depending on the number of projects. That bein said I refactored the functions to handle everything in one single database query.

```elixir
  @spec projects_blocking_credential_transfer(
          Lightning.Credentials.Credential.t(),
          Lightning.Accounts.User.t()
        ) :: [Lightning.Projects.Project.t()]
  defp projects_blocking_credential_transfer(
         %Credential{id: credential_id},
         %User{id: user_id}
       ) do
    from(p in Lightning.Projects.Project,
      join: pc in Lightning.Projects.ProjectCredential,
      on: pc.project_id == p.id and pc.credential_id == ^credential_id,
      left_join: pu in Lightning.Projects.ProjectUser,
      on: pu.project_id == p.id and pu.user_id == ^user_id,
      where: is_nil(pu.id),
      distinct: true,
      select: p
    )
    |> Repo.all()
  end
```

lib/lightning/accounts/user.ex
elias-ba: Thanks, changed it to this

```elixir
recipient_email == sender_email ->
        Ecto.Changeset.add_error(
          changeset,
          :email,
          "You cannot transfer a credential to yourself"
        )
```

lib/lightning/accounts/user.ex
elias-ba: Thanks, this is also changed and I agree that we do not need to generalize too early.

lib/lightning_web/controllers/credential_transfer_controller.ex
elias-ba: That's is already done in the Router

```elixir
pipe_through [:browser, :require_authenticated_user]
```

jyeshe: Sure, I was praising your careful approach (:

lib/lightning/accounts/user.ex
jyeshe: For `projects_blocking_credential_transfer` instead of `get_projects_by_ids` it makes sense.

lib/lightning_web/controllers/credential_transfer_controller.ex
elias-ba: Ah ! Thanks my brother 😊

lib/lightning/accounts/user.ex
elias-ba: Thanks a lot

jyeshe (approved):
Nicely done man!

PR #2947, opened 2025-02-19
allow users to configure their log levels

### LINKED ISSUE #2206: Control log levels in run logs
~As a project admin/owner, I want to control the level of logs printed in my run logs through the workflow settings panel. There are 5 levels of logs (Debug, Info, Warning, Error, and Critical). The default behavior here will be to print all levels of logs but when I choose a level (say Warning), all the levels preceding it will be printed (Debug, Warning, Error).  The image below is a reference for how log levels are printed.~

image

~When printing logs, we need to print the log level that's configured for the workflow.~

~Design:  https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App-(new)?node-id=4538-643&t=gKE781fliu7QFjgA-0~
---

Few ideas from @josephjclark: 
When displaying the log levels, we should: 
- Add colour to the output to reinforce saliency
- Add an icon in the margin to indicate the log level

EDIT:
As a user I want to be able to see only specific logs so that I can be able to debug/find logs easily.

The suggested implementation is a setting on the workflow: https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App-(new)?node-id=4538-643&t=gKE781fliu7QFjgA-0 

We should also provide a way for an individual user to override this directly in the log viewer. This can be stored in the `user.preferences`

midigofrank:
## Description

This PR makes it possible for users to configure their preferred log levels. This preference is persisted and used when they are viewing logs.
The default log level set if there is none configured is `info`

Screenshot 2025-02-25 at 11 46 26

Closes #2206 

## Validation steps
- Open the log viewer, either in the inspector or when viewing the whole run
- The gear/cog icon appears on the top right corner of the viewer. Clicking on it opens up a modal
- When you check/uncheck the levels in the modal, you'll see the logs getting filtered.

## Notes for the reviewer
I have chosen to persist the `level type` as opposed to score, i.e. `debug instead of 0`. This is so that we can have one place for converting the `type` to `score`. Currently the logic is in the js side. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


assets/js/log-viewer/store.ts
```diff
@@ -17,16 +17,37 @@ interface LogStore {
...
+  setDesiredLogLevels: (desiredLogLevel: string[] | undefined) => void;
 }
 
-function findSelectedRanges(logs: LogLine[], stepId: string | undefined) {
+// check if a log matches the desired log level
+function matchesLogFilter(
+  log: LogLine,
+  desiredLogLevels: string[] | undefined
+): boolean {
+  if (!desiredLogLevels) return true;
```
jyeshe, line 29: For the sake of readability I would validate undefined explicitly because of the way this if sounds (not desired => true).

assets/js/log-viewer/store.ts
```diff
@@ -17,16 +17,37 @@ interface LogStore {
...
+) {
   if (!stepId) return [];
 
   const { ranges } = logs.reduce<{
     ranges: { start: number; end: number }[];
     marker: number;
   }>(
     ({ ranges, marker }, log) => {
+      // Skip logs that don't match the desired log level
+      if (!matchesLogFilter(log, desiredLogLevels)) {
```
jyeshe, line 47: For a `findSelectedRanges` I would expect the matching logs to be filtered. If it works down the line, maybe rename to `rejectSelectedRanges`?

assets/js/log-viewer/store.ts
```diff
@@ -121,18 +161,34 @@ export const createLogStore = () => {
...
+      const state = {
+        highlightedRanges: findSelectedRanges(
+          logLines,
+          stepId,
+          desiredLogLevels
+        ),
+      };
+
+      if (prevLogLevels !== desiredLogLevels) {
+        state.formattedLogLines = stringifyLogLines(logLines, desiredLogLevels);
```
jyeshe, line 179: If the log levels change would it be expected that it filter the logs again?

lib/lightning_web/components/viewers.ex
```diff
@@ -55,32 +68,117 @@ defmodule LightningWeb.Components.Viewers do
...
         </div>
       </div>
     <% end %>
     """
   end
 
+  defp selected_log_levels(user) do
+    configured_levels = user.preferences["log_levels"] || []
+
+    if configured_levels == [] do
```
jyeshe, line 110: Please notice it puts the [] to not use it. Would use  a `with` to nil, the default is good.

lib/lightning_web/components/viewers.ex
```diff
@@ -55,32 +68,117 @@ defmodule LightningWeb.Components.Viewers do
...
+          id={"#{@id}-form"}
+          for={to_form(%{}, as: :log_filter)}
+          phx-change="save-log-filter"
+        >
+          <fieldset>
+            <div class="space-y-2">
+              <%= for level <- all_log_levels() do %>
+                <.input
+                  type="checkbox"
+                  label={level}
```
jyeshe, line 162: This is fancy and super flexible, I am just afraid it will become harder to debug because instead of 1 user input we are allowing a permutation of options.

jyeshe (changes requested):
This is very thorough and flexible. The findSelected function might need to be inverted or renamed.

midigofrank (commented):
Hey @jyeshe , thanks for the detailed review. Please note that in the `log store`, there were 2 possible routes to handle the state:

1.  Introduce a different key, `filteredLogLines`
2. Filter `logLines` during each operation 

I decided not to introduce a different key because then we would have 2 sources of truth with regards to `logLines`.

Filtering `logLines` during each operation might seem costly but it is actually trivial. Here's why:

There are 2 possible operations:

- New Log lines are added 
- The user selects a different step.

Before this PR, when new log lines are added we append the lines to the existing `logLines`. The resulting value is the new state of the `logLines`. After this we format the `logLines[]` into a string which is passed to monaco. What this PR does is that during the process of converting `logLines[]` into a string, we skip `logLine`s that don't match the desired filter. Regardless of this PR, we were still iterating over all the log lines while formatting them into our desired string.

This is the reason why I wasn't passing an already filtered subset to the subsequent operations. I've kind of hijacked the existing operations for performance reasons

assets/js/log-viewer/store.ts
jyeshe: What makes it harder to understand is that it it's trying to find ranges and does operations that sounds the opposite like skipping logs when doesn't match `desiredLogLevels` (as if `findSelected` is conflicting to _not_ matching the desired logs).

I understand the intention was super, to abbreviate the reduce, it's more about the names.

Same here on this same function, for findSelected passing a stepId and negating it with:
```
if (log.step_id !== stepId) {
        return { ranges, marker: nextMarker };
      }
```

lib/lightning_web/components/viewers.ex
jyeshe: I am imagining that `user.preferences["log_levels"]` will have either some _non empty_ list or nil (from a combox that has log levels or nil for users that has not set it yet)

lib/lightning_web/components/viewers.ex
jyeshe: It's a comment more about a feature decision not your implementation. I think allowing users to check multiple levels like DEBUG and ERROR makes it very flexible but more complex than rejecting/eliminating all logs that have higher level than the selected. I am not sure how we receive this log levels, if by an integer but would have it as a goal to filter them fastly.

lib/lightning_web/components/viewers.ex
jyeshe: Combobox or checkboxes (:

lib/lightning_web/components/viewers.ex
midigofrank: Valid point

midigofrank:
@josephjclark I have added you as a reviewer to help check the user interaction side of things

midigofrank:
Screenshot 2025-02-25 at 12 24 52
I have updated the copy to mention "success" logs.

stuartc (approved):
@midigofrank this is good, I think we can do better on the "Config" button - but imo I think it should be even more 'hidden'... I think let's roll with this for now.

I'm thinking that this kind of button/functionality should be either small bar at the top or bottom of the editor. But I'm thinking maybe waiting and see what else makes sense to put on a bar (if we actually do that).

josephjclark (approved):

test/lightning_web/live/workflow_live/edit_test.exs
```diff
@@ -3717,6 +3717,93 @@ defmodule LightningWeb.WorkflowLive.EditTest do
...
+      {:ok, view, _html} =
+        live(
+          conn,
+          ~p"/projects/#{project.id}/w/#{workflow.id}?#{%{a: run.id, m: "expand", s: job_1.id}}"
+        )
+
+      run_view = find_live_child(view, "run-viewer-#{run.id}")
+
+      render_async(run_view)
+      assert render(run_view) =~ "Configure log levels"
```
jyeshe, line 3774: 💯

jyeshe (commented):
Hey @midigofrank this looks super neat! I am not very familiar with Runtime logs but I get this when I select either warn or error:
image
Is it something local or expected?

Moreover if I would add something to the feature, my 5 cents would be to add an OK button on the modal that allows selecting the log. It's beautiful and practical but just to give an option for those skeptical users that think, will it be saved/applied or not. It is less important as one can see the result right away on the background but as a matter of giving an option that the user knows that it will definitely apply the config.

midigofrank:
Thanks for the review @jyeshe.

The 2 logs you're seeing in the screenshot are of `success` level. This is the reason why I mention `success` in the filter. In order to make the options brief, I sacrificed `success/always` logs. Those log types will _always_ be displayed anyway.

Regarding the `OK` button. Yes I thought about it, but I decided not to add it because it introduces an "unsaved" state.
Meaning, if the user decides to click on `X` instead of the `Save/Ok` button, we should discard their current  unsaved selection.
The alternative was to _ONLY_ apply the selection after hitting the `Save/Ok` button (on submit), this didn't really feel like a good UX because after clicking save we should probably close the modal but then what if the user realizes that they want another level instead?

Does this now sound okay to you?

midigofrank:
Maybe I should call them `job completion logs` instead of `success`. @josephjclark is there any way to trigger a `success` log apart from job completion log? Is this accurate?
Screenshot 2025-02-25 at 18 47 32

midigofrank:
@jyeshe after consulting Joe, I have updated the copy on the filter modal. We no longer mention ~`success`~, we call them `key logs`

> Maybe I should call them `job completion logs` instead of `success`. @josephjclark is there any way to trigger a `success` log apart from job completion log? Is this accurate? Screenshot 2025-02-25 at 18 47 32

jyeshe:
> Thanks for the review @jyeshe.
> 
> The 2 logs you're seeing in the screenshot are of `success` level. This is the reason why I mention `success` in the filter. In order to make the options brief, I sacrificed `success/always` logs. Those log types will _always_ be displayed anyway.
> 
> Regarding the `OK` button. Yes I thought about it, but I decided not to add it because it introduces an "unsaved" state. Meaning, if the user decides to click on `X` instead of the `Save/Ok` button, we should discard their current unsaved selection. The alternative was to _ONLY_ apply the selection after hitting the `Save/Ok` button (on submit), this didn't really feel like a good UX because after clicking save we should probably close the modal but then what if the user realizes that they want another level instead?
> 
> Does this now sound okay to you?

I tend to prefer more explicit behaviours but I understand having no option other than the X gives a single path to the user. On a second though, naming it `OK` would not be the most clear path as the X becomes dubious. With the `Save` button, on the other hand, the user could click on different levels and it would only close on the `Save`. But it's only a suggestion about design taking a small portion of skeptical thinking in consideration.

jyeshe (approved):
I would go with the `Save` button for explicit action but it's more a matter of choice from the product team. Otherwise, it looks great! 👏

Issue #2914, opened 2025-02-07
Allow for blank log lines


stuartc:
### Details

We currently don't allow blank log lines to be stored.

However a lot of people emit blank `console.log()` and should be able to see the logs like as if they were running the job locally.

### User acceptance criteria

Blank log lines should be persisted, and this [Elixir.LightningWeb.RunChannel.Error error](https://openfn.sentry.io/issues/5878576609/?environment=prod&project=6446500&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=0) should no longer be experienced in the RunChannel:

```
#Ecto.Changeset<action: :insert, changes: %{timestamp: ~U[2025-02-07 09:32:16.249492Z], run: #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #Lightning.Run<>, valid?: true, ...>, source: "ADA", step_id: "162e723f-0fee-4ec8-b40b-8994fb50168d", run_id: "1fe6ee9a-046a-4af9-9471-ae7b7013dbea"}, errors: [message: {"can't be blank", [validation: :required]}], data: #Lightning.Invocation.LogLine<>, valid?: false, ...>
```

PR #2959, opened 2025-02-21
Fix another variant blank log lines

stuartc:
## Description

Re: #2914

Been seeing a bunch of warnings on Sentry saying that a log line couldn't be blank/is rejected.

Really thought I fixed this one.

From what it looks like, `console.log(undefined)` arrives as `%{"message": []}` which was previously considered blank.

This PR converts drops `[]` from the list of 'empty values' when casting, which is valid (from the previous set of changes on this issue).

Now the question is here, is there such a thing as a log line that is "too blank"?

```
> console.log()

undefined
> console.log(undefined)
undefined
undefined
```

So in the node real, an empty console.log, really logs an empty string, and logging `undefined` prints `undefined`. 

Perhaps this should be changed upstream in the worker, either actually send `['undefined']` or skip sending the message?

## Validation steps

See: `test/integration/web_and_worker_test.exs:290`, and run it with `mix test.watch test/integration/web_and_worker_test.exs:290`.

This test checks the different output casting for log lines.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/invocation/log_line.ex
+++ b/lib/lightning/invocation/log_line.ex
@@ -50,7 +50,7 @@ defmodule Lightning.Invocation.LogLine do
   def new(%Run{} = run, attrs \\ %{}, scrubber) do
     %__MODULE__{id: Ecto.UUID.generate()}
     |> cast(attrs, [:message, :timestamp, :step_id, :run_id, :level, :source],
-      empty_values: [[], nil]
+      empty_values: [nil]
     )
     |> put_assoc(:run, run)
     |> validate(scrubber)
--- a/test/integration/web_and_worker_test.exs
+++ b/test/integration/web_and_worker_test.exs
@@ -287,7 +287,7 @@ defmodule Lightning.WebAndWorkerTest do
       # console.log('');        => new line
       # console.log(null);      => null
       # console.log(undefined); => new line
-      assert log =~ "Starting operation 1\n\nnull\n{"
+      assert log =~ "Starting operation 1\n\n\nnull\n{"
 
       assert log =~ ~S[{"password":"***","username":"quux"}]
       assert log =~ ~S"Check state.errors"


josephjclark:
@stuartc something I do a lot in the CLI, to get better log output, is `console.log()`. Just to get the clean line break and make logs more readable in stdout. But this stuff isn't useful at all for Lightning, where ideally we'd have a nice logger UI which lets us filter stuff and there's natural padding of lines so we don't need artificial breaks.

This is different of course to a user (or maybe an adaptor) doing `console.log(x)` where x is undefined.

I could look into a few things upstream:
- In the job logger - the thing that logs user's console.log statements -  I could (probably) detect an `undefined` value and pass an `'undefined'` string instead. Now Lightning will receive `message: ['undefined']`. I don't know if that will confuse anything. I suppose I could also do `message: [undefined]` if an argument was passed, which is probably more useful
- The logger has  `console.break()` function, which I think is mostly what the CLI uses for this. I could have a rule like: when logging json output, ignore any `console.break` statements - because we assume that the JSON will be consumed by something richer that a terminal downstream.
- The worker could refuse to send any logs with `message: []` on the assumption that it's a waste of everyone's time

If I do that, then I think I can promise:
- `message: [undefined]` is useful and even important to log
- You will not be sent meaningless empty lines which are just stdout padding

josephjclark:
bottom line: the contract on `log` is that the worker _always_ sends a message with readable content. No empty lines. Stringified undefineds. Only send it if it has value to lightning.

I should extend the mock to throw for empty lines too, to help us catch this stuff

josephjclark:
This is done in https://github.com/OpenFn/kit/pull/906 and  should be fixed in worker 1.13.0

PR #2968, opened 2025-02-24
Disable jobs with non retriable steps

### LINKED ISSUE #2925: When retrying runs from step, don't show steps without input dataclip
Besides data storage rules, there are situations where the worker does not return output data clips because of the size of the data size limit. This constraint does not break the workflow execution but it means that users might not be able to retry subsequent (steps) from the history page as there is no input data to try with. 

When the work order is expanded, the replay button is replaced by the 🔄 icon but when I tried to use the `retry from step` button, all the steps were listed even though there was no input dataclip. 

The request here is to filter out steps that do not have input data clip, so users are not misled. 

https://drive.google.com/file/d/1bPsUhbt9-LNN8VBt_auYaCJwvXMp4Lzq/view?usp=sharing

jyeshe:
## Description

This PR disable jobs when retrying a work order from a specific step that doesn't have a non-wiped input dataclip. So either if the dataclip was wiped or if by any chance the step doesn't have an input dataclip id, the step is non retriable and the respective job gets disabled. 

Closes #2925 

## Validation steps

1. Run a simple workflow with 2 jobs where the first jobs changes the state;
2. Run it again with a different dataclip;
3. Call `Ecto.Changeset.change(input_dataclip, %{wiped_at: DateTime.utc_now() |> DateTime.truncate(:second)}) |> Repo.update!()`for the input_dataclip of the latest step of the second workorder.
4. On the History page, for the first work order, the Retry from Step shows all jobs enabled and selectable while for the second work order the last job is disabled.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/work_orders.ex
+++ b/lib/lightning/work_orders.ex
@@ -513,7 +513,7 @@ defmodule Lightning.WorkOrders do
           r.inserted_at == lr.last_inserted_at,
       where: s.job_id in ^job_ids,
       order_by: [asc: wo.inserted_at],
-      select: %{rs | step: s}
+      select: %{rs | step: s, run: r}
     )
     |> then(fn query ->
       if Keyword.get(opts, :non_wiped_dataclip?) do
--- a/lib/lightning/accounts/user_notifier.ex
+++ b/lib/lightning/accounts/user_notifier.ex
@@ -10,6 +10,7 @@ defmodule Lightning.Accounts.UserNotifier do
     max_attempts: 1
 
   import Swoosh.Email
+  import LightningWeb.Utils, only: [pluralize_with_s: 2]
 
   alias Lightning.Accounts.User
   alias Lightning.Helpers
@@ -412,9 +413,6 @@ defmodule Lightning.Accounts.UserNotifier do
   defp alternate_storage_message(false = _alternate_storage_enabled) do
     "THIS LIGHTNING INSTANCE DOES NOT HAVE ALTERNATE STORAGE ENABLED, SO THESE FAILED MESSAGES CANNOT BE RECOVERED WITHOUT MAKING THEM AVAILABLE ON THE KAFKA CLUSTER AGAIN."
   end
-
-  defp pluralize_with_s(1, string), do: string
-  defp pluralize_with_s(_integer, string), do: "#{string}s"
 
   @doc """
   Deliver instructions to confirm a credential transfer.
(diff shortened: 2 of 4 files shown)


lib/lightning/work_orders.ex
```diff
@@ -491,6 +468,48 @@ defmodule Lightning.WorkOrders do
...
+        join: r in assoc(rs, :run),
+        join: s in assoc(rs, :step),
+        join: wo in assoc(r, :work_order),
+        join: lr in subquery(last_runs_query),
+        on:
+          r.work_order_id == lr.work_order_id and
+            r.inserted_at == lr.last_inserted_at,
+        where: s.job_id in ^job_ids,
+        order_by: [asc: wo.inserted_at],
+        select: %{rs | step: s}
```
stuartc, line 500: Instead of merging the RunStep struct, I think this should be a `preload: [step: s]` right?

stuartc (commented):
Looks good to me, I did have one idea - is that instead of splitting the concepts of wiped and retryable, we could put a virtual field on `Step`, called `retryable?` and we use a db query to fill that in. That way we bring the concepts together and in the we can avoid bringing the criteria for retryable into the view.

lib/lightning/work_orders.ex
jyeshe: Preloads are tricky, there are times when they produce an additional/detached query like in this case.

jyeshe:
> Looks good to me, I did have one idea - is that instead of splitting the concepts of wiped and retryable, we could put a virtual field on `Step`, called `retryable?` and we use a db query to fill that in. That way we bring the concepts together and in the we can avoid bringing the criteria for retryable into the view.

I think the intention is good but I try to avoid the use booleans without initialisation. Using the virtual field, on the rest of the application this retryable? would need to be set or would have an unknown or inconsistent state. Let me see moving the logic to the context module.

lib/lightning/work_orders.ex
jyeshe: I have confirmed with debug log level that it's really sending one more query down the wire.

stuartc (approved):

midigofrank (changes requested):
Hey @jyeshe , this looks great!
I have tested and noticed that the step only gets disabled when **ALL** the selected work orders have that step with a wiped `input_dataclip`. Is it not supposed be when **ONLY SOME**?

jyeshe:
> Hey @jyeshe , this looks great! I have tested and noticed that the step only gets disabled when **ALL** the selected work orders have that step with a wiped `input_dataclip`. Is it not supposed be when **ONLY SOME**?

Hey @midigofrank it works for disabling mixed states. Please notice you need a change on the `state` on the previous job so that the `input_dataclip` is different:
image
image

image

midigofrank (approved):
Perfect!!! Nicely done

PR #2970, opened 2025-02-25
One Token, Multiple Credentials

### LINKED ISSUE #2908: One Token, Multiple Credentials
Below is a **technical proposal** for how to handle multiple OAuth credentials that share **the same scope set**, as well as how to handle **reauthorization** in a way that ensures old tokens aren’t disabled (or that everything updates gracefully when they are).

**PS: We use Google here as the example OAuth provider. However, this approach is generic and can be applied to other OAuth providers**

## The Core Problem(s)

1. **Multiple Credentials, Same Scopes**  
   - A user tries to create **two** credentials with the **same** scopes (e.g., two “Google Sheets” credentials).  
   - Google does not return a `refresh_token` the second time if the user+client+scopes match a previously granted token. Instead, it may only return a new `access_token` (which expires within an hour).  
   - We either end up with a second credential that **lacks** a `refresh_token` (cannot refresh once `access_token` expires), and usually Google invalidates the previous one.

2. **Reauthorization Overwrites Old Tokens**  
   - When a user “reauthorizes” an existing credential (with the same scopes), Google issues a **new** `refresh_token` + `access_token`, invalidating **all previous tokens** for that same user+client+scopes.  
   - We want all “credentials” in our system that share those scopes to remain in sync so they don’t break.

**Watch these two videos for a quick demo of the problems:**
- [We do support multiple credentials with same client and user but different scopes](https://www.loom.com/share/3d35a3a295d047b68912933bec90f359?sid=9849063a-83a1-49ee-8d96-45bcab490c24).
- [We do not support multiple credentials with same client, user, and scopes](https://www.loom.com/share/bf8f9a87e7aa4b0895250ac666df1f1a?sid=d2d8e7bb-8dea-483f-9b32-82674222e918).

## Goal

- **Allow** the user to create multiple “credentials” in the UI for the same scope set (e.g., maybe they want them named differently, or they belong to different projects).  
- **Internally** store or manage only **one** refresh token per combination of `[user, oauth_client, exact scope set]`.  
- **When reauthorizing**, we replace the old refresh token with the new one for that scope set, and update (or unify) any credentials that share it.

## High-Level Solution

1. **Enforce “One Token Per User + Client + Scope Set”** 
In the database or in the business logic, if a user attempts to create a second credential for the **exact same scope set** (e.g., `[spreadsheets, openid, userinfo.email, userinfo.profile]`), we don’t store a whole new refresh token object. We reference the same underlying token.  
2. **“Token Storage” Decoupled from “Credential Record”**  
   - Instead of each `credentials` record physically storing its own refresh token, we store tokens in a separate table (call it `CredentialToken` or `CredentialBody` or something similar) keyed by `[user_id, oauth_client_id, scope_set]`.  
   - Each `credentials` record can reference that single `CredentialBody` row.  
3. **Reauthorization Updates**  
   - If a user reauthorizes any credential that references that `CredentialBody`, Google issues a brand-new refresh token (invalidating the old one). We then update the `CredentialBody` row with the new refresh token and remove the old one. All `credentials` referencing that token are now effectively updated.  

### Result

- Users can have multiple credentials in OpenFn with the same scopes. **They all point to the same “token record.”**  
- If they reauthorize, we update that single token record. All referencing credentials keep working.

### 1. Data Model Changes

Currently, we have:

```elixir
schema "credentials" do
  field :body, Lightning.Encrypted.Map, redact: true
  ...
end
```

`body` typically stores `access_token`, `refresh_token`, `id_token`, `scope`, etc.

**Proposal**: Introduce a new table (e.g. `credential_bodies`).

```elixir
schema "credential_bodies" do
  belongs_to :oauth_client, OauthClient
  field :body, Lightning.Encrypted.Map, redact: true
  ...
  timestamps()
end
```

Then, in `credentials`, replace or augment:

```elixir
schema "credentials" do
  ...
  belongs_to :credential_bodies, CredentialBody
end
```

Here is an ER representation of the data model:

Image

### 2. Logic for Creating a New Google Credential

When a user tries to create a new “Google Sheets” credential:

1. **Check** if there is an existing `CredentialBody` row for `[user_id, oauth_client_id, exact scope set]`.  
   - If it **exists**, simply create a new `Credential` record that references that existing `credential_body_id`.  
   - If it **doesn’t exist**, proceed with an OAuth flow with `prompt=consent&access_type=offline`, get the tokens from Google, and create a new row in `credential_bodies`.

#### 2.1 Handling the “Missing Refresh Token” from Google

If the user had **already** granted these exact scopes for this user+client combo, Google might return no new refresh token. Typically, that means we actually already have a valid refresh token in `credential_bodies`.

- So if we detect that Google only gave us an `access_token` (no `refresh_token`), we know the user has previously authorized these scopes. In this scenario, we can:
  1. **Look up** any existing `CredentialBody` row for these scopes.
  2. **If** that row is found, we reference it.  
  3. If none is found (very edge case) or if the old token was invalid, we might need to force reauthorization with `prompt=consent` again or instruct the user to “remove access” in their Google Account so that a new refresh token can be obtained.

### 3. Logic for Reauthorization (Updating Tokens)

When a user “reauthorizes”:

1. **We know** the credential references some `CredentialBody`.  
2. On the callback from Google, we get a **new** `refresh_token` + `access_token`.  
3. Because Google invalidates the old refresh token for these scopes, we simply **update** the `credential_bodies` row with the new `refresh_token`, `access_token`, `expires_at`, etc.  
4. **All** credentials referencing that `credential_bodies` row continue to work seamlessly.

### 4. Migration / Transitional Steps

1. Create the `credential_bodies` table.  
2. Run a migration that looks for each `credential` with `schema = "oauth"` + `oauth_client_id`, extracts the `refresh_token`, `scopes`, etc., and **inserts or merges** it into `credential_bodies`.  
3. Update the `credentials.credential_body_id` to reference that record.  
4. Remove the credential_body field from `credentials`.

### 5. “One Refresh Token for Many Access Tokens”

The advantage of this approach is that multiple credentials referencing the same scopes (Sheets) can share one refresh token. Whenever you do the refresh flow, you get a new `access_token`; that’s ephemeral. The shared `refresh_token` is used by the adapter code to request new access tokens as needed.

If Google **does** give us a brand-new refresh token (e.g., upon reauthorization), we just update the shared record. This is exactly what we want for “multiple credentials, same scope set, all remain functional.”

### Scenario: User Creates 2 Google Sheets Credentials

1. **User clicks** “Add Credential → Google Sheets 1”  
   - We do the OAuth flow with scopes `[spreadsheets, userinfo.profile, userinfo.email, openid, ...]`  
   - We get `refresh_token=XYZ1`.  
   - We create `CredentialBody` row : 
     ```
     %CredentialBody{
       oauth_client_id: <google_client>, 
       scopes: ["..."], 
       refresh_token: "XYZ1",
       access_token: "ABC1",
       expires_at: ...
     }
     ```  
   - We create `Credential` 1 referencing `CredentialBody`.  

2. **User clicks** “Add Credential → Google Sheets 2”  
   - We see the **exact same** scope set.  
   - We do the OAuth flow. Google decides not to give a new refresh token (only an `access_token`). 
   - We detect that we already have a `CredentialBody` with that `[user, oauth_client, scopes]`.  
   - We create `Credential` 2 referencing the **same** `CredentialBody`.  

Now both credentials share `refresh_token=XYZ1`. The user sees “Google Sheets 1” and “Google Sheets 2” in the UI, but they actually share the same underlying token record.

### Scenario: User Reauthorizes “Google Sheets 1”

1. The user picks “Reauthorize” on credential 1.  
2. OAuth callback returns a new `refresh_token=XYZ2`, invalidating XYZ1.  
3. We update `CredentialBody` 1’s `refresh_token` to “XYZ2”.  
4. Credential 2 also references `CredentialBody` 1, so it now effectively uses `XYZ2`. Both remain valid.

## Additional Considerations

1. **Partial Overlap vs. Exact Match**  
   - This solution is easiest if we treat **exact** scope sets. If partial overlap is allowed (e.g., `Sheets + userinfo` vs. `Sheets + userinfo + Drive`), that’s more complicated. Usually, if the user is requesting the exact same scope set, they share a token; otherwise, it’s a different record.  
2. **User Uniqueness**  
   - The logic to match `[user_id, oauth_client_id, scope_set]` is critical. If you have multiple users, they each get their own `CredentialBody`.  
3. **UI Warnings**  
   - If the user tries to create a new credential with the exact same scopes, we might want to warn them: “You already have a Sheets token. This new credential will share the same token. Reauthorizing one will update them all.”  
   - Or we can do it silently, if that’s acceptable.  
4. **Security**  
   - Storing tokens in a separate table might have the same encryption approach we use for `credentials.body` (using `Lightning.Encrypted.Map` or a similar approach).  
5. **“One Refresh Token for Many Access Tokens”** is exactly how Google typically wants us to handle it. Each `refresh_token` can yield new `access_tokens` on-demand. If we want to run parallel processes, each can just use the same refresh token or share the ephemeral access token if still valid.  

## Conclusion

By **decoupling** the actual token storage into a separate table keyed by `[user_id, oauth_client_id, scope set]`, we can:

- **Prevent** the creation of a second, incomplete credential (missing refresh_token) for the same scope set.  
- **Easily** manage reauthorization events, any new refresh token overwrites the old one in that single record.  
- **Allow** the UI to show multiple credentials for the same scope set with different names or project affiliations, while in reality they share one refresh token.

### LINKED ISSUE #2945: GoogleSheet Oauth Credential Fails on first attempt (No Refresh token)
During a demo call yesterday, a new user was trying to create a GSheets OAuth credential and it failed with a No Refresh token error and the user had to click "retry here" to authenticate again.

elias-ba:
### Overview

This PR refactors how OAuth tokens are stored and referenced, so that multiple “credentials” with the same scope set (for the same OAuth provider) can share a single refresh token. This prevents errors caused when OAuth providers (e.g. Google) invalidate or omit a refresh token if it has already been issued for the same user+client+scopes combination.

**Why?**  
Previously, each credential row stored its own OAuth tokens in `credentials.body`. This was problematic when:
- A provider _didn’t_ return a refresh token for a new credential if it was equivalent to an existing token, causing incomplete credentials (missing refresh token).
- Reauthorizing one credential could invalidate refresh tokens used by other credentials.

**What’s New?**  
We introduce a new `oauth_tokens` table (previously referred to as `credential_bodies`) where the actual OAuth token data (including `refresh_token`, `access_token`, `expires_at`, etc.) now lives. Each row in `oauth_tokens` is keyed by `[user_id, oauth_client_id, scope set]`. The `credentials` table references this token row via `oauth_token_id`.

### Problems Addressed

1. **Multiple Credentials, Same Scopes**  
   - When a user creates two credentials with identical scopes, Google often returns no new `refresh_token` for the second.  
   - The old system tried to save an incomplete credential lacking a `refresh_token`—which blocked the user.

2. **Reauthorization Overwrites Old Tokens**  
   - When reauthorizing one credential, OAuth providers will issue a brand-new refresh token and invalidate the old one.  
   - Under the old model, that would break other credentials sharing the same scope set.

### Goals & Approach

1. **One Token Per `[User, OauthClient, Exact Scope Set]`**  
   - If a user tries to create multiple credentials for the same scope set, they share one token record in the DB.  
2. **Decouple “Token Storage” from “Credential Record”**  
   - Migrate token data into a new table: `oauth_tokens` (fields: `user_id`, `oauth_client_id`, `scopes`, `body`).  
   - `credentials` references `oauth_tokens` via `oauth_token_id`.  
3. **Reauthorization Updates**  
   - When reauthorizing a credential, any new refresh token overwrites the old one in `oauth_tokens`.  
   - All referencing credentials stay valid automatically.  
4. **Migration**  
   - A migration script finds OAuth credentials, extracts their token data, and inserts/merges it into `oauth_tokens`.  
   - The `credentials.body` is then cleared for OAuth creds, and we set `oauth_token_id`.

**Result:** Multiple credentials in the UI can share a single refresh token behind the scenes. If the refresh token is reissued, we update it once and all referencing credentials keep working.

### Example Flow

1. **User Creates Credential “Google Sheets 1”**  
   - OAuth flow returns `refresh_token=ABC123`.  
   - We create one `oauth_tokens` record with that refresh token, and one `credentials` record pointing to it.

2. **User Creates Credential “Google Sheets 2”**  
   - Same scopes, so Google might _not_ return a new refresh token.  
   - We detect that the user+client+scopes already exists in `oauth_tokens`.  
   - We simply create a second `credentials` record pointing to the existing token row. Both credentials share the token.

3. **Reauthorization**  
   - A new refresh token is issued by Google, invalidating the old one.  
   - We overwrite the `oauth_tokens.body` with the new refresh token.  
   - All credentials that point to this `oauth_tokens` record automatically reference the new token.

### Technical Implementation

1. **`oauth_tokens` Table**  
   - `id`, `user_id`, `oauth_client_id`, `scopes`, `body (encrypted)`, timestamps.  
   - Unique index on `[user_id, oauth_client_id, scopes]`.

2. **`credentials` Table**  
   - New column `oauth_token_id` → references `oauth_tokens`.  
   - Existing `body` column is set to hold other credential fields not directly related to OAuth for example the `apiVersion` and `sandbox` post-migration.
   - Removal of the old `oauth_client_id` column in `credentials` (it now lives in `oauth_tokens`).

One Token - Multiple Credentials

3. **Ecto Changes**  
   - `OauthToken` schema handles creation, updating, and retrieving of tokens.  
   - `Credential` references `OauthToken` for any OAuth credentials.  
   - `Credentials.validate_oauth_token_data/5` ensures we either have a refresh token or detect an existing token for a given scope set.

4. **Migration Script**  
   - Iterates over existing OAuth credentials.  
   - Extracts scope data from `credentials.body`, upserts into `oauth_tokens`.  
   - Updates `credentials` to point to the newly inserted token record.  
   - Drops `oauth_client_id` column from `credentials` once migration is verified.

### Notes / Future Considerations

- **Partial Scope Overlap**: This PR only handles exact scope matches. Overlapping scope sets remain separate tokens.  
- **UI Warnings**: We might add UI messages (“Creating another token with the same scopes will share a refresh token”).

### Conclusion

By splitting out token storage into `oauth_tokens` keyed by `[user_id, oauth_client_id, exact scopes]`, we fix edge cases around missing refresh tokens, and keep multiple credentials in sync upon reauthorization. This streamlines the OAuth process and prevents accidental breakage when providers reuse or invalidate tokens.

Closes #2908
Might close #2945 too

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning/credentials/credential.ex
```diff
@@ -50,33 +51,45 @@ defmodule Lightning.Credentials.Credential do
...
-    |> validate_oauth()
+    |> maybe_validate_oauth_fields(attrs)
   end
 
-  defp validate_oauth(changeset) do
-    if get_field(changeset, :schema) == "oauth" do
-      body = get_field(changeset, :body) || %{}
+  defp maybe_validate_oauth_fields(changeset, attrs) do
+    is_oauth = get_field(changeset, :schema) == "oauth"
+    oauth_token_id = get_field(changeset, :oauth_token_id)
```
jyeshe, line 62: This is used only the condition to see if it's not set. I believe thanks to the pairing I can understand better this as where part of the magic happens. For future maintainance, maybe we should give here a domain variable name to the condition. Something related to the purpose like `reusing_token? = get_field(changeset, :oauth_token_id)`?

lib/lightning/credentials/oauth_token.ex
```diff
@@ -0,0 +1,163 @@
...
+          oauth_client: OauthClient.t() | nil,
+          user_id: Ecto.UUID.t() | nil,
+          user: User.t() | nil,
+          credentials: [Credential.t()] | nil,
+          inserted_at: DateTime.t() | nil,
+          updated_at: DateTime.t() | nil
+        }
+
+  schema "oauth_tokens" do
+    field :body, Lightning.Encrypted.Map, redact: true
```
jyeshe, line 29: 💯

lib/lightning/credentials/oauth_token.ex
```diff
@@ -0,0 +1,163 @@
...
+
+  @doc """
+  Creates a changeset for an OAuth token.
+  """
+  def changeset(oauth_token, attrs) do
+    oauth_token
+    |> cast(attrs, [:body, :scopes, :oauth_client_id, :user_id])
+    |> validate_required([:body, :scopes, :oauth_client_id, :user_id])
+    |> assoc_constraint(:oauth_client)
+    |> assoc_constraint(:user)
```
jyeshe, line 47: 🙏

lib/lightning/credentials/oauth_token.ex
```diff
@@ -0,0 +1,163 @@
...
+    case find_by_scopes(user_id, oauth_client_id, scopes) do
+      nil ->
+        %__MODULE__{}
+        |> changeset(%{
+          user_id: user_id,
+          oauth_client_id: oauth_client_id,
+          scopes: scopes,
+          body: tokens
+        })
+        |> Lightning.Repo.insert()
```
jyeshe, line 81: You probably wanted to move it to a context module. For this and the operations bellow to avoid the insert and querying in the Schema module.

jyeshe (changes requested):
This is awesome work @elias-ba . There are few details left only for future maintainability and it's a beautiful goal!

lib/lightning/credentials/credential.ex
elias-ba: Thanks, that is done

lib/lightning/credentials/oauth_token.ex
elias-ba: This is done, nice catch, thanks

midigofrank (changes requested):
Hey @elias-ba , this looks really great. I've left some comments regarding the migration timestamp and the script.

priv/repo/migrations/20250224080735_create_oauth_tokens_table.exs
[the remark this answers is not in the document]
elias-ba: Done, thanks for the catch

priv/repo/migrate_oauth_tokens.exs
```diff
@@ -0,0 +1,128 @@
+# File: priv/repo/migrate_oauth_tokens.exs
+#
+# Usage: mix run priv/repo/migrate_oauth_tokens.exs
+#
+# This script migrates OAuth credential bodies to the new oauth_tokens structure,
+# then updates credentials to reference these tokens instead of storing the body directly.
+
+# Make sure the application is started
+Mix.Task.run("app.start")
```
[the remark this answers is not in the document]
elias-ba: I moved this to a script in `/priv/repo/migrate_oauth_credentials.exs`. Unfortunately, due to the credentials body encryption we can't read the scopes at the migrations level. We need to run this script. But the whole implementation of this feature is reverse compatible with `main` so no risk of downtime in running this script.

midigofrank (changes requested):
Nicely done @elias-ba.
It seems you'll need ensure that some of the applications are started in that function. I'm getting this error:
```sh
frank@mid-ego lightning % _build/dev/rel/lightning/bin/lightning eval "Lightning.Credentials.OauthMigration.run()"
[info] Starting OAuth credentials migration
** (RuntimeError) could not lookup Ecto repo Lightning.Repo because it was not started or it does not exist
    (ecto 3.12.5) lib/ecto/repo/registry.ex:22: Ecto.Repo.Registry.lookup/1
    (ecto 3.12.5) lib/ecto/repo/supervisor.ex:176: Ecto.Repo.Supervisor.tuplet/2
    (lightning 2.10.16) lib/lightning/repo.ex:2: Lightning.Repo.all/2
    (lightning 2.10.16) lib/lightning/credentials/oauth_migration.ex:28: Lightning.Credentials.OauthMigration.run/0
    nofile:1: (file)
    (stdlib 5.2.3) erl_eval.erl:750: :erl_eval.do_apply/7
```

elias-ba:
Thanks dude, I just pushed, please can you try again.

lib/lightning/credentials/oauth_client.ex
```diff
@@ -35,7 +35,7 @@ defmodule Lightning.Credentials.OauthClient do
 
     belongs_to :user, User
 
-    has_many :credentials, Credential
+    has_many :oauth_tokens, OauthToken
```
stuartc, line 38: I can't see where this relationship is being used from?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -94,7 +95,12 @@ defmodule LightningWeb.RunChannel do
          {:ok, scrubber} <- update_scrubber(scrubber, samples, basic_auth) do
       socket
       |> assign(scrubber: scrubber)
-      |> reply_with({:ok, remove_empty_values(credential.body)})
+      |> reply_with(
+        {:ok,
+         credential
+         |> assemble_body()
+         |> remove_empty_values()}
+      )
```
stuartc, line 103: I think we need to be really careful here:

1. I see we're changing the body we send to the client, that makes sense - but what about the scrubber, does the scrubber still work exactly the same way since it've been given a different structure?
2. I don't think we should manipulating the response body for the worker in this module, can we use or create a module that does rendering for these different structs?

lib/lightning/credentials.ex
```diff
@@ -979,4 +1155,459 @@ defmodule Lightning.Credentials do
...
+
+  Convenience function that calls `process_scopes/2` with a space delimiter.
+
+  ## Parameters
+    - `input`: OAuth scope string or token to extract scopes from
+
+  ## Returns
+    - List of normalized scope strings
+  """
+  def process_scopes(input) do
```
stuartc, line 1577: This function name is a bit vague imo, perhaps `normalize_scopes` or `extract_scopes`?

lib/lightning/credentials/credential.ex
```diff
@@ -50,33 +53,45 @@ defmodule Lightning.Credentials.Credential do
       message: "you have another credential with the same name"
     )
     |> assoc_constraint(:user)
-    |> assoc_constraint(:oauth_client)
     |> validate_format(:name, ~r/^[a-zA-Z0-9_\- ]*$/,
       message: "credential name has invalid format"
     )
-    |> validate_oauth()
+    |> maybe_validate_oauth_fields(attrs)
```
stuartc, line 59: I'm wondering here, could we not use `cast_assoc` for the oauth stuff, I see there are two places that call `Credentials.validate_oauth_token_data`; would it not be better to have the oauth body be validated by the OAuth token exclusively, since `Credential` won't have the body anymore - please correct me if my understanding is wrong.

stuartc (changes requested):
Sho! @elias-ba this is big set of work. Nicely done.

I've left a few comments around code organisation, I feel the `Credentials` module is getting a lot of OAuth specific functions and helpers.

There's also some questions around the changeset stuff, and also some concerns around the RunChannel.

Let me know if you want to pair and/or if I'm missing the point on some of these parts of the implementation.

stuartc (approved):

midigofrank (approved):

PR #2982, opened 2025-02-28
Add empty body as default new input on Job inspector

### LINKED ISSUE #2952: Create work orders with `{}` when input data is not manually set
In the inspector, when I'm editing my job code, I can't run my job unless I have some input.
Well, okay, fair enough, but I also find myself setting `{}` to my input all the time. I’d love it if I could just run the job and, if I haven’t specified an input, for `{}` to be used.

Possible that this touches on #2944 - a bug we've discovered recently.

Note that **_this is not_** [the big request for manual runs](https://community.openfn.org/t/simpler-ui-to-run-workflows-on-demand/713/3), but rather a short-term, high-value improvement to quality of life.

---

Here's more information from @taylordowns2000 :

A really big bug-bear for the group (in Ghana) this week was having to type `{}` in the left side panel to run a job. This was probably the biggest common blocker to the UX.

I’m quite eager to introduce the “better manual runs” stuff (with those diagrams I’ve shared already) [diagrams](https://openfn.slack.com/archives/C0394D0GBUN/p1738170510035209?thread_ts=1738170353.007589&cid=C0394D0GBUN)

But in the meantime, I was expecting that this would work:
```ex
@@ -123,6 +123,7 @@ defmodule LightningWeb.WorkflowLive.ManualWorkorder do
     <div :if={is_nil(@selected_dataclip)} class="grow">
       <.input
         type="textarea"
         value="{}" # 👈 Look here!
         field={@form[:body]}
         disabled={@disabled}
         class="h-full font-mono proportional-nums text-slate-200 bg-slate-700"
```
That one line addition sets the default value to `{}` which I think is great… if you don’t specify an input and just want to run get('https://www.google.com/') we let you type that and then hit run. End of story!

But… _it doesn’t actually work._ The run button is greyed out. Ayo thinks it may be related to the run button listening to an `onchange` event in the input panel. If I click into the input after this default value loads, backspace to remove a `}` and then type the same `{` back, everything is perfect.

jyeshe:
## Description

This PR adds an empty body `{}` for the creation of new input on the job inspector. 

Closes #2952

## Validation steps

1. Access the job inspector without any additional query params.
2. The `{}` is suggested as basic dataclip for start and to create a new work order.
3. Create a new workorder and notice the Rerun button is enabled.
4. Select another job and use an existing _non-empty_ dataclip to create a new work order.
5. After running click on Create a new input and the empty dataclip is suggested.

Notice that if you copy a link of a run to follow it and this run has a wiped input dataclip the empty dataclip is NOT suggested to honor the fact that this run does NOT have any dataclip.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -2068,7 +2068,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
           end
 
         WorkOrders.Manual.new(
-          %{dataclip_id: dataclip && dataclip.id, body: body},
+          %{"dataclip_id" => dataclip && dataclip.id, "body" => body},
           attrs
         )
       end)
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -1627,13 +1627,6 @@ defmodule LightningWeb.WorkflowLive.Edit do
         job: socket.assigns.selected_job,
         created_by: socket.assigns.current_user
       }
-      |> then(fn attrs ->
-        if params["dataclip_id"] == "" do
-          Map.put(attrs, :body, "{}")
-        else
-          attrs
-        end
-      end)
 
     changeset =
       WorkOrders.Manual.new(params, attrs)
@@ -2048,30 +2041,20 @@ defmodule LightningWeb.WorkflowLive.Edit do
       new_manual_run_form_body(
         manual_run_form,
         selected_job,
-        follow_run,
         dataclip
       )
 
     changeset =
-      %{
-        project: project,
-        workflow: workflow,
-        job: selected_job,
-        user: current_user
-      }
-      |> then(fn attrs ->
-        attrs =
-          if body == "{}" do
-            Map.put(attrs, :created_by, current_user)
-          else
-            attrs
-          end
-
-        WorkOrders.Manual.new(
-          %{"dataclip_id" => dataclip && dataclip.id, "body" => body},
-          attrs
-        )
-      end)
+      WorkOrders.Manual.new(
+        %{"dataclip_id" => dataclip && dataclip.id, "body" => body},
+        %{
+          project: project,
+          workflow: workflow,
+          job: selected_job,
+          user: current_user,
+          created_by: current_user
+        }
+      )
 
     selectable_dataclips =
       Invocation.list_dataclips_for_job(%Job{id: selected_job.id})
@@ -2094,7 +2077,6 @@ defmodule LightningWeb.WorkflowLive.Edit do
   defp new_manual_run_form_body(
          prev_manual_run_form,
          selected_job,
-         follow_run,
          selected_dataclip
        ) do
     if is_nil(selected_dataclip) do
@@ -2108,8 +2090,6 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
       if is_struct(prev_job) and prev_job.id == selected_job.id do
         Ecto.Changeset.get_change(prev_manual_run_form.source, :body)
-      else
-        if is_nil(follow_run), do: "{}"
       end
     end
   end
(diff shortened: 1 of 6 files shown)


lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -2013,73 +2022,95 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+      %{
+        project: project,
+        workflow: workflow,
+        job: selected_job,
+        user: current_user
+      }
+      |> then(fn attrs ->
+        attrs =
+          if body == "{}" do
+            Map.put(attrs, :created_by, current_user)
```
midigofrank, line 2065: Why are we adding the `created_by` only if `body == {}` ? 
I can see in the above block we are already passing in a `user: current_user`, is it any different from `created_by`?

midigofrank (changes requested):
Hey @jyeshe , great job!

I'm a little concerned about the multiple `if` statements introduced. I think we should only have this in one place, in the changeset: https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/manual.ex#L35
```elixir
    ...
    |> validate_required([:project, :job, :created_by, :workflow])
    |> then(fn changeset -> 
       if get_change(changeset, :body) |> is_nil() do
          put_change(changeset, :body, "{}")
       else
          changeset
       end
    end)
    |> remove_body_if_dataclip_present()
    ...
```

Also, is there a reason why we can't always just build the manual workorder with the `created_by: current_user`. The user who is logged in is the only person who can initiate this right? I don't understand why we conditionally pass this in.

jyeshe:
> Hey @jyeshe , great job!
> 
> I'm a little concerned about the multiple `if` statements introduced. 

Hi @midigofrank I appreciate your concern once this is a module that handles tens of possibilities. 
However notice that there are only 2 situations where the default body is handled. 
1. for dataclip change (it shall suggest the `{}` when the user selects "Create new input")
2. for the show that also calls `Manual.new` , in the same one that you've recently added the conditions to handle  previous jobs.


> Also, is there a reason why we can't always just build the manual workorder with the `created_by: current_user`. The user who is logged in is the only person who can initiate this right? I don't understand why we conditionally pass this in.

It is passed because the user is not explicitly typing an input but accepting the suggestion when creating a work order with it.

lib/lightning_web/live/workflow_live/edit.ex
jyeshe: When the user accepts the suggestion of the empty body, the user is the creator of the dataclip. This is not needed when the user types the input but without typing it is.
I would refactor the existing validation that checks about this `created_by` but once this is a huge module that handle many cases I am leaving it do it after some prioritization.

midigofrank:
@jyeshe I get that the user is not explicitly setting the input, but does it really matter now that we have a default input anyway.

jyeshe:
> @jyeshe I get that the user is not explicitly setting the input, but does it really matter now that we have a default input anyway.

It's required field. Here is the validation I was referring to:
https://github.com/OpenFn/lightning/blob/default-wo-dataclip/lib%2Flightning_web%2Flive%2Fworkflow_live%2Fedit.ex#L2172-L2173

But I agree with you that we should find a way to make it more maintainable. Not only in regards to the creation but a full refactor breaking the View into components and helpers.

jyeshe:
> Aaah, I see. Can't we just change that line to `Enum.empty?(manual_run_form.source.errors)` ? 😅 . Because from now on, we always have the `created_by` field

That's what I wanted but do you remember the request (: ?

jyeshe:
@stuartc as you're resuming this, please notice my last commit shows that with an empty body on the schema it is not shown on the input and the create button is disabled:
image

stuartc:
Closing because this will be superseded/reimplemented in the React editor rewrite.

https://github.com/OpenFn/lightning/issues/2952#issuecomment-2739599848

PR #3043, opened 2025-03-20
Cont. Workflow Editor migration

doc-han:
## Description

This is a continuation of #2988
1. Moved DataclipViewer via the new way of integrating React from #2988

## Validation steps

1. Check the DataClipViewer in the Editor.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


doc-han:
Incomplete cont. at https://github.com/OpenFn/lightning/pull/3063

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react"
```
josephjclark, line 1: Can we find better names for `JobEditorMain` and `JobEditor`? It's quite untidy

I presume we have the distinction becuse `JobEditorMain` is aware of the phoenix infrastructure, but `JobEditor` is not? So the `Main` file is just like a mount or wrapper for the app

The architecture is great but the name is off :)

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+  source: string,
+  disabled: boolean,
+  disabled_message: string,
+}
+
+export const JobEditorMain: WithActionProps<JobEditorComponentProps> = (props) => {
+  const [metadata, setMetadata] = React.useState<false | object>(false);
+  const [source, setSource] = React.useState('');
+
+  // use workflowstore here
```
josephjclark, line 22: We can tidy this  comment up

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { change, getById } = useWorkflowStore();
+
+  // debounce editor content update
+  const _debouncedPushChange = pDebounce((content: string) => {
+    change({ jobs: [{ id: props.job_id, body: content }] })
+  }, EDITOR_DEBOUNCE_MS);
+
+  // init hook - getting metadata
+  React.useEffect(() => {
+    props.handleEvent('metadata_ready', (payload) => {
```
josephjclark, line 32: Maybe out of scope but I'd love to see a central definition of all the events and their expected payloads somewhere. Can we fix it or raise an issue to do it soon?

assets/js/job-editor/mount.tsx
```diff
@@ -1,230 +0,0 @@
...
-        this.changeEvent = changeEvent;
-      } else {
-        console.warn('Warning: No changeEvent set. Content will not sync.');
-      }
-      this.setupObserver();
-      this.render();
-
-      let end = instrumentEnd(event);
-
-      this.pushEventTo(this.el, 'job_editor_metrics_report', {
```
josephjclark, line 94: Have we lost this event? Do we need it still?

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+      setMetadata(payload as object);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const sortedMetadata = sortMetadata(payload);
+      setMetadata(sortedMetadata as object);
+    })
+    props.pushEventTo('request_metadata', {});
+  }, [props])
+
+  React.useEffect(() => {
+    const foundJob = getById<Lightning.Job>(props.job_id);
```
josephjclark, line 42: Is this behaviour the same as production, or have we changed? I guess props.source is the placeholder text?

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const JobEditorMain: WithActionProps<JobEditorComponentProps> = (props) => {
+  const [metadata, setMetadata] = React.useState<false | object>(false);
+  const [source, setSource] = React.useState('');
+
+  // use workflowstore here
+  const { change, getById } = useWorkflowStore();
+
+  // debounce editor content update
+  const _debouncedPushChange = pDebounce((content: string) => {
```
josephjclark, line 26: Why is this underscored?

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { sortMetadata } from "../metadata-loader/metadata";
+import { useWorkflowStore } from "../workflow-editor/store";
+import pDebounce from "p-debounce";
+import { EDITOR_DEBOUNCE_MS } from "../common";
+import type { Lightning } from "#/workflow-diagram/types";
+
+interface JobEditorComponentProps {
+  job_id: string,
+  adaptor: string,
+  source: string,
```
josephjclark, line 13: This is like `initialSource`, right? This is the source that will be shown until the store is connected. Do you think we should rename the prop or just leave it for now?

assets/js/react/hooks/react-component.tsx
```diff
@@ -148,10 +151,10 @@ export const ReactComponent = {
   },
 
   _onBoundary(element) {
-    this.__view.liveSocket.requestDOMUpdate(() => {
+    this.__view().liveSocket.requestDOMUpdate(() => {
       if (element == null) return;
 
-      this.__view.execNewMounted();
+      this.__view().execNewMounted();
```
josephjclark, line 157: Ah - presumably this was the change that you were stuck on before?

lib/lightning_web/live/job_live/job_builder_components.ex
```diff
@@ -9,53 +11,20 @@ defmodule LightningWeb.JobLive.JobBuilderComponents do
...
-    <div
-      data-job-id={@job_id}
-      data-adaptor={@adaptor}
-      data-source={@source}
-      data-disabled={@disabled}
-      data-disabled-message={@disabled_message}
-      data-change-event={@change_event}
-      phx-hook="JobEditor"
-      phx-update="ignore"
+    <.JobEditorMain
```
josephjclark, line 20: Yeah, here's where the component name is particularly tricky. hm.

assets/js/job-editor/JobEditorMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  // debounce editor content update
+  const _debouncedPushChange = pDebounce((content: string) => {
+    change({ jobs: [{ id: props.job_id, body: content }] })
+  }, EDITOR_DEBOUNCE_MS);
+
+  // init hook - getting metadata
+  React.useEffect(() => {
+    props.handleEvent('metadata_ready', (payload) => {
+      setMetadata(payload as object);
```
josephjclark, line 33: Why are we calling setMetadata twice?

josephjclark (commented):
Love that removes more code than it adds :raised_hands: 

Left some comments - mostly questions but please address :)

assets/js/job-editor/mount.tsx
josephjclark: Just seen it in the other PR, so I think we're good once that's merged in

assets/js/job-editor/JobEditorMain.tsx
doc-han: Yup I agree. any suggestions? 
`JobEditor` & `JobEditorComponent` ?

assets/js/job-editor/mount.tsx
doc-han: Yeah. we wouldn't lose any event. I just ignored it here because I had harder problems to deal with 😂

assets/js/job-editor/JobEditorMain.tsx
doc-han: zero reason.

assets/js/job-editor/JobEditorMain.tsx
doc-han: Exactly. renaming the prop will also require renaming the attr name in phoenix. I'm okay with either tho

assets/js/react/hooks/react-component.tsx
doc-han: Yeah, for some unknown reason. The internal implementation of a phoenix live's view got changed from an object to a function that then returns the object. Seems we were on a previous version during Xandor's implementation & after the version update it started erroring. 
@stuartc sent a git diff which fixed it

lib/lightning_web/live/job_live/job_builder_components.ex
doc-han: This name has to match what the mounting React component is called. Hence if we change the mounting react component name to `JobEditor` then we can simple change this too to that.

assets/js/job-editor/JobEditorMain.tsx
doc-han: Probably an oversight

assets/js/job-editor/JobEditorMain.tsx
doc-han: After the store is connected. JobEditor gets it's source from the store. Hence this is for when `job_id` changes. It searches for the source of a job in the store.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -412,6 +401,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
           centered
         />
         <.form
+          :if={@selection_mode !== "expand"}
```
stuartc, line 404: @doc-han what was the reasoning for this? We've got some failing tests (_happy to work around this change_), but was curious why we're not rendering the form conditionally here.

doc-han: The thought is to not render anything for the WorkflowEditor when on the JobEditor view. 
Hence, if you're in the JobEditor and then decide to refresh your page. nothing from WorkflowEditor gets rendered.

stuartc: Right so it's an optimisation, nothing more? The fact that it was rendering wasn't getting in your way right?

josephjclark: I'm a big fan of this change! And it's only possible because the store has been properly decoupled. I've not tested yet and I can't speak to the condition itself, but I hope we can keep it!

doc-han: > Right so it's an optimisation, nothing more? The fact that it was rendering wasn't getting in your way right?

Yeah. It's just optimization.

config/config.exs
```diff
@@ -114,6 +114,10 @@ config :esbuild,
          js/app.js
          js/storybook.js
          js/editor/Editor.tsx
+         js/react/components/DataclipViewer.tsx
```
josephjclark, line 117: Ouch, so this is needed to allow the `jsx()` macro thing to work in the production build?

We don't need to do it now, but I think we should do a deep dive on this and find a better solution. Because this is just an invitation for nasty bugs to creep in

stuartc: Would love to see a demo, but I reckon this is the lessor of two evils because _if_ you don't up date this list, it won't work in dev or prod.

While we shouldn't _have_ to have everything be it's own entrypoint, we do need to be super mindful of what module imports what - we've had more than once the typescript compiler make it's way into modules that change frequently; resulting in massive assets for users.

assets/js/workflow-diagram/usePlaceholders.ts
```diff
@@ -51,17 +50,15 @@ export const create = (parentNode: Flow.Node) => {
 
 export default (
   el: HTMLElement | null | undefined,
-  store: StoreApi<WorkflowState>,
   requestSelectionChange: (id: string | null) => void // TODO more like changeSelection
 ) => {
+  const { add: addTo } = useWorkflowStore();
```
josephjclark, line 55: what about:
```
const store = useWorkflowStore()
store.add(toWorkflow(placeholders))
```

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -44,10 +41,12 @@ type ChartCache = {
 
 const LAYOUT_DURATION = 300;
 
-export default (props: WorkflowDiagramProps) => {
-  const { selection, onSelectionChange, store, el } = props;
+export default function WorkflowDiagram(props: WorkflowDiagramProps) {
+  const { jobs, triggers, edges, disabled } = useWorkflowStore();
+  const { selection, onSelectionChange, containerEl: el } = props;
```
josephjclark, line 46: What if `usePlaceholder` attaches its listeners to `document.body`? That way we should be able to remove this element reference entirely? Which feels way cleaner

The downside is you can't have two editors in one page. But I think the way this is set up right now you can't really do that anyway?

You could namespace the event with an id if you needed two stores - that would work

assets/js/workflow-diagram/usePlaceholders.ts
doc-han: Yeah makes sense. I think the right pattern is to even also use selectors.

josephjclark:
QA note: make sure unsaved changes which error are handled well. Ie, make a change, kill the server, and see if the change gets sent after reconnect. If not raise issues to fix later.

josephjclark:
QA note: ctrl+s is broken. @stuartc to fix

josephjclark:
#QA 

Issues:

* Save button seems to not work. Load the page, type an edit, click the save button, the red notification doesn't go away and when I refresh, changes are lost. I don't see any event in the log when I click save. Save and run is fine?
* If I load the Editor, then kill my sever, the save button greys out and the red offline icon appears. When I restart the server the offline icon correctly vanishes, but the save button does not come back. When I do ctrl + S to save, I get a comforting "workflow saved succesfully" notification (why "successfully"? Surely it just saved? Are we cracking open the Bollinger every time we manage to successfully save a workflow?), but on refresh my changes are lost
* If I have unsaved changes and hit refresh, I get no warning
* Without Placeholder support the Editors load is a little janky. It really pops in.  I think this is better on main?


Good stuff:
- undo still seems to work nicely
- editing seems stable
- loading the editor directly (without the workflow diagram working behind) is fantastic, so slick
- Copy and paste seem fine
- Inserting from the docs widget seems fine

josephjclark:
I've just rolled back to main and some notes:
- The editor does still pop in :thinking: Maybe it's a little better than the branch? hard to say
- All those save/offline issues are consistent with main. Like If I take the the server offline, when reconnecting the save button is inconsistent and subsequent saves are lost

Note that if I hit `ctrl + s` when the server is offline,I **don't** get a pop-up telling me that the save failed. That's problematic.

josephjclark:
I think this QA reveals some general problems with the editor and saving in offline mode, but the only thing I think needs looking at on this branch is clicking the save button

stuartc:
Thanks dude, you’re talking about the animation as the editor arrives right? If so I know where that regression from. I can fix it.As for the save error (or lack thereof) I can try something to see if I can get an error reply. Will do that too.On 11 Apr 2025, at 13:47, Joe Clark ***@***.***> wrote:﻿
  I think this QA reveals some general problems with the editor and saving in offline mode, but the only thing I think needs looking at on this branch is clicking the save button—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



josephjclark left a comment (OpenFn/lightning#3043)
I think this QA reveals some general problems with the editor and saving in offline mode, but the only thing I think needs looking at on this branch is clicking the save button

josephjclark:
QA note: make sure the global store doesn't get confused if switching between projects. Like when the inspector is unmounted, the component doesn't get unloaded and the store is still in memory. So let's just verify that switching projects doesn't invalidate the store

elias-ba (approved):
@stuartc I am happy I failed my mission of breaking the key handlers, they work like a charm, thanks 👏🏽 ❤️

elias-ba:
> @stuartc I am happy I failed my mission of breaking the key handlers, they work like a charm, thanks 👏🏽 ❤️
Btw @stuartc I really loved the refactor on the Keyhandlers, thanks a lot, that's amazing

doc-han:
**QA note**: I think more tests around creating a new workflow is needed. I assume most tests are around already created workflows. I personally found some issues in the direction of creating new workflows and resolved them before handing over this PR. There might be grey areas I'm not aware of. 
@elias-ba  @stuartc  @josephjclark

josephjclark:
Saving looks good :+1: 

But thanks for flagging @doc-han:

* Create a new workflow
* DON'T SAVE
* Click on `new job`
* The server will hang forever

doc-han:
**Blocking Issue**
When creating a new workflow, clicking on a step in the WorkflowEditor doesn't open up the Form for editing the name of the step & selecting the adaptor. **I can take a look at this**.  🥹

doc-han:
> Saving looks good 👍
> 
> But thanks for flagging @doc-han:
> 
> * Create a new workflow
> * Click on `new job`
> * The server will hang forever
> 
> After a refresh its fine

Yeah. that's correct. it hangs rather

josephjclark (approved):
Thanks @doc-han !

@stuartc I'm happy with this. There's conflicts with main and I don't want to merge it anyway because I think there's a release imminent? Over to you to merge when you're ready :+1:

doc-han:
[13523e9](https://github.com/OpenFn/lightning/pull/3043/commits/13523e9f05028a2fa2fb481c50c826e72b0dce7f) An event `set-disabled` was omitted.

doc-han:
I guess I'm also satisfied with this now. 🎉

**QA:** Any proper tests around **JobEditor metadata** would be helpful. I don't remember that part being checked by anyone. Thanks
@elias-ba @josephjclark @stuartc

josephjclark:
I wouldn't expect metadata to be affected.

It looks to be broken on prod, main and this branch...

josephjclark:
@stuartc @doc-han OK, single-hierarchy is fixed, we need to get this merged down now, ASAP. There are some conflicts to resolve. Let's catch up at  standup

Issue #2284, opened 2024-07-11
Reauthorization fails when credential has not been saved


christad92:
During OAuth credential creation, I run into a crash error when I try to reauthorize the credential without saving it first. On the canvas, this crash returns an Internal Server Error with this [Sentry log](https://openfn.sentry.io/issues/5539533467/events/9557d2e55a394d6691b9ac97d1e8155b/) but on the credentials page, you'll see this modal in a pending state. (See image below). 

Reauthorization works for saved credentials without any errors but if the credential/token hasn't been saved when a user requests to reauthorize, we should treat it like a new authorization process. 

<img width="559" alt="image" src="https://github.com/OpenFn/lightning/assets/5230530/d243f185-1fed-4955-aae7-aeac2081a182">

See Loom video: 
https://www.loom.com/share/57babeaf7fb444319bff89990f897986

theroinaochieng:
1. Investigating the page crash
2. Throw away the state when the modal is closed (currently the state persists and the modal is just hidden)
3. If the action is new, disable reauthorisation
4. Won't fix: what happens when the user discards their settings and credentials.

elias-ba:
@taylordowns2000 I can confirm that the banner/message does appear when the "won’t fix" pathway is reached. However, since we shipped the token sharing feature a few days ago, this pathway has become extremely hard to reach — practically impossible for a regular user. The only real way to trigger it now would be to manually delete all tokens from the database, which is highly unlikely in a production environment.

That said, the design of the banner/message could use improvement. It’s very text-heavy, and the Reauthorize button is easy to miss. This highlights a broader issue: the credentials UI needs a serious design rethink.

I think this deserves a bigger conversation.
cc @theroinaochieng @jyeshe @stuartc

jyeshe:
> [@jyeshe](https://github.com/jyeshe) , does [this message](https://github.com/OpenFn/lightning/blob/main/lib/lightning_web/live/components/oauth.ex#L235-L268) show up for the user when the "wont fix" pathway is reached? If so, I'm happy closing this. If not, please ensure that the message appears and then we can close this.

@taylordowns2000 as we are reusing the token when creating multiple google credentials by changing the scopes, this code seems to be unreachable by a regular use. Via UI interaction it might not be possible to get this message. The only way we could see it was by deleting the oauth token on the database, then when creating a `Google Client` credential (via workflow) Google replies saying "hey you already have a token" and we show an option for reauthorisation. 

[...]

PR #3059, opened 2025-03-25
Throw away the state of oauth and other non saved credentials

jyeshe:
## Description

This PR changes the creation of new credentials to always start from the first step if the user skips it by closing the modal.

It provides a structured way independent on the context where the `CredentialFormComponent` is use to reset the whole component state (including changed by children components) without keeping track of variable names.

Refs #2284 

## Validation steps

1. Select a job
2. Click on New credential
3. Select a credential type (Http for example)
4. Close the modal on the X
5. Click on New credential and it shall display the modal to select a credential type again

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


jyeshe:
Unrelated flaky  
```
1) test Allow low priority access users to retry steps and create workorders Users with low priority access to the workflow canvas will automatically be locked in a snapshot when the high prior uses saves the workflow (LightningWeb.WorkflowLive.EditTest)
     test/lightning_web/live/workflow_live/edit_test.exs:3860
     Expected false or nil, last got true
     code: refute_eventually(low_priority_view |> has_element?("#inspector-workflow-version", "latest"))
     stacktrace:
       test/lightning_web/live/workflow_live/edit_test.exs:3899: (test)
```

lib/lightning_web/live/credential_live/credential_form_component.ex
```diff
@@ -432,10 +432,9 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
             <span class="font-bold"><.modal_title action={@action} /></span>
             <button
               id={"close-credential-modal-form-#{@credential.id || "new"}"}
-              phx-click="close_modal"
-              phx-target={@myself}
+              phx-click={JS.navigate(@return_to)}
```
elias-ba, line 435: Hey @jyeshe I don't have a lot of background context for this work but I was wondering why don't we do this `JS.navigate` thing for the other **X** close buttons in the page ? The OAuth one for example ?

jyeshe: Done

midigofrank:
Hey @jyeshe did you check if the `close_modal` `handle_event` callback is now unused in that `live_component`?

midigofrank:
> Hey @jyeshe , why can't we just patch the liveview instead? This will create another liveview

Okay so I have tested, `JS.patch/1` doesn't do anything because nothing is unchanged. I'm however against terminating the liveview process. This means that if there were unsaved changes in the workflow, all those changes would be thrown away.

@jyeshe I would recommend that we find a way to reset only the state of the `CredentialComponent` and not the liveview process. 

I have done some digging into the `close_modal` event callback and noticed that all it does is `push_event/2`. I know it's out of scope but I think we should get rid of it and instead just hide/remove the modal with `LightningWeb.Components.Modal.hide_modal/1` .
```elixir
  def handle_event("close_modal", _, socket) do
    {:noreply, push_event(socket, "close_modal", %{})}
  end
```

Or maybe we could have a `reset_state` event callback instead. And so on the close button:
```heex
<button phx-click={hide_modal(modal_id) |> JS.push("reset_state", target: @myself)}></button>
```

jyeshe:
> Hey @jyeshe , why can't we just patch the liveview instead? This will create another liveview.

@midigofrank  please notice the core of the issue involves reseting the LV state. As you're aware of from your question using `patch` wouldn't throw away the changes we have in the assigns.

midigofrank:
@jyeshe yes I get that we need to reset the state but can't we _ONLY_ reset the state of the live component? Throwing away the liveview state will impact the users in case they have other unsaved changes in the workflow

jyeshe:
> @jyeshe yes I get that we need to reset the state but can't we _ONLY_ reset the state of the live component? Throwing away the liveview state will impact the users in case they have other unsaved changes in the workflow

@midigofrank the problem is that the Component still has some coupling to the parent state:
```
    send_update(LightningWeb.CredentialLive.OauthComponent,
      id: "inner-form-#{socket.assigns.credential.id || "new"}",
      scopes: selected_scopes
    )
```

jyeshe:
@stuartc I am making this back to draft because we might need to discuss with product team if it's satisfactory loosing the state of other workflow changes. In order to reset the state without a navigate we would need to reset all this assigns:
```
{:noreply, assign(socket,
       page: :first,
       on_save: nil,
       scopes: [],
       scopes_changed: false,
       sandbox_changed: false,
       sandbox_value: nil,
       schema: false,
       project: nil,
       projects: [],
       available_projects: [],
       selected_project: nil,
       selected_projects: [],
       available_projects: [],
       oauth_clients: [],
       selected_oauth_client: nil,
       schema_selection_form: nil,
       allow_credential_transfer: false,
       users: [],
       api_version: nil,
       changeset: nil,
       update_body: nil
       )}
``` 
which is a big thing so I need some time to find a proper solution.
cc @midigofrank I am aborting to ship this on the deployment, thanks.

jyeshe:
This 8f3dea7 change is a structured way to reset the whole component state (including changed by children components) regardless of keeping track of variable names.

jyeshe:
CI Fails on unrelated flaky

jyeshe:
This 26c8960 change is a structured way to reset the whole component state (including changed by children components) regardless of keeping track of variable names.

midigofrank (approved):

lib/lightning_web/live/credential_live/credential_form_component.ex
```diff
@@ -13,25 +13,29 @@ defmodule LightningWeb.CredentialLive.CredentialFormComponent do
   alias Phoenix.LiveView.JS
 
   @impl true
-  def mount(socket) do
+  def mount(%{assigns: init_assigns} = socket) do
```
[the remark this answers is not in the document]
jyeshe: You would find inner/private assigns.

PR #3063, opened 2025-03-26
feat: hoist store & move workflowEditor to new react mount

doc-han:
## Description
I made this a new Branch off of #3043 because #3043 looks quite good and this isn't complete yet. 
In this PR
1. Hoists store
2. Moves entire WorkflowEditor to the new react mount

... **work in progress**

It moves most of the mount initializations (getting current state, pushing navigations, ...etc) into the React world.

@stuartc @theroinaochieng
## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


doc-han:
Over here. the store is mainly used via `useWorkflowStore` hook anywhere it's needed.

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -1,37 +1,33 @@
...
+import usePlaceholders from './usePlaceholders';
 import fromWorkflow from './util/from-workflow';
+import shouldLayout from './util/should-layout';
 import throttle from './util/throttle';
 import updateSelectionStyles from './util/update-selection';
-import { FIT_DURATION, FIT_PADDING } from './constants';
-import shouldLayout from './util/should-layout';
 
-import type { WorkflowState } from '../workflow-editor/store';
+import { useWorkflowStore } from '../workflow-editor/store';
```
josephjclark, line 23: As part of this work, we should move the store out to its own component. So the hook gets imported from `hooks/use-workflow-store` or something, and the component gets mounted directly the owning `.ex` file

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -44,8 +40,11 @@ type ChartCache = {
 
 const LAYOUT_DURATION = 300;
 
-export default (props: WorkflowDiagramProps) => {
-  const { selection, onSelectionChange, store, el } = props;
+export default function WorkflowDiagram(props: WorkflowDiagramProps) {
+  const { jobs, triggers, edges, disabled } = useWorkflowStore((v) => {
```
josephjclark, line 44: But nice - this is exactly what we want to see

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+      source_trigger_id: triggers[0].id,
+      target_job_id: jobs[0].id,
+      condition_type: 'always',
+    },
+  ];
+  return { triggers, jobs, edges };
+};
+
+export const WorkflowEditor: WithActionProps<{ selection: string }> = (props) => {
+  const { add, setState, getItem } = useWorkflowStore();
```
josephjclark, line 37: This is also what we want to see, great stuff

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    if (
+      currentUrl.searchParams.toString() !== nextUrl.searchParams.toString()
+    ) {
+      props.navigate(nextUrl.toString());
+    }
+  }
+
+  React.useEffect(() => {
```
josephjclark, line 72: So all this stuff has moved out of the old `workflow-editor` index into here?

This is stuff that will need moving into the store component, so it's managed by the store itself, not by the WorkflowEditor

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -0,0 +1,107 @@
...
+    props.handleEvent('navigate', (e: any) => {
```
josephjclark, line 73: Can we create a `useWorkflowSelectionHook` to track this? The URL manipulation here is surprising and unclean. Be nice to tidy it away into a hook (and maybe later that hook can use the store rather than the URL)

josephjclark (commented):
Looks like you're on the right track!

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -343,23 +346,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
-            <span class="inline-block top-[50%] relative">
-              <div class="flex items-center justify-center">
-                <.button_loader>
-                  Loading workflow
-                </.button_loader>
-              </div>
-            </span>
-          </div>
-        </div>
+        <.WorkflowEditor :if={@selection_mode != "expand"} />
```
josephjclark, line 349: This is fantastic!!!!

doc-han:
**Disclaimer**: Renaming will be done to several components, variables and so on. Currently not worried about those. 
**Update**:  Introduced a new component `WorkflowStoreSync` which actually handles the syncing of a workflow store between the client(react component) and the backend(phoenix). 

Hence `WorkflowEditor` is decoupled of all it's store interactions. It just used the store and then provides it's updates to it without interacting with phoenix directly. 

1. We can now mount JobEditor without having to mount WorkflowEditor first. 
2. Store is mounted before either JobEditor or WorkflowEditor depending on which needs to be on the screen at the moment. 
3. The share the same store. Hence if JobEditor is mounted, the store would be created but nothing regarding WorkflowEditor would be mounted. When we close the JobEditor, WorkflowEditor then gets mounted. It makes no direct calls to phoenix. It just comes to make use of the existing store component(the store component handles all server communcations)

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -172,6 +176,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
       <div class="relative h-full flex" id={"workflow-edit-#{@workflow.id}"}>
         <div class="flex-none" id="job-editor-pane">
+          <.WorkflowStoreSync />
```
josephjclark, line 179: Yes!!!

doc-han: Yup. works beautiful now. 
1. Gets store updates from phoenix downwards
2. Both WorkflowEditor & JobEditor share the same store regardless of which was mounted first
3. I'm currently not pushing updates back to phoenix yet. **in progress**...

doc-han:
### phx-hook attached to a script tag
With the current phoenix-to-react implementation. phx-hook is attached to a script tag and not an interactive HTML element like a DIV. 

**Downside:** DOM events will never be propagated to a script tag. Hence, a mounted element listening to some propagated DOM events wouldn't receive them. 

**Current Issue:** Unable to intercept the `commit-placeholder` event which is used for creating new workflow nodes.

doc-han:
### Some context

Untitled (1)

Currently, we need to track two DOM elements for every single mount. 
1. A script tag 
2. The actual react mount point

### The script tag
This is where phx-hook does its work. Hence the phx mounted event is called for this element and `this.el` will be attached to this script tag. This script tag will then have to search for it's mount point to mount a react component. It's easy to find the mount point because there's an id attribute passed to the script tag which is also passed to the react mount point div. 
**`id` in the code below**
https://github.com/OpenFn/lightning/blob/081743008fe48ec0efa72a47f1664c97cf7f1dc5/lib/react.ex#L113-L131

### The actual React mount point
The div for this mount is created by phoenix with `phx-update="ignore"`[see the code snippet above]. Now the script tag talked about above has the duty to find this div and mount a react component inside it. 


### Why we need to have reference to both elements instead of just `this.el` as we already do. 
- The script tag now has the responsibility to communicate with phoenix because the livesocket is connected.
- Hence, when we need to do `pushEventTo` it should happen via the script tags reference. 
- When it comes to event propagation inside the DOM(react), It'll never reach the script tag which is now our `this.el`. Hence we need a reference to the mount point container to intercept interested events being propagated. 

### Example Issue
When creating a new node via the workflow diagrammer, we have an event called `commit-placeholder` which would be fired and needs to be listened to on the `this.el` element. The assumption was that `this.el` will always be a parent of the mounted react component. Now it's not. Hence, we need to listen to that event on its sibling(the react mount point). 

**The commit below resolves it & this is a note for us to be aware of this.** 
https://github.com/OpenFn/lightning/pull/3063/commits/d25e5edf9424375bfe294557229f24dbe942c70c

Not sure if there are other affected events too yet.

josephjclark:
Create issues for:
- Allow html placeholders to be added to components in elixirland
- Removing the script/JSON stuff

josephjclark:
QA notes:
- load the inspector directly (from URL) and from the workflow editor
- Pay attention to inspector mounting and unmounting
- look at event stuff - push as hard as you can
- closely check positions and also resize code

assets/js/workflow-editor/WorkflowEditor.tsx
doc-han: moved into store. I doubt we want to wrap the event handlers in a hook.

doc-han:
Repositioning of canvas after a new node is added has been fixed. 
It was because the calculation of width and height were based off of the element attached to `this.el` which is now a script tag with [0, 0] has width and height. 

Using our containerEl doesn't also seem to work. I think it's because of `display: contents`. 

We just use a node reference to the `<ReactFlow />` react component

PR #3084, opened 2025-04-03
Allow users to view and copy their workflow as code

### LINKED ISSUE #3016: View and copy their workflow as code from canvas
Click a button on the workflow canvas which displays the full `workflow.yaml` (minus credentials?!?) for the current workflow and give the user the option to copy to clipboard or download as file to desktop.

The code generated here (or file) will be used to create new workflows from templates in #3013 

**Suggested Approach**

- [ ] The toggle button can appear in the workflow settings modal
- [ ] When clicked a textarea appears with the yaml code as the input
- [ ] A JS Hook should handle converting the `WorkflowParams` json into `YAML` and vice versa
- [ ] We should define a json schema that describes the yaml spec. This can later be ported to the cli

midigofrank:
## Description

This PR allows users to view their workflow as code. They can download and copy this code should they wish

Closes #3016 

## Validation steps

1. Open the workflow settings panel.
2. You should see a link to view the workflow as code. Click on it
3. A panel shows with the yaml code. You can copy and download the code

Here's a loom recording: https://www.loom.com/share/d698fa1b25e349ec92f54a0798214b61?sid=9fa3036c-4687-443e-868f-815e42f02885

This code uses the https://eemeli.org/yaml library to convert the state to yaml. This is the same library used by the cli

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


midigofrank:
Hey @josephjclark , if you get some time to spare could you have a look at the type definitions I made for the yaml exporter. It's almost a copy paste of https://github.com/OpenFn/kit/blob/main/packages/deploy/src/types.ts . 
I am however using a `Map` instead of an object to store the yaml key-values so that we can have some order. 
Question for you, is there a preferred way to define a custom `Map`? Essentially, I want all `WorkflowSpec, EdgeSpec, JobSpec, TriggerSpec` to be of type `Map`

taylordowns2000 (changes requested):
This is great, thanks! From the video I see 3 required changes:

1. fix spacing between buttons - it's not the normal spacing (img 1, img 2)
2. fix button size glitch - it gets small? (img 3)
3. fix button border? something seems off here. like these buttons aren't our buttons. the border gets heavier on click?

## to tight
image

## normal
image

## size jump
image

josephjclark:
@midigofrank regarding Map - I haven't used maps with TypeScript much but I've taken a look

Basically for a basic Map you can type the keys and you can type the values, but like, generically across the map. So "all keys are strings and all values are strings". Ie `new Map<K, V>()`

So that would look like:
```
const workflowSpec = new Map<sting, string | object>();
```
This doesn't capture keys or values in a very useful way. But you can do this for the keys:
```
const workflowSpec = new Map<keyof WorkflowSpec, string | object>();
```
When you do `map.get()` it'll always return `string | undefined`. So probably when you `get` from the map you need to cast the type anyway
```
const target_name = edgeDetails.get('target_job') as string
```
You could get around all this by specifying your own type but I don't think its worth the effort.

Finally: I hear you about preserving order in the yaml but do you really want to use maps? Js objects tend to preserve key order in iteration, but also maybe there's a better way to preserve order in yaml. Like assigning keys manually. Because JS maps are kinda horrible

midigofrank:
> Because JS maps are kinda horrible

Thanks @josephjclark, this is what I wanted to hear 😅. I wasn't really sure that assigning keys one at a time to an object preserves the order. Let me use a normal object

midigofrank:
Hey @taylordowns2000 , I have added a gap between the action buttons
Screenshot 2025-04-07 at 07 19 21

Regarding the glitch, it was because the hook for copying text was replacing the whole text content, which wipes the icon as well. To go around this, I have gotten rid of the icons.

However something still feels off when the button is "active", after being pressed. It still kinda gets smaller and the border/ring boldens. I haven't been able to fix this. By the way, I am using the `NewInputs.button` component, which is what is being used across the app. Therefore this behaviour should be consistent everywhere else. @stuartc might you have a clue?

stuartc:
@midigofrank I've updated the button, what I've done is make a new one - using `variant="secondary"` to split the logic. The issue is that by using the `color_classes` or `class` overrides we still end up inheriting other outline rules from the primary button. I'm sure overriding everyone _is_ possible, but it would be a mess imo.

I'll make an issue for cleaning up these buttons - because we have a button in common that is probably deprecated now - and we have a bunch of places that either extend `NewInputs.button` with classes that should be a variant.

stuartc:
https://github.com/user-attachments/assets/475d24f2-bf5e-4803-8825-04e810d99d62

taylordowns2000 (approved):

PR #3095, opened 2025-04-08
Store the RAG Results and the Apollo Prompt in the Database

elias-ba:
## Description

Closes #__

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/ai_assistant.ex
+++ b/lib/lightning/ai_assistant/ai_assistant.ex
@@ -141,7 +141,15 @@ defmodule Lightning.AiAssistant do
          session
        )
        when status in 200..299 do
-    message = body["history"] |> Enum.reverse() |> hd()
+    message =
+      body["history"]
+      |> Enum.reverse()
+      |> hd()
+      |> Map.merge(%{
+        "rag_results" => body["rag"],
+        "prompt" => body["system_message"]
+      })
+
     save_message(session, message, body["usage"])
   end
 
--- a/lib/lightning/ai_assistant/chat_message.ex
+++ b/lib/lightning/ai_assistant/chat_message.ex
@@ -14,12 +14,16 @@ defmodule Lightning.AiAssistant.ChatMessage do
           role: role(),
           status: status(),
           is_deleted: boolean(),
-          is_public: boolean()
+          is_public: boolean(),
+          rag_results: map() | nil,
+          prompt: String.t() | nil
         }
 
   schema "ai_chat_messages" do
     field :content, :string
     field :role, Ecto.Enum, values: [:user, :assistant]
+    field :rag_results, :map
+    field :prompt, :string
 
     field :status, Ecto.Enum,
       values: [:success, :error, :cancelled],
@@ -42,7 +46,9 @@ defmodule Lightning.AiAssistant.ChatMessage do
       :status,
       :is_deleted,
       :is_public,
-      :chat_session_id
+      :chat_session_id,
+      :rag_results,
+      :prompt
     ])
     |> validate_required([:content, :role])
     |> maybe_put_user_assoc(attrs[:user] || attrs["user"])
--- a/priv/repo/migrations/20250408010653_add_rag_and_prompt_to_chat_messages.exs
+++ b/priv/repo/migrations/20250408010653_add_rag_and_prompt_to_chat_messages.exs
@@ -1,0 +1,10 @@
+defmodule Lightning.Repo.Migrations.AddRagAndPromptToChatMessages do
+  use Ecto.Migration
+
+  def change do
+    alter table(:ai_chat_messages) do
+      add :rag_results, :jsonb
+      add :prompt, :text
+    end
+  end
+end


hanna-paasivirta:
> Are we saving rag and prompt per message, or per session?
> 
> I was expecting to save the last rag and the last message for the session
> 
> I think this implementation saves the results and prompts per _message_?
> 
> I can see value in saving rag per message, because it'll change and you maybe need to change context mid-session. But for the prompt we surely only need to save the last prompt?

Sorry, I told Elias per message would be fine because we're passing it back and forth with each message. The prompt will change if the rag search results change, so if/when we enable rag per message there might be multiple different prompts per session.

josephjclark:
@hanna-paasivirta If we ignore the prompt and just send the rag results, is that enough to rebuild the prompt on the apollo side? Is that a better starting point? Don't worry about project planning - let's just settle the correct architecture

Still a bit concerned with rag per message. It's the easiest and most future proofed architecture, but it will also be quite wasteful - especially if we have lots of rag data (like including several function signatures and examples later on)

Ok, here's a possible solution. What if we store on the session a rag_data, which includes whole rag documents with a key for each, and on each chat we just have a list of relevant keys? Now every rag document used in a session is available, but each message adds negligible duplication.

```json
{
	"history": [{
		{ "role": "user", "content": "what is openfn?"},
		{ "role": "assistant", "content": "what is openfn?", "rag": ["job_writing_guide", "adaptor_common_each"]},
	}],
	"rag_data": {
		"job_writing_guide": "..."
	    "adaptor_common_each": "...",
	}
}
```
You could add versions or dates into the document keys so that a session can take advantage of docs updates without breaking anything.

This feels like a lot of work on the apollo side - so it would mean we deploy with rag on every message (no cache) for now, then top priority after Easter is to plug in this new cache

Thoughts?

hanna-paasivirta:
@josephjclark Currently, the prompt construction step (including RAG, adaptor function description) will run if there is no existing prompt in the incoming call from Lightning. If we can get Lightning to tell us what prompt to use instead of prompt a new one, then prompt prompt only runs once a conversation (when there is no existing prompt).
I think we're only concerned about duplication etc because of the size of the database, right? Is it possible to only save the rag info to the database once per session, but still give that prompt back to Apollo every conversation turn to avoid running prompt construction?

josephjclark:
Well what I'm suggesting as a minimum is that we _always_ run prompt construction, but we skip rag if lightning sends us rag results (unless, later, we decide to ignore this because we think the context has changed). So drop the `prompt` key entirely and switch to a more nuanced prompt construction.

> I think we're only concerned about duplication etc because of the size of the database, right?

Yeah, and the amount of traffic we're needlessly sending across the wire. The more rag results we want to add, the bigger the problem this gets.

> Is it possible to only save the rag info to the database once per session, but still give that prompt back to Apollo every conversation turn to avoid running prompt construction?

Yes, absolutely. But if we're doing that, why not just store the metadata on the session, rather than the message?

josephjclark:
I think there's another risk in storing the whole prompt in lightning

Part of the prompt is, and will always be, hand written. And when the prompt is saved to the chat history, we lose the ability to update the prompt for that chat.

For example, in v1 we build a prompt and send it to lightning. The user replies, we send the prompt back and use it in the next response. Great.

But then we notice a bug in the prompt builder. So we fix it and release v2.

But older sessions are still stuck with the v1 results, so they won't see the bug go away. Users have to start a new chat session to get the benefit.

This is convincing me that we don't even _want_ to send the whole prompt Lightning, We just want to send the rag results, because that's the most expensive bit to compute.

Also, unrelated issue in apollo, maybe we should append the chat version number into the response.

hanna-paasivirta:
> Yeah, and the amount of traffic we're needlessly sending across the wire. The more rag results we want to add, the bigger the problem this gets.

Is sending the prompt string back to Apollo more expensive than generating it? The prompt string might be shorter than e.g. a multiturn conversation history.

> > Is it possible to only save the rag info to the database once per session, but still give that prompt back to Apollo every conversation turn to avoid running prompt construction?
> 
> Yes, absolutely. But if we're doing that, why not just store the metadata on the session, rather than the message?

Sorry I don't understand what storing metadata on the session means exactly, maybe because I'm not sure what the logic is for writing to the database? It's fine to store stuff only once per conversation in the metabase database, and we can minimise the amount of text there as well if needed. For the Apollo side, the key thing is whether we can pass the prompt back so it doesn't have to be regenerated. If yes, no change is needed on the Apollo side.

hanna-paasivirta:
> But older sessions are still stuck with the v1 results, so they won't see the bug go away. Users have to start a new chat session to get the benefit.
Is a session a new conversation with no conversation history? We update the prompt every few months, whereas users should start new conversations multiple times per hour of work on a topic, every time the memory gets bloated or the suggestions get off-track etc.

josephjclark:
- Include a top level `meta` key on the payload
- `meta` is optional and can be empty
- `meta` is saved to the session, not the message
- Apollo will send an dict like `{ rag }` inside the meta key, which lightning will save
- Drop the `prompt` column completely
- We'll worry about doc metadata and chat history later

josephjclark:
Hi @elias-ba I'll try and get to this today :) I want to test this and Hanna's stuff together, just got a few things to sweep up first

stuartc (approved):

josephjclark:
@elias-ba I've run this apollo locally and I don't think the meta object is getting sent back to apollo? I can see it saves and I can see where the code wires it up, so I'm a bit confused

Can we book some time to screenshare? Maybe I'm just doing something wrong

elias-ba:
> @elias-ba I've run this apollo locally and I don't think the meta object is getting sent back to apollo? I can see it saves and I can see where the code wires it up, so I'm a bit confused
> 
> Can we book some time to screenshare? Maybe I'm just doing something wrong

Hey @josephjclark let's talk about this in the AM. I think a quick Google Meet could help. Let's talk in Slack

elias-ba:
Hey @josephjclark, I found the bug and fixed it! 

The issue was that we were losing meta data because `save_message/4` is used for both user messages and assistant responses, but only assistant responses have meta data from Apollo. When saving user messages, we were accidentally setting meta to `nil` (due the defaults in the function params).

My fix is simple - just added `meta || session.meta` in the changeset. This way we preserve existing meta data for user messages while still allowing new meta data from Apollo responses.

Could you do another round of testing with your local Apollo setup? Just want to make sure the meta object is flowing correctly between Apollo and the chat session now. Let me know what you find! 🙏 I tested quickly on my end and everything seems to work fine.

josephjclark (approved):
Thanks @elias-ba - this is perfect!

Issue #3047, opened 2025-03-21
Clickable rows on the workflows (list) table


theroinaochieng:
Referenced [PR](https://github.com/OpenFn/lightning/pull/2889#issue-2825827686): In Ghana this week, users were having a very hard time figuring out how to access their project after signing up. While other changes are required, this is a cheap fix for a behaviour I noticed... even when people move their mouse of the projects item in the table, they can't figure out how to actually access it. (Hint: you've got to move your mouse over the text of the project name.)

This PR makes the whole row in the projects table clickable, while ensuring that individual links and buttons still prevent that row-level behaviour and take you where you want to go.

**_Note that I really wanted to do this for the workflow table, but that table seems to be built in a much different way and I couldn't figure it out with a quick google/chatgpt so moving on. (We don't have access to <.tr > in the workfows table 😔)_**

This 🆙 last bit is what we want to do here. See a video explanation: https://www.loom.com/share/c4a2879cd04b4f7ea73b627b214714e7

Validation steps:
1. Check to see that the whole workflow row is clickable.
2. Check to see that the formatting of the table is the same as the projects table. (This is important because eventually we'll introduce the same sorting logic.) I'm hoping they will actually use the same component, but if they don't/can't the styling should be the same.
3. Check to see that the "overrides" still work. I.e., make sure that if a link within the row is clicked (see the "55", "4", toggle, and delete buttons in the image below) that link still works and you don't get taken to the workflow canvas instead.

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/df3b7e40-cc89-40aa-a6e7-477198a06501" />

taylordowns2000:
@theroinaochieng , i've removed the "needs detail" flag as i've updated the issue and added a figma. holler if you need anything else.

elias-ba:
This is now fixed in this PR: [here](https://github.com/OpenFn/lightning/pull/3109)

Issue #2894, opened 2025-02-03
On hover, display full workflow name in a tool tip on the workflow list table


christad92:
Due to the width of the workflow list table, long workflow names are often truncated. Consequently, users cannot see the full name of the workflow, especially when multiple workflows have similar prefixes. 

To solve this, we should display the Full name of the workflow in a tooltip when a user hovers over the workflow name in the list.

<img width="1068" alt="Image" src="https://github.com/user-attachments/assets/901b83d9-edf5-4e67-83b3-5f9433d5929d" />

christad92:
cc: @theroinaochieng

elias-ba:
This is fixed in this PR [here](https://github.com/OpenFn/lightning/pull/3109). cc @theroinaochieng

PR #3109, opened 2025-04-13
Sort & Filter Workflows

### LINKED ISSUE #3046: Sort & filter workflows
1. Add sort on name, latest, workorders count, total, and enabled/disabled (this may require adding a column for enabled/disabed)
2. Make the default sort workflow name (asc)
3. When sorting by enabled/disabled, always _then_ sort by workflow name (asc)
4. Add filter on nameImage

elias-ba:
## Description

This PR adds real-time search and sorting functionality for workflows, with proper URL parameter integration to maintain state across page refreshes and navigation.

Closes #3046, #3047, & #2894

## Validation steps

1. Visit the workflows list page
2. Notice the search bar next to the 'Create workflow' button
3. Type in it to search workflows by name
4. Notice the sorting arrows in the workflows table headers
5. Click on them to sort the table
6. Notice how both sorting and searching are persisted in the URL
7. Apply search and sorting, copy the URL, paste it in another tab, and notice how state is persisted
8. Click any row in the workflows list to open that workflow (note that this doesn't block the column links and buttons) [#3047]
9. Hover on workflows names to see a little tooltip rendering the full workflow name [#2894]

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


lib/lightning/dashboard_stats.ex
```diff
@@ -74,6 +75,54 @@ defmodule Lightning.DashboardStats do
...
+
+      iex> sort_workflow_stats(workflow_stats, "workorders_count", "desc")
+      [%WorkflowStats{workorders_count: 100, ...}, %WorkflowStats{workorders_count: 50, ...}]
+
+      iex> sort_workflow_stats(workflow_stats, "last_workorder_updated_at", "asc")
+      [%WorkflowStats{last_workorder: %{updated_at: ~U[2023-01-01 00:00:00Z]}, ...}, ...]
+  """
+  def sort_workflow_stats(workflow_stats, sort_field, sort_direction) do
+    sorter = get_sort_function(sort_field)
+    sort_order = String.to_existing_atom(sort_direction)
```
jyeshe, line 102: I believe this should be handled in the `handle_params`, doing the conversion once and everywhere else could deal only with atoms.

jyeshe (commented):
Nice coverage @elias-ba and it was quite readable. Nonetheless, there might be some opportunities to make it more succinct/direct. This would bring in the value praised in Elixir of having less code to maintain while keeping it easy to understand with less cognitive jumps (even when the names are pretty clear).

taylordowns2000 (changes requested):
1. UI looks great. well done.
2. Please remove the errant space after "success" and before the closing parenthesis (see image.)
3. I have not reviewed the code, but i'm worried about the naming here (image 2). at first glance, it looks like we've duplicated the same component twice. i see you haven't, you've used the common.sortable_table_header and then created a couple of private functions that do nothing but pass in the proper parameters to the underlying component. setting aside whether or not there's value in having private functions wrapping around the actual component (i honestly have no idea, happy for you to say, "yes, we all agree that's the best way to do it") it seems weird that they're all named the same. could you rename the underlying functions in a way that justifies their existence? should you?

all honest questions. it strikes me as an odd approach, but I could just not know the proper Phoenix ways! Teach, Cherno!

image

image

~also I think you need an Elixir reviewer. I'm the product reviewer. Who does the code review here?~ Edit: I'm a dummy. I see Roger in there now. Thanks, EB!

elias-ba:
@taylordowns2000 thanks for the review. Item 2 is handled, thanks for catch. Item 3 looks misleading but actually those defps functions you see in that search result are wrappers around the `Common.sortable_table_header/1`. This is super handy because it allows each module to reassign the `Common.sortable_table_header/1` into a smaller function for readability sake. The fact that they all reuse the same name is misleading but actually a known Elixir pattern.

taylordowns2000 (approved):
Thanks @elias-ba . As long as that's the standard practice then I'm on board 😊

lib/lightning/dashboard_stats.ex
elias-ba: Thanks for the catch, this is done.

lib/lightning/dashboard_stats.ex
```diff
@@ -74,6 +75,54 @@ defmodule Lightning.DashboardStats do
...
+  defp get_sort_function("failed_workorders_count") do
+    fn stats -> stats.failed_workorders_count end
+  end
+
+  defp get_sort_function("last_workorder_updated_at") do
+    fn stats ->
+      timestamp = stats.last_workorder.updated_at || ~U[1970-01-01 00:00:00Z]
+      timestamp
+    end
+  end
```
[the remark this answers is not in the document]
elias-ba: The default timestamp of 1970-01-01 is used to ensure workflows with no updates appear at the beginning of the list when sorting by ascending order (oldest first). This was the initial design choice to make these workflows more visible.

I understand your suggestion about moving workflows with no updates to the end of the list - it's a good point about making the sorting logic simpler and it could improve UX by prioritizing active workflows. However, this would be a behavioral change that needs to be validated by the product guys (cc @taylordowns2000). Once we have that validation, we can potentially simplify the sorting logic as you suggested using `Enum.sort_by/3` with a custom comparator.

Thanks for bringing this up - let's hear what Taylor thinks.

PS: @taylordowns2000 to sum up, the question is: Are you happy when we sort the table by Latest Work Order to push the workflows with no work orders at the end of the table ?

lib/lightning_web/live/workflow_live/index.ex
```diff
@@ -87,40 +89,131 @@ defmodule LightningWeb.WorkflowLive.Index do
...
+      end
 
     socket
     |> assign(
       active_menu_item: :overview,
       page_title: "Workflows",
-      metrics: DashboardStats.aggregate_project_metrics(workflows_stats),
-      workflows_stats: workflows_stats
+      metrics: DashboardStats.aggregate_project_metrics(sorted_stats),
+      workflows_stats: sorted_stats
```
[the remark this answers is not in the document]
jyeshe: I value your concern for the use of `db_sortable_fields`. Could we apply also to the condition to the `opts`? This way we would have two separate paths (always sorting only by DB or only by calculated fields).

lib/lightning/dashboard_stats.ex
jyeshe: maybe missed the commit? (: it's still `sort_direction = params["dir"] || "asc"`

jyeshe: After this change we could remove this: 
  defp string_to_atom("asc"), do: :asc
  defp string_to_atom("desc"), do: :desc

lib/lightning/workflows.ex
```diff
@@ -292,18 +292,83 @@ defmodule Lightning.Workflows do
...
-  @spec get_workflows_for(Project.t()) :: [Workflow.t()]
-  def get_workflows_for(%Project{} = project) do
-    from(w in Workflow,
-      preload: [:triggers, :edges, jobs: [:workflow]],
-      where: is_nil(w.deleted_at) and w.project_id == ^project.id,
-      order_by: [asc: w.name]
-    )
+  def get_workflows_for(%Project{} = project, opts \\ []) do
+    include = Keyword.get(opts, :include, [:triggers, :edges, jobs: [:workflow]])
+    include = if :triggers in include, do: include, else: [:triggers | include]
```
jyeshe, line 334: to always have the triggers, let's change the default `opts`?

jyeshe (commented):
Thanks for explaining the requirement Elias. I have added some comments to simplify the code a bit. Worth to mention again that it's already readable but we can prune it a bit more.

lib/lightning/dashboard_stats.ex
elias-ba: Oh man, I am sorry I missed pushing my commits. Could you check again ?

theroinaochieng:
@elias-ba tested locally, runs great. Nothing to add beyond Taylor's spacing feedback. Thank you.

theroinaochieng (approved):

lib/lightning_web/live/workflow_live/index.ex
```diff
@@ -220,4 +324,18 @@ defmodule LightningWeb.WorkflowLive.Index do
...
+  defp to_sort_key("name"), do: :name
+  defp to_sort_key("enabled"), do: :enabled
+  defp to_sort_key("workorders_count"), do: :workorders_count
+  defp to_sort_key("failed_workorders_count"), do: :failed_workorders_count
+  defp to_sort_key("last_workorder_updated_at"), do: :last_workorder_updated_at
+  defp to_sort_key(_), do: :name
+
+  defp to_sort_direction("asc"), do: :asc
+  defp to_sort_direction("desc"), do: :desc
+  defp to_sort_direction(_), do: :asc
```
jyeshe, line 337: @elias-ba For `nil`, I am happy to go with `:asc` but better not use the generic matching because it might hide some bug.

lib/lightning/dashboard_stats.ex
```diff
@@ -74,6 +75,53 @@ defmodule Lightning.DashboardStats do
...
+  defp get_sort_function(:last_workorder_updated_at) do
+    fn stats ->
+      timestamp = stats.last_workorder.updated_at || ~U[1970-01-01 00:00:00Z]
+      timestamp
+    end
+  end
+
+  defp get_sort_function(_) do
+    fn stats -> stats.workflow.name end
+  end
```
jyeshe, line 123: Can we remove the generic match here as well, the default is already set on the `sort_key`.
We should have keep the specific for `defp get_sort_function(:last_workorder_updated_at)` and all others would be and identity function:
`defp get_sort_function(field), do: field`

jyeshe (changes requested):
@elias-ba there was missing related to the default :triggers in opts. I had left a suggestion also on how to avoid the double sorting. I could waive it if you have an urgent issue on the plate but the `:triggers` thing is quite localised and the `get_sort_function` could be easily tackled to remove the boilerplaite (in the sense of repeated definitions because of a field name). If you use only the `get_sort_function` (or `get_sorter` more precisely when returning a field) we could also get rid of the `to_sort_key`. Code-wise to express myself crystal clear:
```
def get_sorter("last_workorder_updated_at") do
  fn stats ->
    timestamp = stats.last_workorder.updated_at || ~U[1970-01-01 00:00:00Z]
    timestamp
  end
end

def get_sorter(field_name), do: String.to_existing_atom(field_name)
```

lib/lightning_web/live/workflow_live/index.ex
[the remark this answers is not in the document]
elias-ba: This is done, now. It was done before but that was in the `Workflows.get_workflows_for/2` function. But I moved it out in the `apply_action/3` function in `index.ex` module as you suggested. Thanks

lib/lightning/workflows.ex
elias-ba: This is fixed now, thanks

lib/lightning_web/live/workflow_live/index.ex
elias-ba: I explicitly added a case for nil

lib/lightning/dashboard_stats.ex
elias-ba: Thanks for the catch this is done

elias-ba:
> @elias-ba there was missing related to the default :triggers in opts. I had left a suggestion also on how to avoid the double sorting. I could waive it if you have an urgent issue on the plate but the `:triggers` thing is quite localised and the `get_sort_function` could be easily tackled to remove the boilerplaite (in the sense of repeated definitions because of a field name). If you use only the `get_sort_function` (or `get_sorter` more precisely when returning a field) we could also get rid of the `to_sort_key`. Code-wise to express myself crystal clear:
> 
> ```
> def get_sorter("last_workorder_updated_at") do
>   fn stats ->
>     timestamp = stats.last_workorder.updated_at || ~U[1970-01-01 00:00:00Z]
>     timestamp
>   end
> end
> 
> def get_sorter(field_name), do: String.to_existing_atom(field_name)
> ```

Hey @jyeshe thanks for the catch on simplifying stuff. I do want to avoid `String.to_existing_atom/1` tho so I will stick with the atom based solution but here is how I implemented your ideas and made it simpler:

```elixir
def sort_workflow_stats(workflow_stats, sort_field, sort_direction)
    when is_atom(sort_field) and is_atom(sort_direction) do
  sorter = get_sorter(sort_field)
  Enum.sort_by(workflow_stats, sorter, sort_direction)
end

defp get_sorter(:last_workorder_updated_at) do
  fn stats ->
    stats.last_workorder.updated_at || ~U[1970-01-01 00:00:00Z]
  end
end

defp get_sorter(field) do
  fn stats -> Map.get(stats, field) end
end
```

lib/lightning/dashboard_stats.ex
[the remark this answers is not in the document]
elias-ba: See answer here: https://github.com/OpenFn/lightning/pull/3109#issuecomment-2807700776

jyeshe (approved):
Well done @elias-ba !

PR #3178, opened 2025-04-30
fix github sync modal in the workflow editor

### LINKED ISSUE #3177: "Save & Sync" button gives no action/feedback
1. Clicking "Save and Sync" from the modal inside the Inspector (see image) does nothing. (No error that I can see in the browser.)
2. Pressing `Command-S` from the Canvas no longer seems to initiative save & sync.
3. No sync appears to be initiated, though a toast message appears at the bottom when saving and syncing from the canvas.

Image

Update: when I then exit to the canvas, save the workflow (command-s), and save & sync from the canvas it seems like it might work, but the modal stays open on the page. 

**Version number**  `v2.12.1 (1.4.28)`

- [ ] Yes
- [x] No

1. Go to inspetor
2. Hit command-s to open the save & sync modal
3. Type your commit message and click save

happen.

I'd expect that a save & sync is initiated and that I see a toast at the bottom when it's been sent off succesfully.

midigofrank:
## Description

This PR fixes the bug on the github sync modal in the workflow editor. The problem was that the modal was embedded inside the `#workflow-form`, which was always present. This PR creates a stand alone form for the github sync modal.

This PR also increases the priority of the `CtrlShiftS` binding. It was being overridden by `CtrlS`.

Closes #3177

## Validation steps

You need to have Github setup to validate this locally. Check the docs for the needed env vars: https://github.com/OpenFn/lightning/blob/main/DEPLOYMENT.md#github 

In the workflow editor, press `CtrlShiftS` or `CmdShiftS` if you're on mac. The github sync modal will appear.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/github_sync_modal.ex
+++ b/lib/lightning_web/live/workflow_live/github_sync_modal.ex
@@ -25,7 +25,7 @@ defmodule LightningWeb.WorkflowLive.GithubSyncModal do
   @impl true
   def render(assigns) do
     ~H"""
-    <div>
+    <form id={@id <> "-form"} phx-submit="save">
       <.modal
         id={@id}
         show={true}
@@ -160,7 +160,6 @@ defmodule LightningWeb.WorkflowLive.GithubSyncModal do
             <.button
               id={"submit-btn-#{@id}"}
               type="submit"
-              form="workflow-form"
               disabled={!@verify_connection.ok?}
             >
               Save and sync
@@ -175,7 +174,7 @@ defmodule LightningWeb.WorkflowLive.GithubSyncModal do
           </div>
         </:footer>
       </.modal>
-    </div>
+    </form>
     """
   end
 end
--- a/lib/lightning_web/live/workflow_live/github_sync_modal.ex
+++ b/lib/lightning_web/live/workflow_live/github_sync_modal.ex
@@ -25,7 +25,7 @@ defmodule LightningWeb.WorkflowLive.GithubSyncModal do
   @impl true
   def render(assigns) do
     ~H"""
-    <form id={@id <> "-form"} phx-submit="save">
+    <form id={@id <> "-form"} phx-submit="save-and-sync">
       <.modal
         id={@id}
         show={true}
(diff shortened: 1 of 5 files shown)


taylordowns2000 (approved):
looks good to me!

image

lib/lightning_web/live/workflow_live/github_sync_modal.ex
```diff
@@ -25,7 +25,7 @@ defmodule LightningWeb.WorkflowLive.GithubSyncModal do
   @impl true
   def render(assigns) do
     ~H"""
-    <div>
+    <form id={@id <> "-form"} phx-submit="save">
```
jyeshe, line 28: Hey @midigofrank nice catch! Would you mind renaming the event for `save_gh`? I find it important because it is handled by the `WorkflowLive.Edit` which we know that needs to be refactored into smaller components.

midigofrank: Hey @jyeshe , good idea, I actually thought about and decided just to leave it as it was under the `def handle_event("save", _, _)`  callback.

The logic is tightly coupled with the usual `save action` with the only difference being the call to github sync and the flash message, so I just decided to leave it coupled given the time. 

Let me see if I can refactor it out easily

midigofrank: Hey @jyeshe , I have refactored it out into a different callback. Let me know your thoughts

jyeshe (approved):
Super, even better! Thanks!

jyeshe:
The decoupling was great because it allows now the modal to handle itsown event, doesn't it? I 've approved as it was already greatly improved but if you have an extra time and could move the handler to the modal, it would be awesome.

midigofrank:
@jyeshe I hear you, but moving the callback to the `live_component` is a little tricky, mostly because of naming. 

 It would have been easy if we could just make the `save_workflow/2` public, but we can't because we already have a `WorkflowLive.Helpers.save_workflow/2` which doesn't take in a socket.

 The alternative is to duplicate the `socket handling logic`, but this isn't ideal since it's directly tied to the `WorkflowLive.Edit` liveview.

PR #3181, opened 2025-05-02
Halve the default chunk size, make configurable, delete more frequently

taylordowns2000:
@jyeshe , how does this look?

1. We say we don't keep data for more than 7 days, but when we only run once a day we could actually keep data for 7 days and 23 hours! This is fixed by running it every 15 min.
2. We shrink the chunk size by default from 1000 to 500
3. We make the chunk size configurable via ENV


jyeshe (approved):
As a global app, used across multiple timezones, this strategy makes sense. @taylordowns2000 along side I would just think preparing a read-replica for the history page to avoid heavy queries competing with the heavy delete.

PR #3187, opened 2025-05-02
Simplified projects activity delete

taylordowns2000:
@stuartc , here's another possible approach to the query


stuartc:
Resolve in https://github.com/OpenFn/lightning/pull/3190

PR #3210, opened 2025-05-13
same search

taylordowns2000:
@doc-han , can you test this out and let me know if the dataclip UUID search is behaving the way you'd expect now?

cc @stuartc , if you have more context on why Roger set up the search differently from the partial UUID search on the history page, please let me know! this is basically a copy paste from how we search for runs, steps, and workorders by parts of their UUIDs.

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -67,16 +67,10 @@ defmodule Lightning.Invocation do
     filters =
       Enum.reduce(filters, dynamic(true), fn
         {:id, uuid}, dynamic ->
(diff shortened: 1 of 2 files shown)


doc-han:
alright

doc-han (approved):

doc-han:
Thanka @taylordowns2000 ❤️
tested & works for me

taylordowns2000:
@stuartc , some additional considerations around byte search here: https://github.com/OpenFn/lightning/pull/3214#issue-3060648365

it sounds like a cool, more performant idea, but maybe has a few small bugs to iron out?

PR #3211, opened 2025-05-13
Allow datetime search

taylordowns2000:
@doc-han and @stuartc , this implements the "before" part of the search. Why was this not here already? Seems fine in my testing

### DIFF
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -81,9 +81,6 @@ defmodule Lightning.Invocation do
         {:type, type}, dynamic ->
           dynamic([d], ^dynamic and d.type == ^type)
 
-        {:date, date}, dynamic ->
-          dynamic([d], ^dynamic and fragment("date(?)", d.inserted_at) == ^date)
-
         {:datetime, ts}, dynamic ->
           dynamic([d], ^dynamic and d.inserted_at == ^ts)
 
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -86,6 +86,9 @@ defmodule Lightning.Invocation do
 
         {:after, ts}, dynamic ->
           dynamic([d], ^dynamic and d.inserted_at > ^ts)
+
+        {:before, ts}, dynamic ->
+          dynamic([d], ^dynamic and d.inserted_at < ^ts)
       end)
 
     Query.last_n_for_job(job_id, limit)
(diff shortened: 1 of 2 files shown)


lib/lightning/invocation.ex
```diff
@@ -81,14 +81,14 @@ defmodule Lightning.Invocation do
         {:type, type}, dynamic ->
           dynamic([d], ^dynamic and d.type == ^type)
 
-        {:date, date}, dynamic ->
-          dynamic([d], ^dynamic and fragment("date(?)", d.inserted_at) == ^date)
-
         {:datetime, ts}, dynamic ->
```
stuartc, line 84: Great! Thank you for removing this, this would require another index to be performant. The `lt` and `gt` combo is much better.

lib/lightning_web/live/workflow_live/new_manual_run.ex
```diff
@@ -69,21 +69,27 @@ defmodule LightningWeb.WorkflowLive.NewManualRun do
...
-        with {:ok, datetime} <- NaiveDateTime.from_iso8601(param) do
-          {:ok, {:after, datetime}}
-        end
+        {:error, :invalid_datetime}
+    end
+  end
+
+  defp parse_param("before:" <> param) do
+    # TODO - Why doesn't the native datetime picker (datetime-local) send seconds?
+    case NaiveDateTime.from_iso8601(param <> ":00") do
```
stuartc, line 84: @taylordowns2000 could you give me some context around why you needed to match and/or drop part of the string here. I don't fully grok the todo, the prior implementation didn't have this matching right? Why now?

taylordowns2000: Yeah totes!

The original implementation made it so that users couldn't search by datetime at all... you could only search by dates—no time. A miscommunication.

This change allows datetime so it behaves like the datetime search on the history page.

The `datetime-local` input, I'm told, doesn't return seconds. In Elixir, we handle that by casting [here](https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex#L66) to a type set [here](https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex#L37)

This is my ugly way of adding seconds to the non-seconds result that comes out of the native html input for `datetime-local`

stuartc: Ok thanks guys, date time local in browsers returns what? The time local to the user without a time zone? 
Or is it left as is?

Our app works exclusively (by design) in UTC. 

@taylordowns2000 i want to check this branch out and just double check there isn't room for improvement, we always need to take extra care when handling user data. The implementation should be able to handle rubbish data as well.

stuartc (approved):

PR #3213, opened 2025-05-13
fix monaco sizing

josephjclark:
Fix Monaco so that:
1) Does not infinitely grow
2) Shrinks when its parent panel shrinks

`overflow-hidden` is the thing, because it tells  monaco's parent not to size to its children

I used https://github.com/OpenFn/lightning/pull/2126 as inspiration

### DIFF
--- a/assets/js/manual-run-panel/views/CustomView.tsx
+++ b/assets/js/manual-run-panel/views/CustomView.tsx
@@ -58,7 +58,7 @@ const CustomView: React.FC<{
         <span className="bg-white px-2 text-sm text-gray-500">OR</span>
       </div>
     </div>
-    <div className="relative h-full flex flex-col">
+    <div className="relative h-full flex flex-col overflow-hidden">
       {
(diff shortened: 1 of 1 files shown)


taylordowns2000 (changes requested):
Looks great except for the error message! Doesn't seem to work when the error message bumps it all down a few pixels:

image

doc-han:
OMG. `overflow-hidden` has really dealt with me.
Thanks you very much for this @josephjclark. I'm happy right now.

josephjclark:
@doc-han @taylordowns2000 let me look at that error thing

doc-han:
@josephjclark I've fixed the issue @taylordowns2000 mentioned

doc-han:
The secret sauce. `overflow-hidden` ❤️

taylordowns2000 (approved):
just tested the last commit (fix jump on error) and it looks great. thanks @doc-han and @josephjclark

doc-han (approved):

PR #3214, opened 2025-05-13
Consideration for the byte search concept

taylordowns2000:
@jyeshe , I'm interested in the byte search idea (reverted back to it here) but i'd want to see an implementation where '7d4' doesn't match '7d866f02'.

These "false positives" were enough for me to revert to the standard UUID search we use on the history page (which does partial string matching quite nicely.) Note the failing test here. I created it to help showcase the false positives.

`mix test test/lightning/invocation_test.exs`

I don't think it's high priority right now, but if @theroinaochieng and @stuartc are interested in following this thread when you're back online, they might consider an implementation like this for all of our UUID searching. Note that right now, UUID search on History is, weirdly, only for runs, steps, and work orders. AFAIK there isn't UUID search for data clips, and your implementation here could be a handy way to standardize all that stuff.

### DIFF
--- a/assets/js/manual-run-panel/views/ExistingView.tsx
+++ b/assets/js/manual-run-panel/views/ExistingView.tsx
@@ -88,7 +88,7 @@ const ExistingView: React.FC<ExistingViewProps> = ({
                 }}
                 type="text"
                 className="focus:outline focus:outline-2 focus:-outline-offset-2 focus:ring-0  disabled:cursor-not-allowed disabled:bg-gray-50 disabled:text-gray-500 border-slate-300 focus:border-slate-400 focus:outline-indigo-600 block w-full rounded-md border-0 py-1.5 pl-10 pr-20 text-gray-900 ring-1 ring-inset ring-gray-300 placeholder:text-gray-400  focus:ring-inset focus:ring-indigo-600 sm:text-sm sm:leading-6"
-                placeholder="Search for inputs by UUID"
+                placeholder="Search by UUID (starts with)"
               />
               <div className="pointer-events-none absolute inset-y-0 left-0 flex items-center pl-3">
                 <MagnifyingGlassIcon className="h-5 w-5 text-gray-400" />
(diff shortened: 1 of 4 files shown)


jyeshe:
Nice catch @taylordowns2000 . It's still a theoretical assumption but I think this approach will save some db cycles as the uuid prefixes with odd number of chars cannot be represented as a _binary_ (uuid byte array) so Postgres is likely encoding each btree node to a string to try to match the string in the where clause. But these guys are no joke so they might also have implemented something really out of a hat on the DB planer.  
PS: My local machine has a fast drive and with my external ssd it's also blazing fast. So instead of spending time on recreating the scenario on staging I have just implemented what would be less likely for the DB planer to do on a very specific case.

stuartc (commented):
Ok I've spotted two issues here.

Since the data clip search requires two chars to actually do a search on the uuid, if you have lots of dataclips and it wasn't one of the most recent 10 _and_ you only enter the first character - you won't get any results.

https://github.com/user-attachments/assets/b19e05f1-37fb-4b54-9085-71dadcefcedb

Is this acceptable?

taylordowns2000:
> Ok I've spotted an issue here.
> 
> Since the data clip search requires two chars to actually do a search on the uuid, if you have lots of dataclips and it wasn't one of the most recent 10 _and_ you only enter the first character - you won't get any results.
> 
>  Screen.Recording.2025-05-14.at.11.55.27.mp4 
> Is this acceptable?

Nice catch. Ehhh.... it's right on the edge. Seems maybe OK to me? If folks are searching by UUID hopefully they'll enter more than 1 char. When we start allowing them to search by UUID _or_ dataclip `label` for the named dataclips maybe then it will be different, but I think this is OK.

What do you think? Merge if you're down with it @stuartc , keep open (unmerged) if not.

jyeshe:
> Ok I've spotted an issue here.
> 
> Since the data clip search requires two chars to actually do a search on the uuid, if you have lots of dataclips and it wasn't one of the most recent 10 _and_ you only enter the first character - you won't get any results.
> 
>  Screen.Recording.2025-05-14.at.11.55.27.mp4 
> Is this acceptable?

Thank you for walking through the details @stuartc . This search really alleviates a lot of pressure on database. This final filtering was supposed to be done in the app when the user types 1 char (or 3, 5 ...) but you've raised an important matter that I was also wondering about, if we shouldn't wait (or expect) the user to type at least 3 or 4 chars. Otherwise the user types only 1 we would need to fetch all to the app. I think we would not be demanding much from the user to remember at least 3 or 4 chars, right? A common pattern on prefix searches...

This point you've spot is worth discussing and refining cc @taylordowns2000 .

PR #3221, opened 2025-05-17
it's ok to not have a filter

taylordowns2000:
not sure if this is the fix, but it seems to highlight and solve the problem here: since https://github.com/OpenFn/lightning/commit/214e8363f4ba0fb8ab01072e7b6ce3f19da09ddc, when no filters are added, we see zero results instead of all the results

### DIFF
--- a/lib/lightning_web/live/workflow_live/new_manual_run.ex
+++ b/lib/lightning_web/live/workflow_live/new_manual_run.ex
@@ -56,10 +56,11 @@ defmodule LightningWeb.WorkflowLive.NewManualRun do
       params,
       [:before, :after, :type, :query]
     )
-    |> Lightning.Validators.validate_one_required(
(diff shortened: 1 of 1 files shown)


doc-han (approved):
This works. `search-selectable-dataclips` event without a filter now returns some dataclips

doc-han:
If we merge this. it mean `get-selectable-dataclip` & `search-selectable-dataclip` will have the same functionality when no filters are passed. Hence we can drop `get-selectable-dataclip`.

taylordowns2000:
if it all behaves as expected and tests are passing, please do merge. we can ask @stuartc about https://github.com/OpenFn/lightning/pull/3221/files#diff-57f0d361ab07b3da41cc7fd7e320aa7926227c11d3d85d496e3766eb7816ebb6L59-R63 later!

Issue #3009, opened 2025-03-12
AI Assistant: Increase response timeout


josephjclark:
We've had mysterious reports of users receiving a "timeout" error message when using the assistant.

We haven't really gotten to the bottom of why this is.

But the timeout to the Apollo server is currently set at 15 seconds. It seems plausible to me that a request with a lot of data, on a slow internet day, could take <15 seconds to return.

Increasing the timeout will help. Maybe 30 seconds?

Note that a user waiting for more than about 5 seconds for a response is terrible, and we really ought to look into a streaming API

josephjclark:
Our new RAG searches are a little lower, so I think this is more important than ever

PR #3230, opened 2025-05-22
Set a timeout on Apollo Client

jyeshe:
## Description

This PR adds a timeout config for Apollo client requests.

Refs #3009 

## Validation steps

1. In one terminal create a TCP listening script or use your own tool to listen on port 3000 with a timeout higher than 5s
`echo ":gen_tcp.listen(3000, [:binary, packet: :line, active: false]); Process.sleep(10_000)" > listen.exs`
2. Run `elixir --no-halt listen.exs
3. Run Lightning on another terminal with `APOLLO_TIMEOUT="3000" iex -S mix phx.server`
4. Execute `{:error, :timeout} = Lightning.ApolloClient.query("foo")` and see the timeout error after 3 seconds

1. This is a palliative solution before implementing the streaming even though it might be needed afterwards to set an expected max timeout for the user.    

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

- [X] I have performed a **self-review** of my code.
- [X] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [X] I have ticked a box in "AI usage" in this PR


config/dev.exs
```diff
@@ -129,4 +129,4 @@ config :phoenix, :plug_init_mode, :runtime
 
 config :lightning, :is_resettable_demo, true
 
-config :lightning, :apollo, endpoint: "http://localhost:3000"
+config :lightning, :apollo, endpoint: "http://localhost:3000", timeout: 5_000
```
elias-ba, line 132: Hey @jyeshe, this is really great! Just thinking out loud — with RAG now integrated into Apollo, don't you think that 5 seconds is a bit short for some AI operations? Curious to hear your thoughts. cc @josephjclark @hanna-paasivirta

test/lightning/apollo_client_test.exs
```diff
@@ -49,4 +50,20 @@ defmodule Lightning.ApolloClientTest do
...
+    Mox.stub(Lightning.MockConfig, :apollo, fn key ->
+      case key do
+        :endpoint -> "http://localhost:3000"
+        :ai_assistant_api_key -> "api_key"
+        :timeout -> 100
+      end
+    end)
+
+    stub(Lightning.Tesla.Mock, :call, fn _env, _opts ->
+      Process.sleep(200)
```
elias-ba, line 64: Is it okay to have process sleep operations in our test suite ? How would this impact the speed of our tests ? I know that tests are ran async but just wanna make sure

elias-ba (commented):
Hey @jyeshe this looks great to me. Nice job. I have left few comments that I would love to hear your thoughts on and maybe @josephjclark and @hanna-paasivirta thoughts too.

config/dev.exs
jyeshe: Please notice this is for dev. We can leave a high one here as well. In the deployment PR it's set to 30s.

test/lightning/apollo_client_test.exs
jyeshe: A very short one is acceptable to cause the situation where the call takes more time than the timeout set. The call takes 200ms for a timeout config of 100ms so that in this test case in particular gonna return the timeout error.

stuartc (approved):
Nice on the config cleanup on the tests!

config/dev.exs
stuartc: I think in dev we will need it much higher as well, even in dev the RAG and the LLM calls are both hosted services. What is the default timeout for Tesla? If we don't have an agreed timeout specifically for dev, we should set this to whatever Tesla would have as a default.

stuartc (changes requested):
@jyeshe looks like we've hit a snag, it looks like the timeout middleware doesn't override the Finch pool one.

```exs
Mix.install([
  {:plug_cowboy, "~> 2.5"},
  {:tesla, "~> 1.14"},
  {:finch, "~> 0.19"}
])

children = [
  {Finch, name: MyFinch}
]

Supervisor.start_link(children, strategy: :one_for_one)

defmodule SlowServer do
  use Plug.Router

  plug :match
  plug :dispatch

  match _ do
    # Never send a response - client will timeout
    Process.sleep(:infinity)
  end

  def start do
    Plug.Cowboy.http(SlowServer, [], port: 3000)
  end
end

defmodule Test do
  def timed_request(timeout) do
    start_time = System.monotonic_time()

    result =
      Tesla.client(
        [{Tesla.Middleware.Timeout, timeout: timeout}],
        {Tesla.Adapter.Finch, name: MyFinch}
      )
      |> Tesla.get("http://localhost:3000")

    end_time = System.monotonic_time()

    duration =
      System.convert_time_unit(end_time - start_time, :native, :millisecond)

    IO.puts("Request took #{duration}ms")
    {result, duration}
  end

  def test_timeout(timeout) do
    {result, duration} = timed_request(timeout)
    IO.puts("Result: #{inspect(result)}")
    IO.puts("Duration: #{duration}ms")
    if duration < timeout do
      IO.puts("⚠️ Warning: Request timed out faster than expected timeout of #{timeout}ms")
    else
      IO.puts("✅ Request timed out within expected timeout of #{timeout}ms")
    end
  end
end

IO.puts("Starting server")
SlowServer.start()

Test.test_timeout(3_000)
Test.test_timeout(40_000)
```

Results in

```
$ elixir testing_finch_timeout.exs

Starting server
Request took 3005ms
Result: {:error, :timeout}
Duration: 3005ms
✅ Request timed out within expected timeout of 3000ms
Request took 15002ms
Result: {:error, :timeout}
Duration: 15002ms
⚠️ Warning: Request timed out faster than expected timeout of 40000ms
```

Looks like the default timeout value for each chunk (or response in the case of HTTP1.1) in Finch is 15s (https://hexdocs.pm/tesla/Tesla.Adapter.Finch.html#module-finch-options)

Since we don't have streaming from Apollo yet, we will need to adjust the Finch start opts, determine which (if not both) options need increasing; `:pool_timeout` and/or `:receive_timeout` and increase this significantly.

stuartc:
> @stuartc please notice the ApolloClient doesn't use Finch explicitly on the client. Following the validation steps, with an external server, the timeout went well.

It doesn't use it explicitly, but it defaults to Finch. See `config.exs:176`:

```
config :tesla, adapter: {Tesla.Adapter.Finch, name: Lightning.Finch}
```

I can confirm with this example:

`RTM=false APOLLO_TIMEOUT="45000" iex -S mix phx.server`

```elixir
start_time = System.monotonic_time()
Lightning.ApolloClient.query("foo")
end_time = System.monotonic_time()

duration =
  System.convert_time_unit(end_time - start_time, :native, :millisecond)

IO.puts("Request took #{duration}ms")
```

That it times out in 15s, not the 45s specified in the config.

opragmatico:
> > @stuartc please notice the ApolloClient doesn't use Finch explicitly on the client. Following the validation steps, with an external server, the timeout went well.
> 
> It doesn't use it explicitly, but it defaults to Finch. See `config.exs:176`:
> 
> ```
> config :tesla, adapter: {Tesla.Adapter.Finch, name: Lightning.Finch}
> ```
> 
> I can confirm with this example:
> 
> `RTM=false APOLLO_TIMEOUT="45000" iex -S mix phx.server`
> 
> ```elixir
> start_time = System.monotonic_time()
> Lightning.ApolloClient.query("foo")
> end_time = System.monotonic_time()
> 
> duration =
>   System.convert_time_unit(end_time - start_time, :native, :millisecond)
> 
> IO.puts("Request took #{duration}ms")
> ```
> 
> That it times out in 15s, not the 45s specified in the config.

@stuartc Documenting it so far, my hypothesis was not confirmed as the difference of being a programatic or declarative  adapter (via config) had no influence.

The problem was with the Tesla Timeout Middleware that only works when the timeout is smaller than the Finch timeout. It gives this false impression that it overrides the adapter timeout when you set APOLLO_TIMEOUT for 10 seconds. But if you set APOLLO_TIMEOUT for 30 seconds it doesn't work. 

Another issue using Tesla here is that the adapter opts are not forwarded. From the [docs](https://hexdocs.pm/tesla/Tesla.Env.html) we could use:
`:opts - list of options. Example: [adapter: [recv_timeout: 30_000]]`

but both `recv_timeout` and Finch `receive_timeout` do not take effect.

Calling: 
```
    client()
    |> Tesla.post("/services/job_chat", payload,
       adapter: [receive_timeout: 30000]
      # adapter: [recv_timeout: 30000]
      # adapter: [request_timeout: 30000]
      # recv_timeout: 30000]
      # receive_timeout: 30000]
    )
``` 
results in
```
[iex:4: (file)]
end_time - start_time #=> 15
```

opragmatico:
@stuartc it had an acceptable effect, I guess, using the opts on the client:
```
  defp client do
    Tesla.client([
      {Tesla.Middleware.BaseUrl, Lightning.Config.apollo(:endpoint)},
      Tesla.Middleware.JSON,
      Tesla.Middleware.KeepRequest
    ], {Tesla.Adapter.Finch, name: Lightning.Finch, receive_timeout: 30000})
  end
```
producing sometimes few seconds of variation on the timeout result:
image

mostly right after 30 seconds:
image

config/dev.exs
opragmatico: Done. Increased for 30 seconds as the default timeout was determined by Finch (15 seconds)

test/lightning/apollo_client_test.exs
opragmatico: It was needed because the :tesla app :adapter for test is the `Lightning.Tesla.Mock`. But now the timeout takes effect only on Finch so I have removed the test case.

I could put a different adapter only for this test case but we would risk introducing a flaky as the temporary config, with Finch adapter during the test, could be read by another test file.

stuartc (approved):
👌 nice!

It's a shame we need the if statement for the adaptor choice - but I don't think either of us are invested in hunting down a "perfect" way.

opragmatico:
> 👌 nice!
> 
> It's a shame we need the if statement for the adaptor choice - but I don't think either of us are invested in hunting down a "perfect" way.

Yeah, unfortunately the Mock adapter needed it. And thanks for putting back the stubs cleanup.

PR #3236, opened 2025-05-28
Optimize claim query performance

taylordowns2000:
@stuartc , @jyeshe , here's the concept from Cursor to discuss today

Optimize eligible_for_claim query performance

The query was taking too long with large queues (>1000 runs). Changes:

1. Simplified query structure:
   - Removed expensive window function that was calculating row numbers for all runs
   - Replaced multiple subqueries with a single in-progress count query
   - Eliminated redundant joins and intermediate result sets
   - Maintains strict FIFO ordering by inserted_at across all workflows/projects

2. Added covering index on runs table for available runs:
   - Partial index (WHERE state = 'available')
   - Includes work_order_id to avoid table lookups
   - Ordered by inserted_at for efficient oldest-first retrieval

This ensures runs are processed in strict chronological order while significantly improving query performance.

### DIFF
--- a/lib/lightning/runs/query.ex
+++ b/lib/lightning/runs/query.ex
@@ -127,17 +127,29 @@ defmodule Lightning.Runs.Query do
   """
   @spec eligible_for_claim() :: Ecto.Queryable.t()
   def eligible_for_claim do
-    Run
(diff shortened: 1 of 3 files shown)


taylordowns2000:
done by @stuartc

PR #3242, opened 2025-05-28
Free node positioning on canvas

### LINKED ISSUE #2175: workflow diagram: create nodes and edges with the same button
I've just added a second hover button to a node, so that now you can add a new child node or an edge

An improvement would be to only have a single `+` button (as originally so), which creates an edge with a ghost node on the end. If you release the mouse on the canvas, a new placeholder node will be created. If you drag to valid node, a new edge will be created.

This should be a nice slick UX. New placeholders are easy to dismiss (just click the canvas again).

One issue is that the  node won't be created where you drop it, it'll be subject to the layout. That could be annoying.

### LINKED ISSUE #3123: Allow users to edit the position of the nodes/steps by moving them around in the canvas. In this case, users should be able to turn on/off the auto-layout feature.

### LINKED ISSUE #3125: For large workflows it would be nice to have a minimap for navigation

midigofrank:
**EDIT FROM JOE**

This PR was heavily edited from here to resolve a number of issues and improve the UI/UX

This includes a version bump to reactflow 12, which caused a few issues

## Original Description

This PR makes it possible to toggle free node positioning. 
It introduces 2 buttons (`MnL - Manual Layout` and `FrcL - Force Layout` as seen below.

Clicking on `MnL` makes it possible to freely position the nodes. After clicking, it changes to `AtL - Auto Layout`. While in auto layout, you cannot move the nodes.

Screenshot 2025-05-28 at 14 54 22

Closes #3123 
Closes #3125
Closes #2175 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,13 @@ and this project adheres to
   [#3123](https://github.com/OpenFn/lightning/issues/3123)
 - Minimap for easier workflow navigation
   [#3125](https://github.com/OpenFn/lightning/issues/3125)
+- Added icons to control layout in the workflow
+  [PR #3242](https://github.com/OpenFn/lightning/pull/3242)
 
 ### Changed
 
+- Update React Flow to version 12
+  [PR #3242](https://github.com/OpenFn/lightning/pull/3242)
 - Create nodes and edges with the same button in the workflow
   [#2175](https://github.com/OpenFn/lightning/issues/2175)
 


josephjclark:
Just an observation: in auto layout mode, the behaviour of dragging out a new node onto the canvas doesn't make any sense. No matter where I drop it, the new node will be created wherever the layout wants.

Possible solutions:
1. In autolayout mode, show the existing Plus and Link icons. No drag and drop. But in manual layout, only show the drag icon.
2. Err, that's all I can think of right now?
3. The only other thing you can do is manual layout by default, and force users to hit the Layout button to tidy up

taylordowns2000 (changes requested):
1. cleanup tippy implementation (if necessary) and merge new icons and tooltips: #3304 
2. node drop not working in manual layout mode
3. make sure that `positions` are added to the template when a template is published

josephjclark:
Some QA issues and general concerns:

* Please don't merge with the `sparkle` emoji (right now sparkle is used for magic functions, AI usage, and layout)
* I know it was requested to remove neighbouring edge highlights, but I really think people are going to miss it when it's gone! I wonder if we should use a semitone highlight instead to make the primary and neighbouring selections easier to read.
* If you drag a new node onto an invalid place - like an existing child - you'll see an error, and when you release the button, a new node will be created. We need to cancel the drop if the target is an invalid node (also, in manual layout, when dropping it might be hard to drop in between nodes if we do this, because the drop area is too big) @doc-han 
* When you drag a new node in manual mode, it never quite drops where you expect it to. There's an offset.
* We're showing a dotted background but there's no snap-to-grid behaviour in manual layout

josephjclark:
Other bugs on this branch right now (with help from farhan)

- Store is not reset properly on unmount, so sometimes opening a new workflow has the wrong data @doc-han 
- Templates break for workflows which don't have positions (`can't read properties of undefined 'x'`) @doc-han 
- Switching quickly between workflows sometimes causes a workflow to load without edges (pretty sure this is happening on main right now) @doc-han 
- sometimes when you crate a new placeholder, on drop, the placeholder vanishes. Kinda hard to repro. This is either a re-render (this is what I used to think, it keeps coming back to this branch) or it's cause by a drag and a click event running at the same time (possible new theory - maybe this only occurs on auto layout)

assets/js/workflow-diagram/nodes/Job.tsx
```diff
@@ -13,7 +14,10 @@ const JobNode = ({
   ...props
 }: NodeProps<NodeData>) => {
   const toolbar = () => [
-    props.data?.allowPlaceholder && [<PathButton key="+path" />],
+    props.data?.allowPlaceholder && [
+      <PlusButton key="+step" />,
```
josephjclark: @doc-han I wasn't expecting this to be reverted - was this deliberate?

doc-han: Yeah deliberate. putting it back in.

doc-han:
There's an issue where after `onConnectEnd`, the event `onClick` is always triggered on the panel. 

We currently use this onClick to reset selection to `null` when the user clicks an empty space on the Editor. 
Dealing with this using a flag.

josephjclark:
Next from Farhan:
- Fix node bounds (so that edge selection works properly)
- Maybe use semi tone on neighbouring edge highlighting
- Don't create node on illegal drop

doc-han:
**Note:** Positions aren't passed down in `workflow_params` for templates during `state-applied` event. 
We need that, to have manual layouts saves as templates to be working.

taylordowns2000:
> **Note:** Positions aren't passed down in `workflow_params` for templates during `state-applied` event. 
> 
> We need that, to have manual layouts saves as templates to be working.

Thanks for catching this! I've added it to the top of your todo list on the working canvas. (This is a higher priority bug than the bounce.)

josephjclark:
The minimap has a tendency to clip over nodes in certain layouts and certain screen sizes

image

I'm going to increase the view padding (so charts fill less space) which might help in some cases

doc-han:
Can't the minimap be a bit transparent till you hover on it? 
Example 😀

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -45,7 +45,7 @@ const Label: React.FC<LabelProps> = ({ children, hasErrors = false }) => {
   if (children && (children as any).length) {
     return (
       <p
-        className={`line-clamp-2 align-left text-m max-w-[275px] text-ellipsis overflow-hidden ${textColorClass}`}
+        className={`line-clamp-2 align-left text-m max-w-[120px] text-ellipsis overflow-hidden ${textColorClass}`}
```
doc-han, line 48: checking this out. looks a bit tricky. 
if I remember clearly. this value has to be inline with the value used for dagre calculations.

Hence if this has reduced. the value for width in dagre calculations can also be reduced a bit to make good use of the space if we want to. 
Screenshot 2025-06-24 at 12 35 09 PM


That would close up the space between the node with title "This is a sill...." and the one next to it

cc: @josephjclark

josephjclark: This is fine - I needed to reduce the label width. It's breaking on whitespace in your screen and then sliding the rest.

We can increase the label length a bit if we want to.

Meanwhile the dagre layout sizes and distancing are controlled elsewhere

PR #3246, opened 2025-05-28
Optimize run claim query performance with per-workflow limiting strategy

### LINKED ISSUE #3245: Optimize run claim query performance with per-workflow limiting strategy
The current run claiming query suffers from performance issues in high-load scenarios due to expensive window function operations applied to large datasets. This is especially problematic with singles workflows having 1000s of available runs, the query can become a bottleneck, leading to delayed run processing or claim messages being dropped by the worker.

We need to improve the performance of the claim query.

stuartc:
## Description

- Add configurable per-workflow limit to prevent query performance degradation
- Implement multi-step optimization: workflow limiting → window functions → concurrency enforcement
- Maintain API compatibility while achieving ~10x performance improvement
- Prevent workflow starvation through fair resource distribution
- Add PER_WORKFLOW_CLAIM_LIMIT config (default: 50)

~~No indexes have been added yet, testing the query against real databases shows it's an order of magnitude faster than the current query. So I'm tempted to add indexes in another PR.~~

Also added an index which in theory negates the expensive sorting operation at the end of the select part of the query.

Closes #3245

## Problem

The current run claiming query applies expensive window functions to the entire dataset of runs, which creates performance issues in scenarios where individual workflows have many available runs. This can lead to query timeouts and delayed run processing.

## Solution

This PR implements a performance optimization that pre-filters the dataset before applying window functions, while maintaining exactly the same claiming behavior and results.

> The `PER_WORKFLOW_CLAIM_LIMIT` _MUST_ be greater than the max concurrency of any workflow. It's currently defaulted to 50, which is way higher than any deployment I'm aware of.

### Implementation

The optimization uses a multi-step approach:

1. **Pre-filtering**: Limit runs per workflow (configurable, default: 50) before applying window functions
2. **Window processing**: Apply row numbering to the smaller, filtered dataset  
3. **Final filtering**: Select claimable runs using the same concurrency logic

This prevents performance degradation in scenarios where single workflows have large numbers of pending runs, while ensuring the query returns identical results to the previous implementation.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


taylordowns2000 (approved):
approved in that the limit of 50 works just fine for me and is a fair caveat. if the behaviour is the same and the performance is better, let's do it.

jyeshe (approved):
I like the promising list specially by saving a removal of the dynamic concurrency (sometimes by workflow other times by project).
It looks as a good strategy and I am looking forward to see this on commodity/limited cloud HW.

PR #3248, opened 2025-05-28
Icons in minimap

taylordowns2000:
joe, this feels like more trouble than it's worth. they jump because they're waiting for a initialization or something.

https://claude.ai/share/4e0fc865-9d66-4efe-b7ff-3d7e207454d1


josephjclark:
Well it was just an idea, and probably an improvement. But I'm not in a position to help fix it right now :(

Issue #3265, opened 2025-06-02
Secondary button on Inspector appears on page load


taylordowns2000:
![Image](https://github.com/user-attachments/assets/b5c40be4-64d6-4c50-ba5c-0e2bbe9efb25)

The secondary option button is appearing automatically on page load. This might be because the new button component uses `inline-block` which conflicts with `hidden`. (So says Claude?)

**Version number** The version of Lightning you have seen the bug on.

`v2.13.0-pre`

I have reproduced this locally on main:

- [x] Yes
- [ ] No

**To Reproduce** Steps to reproduce the behavior:

1. Go to the Inspector
2. Hard refresh
3. Note that the secondary buttons are open

PR #3259, opened 2025-05-30
Fix Paddings in Settings Page

elias-ba:
## Description

This PR fix a regression on the settings page where all paddings between sections were broken

Closes #3257 

## Validation steps

1. Go to the settings page
2. Note how paddings are back to normal

## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to know!):

- [ ] Code generation (copilot but not intellisense)
- [ ] Learning or fact checking
- [ ] Strategy / design
- [ ] Optimisation / refactoring
- [ ] Translation / spellchecking / doc gen
- [ ] Other
- [ ] I have not used AI

You can read more details in our [Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Pre-submission checklist

- [ ] I have performed a **self-review** of my code.
- [ ] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [ ] I have updated the **changelog**.
- [ ] I have ticked a box in "AI usage" in this PR

PR #3266, opened 2025-06-02
Keep secondary button from appearing on page load

taylordowns2000:
Closes a strange issue (https://github.com/OpenFn/lightning/issues/3265) where the secondary action button was _OPEN_ during page refreshes. (It should only open when you click the action dropdown.)

## Validation steps

1. Go to the Inspector
2. Hard refresh
3. Note that the secondary buttons are not open anymore, but that they work!

I can't figure out why this `style="display: none;"` works when `hidden` doesn't. And I'm not even entirely sure that the `new_input` standardization of the buttons was the culprit. So strange that it only happens on page load and goes back to what it should be when you click somewhere on the page. Makes me think there's some strange JS going on elsewhere.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -2998,10 +2998,11 @@ defmodule LightningWeb.WorkflowLive.Edit do
             phx-click="toggle_github_sync_modal"
             theme="secondary"
             class={[
-              "hidden absolute right-0 z-10 w-max",
+              "absolute right-0 z-10 w-max",
               if(@dropdown_position == :top, do: "bottom-9 mb-2"),
               if(@dropdown_position == :bottom, do: "top-9 mt-2")
             ]}
+            style="display: none;"
             disabled={@disabled}
             phx-hook="OpenSyncModalViaCtrlShiftS"
           >


doc-han (commented):
This issue was caused because `inline-block` was added to base button styling by the PR #3259
https://github.com/OpenFn/lightning/blob/2c2c4ddfe2bd07884f53a51bb400c8a8d799dbce/lib/lightning_web/components/new_inputs.ex#L188

Tailwind defines both styles for `inline-block` and `hidden` in the same `@layer` which is `Utilities`. Due to this, the order in-which they appear in classes doesn't really matter to the browser. what matters is which was defined last in their source files? that's why in the classes we have `"inline-block ... ... hidden"` but still `inline-block` has precedence. This is because in the final generated file from tailwind. `hidden` is declared before `inline-block` is defined.

Why does using the style attribute with `display: none` work? it takes precedence over layers. 

**My Question:** What did adding `inline-block` to the base button solve? I checked a bit on main and couldn't figure that out.

This PR resolves it but then it means this trick of using `style="display: none;"` will become the new norm.

doc-han:
Here
https://play.tailwindcss.com/Bbj129bVvz

In the example shared. `bg-amber-800` always takes precedence regardless of it's order in the class attribute. This is because in tailwinds final generated css. `bg-amber-800` appears below `bg-amber-100`. Because they're on the same @layer their order in the source is used rather than their order in the class attribute.

taylordowns2000:
this will be fixed by removing the `inline-block` style from the new base button

PR #3289, opened 2025-06-10
Do not validate edge js condition expression in lightning

### LINKED ISSUE #3028: Incorrect edge condition validation
## The Problem

When users write edge conditions in the app, there's a little bit of app-side validation

Check this out:

Image

Here's the validation function in the app:
```
  defp validate_condition_expression(changeset) do
    js_expr = get_field(changeset, :condition_expression)

    if String.match?(js_expr, ~r/(import\b|require\b|process\b|await\b|eval\b)/) do
      add_error(
        changeset,
        :condition_expression,
        "contains unacceptable words"
      )
    else
      changeset
      |> validate_length(:condition_label, max: 255)
      |> validate_length(:condition_expression, max: 255)
    end
  end
```

This short list of banned strings (not words, strings) will prevent users writing certain expressions.

## Should the app validate?

I'm not keen on this at all 😐️

The worker does its own validation of edge conditions. It has to before it can run. And the worker will impose the validations that it considers necessary.

I would let the worker validate, especially for deep stuff. And no-ones going to try and run an import or eval anyway (why would you??).

This frontend validation isn't even comprehensive. It won't check for syntax errors, or ensure you're inputting an expression (not a statement). We'd be better off using a code editor for that, like Monaco.

## A Caveat

One possible hiccup to my  flawless critique of this validation code:

I think, right now, when there's an error in an edge condition, we're not very good at reporting it. I think the runtime throws a decent error but I don't think we report this back to Lightning properly?

I need to investigate.

There's some notes around this here: https://github.com/OpenFn/kit/issues/531

midigofrank:
## Description

This PR removes the edge validation on js conditions, instead it adds a warning in the UI
Screenshot 2025-06-10 at 11 34 14

Closes #3028

## Validation steps

1. Open a workflow
2. Click on an edge connecting 2 jobs
3. Change the edge condition to `Matches a JS Expression`
4. In the JS Expression, type in `eval`. Notice that a warning circle appears. When you hover it, a tooltip appears with a message: `Your expression contains potentially unsafe functions (eval, require, import, process, await) that may cause your workflow to fail`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/components.ex
+++ b/lib/lightning_web/live/workflow_live/components.ex
@@ -554,7 +554,7 @@ defmodule LightningWeb.WorkflowLive.Components do
               name="hero-exclamation-circle-solid"
               class="size-5 text-yellow-600"
               phx-hook="Tooltip"
-              aria-label="Warning: Your expression contains potentially unsafe functions (eval, require, import, process, await) that may cause your workflow to fail"
+              aria-label="Warning: this expression appears to contain unsafe functions (eval, require, import, process, await) that may cause your workflow to fail"
             />
           </.label>
           <.input
@@ -598,7 +598,10 @@ defmodule LightningWeb.WorkflowLive.Components do
   end
 
   defp js_expression_safe?(js_expr) do
-    !String.match?(js_expr, ~r/(import\b|require\b|process\b|await\b|eval\b)/)
+    !String.match?(
+      js_expr,
+      ~r/(\bimport\b|\brequire\b|\bprocess\b|\bawait\b|\beval\b)/
+    )
   end
 
   slot :inner_block, required: true
--- a/lib/lightning_web/live/workflow_live/components.ex
+++ b/lib/lightning_web/live/workflow_live/components.ex
@@ -548,6 +548,14 @@ defmodule LightningWeb.WorkflowLive.Components do
         <div>
           <.label>
             JS Expression
+            <.icon
+              :if={!js_expression_safe?(@form[:condition_expression].value)}
+              id="edge-js-expression-unsafe-warning"
+              name="hero-exclamation-circle-solid"
+              class="size-5 text-yellow-600"
+              phx-hook="Tooltip"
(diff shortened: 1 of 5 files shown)


midigofrank:
@stuartc thoughts?

lib/lightning_web/live/workflow_live/components.ex
```diff
@@ -589,6 +597,10 @@ defmodule LightningWeb.WorkflowLive.Components do
     """
   end
 
+  defp js_expression_safe?(js_expr) do
+    !String.match?(js_expr, ~r/(import\b|require\b|process\b|await\b|eval\b)/)
```
josephjclark: What is the relevance of `\b` ?

lib/lightning_web/live/workflow_live/components.ex
```diff
@@ -548,6 +548,14 @@ defmodule LightningWeb.WorkflowLive.Components do
         <div>
           <.label>
             JS Expression
+            <.icon
+              :if={!js_expression_safe?(@form[:condition_expression].value)}
+              id="edge-js-expression-unsafe-warning"
+              name="hero-exclamation-circle-solid"
+              class="size-5 text-yellow-600"
+              phx-hook="Tooltip"
+              aria-label="Warning: Your expression contains potentially unsafe functions (eval, require, import, process, await) that may cause your workflow to fail"
```
josephjclark: I think I'd tweak the language to `Warning: this expression appears to contain unsafe functions (eval, require, import, process, await) that may cause your workflow to fail`

lib/lightning_web/live/workflow_live/components.ex
midigofrank: Good question. I just plucked it as it was from the previous logic. 
A quick search says it's to denote word boundary, it seems we need to have it before and after to have it exclude something like `retrieval`

josephjclark: Ah great, nice. I did a quick search and it looked like backspace. Very good, carry on!

PR #3298, opened 2025-06-10
Inspector Input panel not showing current run's dataclip for older runs

### LINKED ISSUE #3288: Input for existing run not loaded properly (sometimes)
When clicking the "Inspect" icon from the history page, we show users the Inspector view with 3 panels loaded:

1. Input (the input that was used for that run-step)
2. Editor (the job code)
3. Logs (the logs for that run—with the run-step highlighted)

Right now this is how it works in _most_ of our testing, but in at least one instance the input isn't being loaded, no matter how you navigate to the Inspector and no matter how many times you hard-refresh: 

`https://app.openfn.org/projects/d78b8cef-95c5-4f94-93e3-5fa28047dc59/w/51e36328-5cfb-4063-a2b2-4d11d312ac9d?m=expand&s=3cef6846-e9f4-41ff-967d-e29d76faab0b&a=55484460-0174-46cf-83f1-531a9429a65b#log`

Instead, the "Empty" tab is selected and retrying is not possible: ("An empty JSON object will be used as the input for this run.")

@aleksa-krolls and @AishaHassen may be able to provide additional information for recreation.

**Version number** `v2.13.0 (1.5.5)`

- [ ] Yes
- [x] No

1. Go to this URL: https://app.openfn.org/projects/d78b8cef-95c5-4f94-93e3-5fa28047dc59/w/51e36328-5cfb-4063-a2b2-4d11d312ac9d?m=expand&a=55484460-0174-46cf-83f1-531a9429a65b&s=3cef6846-e9f4-41ff-967d-e29d76faab0b#log
2. Note that the Input appears in the **right side** Input tab, but not in the **left side** Input panel

stuartc:
## Description

The Inspector's Input panel fails to load when viewing older workflow runs. Users can see input data in the right-side "Input" tab, but the left-side Input panel appears empty, preventing them from using that data to create new manual runs.

Fixes #3288 

**Root Cause:** The Input panel only searches through the 10 most recent dataclips for a job. When viewing older runs (typically runs 11+), their input data isn't included in this limited search, so the panel can't find and display it.

**Fix:** When viewing a specific run (indicated by the `a` parameter in the URL), the component now:
- Fetches the current run's input dataclip separately via a new `get-run-input-dataclip` event
- Displays it at the top of the dataclip list with a star icon to indicate it's the current run's input
- Automatically selects it if it matches the expected `selected_dataclip_id`
- Prevents duplication by filtering it out of the regular search results

## Validation Steps

1. **Test the original issue:**
   - Open Inspector for a workflow job with >10 runs
   - Navigate to run #11 or later (URL will have `a=<run-id>` parameter)
   - Verify the Input panel now shows the dataclip content with a star icon

2. **Test recent runs still work:**
   - Navigate to runs 1-10 for the same job
   - Verify the Input panel continues to work as before

3. **Test search functionality:**
   - In the Input panel's "Existing" tab, perform searches with filters
   - Verify search results are limited to 10 items and don't include the current run dataclip to avoid duplication

## Additional Notes

I've taken the liberty of changing our UX a bit here, there was no way to unselect and then reselect the current Run/Step input dataclip.

So the change here is that there is a star representing the current Run/Steps input, and it's "pinned" to the top.

image

Feedback welcome.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/manual-run-panel/ManualRunPanel.tsx
+++ b/assets/js/manual-run-panel/ManualRunPanel.tsx
@@ -27,14 +27,16 @@ export const ManualRunPanel: WithActionProps<ManualRunPanelProps> = props => {
     before,
     after,
     query: urlQuery,
-  } = useQuery(['active_panel', 'type', 'before', 'after', 'query']);
-  const { pushEvent, job_id, selected_dataclip_id, navigate } = props;
+    a: runId,
+  } = useQuery(['active_panel', 'type', 'before', 'after', 'query', 'a']);
+  const { pushEvent, pushEventTo, job_id, selected_dataclip_id, navigate } = props;
   const [selectedOption, setSelectedOption] = React.useState<SeletableOptions>(
     active_panel
       ? (Number(active_panel) as unknown as SeletableOptions)
       : SeletableOptions.EMPTY
   );
   const [recentclips, setRecentClips] = React.useState<Dataclip[]>([]);
+  const [currentRunDataclip, setCurrentRunDataclip] = React.useState<Dataclip | null>(null);
   const [query, setQuery] = React.useState(urlQuery ? urlQuery : '');
   const [selectedDataclip, setSelectedDataclip] =
     React.useState<Dataclip | null>();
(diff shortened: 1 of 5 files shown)


assets/js/manual-run-panel/ManualRunPanel.tsx
```diff
@@ -104,17 +106,17 @@ export const ManualRunPanel: WithActionProps<ManualRunPanelProps> = props => {
...
+      type: selectedClipType || undefined,
+      before: selectedDates.before || undefined,
+      after: selectedDates.after || undefined,
+    };
+
+    const cleanFilters = Object.fromEntries(
+      Object.entries(filters).filter(([_, value]) => value !== undefined)
+    ) as Record<string, string>;
+
+    return { query, filters: cleanFilters };
```
doc-han, line 131: ❤️ lovely

doc-han (approved):
sweet.

doc-han:
Hmmmm. was checked out here and mistakenly pushed a commit 🥲

taylordowns2000 (changes requested):
looks great. two small things:
1. when you create a new custom dataclip and run with it, after deselecting that dataclip it doesn't appear as the starred run on the list (works after refresh)
2. retry from here is available when a _different_ input from the run's input is selected. this shouldn't be possible. you're only allowed to create a new workorder in this case. (see image.) another way to think of this is, "if you're creating a new custom input, it's NEVER a retry, it's always a new work order."
3. (maybe related to number 2?) after you create a new work order with a custom input for a workflow when a run was previously loaded, it seems to throw out your custom input. (see second image.)

image

image

Issue #3294, opened 2025-06-10
Play Button: Add to Step editor


josephjclark:
Once we've got a top play button to run the whole workflow, we need to add  a button to each step and trigger

Sort of like this:

![Image](https://github.com/user-attachments/assets/0ca86a1e-775b-4142-813b-60fcbf857029)

But instead of just the icon I'd have a label like "Run from here"

From each step and trigger, I should be able to run the whole workflow from this point. Just like I can in the inspector.

Clicking the Run button should open the manual runs component - very similar to how the top Workflow run works.

Because the dialog will be react,  think we might need a bit of fancy footwork when navigating between the LiveView and React dialogs. This needs to be seamless for users.

doc-han:
This is done. by #3299

PR #3299, opened 2025-06-11
feat: put ManualRunPanel into canvas

### LINKED ISSUE #3290: Play button: add Run Workflow button to top panel
Add a Run Workflow button to the top navigation bar in the Workflow Diagram

Image

Clicking this should open a new React dialog, which looks _exactly_  the same as the current Live View step editor dialog. It contains the Manual Run UI and a button called Run Workflow (not sure about this text yet, will refine it later)

Clicking Run Workflow should navigate to the Run View.

The toolbar is going to be very cluttered after this, but don't worry about that for now.

doc-han:
## Description

Adds a Play button to the navbar of the WorkflowEditor view which allows users to run the entire workflow straight from the canvas
Screenshot 2025-06-11 at 8 28 02 AM

Closes #3290 and #3294

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


doc-han:
work in progress. trying to reuse the "create new order" button but currently, so much logic for `manual_run_form` is tied to this button. which makes it functional in JobEditor view.

josephjclark:
1. Note that the dialog is still in Liveview - Han hasn't created react component because there's quite a lot of management logic we need (open debate as to what we should do here - at this point I'm inclined to keep the panel in liveview in the interests of moving quickly)

1. Yes, the existing play button (Create New Work Order) has  lot of its logic tied up in elixirland. I think to get this work - and to futureproof for later - we need an API on the server to allow us to request a new workorder via POST. Like we just post a JSON object with whatever data we need to trigger a run. The API would return a 201 if the work order is added to the queue, and then the clientside app can re-navigate to the Run Viewer.

Let's discuss this in the morning - I think we should take @stuartc's guidance here

doc-han:
**Engineering sync 12-06-2025**
@stuartc  - with a case statement we can turn the entire panel into a react component. and lean towards @josephjclark's suggestion of opening up an endpoint for the "Create Work Order" button.

Hence, that entire panel will be react when this approach is achieved. ❤️

doc-han:
**update:** The entire panel is now in react. 
**way forward:** an endpoint where we can push that state of our `manual_run_form`
This endpoint would
1. run checks on whether the workflow can be run (deleted, ...)
2. create a run from the provided input and return redirection endpoint to the run page. 

cc: @midigofrank (around the endpoint)

doc-han:
@midigofrank thanks man. this is now working as expected. ❤️

doc-han:
**What's left:** Some few UI not fitting issues

josephjclark:
@doc-han there are conflicts against main which need resolving

Should we go ahead and add #3294 here? I don't think it's a lot more work, especially now that the backend API is done

doc-han:
> @doc-han there are conflicts against main which need resolving
> 
> Should we go ahead and add #3294 here? I don't think it's a lot more work, especially now that the backend API is done

done implementing... 
@josephjclark check the experience and let me know of any changes.

josephjclark:
TODO: communicate that hitting Run will also save

TODO: when manual runs panel is open, is it clear which step is actually going to run?

josephjclark:
This looks great @doc-han. Based on your own comments, I'd like to try one major UX change (maybe on a new branch?). I've also got some other, smaller feedback (although it's a long post, sorry!)

# Selection & Context

We've got two big UX problems here:

1. As you say, when the Run panel is open on a step, it's maybe not always super clear which step will run
2. When the Run panel is open and you click a note on the chart, the Run panel stays open and that node is not selected.

I suggest we do this:

1. Change the run panel title to include context (and move the `Select Input...` instruction down into the panel). I've mocked this up:
play_run-from-trigger
play-from-step
1. Clicking the top Run button goes to "Run from Trigger". Clicking the Run button at any time should select back to this state
1. Selecting a new node in the chart should update the Run panel to run from here
1. When the selection changes, keep whatever state is currently active. Don't reset the panel.
1. Add a "Back" icon (chevron) to the title. When a Trigger is selected, Back will close the panel. When a step is selected, Back will go back to the edit panel. 

# Unsaved Changes & Inspector Consistency

If you edit the workflow and hit Run, your changes will be saved forever - and we don't tell you.

This is a question raised by Taylor this morning.

The inspector has exactly the same behaviour though: if you edit your code and hit the run button, the app will secretly save changes before running. 

I propose we do nothing about this for now. We haven't had a complaint yet.

I'm more concerned that the Inspector says `Create New Work Order` and the Workflow Diagram says `Run Workflow Now`. Totally different language for _exactly_ the same behaviour. I think this is a landmine so I've asked Taylor about it. Suggest we do nothing for now.

# Selecting Edges

When the Run panel is open, and you select an Edge, the Panel seems to jump around the screen.

And if we do the new selection logic like above, what happens when the user clicks an Edge?

I suggest that if the Run panel is open, and an Edge is selected, we show an error message like "Select a Step or Trigger to start a Run from"

# Run Panel Clutter

Something about this bugs me:

play-icons

Using the Play icon is smart. But I don't like the rainbow of button  colours in this panel. I also don't like that Play is emphasised over the Inspector (and neither has a tooltip).

Can we make the Inspector icon blue and add a tooltip to both so that the weighting is even?

I think I'd also vote for `Edit`, `Run` and `Delete` wording on the buttons, rather than the icons.

# Custom Panel Responsiveness

Because the panel is quite narrow, the Custom panel doesn't look very good:

image

We just need to add some break logic so that the panel draws in a single column (I've just looked at the code and this  isn't as trivial as I first  assumed). We want to end up with something like this:

image

assets/js/panel/panels/WorkflowRunPanel.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface WorkflowRunPanel {
+  job_id: string
+  cancel_url: string
+}
+
+export const WorkflowRunPanel: WithActionProps<WorkflowRunPanel> = (props) => {
+  const { job_id, cancel_url, pushEvent, ...actionProps } = props;
+  const [manualContent, setManualRunContent] = React.useState<ManualRunBody>({ manual: { body: null, dataclip_id: null } })
+
+  const runDisabled = React.useMemo(() => {
```
josephjclark, line 26: I'm sure this can be simplified a lot? If we had a unit test framework here I'd push this a bit harder.

It looks like an important function and it's very hard for me to see the rules of when a run is disabled

josephjclark:
Having spoken with Taylor, here's the solution to `Inspector Consistency`:

- The Canvas always says Run or Run from here
- (Later, when we superimpose an old run on the canvas, we'll say `Run (retry)` and `Run (new Work Order)`)
- Change the Inspector so that `Create new Work Order` becomes `Run`
- Change the Inspector so that `Retry from Here` becomes `Run (retry)`, and the drop down option becomes `Run (New Work Order)`
- Change the tooltip below to say "Run this step..." rather than "Rerun". Because rerun isn't a verb we have.
image

josephjclark:
Maybe this instead of Back to get back to edit mode?
image

doc-han:
All green!

josephjclark:
Looking great @doc-han! I've left feedback on slack but the summary is:
- I was wrong about Icons, we need to move them to the final Start The Actual Run Step
- There's some flicker on the panel when going back
- A couple of label adjustments

I haven't deep tested the manual runs panel and all its options. I'm sort of assuming that Just Works. But I'll try and do some deeper testing on that tomorrow, just to be sure

josephjclark:
Raised https://github.com/OpenFn/lightning/issues/3349 for the flicker issue. It exists on prod too, let's do this elsewhere

doc-han:
Why shouldn't the back arrow on "Run from Trigger" work just like the back arrow when running from a job?

**"Run from Trigger" back arrow**
Closes the entire panel

**Job Run back arrow**
Goes back to the edit panel for the job. 

Note: "Run from Trigger" Panel has to action buttons on the title. 
<- (back arrow) and x (close)
they now both work the same way. I vote for making the back arrow behave just like that of the job run.

josephjclark:
I'm relaxed on the "Back" step.

If you click the Trigger, then hit Run, then back should take you back to edit. If you hit the top Run and back, it should probably close the dialog.

But I think this is a bit fussy and hard to do. I suggest we leave it as-is and wait and see if we get any feedback. It's not a big deal.

But one showstopper issue just came up in testing.

Click a step. Hit run. Filter to Existing. Click a UUID. And the editor will popup and scroll infinitely behind the panel :sob: :sob: :sob: 

`overflow-hidden` ?

josephjclark:
Less important, a possible regression. Here's the manual input panel on main:

image

But here it is on this branch, with the input box too long and the search button wrapped down:

image

This is because we've added more responsiveness generally I think - so is this by design?

doc-han:
> I'm relaxed on the "Back" step.
> 
> If you click the Trigger, then hit Run, then back should take you back to edit. If you hit the top Run and back, it should probably close the dialog.
> 
> But I think this is a bit fussy and hard to do. I suggest we leave it as-is and wait and see if we get any feedback. It's not a big deal.
> 
> But one showstopper issue just came up in testing.
> 
> Click a step. Hit run. Filter to Existing. Click a UUID. And the editor will popup and scroll infinitely behind the panel 😭 😭 😭
> 
> `overflow-hidden` ?

nice catch. did some style changes at the end. checking it out

doc-han:
resolved @josephjclark

josephjclark:
@stuartc @taylordowns2000 This is done and I'm happy to merge. @stuartc are the failing tests OK, do we need to take action? @taylordowns2000 would you like to take a look?

taylordowns2000 (approved):
looks great! also tests are all passing. just codecov complaining. please go ahead and merge

doc-han:
Hmmm, after hitting merge. I'm noticing there's no changelog on this.
@midigofrank @taylordowns2000

taylordowns2000:
well darn. yeah. this is not good.

image

i will write the changelog entry now on main, but please update the PR checklist now for posterity. thanks @doc-han !

doc-han:
> i will write the changelog entry now on main, but please update the PR checklist now for posterity. thanks @doc-han !

Thanks

Issue #3300, opened 2025-06-11
Workflow Diagram: node not selected on load


josephjclark:
* Open any workflow
* Click any step and notice the selection style on the workflow diagram
* Refresh the page
* The step editor should be open, but the node on the workflow is unstyled

Or:
* Select a node
* Copy the URL
* Paste the URL in a new window
* The step editor should be open, but the node on the workflow is unstyled
* 
Spotted this in dev today - can't believe I didn't see it earlier. I don't intend to fix today so raising it out as a standalone issue.

PR #3301, opened 2025-06-11
Free node positioning fixes

josephjclark:
This PR attempts to fix logic and behaviour of the free node positioning PR.

The base PR has a fundamental problem in how positions are loaded and tracked.  There are just too many `positions` coming from too many places.

The logic now is:

- If `positions` for a workflow is null, we auto layout
- If `positions` is set, then we manually layout
- The last known positions are always cached, so that we can tell if something changed

I'm also noticing that the selection logic keeps triggering race conditions which cause problems in the workflow. Since we've had a request recently to remove the neighbour selection behaviour, I've just cut this code out. At the time of writing some code still needs deleting (you'll see it in the diff). If we decide we actually want to keep this code I'll need to rethink how that stuff works.

This needs testing really thoroughly - but I think it's OK?

Note that there is an existing bug on selection #3300 . This exists in production too and no-one else has complained. Not going to fix it here.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -47,23 +47,30 @@ type ChartCache = {
 const LAYOUT_DURATION = 300;
 
 export default function WorkflowDiagram(props: WorkflowDiagramProps) {
-  const { jobs, triggers, edges, disabled, positions, updatePositions } =
-    useWorkflowStore();
   const { selection, onSelectionChange, containerEl: el } = props;
 
+  const {
+    jobs,
+    triggers,
+    edges,
+    disabled,
+    positions: fixedPositions,
+    updatePositions,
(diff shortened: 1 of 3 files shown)


doc-han (approved):
Nice. I think this PR clears the ambiguity around positions. which seems to have solved the bug.

**One more thing to deal with on the main PR.** 
When a workflow comes with position values from the backend(manual layout), whenever I use the toggle to go from manual -> auto. if I go back to manual. I expect the recent manual layout to be restored. not me starting from scratch to reposition the nodes.

doc-han:
@midigofrank @josephjclark 
**Note:** the + on nodes which is supposed to be available during auto-layout seems to not be available

midigofrank:
Hey @doc-han , the `+` isn't supposed to be present at all according to #2175

PR #3310, opened 2025-06-17
fix: free node issues

doc-han:
## Description

Closes #__

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/workflow-diagram/util/to-workflow.ts
+++ b/assets/js/workflow-diagram/util/to-workflow.ts
@@ -11,6 +11,7 @@ const model = (model: Flow.Model) => {
     triggers: [],
     jobs: [],
     edges: [],
+    positions: {},
   };
 
   model.nodes.forEach(node => {
@@ -27,6 +28,7 @@ const model = (model: Flow.Model) => {
       workflow.triggers.push(wfNode as Lightning.TriggerNode);
     } else {
       workflow.jobs.push(wfNode as Lightning.JobNode);
+      workflow.positions[node.id] = { x: node.position.x, y: node.position.y };
     }
   });
 
--- a/assets/js/workflow-store/WorkflowStore.tsx
+++ b/assets/js/workflow-store/WorkflowStore.tsx
@@ -1,13 +1,13 @@
 import type { WithActionProps } from '#/react/lib/with-props';
 import React from 'react';
+import { randomUUID } from '../common';
+import { DEFAULT_TEXT } from '../editor/Editor';
 import {
   useWorkflowStore,
   type ChangeArgs,
   type PendingAction,
-  type WorkflowProps,
+  type WorkflowProps
 } from './store';
-import { randomUUID } from '../common';
-import { DEFAULT_TEXT } from '../editor/Editor';
 
 const createNewWorkflow = (): Required<ChangeArgs> => {
   const triggers = [
@@ -38,6 +38,7 @@ const createNewWorkflow = (): Required<ChangeArgs> => {
 
 // This component renders nothing. it just serves as a store sync to the backend
 export const WorkflowStore: WithActionProps = props => {
+  const { pushEventTo, handleEvent, children } = props;
   const pendingChanges = React.useRef<PendingAction[]>([]);
   const {
     applyPatches,
@@ -47,6 +48,7 @@ export const WorkflowStore: WithActionProps = props => {
     subscribe,
     setDisabled,
     setForceFit,
+    reset
   } = useWorkflowStore();
 
   const pushPendingChange = React.useCallback(
@@ -54,7 +56,7 @@ export const WorkflowStore: WithActionProps = props => {
       return new Promise((resolve, reject) => {
         console.debug('pushing change', pendingChange);
         // How do we _undo_ the change if it fails?
-        props.pushEventTo('push-change', pendingChange, response => {
+        pushEventTo('push-change', pendingChange, response => {
           console.debug('push-change response', response);
           // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
           if (response && response.patches) applyPatches(response.patches);
@@ -62,7 +64,7 @@ export const WorkflowStore: WithActionProps = props => {
         });
       });
     },
-    [props, applyPatches]
+    [pushEventTo, applyPatches]
   );
 
   const processPendingChanges = React.useCallback(async () => {
@@ -81,33 +83,33 @@ export const WorkflowStore: WithActionProps = props => {
   }, [processPendingChanges, subscribe]);
 
   React.useEffect(() => {
-    props.handleEvent('patches-applied', (response: { patches: Patch[] }) => {
+    handleEvent('patches-applied', (response: { patches: Patch[] }) => {
       console.debug('patches-applied', response.patches);
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unnecessary-condition
       if (response.patches) applyPatches(response.patches);
     });
-  }, [applyPatches, props]);
+  }, [applyPatches, handleEvent]);
 
   React.useEffect(() => {
-    props.handleEvent('state-applied', (response: { state: WorkflowProps }) => {
+    handleEvent('state-applied', (response: { state: WorkflowProps }) => {
       console.log('state-applied', response.state);
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unnecessary-condition
       if (response.state) setState(response.state);
     });
-  }, [setState, props]);
+  }, [setState, handleEvent]);
 
   React.useEffect(() => {
-    props.handleEvent('navigate', (e: any) => {
+    handleEvent('navigate', (e: any) => {
       const id = new URL(window.location.href).searchParams.get('s');
 
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       if (e.patch) setSelection(id);
     });
     // force fitting
-    props.handleEvent('force-fit', () => {
+    handleEvent('force-fit', () => {
       setForceFit(true);
     });
-  }, [add, props.handleEvent, setSelection, setForceFit]);
+  }, [add, handleEvent, setSelection, setForceFit]);
 
   // Fetch initial state once on mount
   React.useEffect(() => {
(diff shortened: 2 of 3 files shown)


assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -85,39 +85,52 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+  // state to keep model & flow instance
+  // model - contains edges & nodes for the diagrammer
+  // flow - instance of the diagrammer
   const [model, setModel] = useState<Flow.Model>({ nodes: [], edges: [] });
+  const [flow, setFlow] = useState<ReactFlowInstance>();
 
   const workflowDiagramRef = useRef<HTMLDivElement>(null);
 
+  const forceLayout = useCallback(
+    (_model: Flow.Model, _flow: ReactFlowInstance) => {
```
josephjclark, line 100: Why are model and flow needed here?

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -85,39 +85,52 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+    },
+    [flow, model]
+  );
+
   const toggleAutoLayout = useCallback(() => {
     if (fixedPositions) {
       // set positions to null to enable auto layout
       updatePositions(null);
-      forceLayout();
+      forceLayout(model, flow);
```
josephjclark, line 120: Don't think this state is needed here?

assets/js/workflow-diagram/usePlaceholders.ts
```diff
@@ -79,20 +85,26 @@ export default (
...
       placeholders.nodes[0].data.name = name;
-      addTo(toWorkflow(placeholders));
+      placeholders.nodes[0].id = id; // hmmm
+
+      // commit this placeholder to the store
+      const modelWithPlaceholder = toWorkflow(placeholders);
+      addTo(modelWithPlaceholder);
+
+      // clear placeholder state after it's been commited to the store.
+      setPlaceholders(reset_data);
```
josephjclark, line 97: I would default this value inside set placeholders (wrap the function), then call it with no arguments

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -136,85 +149,41 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
     lastSelection: selection,
   });
 
-  const [flow, setFlow] = useState<ReactFlowInstance>();
+  const modelFromWorkflow = useCallback((placeholders: Flow.Model) => {
+    const _model = fromWorkflow(workflow, fixedPositions, placeholders, selection);
+    if (!fixedPositions) forceLayout(_model, flow);
```
josephjclark, line 154: Are we sure we want to force a layout here? This doesn't feel right . How are we able drop the layout cache stuff?

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -136,85 +149,41 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+    else setModel(_model)
```
josephjclark, line 155: I think you need to:
a) Have prettier installed and configured, because I keep getting silly formatting diffs
b) Always include braces `{}` around if clauses. The indenting makes code much easier to read.

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -136,85 +149,41 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
     lastSelection: selection,
   });
 
-  const [flow, setFlow] = useState<ReactFlowInstance>();
+  const modelFromWorkflow = useCallback((placeholders: Flow.Model) => {
+    const _model = fromWorkflow(workflow, fixedPositions, placeholders, selection);
```
josephjclark, line 153: Not keen on underscored variable names. I'd call it `flowModel` or `m`

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -136,85 +149,41 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
-    );
-    if (fixedPositions) {
-      if (layoutId) {
-        updatePositions(fixedPositions);
+    // when store is not initialized. don't create a model yet.
+    if (!init) return;
+    modelFromWorkflow()
+  }, [init, workflow, selection])
+
+  // function being passed to `usePlaceholders` can be stale because it's a useCallback.
```
josephjclark, line 164: I don't understand any of this stuff

We've removed `{ placeholders } = usePlaceholders()`, which use to trigger a re-render of the main effect.

And instead, we've pushed a callback into `usePlaceholders` to explicitly request a render through this backdoor. Which isn't really how react works, and so we've not got all this weird ref stuff to make it work.

Either you've over-engineered this or I've misunderstood it. Surely this will work:
```
  useEffect(() => {
     const _model = fromWorkflow(
         workflow,
         fixedPositions,
         placeholders,
         selection
      );
      if (!fixedPositions) forceLayout(_model, flow);
      else setModel(_model);
  }, [init, workflow, selection, placeholders]); // note placeholders is here, so it'll force a re-render

```

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -136,85 +149,41 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
-    // This will tell us if there's been a structural change
-    // in the model and force an updaet
-    const layoutId = shouldLayout(
-      newModel.edges,
-      chartCache.current.lastLayout
-    );
-    if (fixedPositions) {
-      if (layoutId) {
-        updatePositions(fixedPositions);
+    // when store is not initialized. don't create a model yet.
```
josephjclark, line 159: I don't think I understand why this is here. Can't we just initialise with an empty model?

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -85,39 +85,52 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+
   const toggleAutoLayout = useCallback(() => {
     if (fixedPositions) {
       // set positions to null to enable auto layout
       updatePositions(null);
-      forceLayout();
+      forceLayout(model, flow);
     } else {
       // fix positions to enable manual layout
       updatePositions(chartCache.current.positions);
```
josephjclark, line 123: If we continue down this road, you shouldn't use the chart cache for this. Just take the positions from the model and set them. Then you can remove chartcache completely

assets/js/workflow-diagram/util/to-workflow.ts
```diff
@@ -27,6 +28,7 @@ const model = (model: Flow.Model) => {
       workflow.triggers.push(wfNode as Lightning.TriggerNode);
     } else {
       workflow.jobs.push(wfNode as Lightning.JobNode);
+      workflow.positions[node.id] = { x: node.position.x, y: node.position.y };
```
josephjclark, line 31: Not `workflow.positions[node.id].position = node.position`?

josephjclark (changes requested):
I'm really not sure about this :thinking: I think this PR currently breaks more than it fixes. We're way off topic with this work and we're in serious danger of coding around in circles.

It simplifies some stuff yes, but only because it removes required behaviour. And it actually introduces some other quirks which I really don't love (the ref stuff)

Animation in auto layout is totally broken. I think you're working on a fix for that but I don't think it's going to be easy because you've dropped all the code that made it work.

The other problem is that layout is calculated every time that effect is run: on non-structural model edits, on selection changes, and on placeholder addition. This is wasteful, but actually it's worse than that because the dagre layouts aren't terribly stable, and bigger workflows tend to jump around, changing node order, as you click around.

I've left some smaller, pickier comments, some of them a bit speculative. But the animation and redundant layout issues are the two blockers for me right now.

One thing I will say is I _do_ think we can simplify that main useEffect. Somewhere between this PR and my scrappy PR I think there's a simpler, cleaner solution that's a bit less fussy about positions. But I'm not convinced it's worth the time to find it.

assets/js/workflow-diagram/util/to-workflow.ts
doc-han: after `to-workflow` function is used, it's being added to the store. didn't want it to contain `_default`

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: Yeah. we wouldn't. still several things to remove

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: Yes. when there's no position values(auto layout). we want forceLayout to provide position values to our nodes. 
Else, when there's position values. we go ahead and set model directly.

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: I'm planning to do another pass to correct the naming. didn't want to spend time thinking of variable names yet

josephjclark: Fair enough! I'll leave this comment unresolved for now

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: probably might be simplified soon. this was to make sure I know what model is passed to forceLayout.

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: didn't want the memoized way yet. explicitly passing them for the mean time.

assets/js/workflow-diagram/WorkflowDiagram.tsx
doc-han: Yeah we can. having an init, solves some issues that may appear on the templates side. 
On that end, we never leave the WorkflowEditor, we just hotswap the workflow information which creates so many bugs. 

Currently, we reset, which sets init to false. 
then during the swap we set init to true. 

we actually might not need the init state actually. will review that.

taylordowns2000:
closing in favor of https://github.com/OpenFn/lightning/pull/3242

PR #3319, opened 2025-06-18
AI Assistant For Workflow Edit

### LINKED ISSUE #3247: Edit existing workflows via AI chat
Currently, once a template is chosen via the "Create" button, the user is taken to the canvas and the chat panel disappears. From this point onwards, they're not able to modify their workflow using the AI chat.

But... the AI chat already takes the _current_ workflow.yaml and the chat history... in other words, it's set up to edit workflows already and we should extend this functionality to the canvas.

**But here's the catch:** Right now, the workflow chat agent does _not_ return job code. This is _probably_ a good thing, as we don't want a high level workflow agent to mess with a user's existing job code _until_ we have a nice code suggestion interface (coming next, like cursor) that allows the user to view each change and accept/reject it in the inspector.

So... for this issue, we'd want:

- [x] the chat panel to stay available (but be collapsible) on the left side of the canvas
- [x] the workflow chat to be able to modify YAML but _not_ modify job code.
- [x] to make sure that for existing jobs it _preserves_ job code. 
- [x] to make sure that if you went back to a previous message in the chat you could still view the workflow as it looked at that point in time. (Without saving.) maybe related to #2177 

Note that deleting a step entirely _will_ result in deleting job code, but that's OK as we'll have the previous version of the job code in the chat and/or in the snapshot history and/or in GitHub.

stuartc:
## Description

This PR extends the AI Assistant functionality to support workflow editing, building upon its existing workflow creation capabilities. Users can now access the AI Assistant at any stage of their workflow development process, from initial creation through ongoing modifications.

Closes #3247, #3442

### 1. Basic AI Panel Functionality
- Navigate to an existing workflow in edit mode
- Click the floating AI Assistant button (top-left corner with chat icon)
- Verify the AI panel slides in smoothly from the left side
- Confirm the workflow adjust it's placement in the canvas to avoid being hidden by the AI panel
- Click the close button (chevron icon) on the panel and verify it slides out smoothly

### 2. AI-Assisted Workflow Modifications
- With the AI panel open, type a workflow modification request (e.g., "Add a new job that transforms data after Step-1")
- Submit the request and verify the AI generates appropriate modifications
- Confirm the workflow canvas updates in real-time with the suggested changes
- Verify the new elements appear correctly on the canvas with proper connections
- Verify that new changes are not directly saved
- Verify that you can hit CTRL+S to save the modifications
- Verify that you can undo the AI changes using the UNDO / REDO buttons
- Add manually changes to the workflow and confirm that the AI modifications integrates well with your changes

### 3. URL Parameter Persistence
- Open the AI panel and note the URL changes to include `?method=ai`
- Refresh the page and verify the AI panel remains open
- Copy the URL, open in a new tab, and confirm the AI panel state is preserved
- Close the panel and verify the `method=ai` parameter is removed from the URL
- Open the other panels, see that they can coexist with the AI Assistant panel
- Open the inspector, close it, note that the AI assistant panel's state hasn't changed
- Open a session in the AI assistant, close the AI Assistant panel, reopen and see that your session is loaded by default
- Try as many combinations of panels open / closed

### 4. Workflow Creation with AI
- Click "New Workflow" from the workflows list
- Pick a template and save to see that you can still create workflows using templates
- Do the same with importing workflow YAMLs
- Click the Build With AI ✨ button and ask the AI to create a workflow for you
- Save your workflow and note that the AI Assistant stays loaded for your modifications
- Note that sessions are now associated to workflows.

### 5. Error Handling
- In the AI panel, request an invalid workflow modification (e.g., "Create a job with invalid syntax")
- Verify error messages appear in the AI chat with helpful feedback
- Confirm the workflow canvas doesn't update with invalid changes
- Check that the workflow remains in a valid state

### 6. Multi-User Scenarios
- Open the same workflow in two different browser sessions
- Make AI-assisted changes in one session
- Verify the presence indicators update correctly
- Confirm read-only mode activates appropriately when another user is editing

### 7. Integration with Existing Features
- With AI panel open, try manual workflow editing on the canvas
- Add a job manually and verify the AI assistant remains functional
- Test switching between job inspector and AI panel
- Verify save functionality works correctly with AI-generated changes

### 8. Edge Cases
- Test AI panel with workflows in snapshot (read-only) mode
- Verify AI panel is disabled appropriately for users without edit permissions
- Test behavior when switching between different workflows with AI panel open
- Confirm proper cleanup when navigating away from the workflow editor

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/ai_assistant.ex
+++ b/lib/lightning/ai_assistant/ai_assistant.ex
@@ -99,7 +99,6 @@ defmodule Lightning.AiAssistant do
 
     case read_at && DateTime.from_unix(read_at) do
       {:ok, datetime} ->
-        # Disclaimer is valid for 24 hours
         DateTime.diff(DateTime.utc_now(), datetime, :hour) < 24
 
       _error ->
--- a/lib/lightning_web/live/ai_assistant/modes/job_code.ex
+++ b/lib/lightning_web/live/ai_assistant/modes/job_code.ex
@@ -122,13 +122,11 @@ defmodule LightningWeb.Live.AiAssistant.Modes.JobCode do
       ) do
     options = extract_form_options(changeset)
 
-    # Preserve existing meta and add message options
     updated_meta =
       session.meta
       |> Kernel.||(%{})
       |> Map.put("message_options", Enum.into(options, %{}))
 
-    # Add follow_run_id if follow_run exists in assigns
     updated_meta =
       case assigns do
         %{follow_run: %{id: run_id}} ->
--- a/lib/lightning_web/live/ai_assistant/quotes.ex
+++ b/lib/lightning_web/live/ai_assistant/quotes.ex
@@ -119,12 +119,6 @@ defmodule LightningWeb.AiAssistant.Quotes do
   ]
 
   @doc """
-  Returns all quotes.
-  """
-  @spec all() :: [quote()]
-  def all, do: @quotes
-
-  @doc """
   Returns enabled quotes only.
   """
   @spec enabled() :: [quote()]
@@ -142,37 +136,4 @@ defmodule LightningWeb.AiAssistant.Quotes do
     enabled()
     |> Enum.random()
   end
-
-  @doc """
-  Returns a random enabled quote with error handling.
-  """
-  @spec safe_random_enabled() :: {:ok, quote()} | {:error, :no_quotes_available}
-  def safe_random_enabled do
-    case enabled() do
-      [] -> {:error, :no_quotes_available}
-      quotes -> {:ok, Enum.random(quotes)}
-    end
-  end
-
-  @doc """
-  Returns quotes by author.
-  """
-  @spec by_author(String.t()) :: [quote()]
-  def by_author(author_name) do
-    Enum.filter(@quotes, fn quote ->
-      quote.author == author_name
-    end)
-  end
-
-  @doc """
-  Counts total quotes.
-  """
-  @spec count() :: non_neg_integer()
-  def count, do: length(@quotes)
-
-  @doc """
-  Counts enabled quotes.
-  """
-  @spec enabled_count() :: non_neg_integer()
-  def enabled_count, do: length(enabled())
 end


assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -1,10 +1,9 @@
-
 import type { WithActionProps } from '#/react/lib/with-props';
 import WorkflowDiagram from '../workflow-diagram/WorkflowDiagram';
 import { useWorkflowStore } from '../workflow-store/store';
 
-export const WorkflowEditor: WithActionProps<{ selection: string }> = (props) => {
-  const { getItem, forceFit } = useWorkflowStore();
```
doc-han: selection prop shouldn't be removed. we need it.

assets/js/workflow-store/WorkflowStore.tsx
```diff
@@ -154,10 +156,15 @@ export const WorkflowStore: WithActionProps = props => {
   }, [props.pushEventTo, setState, add]);
 
   React.useEffect(() => {
-    return props.handleEvent('set-disabled', (response: { disabled: boolean }) => {
+    props.handleEvent('set-disabled', (response: { disabled: boolean }) => {
```
doc-han: The `return` shouldn't be removed. 
Whenever you use `props.handleEvent` it now returns its cleanup function. and we need to return that in the `useEffect` to be sure we get a proper cleanup of the handler. 

Hence if you have two `props.handleEvent` in one useEffect. then you'll want to keep the cleanup functions in variables to call them later. 

eg. 

```
React.useEffect(() =>  {
	const hanCleanup = props.handleEvent...
	
	const eliasCleanup = props.handleEvent...
	
	return () => {
		hanCleanup();
		eliasCleanup();
	}
}, [])
```

doc-han (changes requested):
Just a few things that got changed during the merge conflict resolution.
Please request a review from me once it's done. I'll like to take a closer look later on. 

Quickly tested undo/redo when building with the AI. there seems to be issues with it when clicking undo after the AI has painted its workflow on the canvas. edges disappear and some new nodes show up at times.

elias-ba:
@doc-han thanks a lot for this review. This is super helpful, and I would love you to continue dissecting this (if you have time).

Could you help me understand what you mean here:
> Quickly tested undo/redo when building with the AI. there seems to be issues with it when clicking undo after the AI has painted its workflow on the canvas. edges disappear and some new nodes show up at times.

Ain't these nodes appearing / disappearing because the AI added / removed them so Undo / Redo would do the inverse ? Isn't this how it should work ?

assets/js/workflow-store/WorkflowStore.tsx
```diff
@@ -131,10 +151,11 @@ export const WorkflowStore: WithActionProps = props => {
...
         setState(workflow_params);
-        if (!workflow_params.triggers.length && !workflow_params.jobs.length) {
-          const diff = createNewWorkflow();
-          add(diff);
-        }
+        // if (!workflow_params.triggers.length && !workflow_params.jobs.length) {
+        //   const diff = createNewWorkflow();
+        //   reset();
+        //   setState({ jobs: diff.jobs, edges: diff.edges, triggers: diff.triggers })
+        // }
```
doc-han: are we taking out the `createNewWorkflow()` code? that's how we initialize the canvas with our init workflow

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -1,48 +1,60 @@
...
+const onSelectionChange = (id?: string) => {
+  console.debug('onSelectionChange', id);
+
+  const currentUrl = new URL(window.location.href);
+  const nextUrl = new URL(currentUrl);
+
+  const idExists = getItem(id);
+  if (!idExists) {
+    nextUrl.searchParams.delete('s');
+    // Don't delete 'm' here - let the mode be managed separately
```
doc-han: Removing the query param `m` here will have a side-effect on user experience. We'll need to verify whether the new experience is accepted. I know of one scenario that will be affected (there might be more we're not aware of). 

1. When on the canvas, select a node - the step editor panel shows up
2. Click the Run button to show the dataclips panel
3. Click an empty space on the canvas to close the panel
4. Click back on any node to select it
5. **The dataclip panel shows up. Before this branch, we take you back to the step editor panel**

Also, for approx. all panels, clicking an empty space on the canvas should close the panel. This branch removes that. 

eg. 
1. Click on the settings button in the navbar next to save to open settings panel. 
2. Click an empty space on the canvas
3. On main, the settings panel closes but on this branch it stays open. 

Hence, removing `m` means all panels can no more be closed by clicking an empty space on the canvas

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -1,48 +1,60 @@
...
-      if (!id) {
-        console.debug('Unselecting');
-
-        nextUrl.searchParams.delete('s');
+      console.debug('Selecting', id);
+      nextUrl.searchParams.set('s', id);
+      // When selecting something, check if we're in settings or code mode
+      // If so, remove the mode (since they can't coexist)
+      const currentMode = nextUrl.searchParams.get('m');
+      if (currentMode === 'settings' || currentMode === 'code') {
```
doc-han: I think currently, most of this mode logic is fully handled in elixir. We even do this kind of mode removals there too. 
From how I think about it. elixir has full control of query params. currently, in the react world we only `push_patch` and let elixir manage the params. 

A function called `handle_selection_with_mode` in edit.ex

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -3002,231 +3257,230 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+  defp run_workflow_button(assigns) do
+    params = [
+      Helpers.param("s", assigns.trigger_id),
+      Helpers.param("m", "workflow_input"),
+      Helpers.param("a", fn a, _ -> a.query_params["a"] end),
+      Helpers.param("v", fn a, _ -> a.query_params["v"] end),
+      Helpers.param("w-chat", fn a, _ -> a.workflow_chat_session_id end),
+      Helpers.param("j-chat", fn a, _ -> a.job_chat_session_id end),
+      Helpers.param("method", fn a, _ -> a.query_params["method"] end)
+    ]
```
doc-han: Interesting. it seems every event or function that has something to do with params now has this kind of params mapping. I don't think this looks good.

assets/js/workflow-store/store.ts
```diff
@@ -38,6 +38,10 @@ export type WorkflowProps = {
 export interface WorkflowState extends WorkflowProps {
   forceFit: boolean;
   setForceFit: (v: boolean) => void;
+  showAiAssistant: boolean;
+  setShowAiAssistant: (v: boolean) => void;
+  aiAssistantId: string;
+  setAiAssistantId: (v: string) => void;
```
doc-han: We need to strongly define what goes through the state and also what doesn't. I don't think these values should go through the store. Also, I don't see the setters being used anywhere. I think it's because that is being managed entirely in elixir. Hence I think it'll be nice to leave it in elixir and just pass down the values as a react prop.

We can actually pass these values from elixir to react via props. That's what we currently do with `selection`. Selection passes through props to reach the canvas. 

https://github.com/OpenFn/lightning/blob/4d56a6f2e3280fe68d9f0bf745823be5d3340d10/lib/lightning_web/live/workflow_live/edit.ex#L407-L415

doc-han:
Creating a workflow from a copied workflow YML doesn't work anymore. 
1. Go to a workflow
2. Open setting and copy the workflow YML
3. Go to create a new workflow with this YML
4. It doesn't work anymore

Possible reason from @elias-ba is that, because we're keeping the ids for jobs and triggers on YML now, Ecto is rejecting the ids since they have to be unique.

josephjclark:
We need to make the chat icon the same collapsible panel to match the Workflow History

image

That probably means we need to merge the history overlay first

elias-ba:
> Creating a workflow from a copied workflow YML doesn't work anymore.
> 
> 1. Go to a workflow
> 2. Open setting and copy the workflow YML
> 3. Go to create a new workflow with this YML
> 4. It doesn't work anymore
> 
> Possible reason from @elias-ba is that, because we're keeping the ids for jobs and triggers on YML now, Ecto is rejecting the ids since they have to be unique.

This is fixed now, thanks

assets/js/workflow-store/WorkflowStore.tsx
```diff
@@ -48,7 +48,9 @@ export const WorkflowStore: WithActionProps = props => {
     subscribe,
     setDisabled,
     setForceFit,
-    reset
+    setShowAiAssistant,
+    setAiAssistantId,
```
doc-han: code to be deleted I guess

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -224,6 +227,79 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+            const bounds = getNodesBounds(model.nodes);
+            void flow.fitBounds(bounds, {
+              duration: FIT_DURATION,
+              padding: FIT_PADDING,
+            });
+          }, 510);
+        }
+      }
+    }, 50);
+
```
doc-han, line 282: Quite interesting that we're doing this to move the canvas controls with the open drawer. 

I seem lost. I finding it hard to wrap my head around the timeouts.
1. `50ms` timeout.
2. `510ms` when fitting view

Could we have changed the HTML structure to get the controls to move with the chat panel instead of relying on JS and resizeObserver?

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -224,6 +227,79 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
+        observer.disconnect();
+      }
+    };
+  }, [props.showAiAssistant, props.aiAssistantId, flow, model.nodes]);
+
+  useEffect(() => {
+    if (props.forceFit && flow && model.nodes.length > 0) {
+      // Immediately fit to bounds when forceFit becomes true
+      const bounds = getNodesBounds(model.nodes);
+      flow.fitBounds(bounds, {
```
doc-han, line 295: We're adding 3 new calls to `flow.fitBounds` on lines 238, 273 and 294. Makes it very tricky. I think we can find a way to make it single call instead.

**Bug:** The canvas does a lot of movement when + button is clicked to add a new node cc: @elias-ba

elias-ba:
> Hey when running locally, I typed on message and got 4 replies. Has this happened to anyone else? I'm on `eaaa0d253dc9917a18388ec0587549bfca81007d` here...
> 
> Image
> 
> Then I asked again, and got a purple box (as if it thought it replied with a workflow) but nothing changed on the canvas...
> 
> Image
> 
> Then when I asked why I can't see anything, it gave me this...
> 
> Image
> 
> **This feels like either there's a setup step I'm missing (did something change in how we run this locally?) or a regression in Apollo. Any ideas?**

Hey @taylordowns2000, whenever you are free, please give this another test. It should be fixed now. Let me know if you bump into any other weirdness.

elias-ba:
@midigofrank wanna look at the Oban implementation here ? Look for this

```elixir
defmodule Lightning.AiAssistant.MessageProcessor do
...
end
```

midigofrank (changes requested):
Hey @elias-ba , the Oban Job for querying Apollo is a brilliant idea. I have 2 concerns:

1. What if we want to send the unsaved job body?
2. I see the concurrency for that queue is `1`. Do we want to stick with that? This will be a bottleneck  incase multiple users are trying to chat with the assistant. Should we maybe set it to `10`?

elias-ba:
> Everything looks smooth, nice work. A regression on the Run button seems to have crept in: the "Run" action when no nodes are selected seems to be broken. It's now opening the trigger node, rather than opening the "Run from trigger" node.
> 
> ## Current (on branch)
> Image
> ## Desired (on production)
> Image

Hey @taylordowns2000 thanks a lot for your review and for this catch. It's now fixed and I compared with prod and everything looks fine. Would you mind checking again ? Thanks a lot

elias-ba:
> Hey @elias-ba , the Oban Job for querying Apollo is a brilliant idea. I have 2 concerns:
> 
> 1. What if we want to send the unsaved job body?
> 2. I see the concurrency for that queue is `1`. Do we want to stick with that? This will be a bottleneck  incase multiple users are trying to chat with the assistant. Should we maybe set it to `10`?

Hey @midigofrank thanks for your review:
1. This seems to be a feature request ? Right ? I did read the code in main quickly but we are accessing the job code thru the selected_job variable in workflow edit. Which makes me think that we only send saved job body ? am I wrong or misunderstanding here ?
2. Thanks for this catch, I raised the queues for the AI Assistant to 10.

taylordowns2000 (approved):

midigofrank:
> This seems to be a feature request ? Right ? I did read the code in main quickly but we are accessing the job code thru the selected_job variable in workflow edit. Which makes me think that we only send saved job body ? am I wrong or misunderstanding here ?

Hey @elias-ba , yeah it was more of a future question. I'm happy with maintaining it as it was

midigofrank (approved):

elias-ba:
> > This seems to be a feature request ? Right ? I did read the code in main quickly but we are accessing the job code thru the selected_job variable in workflow edit. Which makes me think that we only send saved job body ? am I wrong or misunderstanding here ?
> 
> Hey @elias-ba , yeah it was more of a future question. I'm happy with maintaining it as it was

Thanks @midigofrank, that's a brilliant idea, I have btw opened [this issue](https://github.com/OpenFn/lightning/issues/3471) so that we can track it

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -224,6 +227,80 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
     }
   }, [workflow, flow, placeholders, el, isManualLayout, fixedPositions, selection]);
 
+  // This effect only runs when AI assistant visibility changes, not on every selection change
+  useEffect(() => {
+    if (!props.showAiAssistant) {
+      setDrawerWidth(0);
+      
+      // Fit view when AI assistant panel closes
```
josephjclark, line 235: This should probably be a recenter rather than a fit. If the user is zoomed in then the view will reset and they'll lose context.

We could use a heuristic like "fit to all nodes that are currently visible". This is a trick we use elsewhere and will probably work well here

assets/js/hooks/index.ts
```diff
@@ -682,21 +682,52 @@ export const BlurDataclipEditor = {
...
+      this.scrollToBottom();
+    }
   },
-  scrollToLastElement() {
-    this.el.lastElementChild &&
-      this.el.lastElementChild.scrollIntoView({
-        behavior: 'smooth',
-        block: 'start',
+  
+  scrollToSpecificMessage(messageId: string) {
```
doc-han, line 704: Just thinking out loud. letting this function take the selector instead will make it more generic. that's if we'll want to reuse this same scrolling.

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -35,6 +35,8 @@ type WorkflowDiagramProps = {
   selection: string | null;
   onSelectionChange: (id: string | null) => void;
   forceFit?: boolean;
+  showAiAssistant?: boolean;
+  aiAssistantId?: string;
```
doc-han, line 39: If `aiAssistantId` is only set when `showAiAssistant` is true then we can drop one of the props and use just `aiAssistantId` and then derive show from it.

PR #3338, opened 2025-06-23
Fix snapshot cleanup cascade deletion

### LINKED ISSUE #3313: Data Retention Job Incorrectly Deletes Active Runs via Cascade Deletion of Snapshots
Sentry Issue: [LIGHTNING-RD](https://openfn.sentry.io/issues/6152001045/?referrer=github_integration)

### Problem Summary
The data retention cleanup job (`delete_unused_snapshots`) is incorrectly deleting workflow snapshots that are still referenced by active runs and steps, causing a cascade deletion that removes the runs/steps from the database. This leads to 500 errors when users try to access these runs via deep links.

#### The Issue
The `delete_unused_snapshots` function in `lib/lightning/projects.ex` only checks if snapshots are referenced by:
1. Current workflows (matching lock_version)
2. Work orders

**However, it fails to check if snapshots are referenced by:**
- **Runs** (`runs.snapshot_id`)  
- **Steps** (`steps.snapshot_id`)

#### Database Schema Impact
The database has `ON DELETE CASCADE` constraints:
```sql
TABLE "runs" CONSTRAINT "runs_snapshot_id_fkey" 
  FOREIGN KEY (snapshot_id) REFERENCES workflow_snapshots(id) ON DELETE CASCADE

TABLE "steps" CONSTRAINT "steps_snapshot_id_fkey" 
  FOREIGN KEY (snapshot_id) REFERENCES workflow_snapshots(id) ON DELETE CASCADE
```

**Result**: When a "unused" snapshot is deleted, all runs and steps referencing it are automatically deleted by the database.

#### User Experience
- User receives deep link to inspect a run (e.g., `?a=d6550a80-4163-46cf-b0b9-b9976e4db1ed`)
- Data retention job runs and deletes the snapshot referenced by that run
- Run gets cascade-deleted from database
- User clicks link → `Lightning.Runs.get(run_id)` returns `nil`
- Code tries to call `Lightning.Runs.Events.subscribe(nil)` → **500 FunctionClauseError**

### Current Status
**Temporary Fix Applied**: [PR #3318](https://github.com/OpenFn/lightning/pull/3318) disabled the `delete_unused_snapshots` call to stop the cascade deletions.

### Proposed Solution

1. Start with a failing test for runs and steps that are deleted following an unused snapshot purge.
2. Remove or change the cascade deletes.
3. We need to fix the `delete_unused_snapshots` function to also check for runs and steps before considering a snapshot "unused":

```elixir
# Add these something like this for additional checks to the unused_snapshots_query:
where:
  not exists(
    from(r in Run,
      where: r.snapshot_id == parent_as(:snapshot).id,
      select: 1
    )
  ),
where:
  not exists(
    from(s in Step, 
      where: s.snapshot_id == parent_as(:snapshot).id,
      select: 1
    )
  )
```

### Additional Considerations
1. **Defensive Programming**: The WorkflowLive.Edit should also handle missing runs gracefully
2. **Cascade vs Restrict**: Consider changing `ON DELETE CASCADE` to `ON DELETE RESTRICT` to prevent accidental data loss

stuartc:
## Description

- Prevent cascade deletion of runs/steps when snapshots are cleaned up
- Add FK constraint migration to change CASCADE to RESTRICT
- Fix range warnings in data retention batch processing with Stream approach
- Consolidate redundant dataclip deletion tests into comprehensive test

Closes #3313

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/workflows/query.ex
+++ b/lib/lightning/workflows/query.ex
@@ -8,11 +8,11 @@ defmodule Lightning.Workflows.Query do
   alias Lightning.Invocation.Step
   alias Lightning.Projects.Project
   alias Lightning.Run
-  alias Lightning.WorkOrder
   alias Lightning.Workflows.Edge
   alias Lightning.Workflows.Job
   alias Lightning.Workflows.Snapshot
   alias Lightning.Workflows.Workflow
+  alias Lightning.WorkOrder
 
   @doc """
   Returns all jobs accessible to a user, via their projects
--- a/test/lightning/workflows/query_test.exs
+++ b/test/lightning/workflows/query_test.exs
@@ -141,10 +141,25 @@ defmodule Lightning.Workflows.QueryTest do
       dataclip = insert(:dataclip, project: workflow.project)
       job = insert(:job, workflow: workflow)
 
-      insert(:workorder, workflow: workflow, snapshot: old_referenced_by_workorder, dataclip: dataclip)
+      insert(:workorder,
+        workflow: workflow,
+        snapshot: old_referenced_by_workorder,
+        dataclip: dataclip
+      )
 
-      work_order_for_run = insert(:workorder, workflow: workflow, snapshot: old_referenced_by_run, dataclip: dataclip)
-      insert(:run, work_order: work_order_for_run, snapshot: old_referenced_by_run, dataclip: dataclip, starting_job: job)
+      work_order_for_run =
+        insert(:workorder,
+          workflow: workflow,
+          snapshot: old_referenced_by_run,
+          dataclip: dataclip
+        )
+
+      insert(:run,
+        work_order: work_order_for_run,
+        snapshot: old_referenced_by_run,
+        dataclip: dataclip,
+        starting_job: job
+      )
 
       insert(:step, snapshot: old_referenced_by_step)
 


stuartc:
@taylordowns2000 this crossed my mind, thanks for raising it. I'll do an explain analyze tomorrow am and see how the select performs with and without the extra where clauses. 

~~I'll change this to a draft PR until I have some findings.~~

Ergh, don't think I can change this to a draft after it's been a grown up PR.

cc/ @midigofrank, you can review this at your leisure, but lets hold off on merging until I've verified the performance impacts of the updated query.

lib/lightning/projects.ex
```diff
@@ -1000,67 +1002,81 @@ defmodule Lightning.Projects do
...
+      )
+
+    total_deleted =
+      Stream.repeatedly(fn ->
+        unused_ids =
+          unused_snapshots_query |> limit(^batch_size) |> Repo.all()
+
+        if unused_ids == [], do: nil, else: unused_ids
+      end)
+      |> Stream.take_while(& &1)
```
midigofrank, line 1037: 🤩

midigofrank (approved):
Nicely done. I love how you've used `Stream. repeatedly/1 |> Stream.take_while/1`

lib/lightning/projects.ex
stuartc: ❤️ yeah it's pretty nifty!

I'm particularly fond of: 

```elixir
Stream.iterate(0, &(&1 + 1))
|> Stream.take(ceil(dataclips_count / batch_size))
```

The previous implementation did this:

```elixir
for _i <- 1..ceil(workorders_count / batch_size) do
   ...
end
```

I was getting warnings sometimes because the ceiling could end up being `0`, which means the range would be `1..0`, and that would be a "reverse range" which I think you need to tell it that beforehand. Another issue is that even a range of `0..0` will still iterate once (turned to a list it's `[0]`). So this is now a fancy way of going:

```elixir
if count > 0 do
  for _i <- 1..count do
    ...
```

stuartc:
## Future Optimization Option: TABLESAMPLE Approach for Snapshot Cleanup

During performance testing of the `unused_snapshots` query, I discovered that PostgreSQL's optimizer was making some really poor decisions, leading to 3.7+ second execution times. The main issue was nested loop anti-joins causing the steps table to be scanned 58+ times.

I found that using `TABLESAMPLE` as an "optimization fence" forces PostgreSQL into a much better execution plan. Here's the optimized query that achieved **37x performance improvement** (3.7s → 412ms):

```sql
WITH candidate_snapshots AS (
  SELECT ws.id, ws.workflow_id, ws.lock_version
  FROM workflow_snapshots ws
  INNER JOIN workflows w ON w.id = ws.workflow_id
  WHERE w.project_id = $1
    AND ws.lock_version <> w.lock_version
  LIMIT 1000
),
step_sample AS (
  SELECT DISTINCT snapshot_id
  FROM steps TABLESAMPLE SYSTEM(100)  -- 100% sample but forces different execution path
  WHERE snapshot_id IS NOT NULL
)
SELECT cs.id
FROM candidate_snapshots cs
WHERE NOT EXISTS (SELECT 1 FROM work_orders wo WHERE wo.snapshot_id = cs.id)
AND NOT EXISTS (SELECT 1 FROM runs r WHERE r.snapshot_id = cs.id)
AND cs.id NOT IN (SELECT snapshot_id FROM step_sample)
LIMIT $2
```

**How it works:**
- `TABLESAMPLE SYSTEM(100)` returns 100% of rows but prevents PostgreSQL from rewriting the query
- This forces a hash-based approach instead of nested loops
- Changes complexity from O(M × N) to O(M + N) where M = steps, N = snapshots

**Trade-offs:**
- I'm not 100% sure we can write this in Ecto, so using raw SQL makes this less maintainable
- Depends on TABLESAMPLE behaviour remaining consistent across PostgreSQL versions
- May not scale well if steps table grows beyond ~10M rows

**When to consider:**
- If snapshot cleanup becomes a performance bottleneck again
- For projects with large numbers of steps and snapshots

Worth keeping in our back pocket for when we need that extra performance boost.

stuartc:
@taylordowns2000 so I've done some performance tests, worst case scenario is around 7s per query. I've added two indexes that should take that down to 3.5s. But while doing testing I found that the query planner wasn't using the indexes for steps no matter what I did; but it used the one for runs. Anyway, see the comment above for a "when in need" path we can take.

taylordowns2000 (approved):

PR #3343, opened 2025-06-24
Visible/usable "next" cron input

### LINKED ISSUE #3335: Cron triggered runs run from final state
As a part of the requirements for [better manual runs for all trigger types](https://github.com/OpenFn/lightning/issues/1880), we were to make available an experience where (cron-triggered only) run from the last final state.

taylordowns2000:
## Description

Allow users to see the next input state that will be used for a cron-triggered workflow to close #3335 

## Validation steps

1. Create a new workflow from the event template, go to the inspector for the first step, and note that "Empty" is the default.
2. Write some code that returns something to state: `fn(state => ({something: true}));`
3. Run it.
4. Quit the inspector.
5. Go back to the Workflows page to _unload_ the run.
6. Go back to the workflow.
7. Go back to the inspector for the first step.
8. Note that you now see the step_result for the first run as your default input state.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


doc-han:
Review 01

1. our new cron result dataclip is always selected by default even when I'm not coming from the history page with the inspector icon. 
2. Does it really contain the result of the entire workflow run? or the input to the last step? I think it's the input to last step not the result

cc: @taylordowns2000

taylordowns2000:
Hey @doc-han ! Thanks for the review. To your Qs:

> 1. our new cron result dataclip is always selected by default even when I'm not coming from the history page with the inspector icon.

Yebo yes! The requirement is for the _default_ input dataclip when generating a new work order (coming from nothing) to be **empty** for webhook and kafka triggered jobs, and to be **next cron state** for cron triggered jobs. Note that this will be the same for your "run button" on the canvas—the default behaviour when you click play on a cron triggered workflow is to "pick up where you last left off".

> 2. Does it really contain the result of the entire workflow run? or the input to the last step? I think it's the input to last step not the result

No, it contains the _**output**_ of the last run of the _FIRST_ step. I.e., of _that_ step. (Note that there isn't always a single _last step_... there could be multiple "last steps" if the workflow has branching logic.)

assets/js/manual-run-panel/ManualRunPanel.tsx
```diff
@@ -168,12 +169,23 @@ export const ManualRunPanel: WithActionProps<ManualRunPanelProps> = props => {
...
+        
+        // Auto-select the next cron run dataclip if it exists (for cron jobs)
+        if (typedResponse.next_cron_run_id && !runId) {
+          const nextCronDataclip = typedResponse.dataclips.find(
+            clip => clip.id === typedResponse.next_cron_run_id
+          );
+          if (nextCronDataclip) {
+            selectDataclipForManualRun(nextCronDataclip);
+          }
+        }
```
doc-han: Okay. saw on the elixir end that the cron_run_id dataclip would always be available. cool

doc-han:
No matter the filter I pick. the cron based dataclip still shows. I guess this is because we just prepend it regardless. 
cc: @taylordowns2000 

Screenshot 2025-07-01 at 1 50 16 PM

lib/lightning_web/live/workflow_live/new_manual_run.ex
```diff
@@ -24,7 +31,56 @@ defmodule LightningWeb.WorkflowLive.NewManualRun do
              limit: limit,
              offset: offset
            ) do
-      {:ok, dataclips}
+      # Check if this job is cron-triggered and include next run state if so
+      {enhanced_dataclips, next_cron_run_id} =
+        maybe_add_next_cron_run(job_id, dataclips)
+
+      {:ok, %{dataclips: enhanced_dataclips, next_cron_run_id: next_cron_run_id}}
```
doc-han: Filter doesn't affect the cron dataclip. Even the date filter.

lib/lightning_web/live/workflow_live/new_manual_run.ex
```diff
@@ -3,28 +3,216 @@ defmodule LightningWeb.WorkflowLive.NewManualRun do
...
+          id: d.id,
+          type: d.type,
+          inserted_at: d.inserted_at,
+          project_id: d.project_id,
+          wiped_at: d.wiped_at,
+          is_next_cron_run: false
+        },
+        distinct: [desc: d.inserted_at],
+        order_by: [desc: d.inserted_at]
+      )
```
doc-han: Couldn't we have extended the logic for `Invocation.list_dataclips_for_job` to support the cron dataclip. Seems we've duplicated the code for filtering.

doc-han:
Feels like datetime on the dataclips got messed up.
It's all `1970-01-01 00:00:00`

Screenshot 2025-07-02 at 10 32 42 AM

But when I pick one I see the actual datetime.  
Screenshot 2025-07-02 at 10 33 38 AM

doc-han (changes requested):
1. The first item(step_result) is every search result is treated as a cron dataclip. This is becuase we do `[cron_dataclip | dataclips]` somewhere in elixir. The other side is when the entire result actually doesn't contain the cron_dataclip. We then treat whatever the first item in the result is as a `cron_dataclip` which is wrong.

taylordowns2000:
Thanks @doc-han , I can't replicate the bug you describe in (1) but am happy that you've given a 👍 to all the heavy code changes. I'll rebase and merge now.

Also, thanks for the solid click testing! I can't tell you how much I appreciate it.

midigofrank:
@taylordowns2000 please hold off on merging, need to take a look on the elixir side

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1537,11 +1537,17 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
-        {:reply, %{dataclips: dataclips}, socket}
+      {:ok, %{dataclips: dataclips, next_cron_run_id: next_cron_run_id}} ->
+        {:reply, %{dataclips: dataclips, next_cron_run_id: next_cron_run_id},
+         socket}
 
       {:error, changeset} ->
-        {:reply, %{errors: LightningWeb.ChangesetJSON.errors(changeset)}, socket}
+        {:reply,
+         %{
+           dataclips: [],
```
taylordowns2000, line 1554: @midigofrank , what's the standard for returning `[]`

or... the deeper question, why is this failing instead of returning an empty list? i think it's because it's not a valid UUID that's being entered. but we don't need to tell this to the user. we should still return no results, rather than erroring.

doc-han:
> Thanks @doc-han , I can't replicate the bug you describe in (1) but am happy that you've given a 👍 to all the heavy code changes. I'll rebase and merge now.
> 
> Also, thanks for the solid click testing! I can't tell you how much I appreciate it.

Welcome. I just pulled the updates and can still replicate the bug tho. (The first step of a cron triggered workflow. That has previously been run and now has a cron dataclip)

lib/lightning_web/live/workflow_live/edit.ex
midigofrank: From how it is structured, this only happens when the server is hit by unknown filters like invalid `timestamp`. Ideally, these errors should not occur but if they do, returning these errors gives us an opportunity to trace this errors because it would mean a bug has popped in somewhere. 

@taylordowns2000 were you able to reproduce it?

taylordowns2000: it's because it's not a valid UUID, which shouldn't bother the user I suppose. they should simply know that nothing matches what they've put in. no error.

midigofrank: Aaah, yeah, good point

lib/lightning/invocation.ex
```diff
@@ -765,6 +834,46 @@ defmodule Lightning.Invocation do
...
+    dataclips =
+      if include_next_cron?,
+        do: [next_cron_dataclip | regular_dataclips],
+        else: regular_dataclips
+
+    {dataclips, next_cron_run_dataclip_id}
+  end
+
+  # Check if a dataclip matches the user filters (applied in Elixir)
+  defp dataclip_matches_filters?(dataclip, user_filters) do
```
midigofrank, line 866: Hey @taylordowns2000 , I'm planning on getting rid of this in favor of reusing the query filter. 
The downside is we will potentially make 1 more trip to the db but I think it's fine compared to maintaining the filter in 2 different places

midigofrank:
Hey @doc-han , would you mind trying to reproduce the bug you experienced with these new changes? Because they're sharing the same logic, I presume it should _just work_. Let me know what you find

taylordowns2000:
ah sorry @midigofrank , your change does not work; well need to revert. I'll send you a quick video explaining: https://www.loom.com/share/b3618d2331574971900aef0fda7d9b1d

midigofrank:
Great catch @taylordowns2000 . i have reverted the filter commit

Issue #3130, opened 2025-04-16
React cleanup


doc-han:
### Description
The react side needs some cleanup exercise.
1. Proper typescript typing (priority)
2. Cleanups for all registered event handlers. (**mainly props.handleEvent**)
currently, none of these handlers are being cleaned, hence, a single useEffect can register a handler multiple time for a single component.
4. Update some eslint rules.

doc-han:
Major task here being
>Cleanups for all registered event handlers. (mainly props.handleEvent)
currently, none of these handlers are being cleaned, hence, a single useEffect can register a handler multiple time for a single component.

Has be resolved in #3358

PR #3358, opened 2025-06-30
feat: Undo/Redo 

### LINKED ISSUE #2177: Workflow Diagram: add undo/redo
Allow undo and redo of changes in the workflow diagram.

https://reactflow.dev/examples/interaction/undo-redo

This feels a bit more important with the addition of edge creation. I feel like I want to easily be able to undo a new edge. Ok, I can go and delete it ,but as a user it FEELS like I should have an undo.

Does this include changes made through the forms, like step names and stuff? Or is it it just changes to the diagram? It could be confusing both ways.

Undo/redo would only affect local changes.

It is independent of save. If you save, then undo, the undone changes will not be saved until you hit Save. Simple.

Undo/redo fires on ctrl/cmd + z/y. It likely appears on the toolbar

doc-han:
## Description

This PR Implements undo/redo on the workflow canvas

Closes #2177 #3130

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/workflow-store/store.ts
+++ b/assets/js/workflow-store/store.ts
@@ -101,7 +101,7 @@ export type ReplayAction = {
 };
 
 const undos: ReplayAction[] = [];
-const redos: ReplayAction[] = [];
+let redos: ReplayAction[] = [];
 
 // Calculate the next state using Immer, and then call the onChange callback
 // with the patches resulting from the change.
@@ -121,7 +121,7 @@ function proposeChanges(
     (p: ImmerPatch[], _inverse: ImmerPatch[]) => {
       if (!skipUndoStack) {
         undos.push({ patches: p, inverse: _inverse });
-        redos.splice(0, redos.length);
+        redos = [];
       }
       patches = p.map(toRFC6902Patch);
     }
--- a/assets/js/workflow-store/store.ts
+++ b/assets/js/workflow-store/store.ts
@@ -103,6 +103,28 @@ export type ReplayAction = {
 const undos: ReplayAction[] = [];
 let redos: ReplayAction[] = [];
 
+// simple squash function
+// I think after squashing. we can actually ignore in-between states.
+function pushUndo(action: ReplayAction) {
+  const lastAction = undos[undos.length - 1];
+  if (lastAction) {
+    const firstPatch = lastAction.patches[0];
+    const patch = action.patches[0];
+    if (
+      firstPatch &&
+      firstPatch.path.join('.') === patch?.path.join('.') &&
+      patch.path.length >= 3
+    ) {
+      lastAction.patches = [...lastAction.patches, ...action.patches];
+      lastAction.inverse = [...action.inverse, ...lastAction.inverse];
+      redos = [];
+      return;
+    }
+  }
+  undos.push(action);
+  redos = [];
+}
+
 // Calculate the next state using Immer, and then call the onChange callback
 // with the patches resulting from the change.
 function proposeChanges(
@@ -120,8 +142,7 @@ function proposeChanges(
     },
     (p: ImmerPatch[], _inverse: ImmerPatch[]) => {
       if (!skipUndoStack) {
-        undos.push({ patches: p, inverse: _inverse });
-        redos = [];
+        pushUndo({ patches: p, inverse: _inverse });
       }
       patches = p.map(toRFC6902Patch);
     }
@@ -163,6 +184,7 @@ export const store: WorkflowStore = createStore<WorkflowState>()(
           state,
           draft => {
             const lastPatch = undos.pop();
+            console.log('undoing:', lastPatch);
             if (!lastPatch) return draft;
             const newState = immerApplyPatches(draft, lastPatch.inverse);
             redos.push(lastPatch);
@@ -374,7 +396,7 @@ export const store: WorkflowStore = createStore<WorkflowState>()(
             path: patch.path.split('/').filter(Boolean),
           })
         );
-        undos.push({ patches: immerPatches, inverse: inverseImmerPatches });
+        pushUndo({ patches: immerPatches, inverse: inverseImmerPatches });
       }
 
       set(state => immerApplyPatches(state, immerPatches));
--- a/assets/js/workflow-store/store.ts
+++ b/assets/js/workflow-store/store.ts
@@ -1,6 +1,10 @@
 import type { Patch as ImmerPatch } from 'immer';
 
-import { applyPatches, enablePatches, produce } from 'immer';
+import {
+  applyPatches as immerApplyPatches,
+  enablePatches,
+  produce,
+} from 'immer';
 import { createStore, useStore, type StoreApi } from 'zustand';
 import type { Lightning, Positions } from '../workflow-diagram/types';
 import { randomUUID } from '../common';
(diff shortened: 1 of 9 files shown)


josephjclark:
We've hit some really big problems in undo redo:
- We ideally don't want everything to be undoable - edits to the job expression, for example (but also I think adaptor changes, version changes, edge state changes, label changes)
- This means we can't do a simple checkpointing system and revert to the whole model (because on undo, some "non undoable" actions will get reverted to whatever is in the checkpoint)
- There's also a complication that some edits are handled server-side (removing edges) and some client-side (adding nodes). So the app isn't actually entirely in control. This means we can't track our own undo/diff objects which we re-generate JSONPatches from (like `{ type: 'CHANGE', id: '<some-node-id>', props: { label: 'new node name' } }`
- Farhan has a nice elegant solution to cache a stack of patches in the store, and on undo-redo just replay those patches. Some patches are ignored.
- But this breaks on delete! Because delete patches aren't like "delete node 3", they're like "delete node 3 and move all the properties of node 4 to whatever is now in position 3". It's a really messy diff and it caches a whole bunch of state, so that when we restore this patch, changes can be lost
- So as a first-pass we're going to make all edits undoable, because at least this gives us a clean history and users will understand what's happening (even if they didn't mean it). Similar edits (like several keystrokes on a label) will be squashed together into a single state
- Some of these undos will be totally invisible to the user (like undoing a change to job code). So I think that once this is working, we'll have to add a little notification to tell users about what certain undos did. We'll have to think about this later

stuartc:
Hey @doc-han,

Great work on the undo/redo implementation! While testing, I noticed a potential
issue that isn't affecting things right now but could bite us later.

(I am fairly sure you are aware of this, but for the sake of remembering.)

Currently, `applyPatches()` adds server-initiated changes to the undo stack,
which means users could theoretically undo server responses (like validation
corrections or default value additions) rather than their own actions. Right now
this seems okay since user actions that trigger server responses are triggered
by LiveView components. And the rest through `proposeChanges()` first, so the
flow works.

But as we add more features, we might have cases where server patches become
undoable and undoing would reply the response etc.

For the future, it might be worth considering having all user actions flow
through handlers that would then go through `proposeChanges()`.

> After reading @josephjclark's message above, I think we've spotted the same
> thing, I didn't know it went that deep. I can't guage how big of a deal this
> is, I mean it sounds like a problem. So it's worth thinking about how to solve
> this, either by adding action wrappers around `proposeChanges()`.

The two-stack undo/redo approach is really clean and easy to follow - nice
choice for getting it working quickly!

For the future, we might want to consider the more traditional single history
array + position pointer approach. While it's a bit more complex to implement,
it unlocks some pretty useful features:

- Could show users a timeline of their changes
- Jump to any point, skip multiple undos to go directly to a specific state
- Full history is always available for troubleshooting
- Could show "what changed" for each history entry
- History compression, branching, undo limits, etc.

As far as I can tell it matches how Git, text editors, design tools work.

The current approach works great though, so this is just a "nice to have" for
future iterations if we want to get fancy with undo/redo features.

josephjclark:
hmm, re undoing server responses that's a fair concern. I'm not really happy with this "let's make everything undoable" approach, but I think it's quick win that gets us 80% of the behaviour we need right now. As you say I don't think server changes are an immediate concern.

Re the array + pointer approach well, maybe later. But that feels over complicated right now. I can't see us providing a git-style checkpoint history any time soon?

doc-han:
I’m thinking. 
if server info is considered in patch then can't we scope the `workflow_params` to just the properties the diagrammer store needs before creating the patch. Would that resolve the issue?

Also, with array + position pointer based undo. I think we can actually easily simulate that with the two stacks too. (I used stacks because I wasn't comfortable moving a pointer around when the array itself can be changed in-place). 
[**undo_stack**] + current_state(pointer) + [**redo_stack**]

Hence, to move the pointer forward. we pop from redo_stack and push unto undo_stack and then we do the opposite to move backwards. with the representation above we can also show that entire array as the undo-redo history. When a user wants to jump to a specific point in time. we apply all patches between pointer and where the user wants to be.

doc-han:
Actually, server responses wouldn't be undone even when a patch is returned. Server responses might return a patch (which we've not confirmed yet. currently return empty patch array) but it never returns and inverse patch. Our logic makes sure to check whether there exists an inverse patch before adding to to the undo stack. 
https://github.com/OpenFn/lightning/blob/ef6310197f0dde7d29bcdb91e4248802fd3825f5/assets/js/workflow-store/store.ts#L391
cc: @stuartc

doc-han:
> ... it behaved exactly as I'd expect it to with the exception of not being able to undo the text I wrote in an edge label field. 

Currently, if you have the text field in focus, the undo doesn't update the content of the text field. but if it's not in focus then it's able to update the content. will figure that out. 
cc: @taylordowns2000

josephjclark:
> but i'm having a hard time figuring out why we wouldn't want to be able to undo/redo everything

@taylordowns2000 There's loads of stuff actually
- Changes to code / expressions. These should be undoable inside the editor
- Changes to JS expressions on edges (maybe switching to true/false/expression should be undoable? Maybe? Not sure)
- When you switch from auto layout to manual layout. I believe this creates an undo state but it's really not one we want, it's totally invisible
- I don't think that adaptor version changes should be undoable. I'm not really keen on label changes or adaptor changes being undoable either
- Disabling nodes/edges/triggers? No idea tbh. They don't feel like good candidates for undo because there's no real consequence to disabling them. It's less an edit and more an option/settings change?
- Switching credentials?
- We're not ready to look at this yet but I have no idea how undo/redo will work with the AI model suggestions. If you click to rollback to an earlier history version and then Undo... what do you expect to happen? Depending on how we handle AI edits that might be a little hairy to resolve

The story here is supposed to be to allow undo on structural stuff - moving nodes around,  creating or removing edges. Mostly it's to undo positional/layout changes

One BIG problem is that some edits aren't very visible from the workflow diagram. If you undo a change to your code, for example, you can't see it from the canvas. You could easily undo an hour's work (all your job edits are squashed to a single unto state) and not realise it :shrug:

josephjclark:
From call wit Han: this is probably good to go but  here are the  current outstanding issues:
1. No visual feedback on some undo steps (this is shelved right now but I think we need a little toast notification)
2. The edit UI doesn't undo while focused (but the diagram does)
3. Probably later want to debounce history squashing because it can go back a bit too far in one step now)

assets/js/workflow-store/store.ts
```diff
@@ -89,12 +95,43 @@ function toRFC6902Patch(patch: ImmerPatch): Patch {
...
+  inverse: ImmerPatch[];
+};
+
+const undos: ReplayAction[] = [];
+let redos: ReplayAction[] = [];
+
+// simple squash function
+// I think after squashing. we can actually ignore in-between states.
+function pushUndo(action: ReplayAction) {
+  const lastAction = undos[undos.length - 1];
```
josephjclark, line 109: No need to change right now, but an easier way to do this is `undos.at(-1)`

assets/js/workflow-store/store.ts
```diff
@@ -89,12 +95,43 @@ function toRFC6902Patch(patch: ImmerPatch): Patch {
...
+  if (lastAction) {
+    const firstPatch = lastAction.patches[0];
```
josephjclark, line 111:
```
const [firstPatch] = lastAction.patches;
const [patch] = action.patches;
```

josephjclark (approved):
I'm approving but I am raising the following issues as a result of QA:

- #3377
- https://github.com/OpenFn/lightning/issues/3378
- https://github.com/OpenFn/lightning/issues/3379
- https://github.com/OpenFn/lightning/issues/3380
- https://github.com/OpenFn/lightning/issues/3376 (unrelated to this work)

I don't think these are blockers, so happy to release and come back to them later.

I think the invisible edits stuff is _probably_ an edge case and won't bother most users. And if it does, we'll escalate the fix.

PR #3364, opened 2025-07-02
OAuth: Exclude offline_access from granted scopes validation

### LINKED ISSUE #3363: OAuth credentials: `offline_access` scope shouldn't be checked in the granted scopes
Some OAuth providers, like Microsoft, do not returned the `offline_access` scope in the list of granted scopes when requesting an oauth token. This is because the `offline_access` scope is a special scope used to return a refresh token. In the app we do have a validation rule that checks whether the scopes that the user has requested are granted by the user during authorization. And that validation rule is currently including the `offline_access` scope in the list of items it checks. That leads to a bug where the app returns a missing scope error when the provider doesn't return the `offline_access` scope. That confuses and blocks the user because they have granted all scopes.

**v2.13.3 (1.5.11)**

- [x] Yes
- [ ] No

1. Create a MS Graph OAuth credential
2. Check all scopes and click authorize
3. Go thru the process of authorization
4. See the red error rendered in OpenFn

happen.

The user shouldn't be blocked about the missing `offline_access` scope.

Image

elias-ba:
## Description

This PR fixes a bug in the OAuth scope validation logic where the offline_access scope was incorrectly treated as a mandatory granted scope. Some providers, like Microsoft, do not include offline_access in the list of granted scopes, even when it was requested, because it is a special-case scope used only to request a refresh token, not a permission per se.

Previously, our validation logic checked for the presence of all requested scopes in the granted scopes, including offline_access, which led to false negatives. This PR updates the validation logic to exclude offline_access from the scope comparison, preventing unnecessary blocking errors when a user successfully authorizes but the provider omits this scope from the returned list.

Closes #3363 

## Validation steps

1. Create a Microsoft Graph OAuth credential.
2. Select all available scopes, including offline_access.
3. Complete the authorization flow.
4. Confirm that no error is shown, and the credential is saved successfully

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


elias-ba:
@stuartc I 100% understand you concern about whitelisting the offline_access scope. The only hypothesis that makes me feel safe about it is we do know that offline_access allows to garantee having a refresh token back from the provider. And in our system we do have a validation that prevents saving oauth tokens without a refresh token. So even if we do not have a offline_access scope back we still do check whether we have a refresh token or not before saving.

taylordowns2000:
hey @elias-ba , if i'm understand this change, it's _not_ whitelisting `offline_access`, it's removing it from the list of "expected" scopes that we check against.

That seems OK, and if there are clients that require `offline_access` then we can add it to the list of mandatory scopes on the other clients. Is that the safe thing to do? (CC @stuartc ?)

Then for the code itself, why are we looking at the expected_scopes and dropping "offline_access". From an English language perspective, that doesn't seem logical. Do you _not expect_ `offline_access` for certain credentials? If so, it shouldn't be added to the list of `expected_scopes` in the first place. Which, naturally, leads to the question: how is it getting there? Does _OpenFn_ expect it to be there? Does the provider expect it to be there? Is there another party?!

Sorry that I can't accept/request changes right now, need more information on this before giving a thumbs up.

elias-ba:
@taylordowns2000 I think there's a misunderstanding here. Let me clarify what's actually happening:
This isn't about whitelisting or provider-specific behavior. The `offline_access` scope is fundamentally different from permission scopes like `read` or `write`:

`offline_access` is a protocol-level scope defined in OAuth 2.0 specs that means "give me a refresh token". It's not a permission to access resources. Some OAuth providers don't return it in the scope list even when they honor it and provide a refresh token. We already validate the actual outcome, We check that a refresh token exists in `validate_refresh_token`. That's the proof that offline_access worked.

Without this exclusion, users would see "Missing required scopes: offline_access" errors even when they successfully received a refresh token. That would be incorrect validation.

To your specific question: No, we're not adding `offline_access` to any list. We're excluding it from validation because validating its presence in the returned scope list is meaningless and can lead to situations like the one Aicha and I faced, where we couldn't save our credential. What matters is whether we got a refresh token.

taylordowns2000 (approved):

PR #3370, opened 2025-07-03
Fix handling of metadata-related errors

### LINKED ISSUE #3367: Lightning displays incorrect message in job editor when adaptor does not support metadata
**Describe the bug**

Given a job that uses an adaptor that does not support metadata (e.g. http), when you view the metadata tab in the job editor, you are presented with `An error occurred while loading metadata: enoent`. This is because Lightning is not correctly handling the case where the CLI is reporting that the adaptor does not support metadata.

**Version number** 

https://github.com/OpenFn/lightning/commit/7641d342408c3d0361c96455958abde4bc3e49bf

- [x] Yes
- [ ] No

See the description above.

**Expected behavior** 

To show the message that the adaptor does not support 'magic functions'.

**Screenshots**

**Additional context**

Pretty sure we need to look here: https://github.com/OpenFn/lightning/blob/main/lib/lightning/metadata_service.ex#L112

rorymckinley:
## Description

When Lightning checks the metadata response for an adaptor, it does not correctly deal with two cases:

- When metadata is not supported, it treats the error message as a path, resulting in:

Screenshot From 2025-07-07 16-37-02

- When there is a credential error (URL is borked, bad creds), it should show a message indicating that there may be an issue with the credential, but instead it shows that the adaptor does not support metadata:

Screenshot From 2025-07-07 16-37-41

Closes #3367 

## Validation steps

- In local Lightning, setup a dummy http credential and then modify a job to use the `http` adaptor with this credential.
- Click to 'Edit' the Job, wait a few seconds for the backend to fetch the adaptor and then navigate to the Metadata tab in the job editor.
- You should see a message similar to the below:

image

- Return to the Workflow editor.
- Create a DHIS2 credential with broken creds (e.g. `https://play.im.dhis2.org/stable-2-42-0` with `foo` and `bar` as the username and password respectively).
- Change your job to use the dhis2 adaptor (version earlier than 7.x.x) and then edit the job.
- Wait for Lighting to fetch the adaptor and then click on the Metadata tab and you should see the below error:

image

Colonel Sanders (of KFC fame) was an "actual" colonel. He was given the honorific title of "Kentucky Colonel" in 1935. He would later capitalise on this by adjusting his appearance and dress to match that of a Confederate colonel as he identified the branding potential of being 'Colonel Sanders'.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -2856,10 +2856,14 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
     patches =
       WorkflowParams.to_patches(initial_params, next_params)
+
     inverse_patches = WorkflowParams.to_patches(next_params, initial_params)
 
     socket
-    |> push_event("patches-applied", %{patches: patches, inverse: inverse_patches})
+    |> push_event("patches-applied", %{
+      patches: patches,
+      inverse: inverse_patches
+    })
   end
 
   defp maybe_push_workflow_created(socket, workflow) do


lib/lightning/metadata_service.ex
```diff
@@ -110,13 +110,17 @@ defmodule Lightning.MetadataService do
...
   defp get_output_path(result) do
-    path =
+    last_message =
       result
       |> CLI.Result.get_messages()
       |> List.last()
 
     cond do
-      path ->
+      is_map(last_message) ->
```
rorymckinley, line 119: There is arguably some room to refactor the cond here, but this fix has already taken a lot of time (due to my ignorance) so I thought I would leave that for another day as I think this PR already complies with the Boy Scout principle :).

elias-ba (approved):
Nice one @rorymckinley, thanks

stuartc (approved):

Issue #3103, opened 2025-04-10
Visualize runs on the canvas


taylordowns2000:
When a run is _not_ already loaded (see "spooky runs loaded on canvas") the user can view history:

<img width="1082" alt="Image" src="https://github.com/user-attachments/assets/7b94e637-6252-44fd-be04-ead0508c7f40" />

When a run is loaded (run = attempt in this image) you can visualize the run-steps on the canvas:

<img width="1032" alt="Image" src="https://github.com/user-attachments/assets/209898f0-3165-43ca-ad27-5c32a61d401b" />

When a step is selected with a run loaded, you see the logs for that step:

<img width="425" alt="Image" src="https://github.com/user-attachments/assets/8a8e00d5-3c90-4914-9c8e-b4c08891b8ba" />

**Here's the [OG demo on Figma](https://www.figma.com/proto/kl4He7PX8Glg8fxqTbn38f/OFG-Sonder-Design-Sprint--new-?page-id=182%3A17117&node-id=182-25856&p=f&viewport=986%2C-2471%2C0.71&t=amqFJLuyMr9EZtHL-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=182%3A28215).**

taylordowns2000:
@stuartc , @theroinaochieng , we added 21 here (stu will break this into a 5, 8, 8 later and we do in next sprint!) thanks

taylordowns2000:
@theroinaochieng , @stuartc , why did this get a 'needs detail' tag? is there something (other than kicking off design with an engineer/designer combo) that you'd like to see on this issue? Stu has his 3 sub-issues for this.

josephjclark:
On the Canvas, if a run is loaded, you have two options: `Run (new Work Order)` or `Run (retry)`

If a run is not loaded, the only option is `Run` (which means "run new work order")

The inspector is exactly the same: `Run`, `Run (new work order` and `Run (try)`, depending on the context

[scribble from taylor](https://openfn.slack.com/archives/C0394D0GBUN/p1750341533286449)

![Image](https://github.com/user-attachments/assets/b8382b3e-8b05-490d-a8c3-07a53f0334fc)

josephjclark:
* How do we transition into this view?
* What happens when you click the new Run button?
* How should we show the logs? In the editor panel or another?


States of nodes:
- Success (green)
- Fail (red)
- Pending (grey)
- In progress (blue)

Consider animated edges

[...]

PR #3382, opened 2025-07-09
feat: visualize runs on the canvas

### LINKED ISSUE #3387: Visualize existing(complete) runs on the canvas
### User story
When a user views a run via the history page. We want to allow them to visualize the run on the canvas. 

### Details
1. Doesn't support live logs or streaming of run statuses. It only supports completed workorders.

doc-han:
## Description

Closes #3387 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -1675,8 +1675,6 @@ defmodule LightningWeb.WorkflowLive.Edit do
   end
 
   def handle_event("switch-version", %{"type" => type}, socket) do
-    dbg(type)
-
     updated_socket =
       case type do
         "commit" -> commit_latest_version(socket)


josephjclark:
TODO: to colour edges properly we need to resolve this: https://github.com/OpenFn/kit/issues/956

josephjclark:
Couple of early TODOs:
1. Stub in some kind of tooltip on the pass/fail icon. This should show output for multiple executions of a single node
1. A node is only a success if it never fails (across multiple executions)
1. Nodes that were not executed need to be greyed out
1. Disable edit mode while looking at a run (don't show the + icon on a node)
1. Some kind of indicator on the start node (maybe an icon top right for now?)

Selection doesn't work very well when we introduce many more colours like this: see https://github.com/OpenFn/lightning/issues/3386

Soon we'll spike out an Edit mode toggle (more comments coming in on that)

josephjclark:
We've encountered an issue against main which I think we can really only resolve here - sort of stubbed in in https://github.com/OpenFn/lightning/issues/3385

Basically what it is is: 

When looking at an old run / a snapshot, the snapshot should be read-only. No edits allowed. On prod right now, the `+`icon is actually live and interactive, sort of, so you can fool around a bit.

We need to sort that out and ensure snapshots are read only.

A wrinkle on this is that the most recent snapshot and the latest version are the same thing. So how do we differentiate? How do we make latest editable but the last run read only??

This PR might help us out.

By having a History Mode and Edit Mode, as a top-level flag on the workflow diagram, we should have clearer seperation of concerns. Old runs - a workflow with an overlay - are never editable. The latest workflow (no run loaded) has no history. Use buttons in the nav bar to switch between states.

josephjclark:
Thoughts on showing multiple runs of the same node:
- render multiple edges?
- render the execution graph, not the workflow graph? Totally different tree?
- duplicate nodes in the workflow (is this the same as above?)
- show multiple status icons (3 ticks on cross)

For the moment, I want to do the simplest possible thing and show show as single status and tooltip.

doc-han:
AI chat icon already at the top-left corner of the screen. same place we want workflow history to be. 
Screenshot 2025-07-15 at 9 49 28 PM

doc-han:
**Note**
We have a **trigger** and **jobs** for a workflow. 
After a Run, we keep the results of each job run in an an array called **steps** on the run but then, we have no run results for a trigger. 

**Situation**
Some Runs do crash even before a single jobs gets to execute. Hence, the **steps** array becomes empty. In this case, it becomes quite tricky to display on the UI. because you can't associate the crash to a node or the trigger too I guess. 

cc: @taylordowns2000

doc-han:
I have a situation where `starting_job_id` doesn't seem correct. strange

josephjclark:
Some UX feedback notes from me:

- `Recent Activities` -> `Recent Activity`
- There are currently two collapse icons in the history panel - can we remove the "open in new window" one? EDIT: actually, because the whole titlebar take the click and collapses the UI, _both_ icons are a bit redundant. I suggest we keep the arrow and drop the "open" icon

image

- The amount of repetition in this list bothers me. I know we're looking at Work Order and Runs here, but a) it's a lot of information, b) I wonder whether users understand this. Not convinced the UUIDs actually help anything either.
image

- Generally, the UI indicators that say "You are currently looking at THIS historical run" are too subtle. It can be hard to know whether a run is loaded, and if so which. Especially if the history panel is collapsed. Maybe the Panel Title needs to be way more obvious? Maybe we need a banner above the diagram? Maybe we need more information alongside the snapshot id? I'm tempted to leave this until later, but I'm really worried that users landing on this page for the first time are going to have no idea what's happening
- We're currently showing the run duration on each step, in the label, but this looks horrible and upsets all the label layout. Can we put that in the tooltip or something?
- I think the "started from here" icon should be a star, not a play button, because it's not a verb and the play button looks like a verb
- As noted in a call recently, some alignments in the history panel look a little off
- There is no way to go  from the Run Viewer straight to the canvas. You have to Inspect a node and close the inspector

What do you think @taylordowns2000 ? I can make some of these changes directly. Maybe others we defer until later.

doc-han:
Flow -

**Collapsing the history panel**
1. When on latest + some highlighted steps -> we remove all the runs and just allow you to edit because you're on latest. 
2. When on snapshot -> 
	a) we do nothing and keep the highlights? because if we remove the highlights then we can't allow you to edit. If we remove the highlights. then what state is the user currently in?
	b) we remove the step highlights, take the user to latest, allow edit
3. Don't allow collapsing the history panel when user is on a snapshot

josephjclark:
- Update inspector close to not remove the run from the URL (Maybe)

josephjclark:
To close this out:
* [x] Put an X on the Run in the Recent Activities list
* [ ] Make sure icons are exactly the same (I'll give you a pass on sizing, just align the colours)
* [x] On refresh, with a snapshot open, the history view should be open (not collapsed)
* [x] When editing a node name on a run (which is fine!), the server is sending an update and the run is getting dropped.
* [x] Open a step from old run (you're in read only mode), go to latest in inspector (you can now edit but the run is still loaded), then close the inspector - the run has gone! But it's in the URL


To do later:
- #3456
- #3457
- better support when the Run and Workflow are inconsistent (ie, steps in the Run are not visible in the Workflow)
- https://github.com/OpenFn/lightning/issues/3388
- One undo redo issue from Farhan?

josephjclark:
Two more problems: 

When viewing a Run, when clicking a Node, the Run UI on the canvas:
- Needs to say Retry, rather than Run
- Needs to preload the input that was used in the Run

Basically when a Run is loaded, we want to Retry the work order, not create a new one

Probably spin this out into a new issue?

doc-han:
"Switch to Latest" may be a toggle. 

reason: once you hit "Switch to Latest", it's difficult to get back to the snapshot for the run. You can do so now by deseleting the run and reselecting it.

doc-han:
I've spotted one bug I'm currently trying to resolve. 
For some unknown reason. view old snapshot -> switch to latest -> try to connect two nodes. 
immer patching doesn't work in that scenario but as soon as you remove the run history highlights. it works.

doc-han:
### User Journey

#### 1. Wokflow Page to Mini-History
- Go to **Workflows Page**
- Select a **Workflow**
- Find the **Mini-History Panel** (top-left) **collapsed** because no run is selected

#### 2. Pick & Visualize a Run
- Click a run → it overlays on the workflow
- Click the run again to deselect and remove overlays

#### 3. Edit Enabled/Disabled
- If on **Latest Snapshot** → Editing is **enabled**
- If on an **Old Snapshot** → Editing is **disabled**

#### 4. Switch to Latest (Old Snapshots)
- Click **“Switch to Latest Version”** → Editing is now **enabled**
- Also, the runs(that happened on an old snapshot) are visualized on the latest snapshot

---
#### 1. History Page to Mini-History
- Go to **History Page**
- Select a **Run** from a Workorder and enter the canvas through there.
- Find the **Mini-History Panel** (top-left) **expanded** because a run is already selected.
- A run history should already be overlayed on the workflow.

Continue from **2.** in the First flow above...

assets/js/utils/timeSpent.ts
```diff
@@ -0,0 +1,15 @@
+export function timeSpent(dateStr1: string, dateStr2: string) {
```
josephjclark: I would call this a `duration`. I'd expect the `date-fns` library to handle this well but taking a quick look at the docs I don't really think they go below seconds.

We show durations currently on the Run Viewer. I guess there's no way to share logic? Are we using the same units and conventions as live view?

image

assets/js/utils/timeSpent.ts
```diff
@@ -0,0 +1,15 @@
...
+  const date1 = new Date(dateStr1).getTime();
+  const date2 = new Date(dateStr2).getTime();
+  const diffMs = Math.abs(date2 - date1);
+
+  if (diffMs >= 3600000) {
```
josephjclark: I think I'd expect the units to be `ms/s/m/hrs`. But that's subject to how LiveView renders durations. Consistency is king.

assets/js/workflow-diagram/MiniHistory.tsx
```diff
@@ -0,0 +1,182 @@
...
+  // at time this component will be rendered before data reaches store. that makes the panel collapse
+  const selectedItem = history.find(w => w.selected)?.id;
+  React.useEffect(() => {
+    if (selectedItem) {
+      setIsCollapsed(false);
+    }
+  }, [selectedItem]);
+
+  const expandWorkorderHandler = (workorder: WorkflowRunHistory[number]) => {
+    if (workorder.runs.length === 1 && workorder.runs[0]) {
```
josephjclark, line 79: Why `workorder.runs[0]` ?

assets/js/workflow-diagram/MiniHistory.tsx
```diff
@@ -0,0 +1,182 @@
...
+      selectRunHandler(workorder.runs[0].id, workorder.version);
+    }
+    setExpandedWorder(prev => (prev === workorder.id ? '' : workorder.id));
+  };
+
+  const historyToggle = () => {
+    setIsCollapsed(p => !p);
```
josephjclark, line 86: Doesn't `p` go stale?

assets/js/workflow-diagram/MiniHistory.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+
+  return (
+    <div
+      className={`absolute left-2 top-2 bg-white border border-gray-200 rounded-lg shadow-sm overflow-hidden ${isCollapsed ? 'w-44' : 'w-88'
+        }`}
+    >
+      {/* Header */}
```
josephjclark: Better to abstract this chunk of JSX into a `<Header>` component than to leave a comment lying around like this

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -83,15 +97,32 @@ const Node = ({
...
   errors,
+  type,
 }: BaseNodeProps) => {
+  const isTriggerNode = type === 'trigger';
+  const runData = data?.runData as RunStep | undefined;
+  const startInfo = data?.startInfo as
+    | { started_at: string; startBy: string }
+    | undefined;
+  const isErrorRun = runData?.exit_reason !== 'success';
+  // TODO: remember triggers
```
josephjclark, line 98: Does this TODO need looking at?

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -83,15 +97,32 @@ const Node = ({
   secondaryIcon,
 
   errors,
+  type,
 }: BaseNodeProps) => {
+  const isTriggerNode = type === 'trigger';
+  const runData = data?.runData as RunStep | undefined;
+  const startInfo = data?.startInfo as
```
josephjclark, line 94: Why do we have this explicit typing here? The `RunStep` type should have everything you need

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -83,15 +97,32 @@ const Node = ({
...
+    | { started_at: string; startBy: string }
+    | undefined;
+  const isErrorRun = runData?.exit_reason !== 'success';
```
josephjclark, line 97: VSC tells me that `isErrorRun` is not used

(and if it is actually needed, if there's no run loaded then the Node would render in an error state right? So hopefully we can just remove this)

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -83,15 +97,32 @@ const Node = ({
...
+  // TODO: remember triggers
+  const didRun = data.isRun
+    ? !!runData || (!!data?.startInfo && isTriggerNode)
```
josephjclark, line 100: This ternary is wild. Can we not just say `const didRun = runData?.exitReason`? Because if there's a truthy exit reason we're dealing with a run, and that's about it

assets/js/workflow-diagram/nodes/Node.tsx
```diff
@@ -83,15 +97,32 @@ const Node = ({
   secondaryIcon,
 
   errors,
+  type,
 }: BaseNodeProps) => {
+  const isTriggerNode = type === 'trigger';
```
josephjclark, line 92: I'm a bit uncomfortable with this because we have a common Node, and then a typed Step and Trigger node. Really, all the trigger node logic should be handled by these child nodes.

But this requires a bigger rethink than we have time for, so we'll have to just roll with it for now.

assets/js/workflow-diagram/styles.ts
```diff
@@ -143,20 +156,48 @@ export const styleEdge = (edge: Flow.Edge) => {
...
+const BORDER_GREEN_600 = '#00a63e';
+const BORDER_RED_600 = '#e7000b';
+const BORDER_ORANGE_600 = '#f54a00';
+
+export const nodeIconStyles = (
+  selected?: boolean,
+  hasErrors?: boolean,
+  runReason: RunStep['exit_reason'] = null
+) => {
+  const getNodeBorderColor = (reasons: RunStep['exit_reason']) => {
```
josephjclark, line 177: These would be better implemented as lookup objects, rather than these syntax-heavy functions

assets/js/workflow-diagram/util/update-selection.ts
```diff
@@ -16,20 +16,20 @@ import type { Flow } from '../types';
  * TODO: as an optimisation, consider exiting early once we've updated both selected items
  */
 export default (model: Flow.Model, newSelection: string | null) => {
-  let neighbours = {};
+  // let neighbours = {};
```
josephjclark, line 19: If we're not going to delete this, then I would leave `let neighbours = {}` and `data: { ...item.data, neighbour: item.id in neighbours },` as they are, and then comment out the body. And also add a comnent to say why it's been commented out.

Alternatively we just remove all the code we're not using :shrug:

assets/js/workflow-editor/WorkflowEditor.tsx
```diff
@@ -1,10 +1,37 @@
...
 import WorkflowDiagram from '../workflow-diagram/WorkflowDiagram';
-import { useWorkflowStore } from '../workflow-store/store';
+import { RUNS_TMP, useWorkflowStore } from '../workflow-store/store';
+import tippy, { type Placement } from 'tippy.js';
 
 export const WorkflowEditor: WithActionProps<{ selection: string }> = (props) => {
-  const { getItem, forceFit } = useWorkflowStore();
+  const { getItem, forceFit, updateRuns } = useWorkflowStore();
+
+  // TODO to be moved to a higher level
```
josephjclark: Can we deal with this nor or remove the commnet?

assets/js/workflow-store/WorkflowStore.tsx
```diff
@@ -151,10 +161,11 @@ export const WorkflowStore: WithActionProps = props => {
         });
       }
     );
-  }, [props.pushEventTo, setState, add]);
+  }, [props.pushEventTo, setState, add, updateRuns]);
 
   React.useEffect(() => {
     return props.handleEvent('set-disabled', (response: { disabled: boolean }) => {
+      console.log("disabled:", response)
```
josephjclark: We can probably remove this log?

josephjclark (changes requested):
Left a few small things which I think ought to be tidied up. Not every comment needs dealing with now - please just fix the small and easy stuff.

A wider note (for @taylordowns2000 really): most of my [UX comments](https://github.com/OpenFn/lightning/pull/3382#issuecomment-3113803963) are outstanding

Basically I am still concerned about the amount of confusion this UI will introduce to users. And sadly there's a lot of work that still needs to be done to get this feature really tight. I think we should aim to phase this feature in a bit more gently - but of course that will require more dev time.

assets/js/workflow-diagram/MiniHistory.tsx
doc-han: makings ts compiler happy. it wouldn't keep quiet when I later try to access `workorder.runs[0]`

assets/js/workflow-diagram/MiniHistory.tsx
doc-han: I doubt it goes stale. since I'm getting it from the setAction itself.

josephjclark: It seems to work OK from my click testing, I'm just surprised it's so stable without a `useCallback` call or anything. But if it's working its working!

lib/lightning/invocation.ex
```diff
@@ -915,4 +915,25 @@ defmodule Lightning.Invocation do
...
+          preload: [:steps]
+        )
+    )
+    |> Repo.one()
+    |> case do
+      nil -> []
+      %{runs: [last_run]} -> last_run.steps
+      _ -> []
+    end
+  end
```
elias-ba: I love this function! I have a few suggestions that might help:

1. Could we consider a more descriptive name like `get_latest_run_steps_for_workflow`? The current name is a bit hard to parse and might be clearer if it indicates we're getting steps from the most recent run.

2. Can I suggest adding a brief doc comment to explain the logic? Something like:
```elixir
@doc """
Fetches steps from the most recent run of the latest work order for a given workflow.
Returns an empty list if no work order or runs exist.
"""
```

3. Would it be worth considering a slightly cleaner pattern match? Maybe something like:
```elixir
|> case do
  nil -> []
  %{runs: [run | _]} -> run.steps
  %{runs: []} -> []
end
```

These are just suggestions, the current implementation seems to work well! The use of preload with the subquery is actually quite elegant for getting exactly what we need in a single database round trip. 

What do you think?

lib/lightning/work_orders.ex
```diff
@@ -484,6 +484,89 @@ defmodule Lightning.WorkOrders do
...
+          start_from: trigger_id,
+          steps: run_steps,
+          isTrigger: true,
+          inserted_at: data.inserted_at
+        }
+
+      _ ->
+        %{start_from: nil, steps: [], isTrigger: true}
+    end
+  end
```
elias-ba: 1. I noticed the function returns different keys depending on the case - sometimes we get `run_by` and `inserted_at`, sometimes not. Could we consider making the return structure consistent across all cases? This would make it more predictable for code consuming this function. Maybe something like:

```elixir
%{
  start_from: ...,
  steps: ...,
  isTrigger: ...,
  inserted_at: nil,  # always include, even if nil
  run_by: nil        # always include, even if nil
}
```

2. Would it be clearer if we simplified the case statement? Perhaps we could consolidate the logic like:

```elixir
case Repo.get(Run, run_id, preload: [:created_by, :steps]) do
  nil ->
    # return default structure
  run ->
    %{
      steps: run.steps,
      inserted_at: run.inserted_at,
      run_by: run.created_by && run.created_by.email,
      isTrigger: !is_nil(run.starting_trigger_id),
      start_from: run.starting_job_id || run.starting_trigger_id
    }
end
```

3. Can I ask about the `_ ->` pattern? It seems like this would only match if a run has both `starting_trigger_id` and `starting_job_id` set. I don't think this a valid state in our data model

4. Would it help to add a `@spec` to document what this function returns? It could help future developers understand the expected output structure.

Just some thoughts! The logic for determining trigger vs job runs is clear, which is great. What do you think about these suggestions?

lib/lightning/work_orders.ex
```diff
@@ -484,6 +484,89 @@ defmodule Lightning.WorkOrders do
...
+          |> Enum.map(fn run ->
+            Map.take(run, [:id, :state, :error_type, :started_at, :finished_at])
+          end),
+        version: worder.snapshot.lock_version,
+        state: worder.state,
+        last_activity: worder.last_activity,
+        id: worder.id
+      }
+    end)
+  end
```
elias-ba: Ah ! I remember this big boy 😆. It is handling some complex logic, and I have a few thoughts that might help make it clearer and more maintainable:

1. I'm finding the UNION logic a bit hard to follow. Could we add a comment explaining the business requirement? It seems like when `run_id` is provided, we want to ensure that specific work order is included plus up to 19 others. Is that correct? Would it be possible to simplify this approach?

2. I noticed we're loading all runs and all steps for each work order. Do we need all this data upfront? If work orders have many runs, this could become a performance bottleneck. Could we consider:
- Loading only recent runs (maybe last 5-10)?
- Or lazy-loading runs when needed?
- Adding a comment about why we need all runs if it's a business requirement?

3. Could we replace `fragment("last_activity")` with `[wo], desc: wo.last_activity`? Using the Ecto syntax would be safer and more maintainable.

4. What do you think about extracting the formatting into helper functions? Something like:

```elixir
|> Enum.map(&format_workorder_summary/1)

defp format_workorder_summary(workorder) do
  %{
    id: workorder.id,
    state: workorder.state,
    last_activity: workorder.last_activity,
    version: workorder.snapshot.lock_version,
    runs: Enum.map(workorder.runs, &format_run_summary/1)
  }
end

defp format_run_summary(run) do
  Map.take(run, [:id, :state, :error_type, :started_at, :finished_at])
end
```

5. Have you considered if we could handle the "highlight specific run" logic differently? Maybe we could:
1. Always fetch the 20 most recent work orders
2. If a `run_id` is provided, check if its work order is in the list
3. If not, fetch it separately and prepend it to the results

Just thinking this might be simpler than using UNION?

These are just suggestions to consider! The current implementation does work, and this business logic here is not quite easy. What are your thoughts on any of these points?

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -517,9 +517,9 @@ defmodule LightningWeb.WorkflowLive.Edit do
                           phx-value-id={@selected_job.id}
                           theme="danger"
                           disabled={
-                            (!is_nil(@workflow.deleted_at) or !@can_edit_workflow or
-                               @has_child_edges or @is_first_job or
-                               @snapshot_version_tag != "latest") ||
+                            !is_nil(@workflow.deleted_at) or !@can_edit_workflow or
+                              @has_child_edges or @is_first_job or
+                              @snapshot_version_tag != "latest" ||
```
elias-ba: I think it's time we start extracting these disabled logic into functions for maintainability. But that is out of scope of this work. Just wondering if we should have an issue to track it.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1520,7 +1522,26 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+      socket
+      |> maybe_disable_canvas()
+
+    {:reply,
+     %{
+       workflow_params: socket.assigns.workflow_params,
+       run_steps: run_steps,
+       run_id: run_id,
+       history: history
+     }, socket}
```
elias-ba: We could do this, it's more elixir than updating the socket

```elixir 
    {:reply,
     %{
       workflow_params: socket.assigns.workflow_params,
       run_steps: run_steps,
       run_id: run_id,
       history: history
     }, maybe_disable_canvas(socket)}
  ```

File: lib/lightning_web/live/workflow_live/edit.ex [hunk 20]
```diff
@@ -2168,6 +2189,30 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+            started_at: step.started_at,
+            finished_at: step.finished_at
+          }
+        end)
+      end)
+
+    history = WorkOrders.get_workorders_with_runs(workflow_id, run_id)
+
+    %{run_steps: run_steps, history: history}
+  end
```
elias-ba: 1. I'm a bit concerned about the `Map.update!` call when `run_id` is nil. Since we return `%{steps: []}` in that case, the `Map.update!` will work, but we might be missing other expected keys. If `WorkOrders.get_run_steps/1` returns additional fields (like `:isTrigger`, `:start_from`, etc.), could this cause issues elsewhere?  Would something like this be safer?

```elixir
run_steps = 
  if run_id do
    WorkOrders.get_run_steps(run_id)
    |> Map.update!(:steps, &format_steps/1)
  else
    %{steps: [], isTrigger: nil, start_from: nil} # include all expected keys
  end
```

2. Could we extract the step transformation into a helper function for better readability?

```elixir
defp format_step(step) do
  %{
    job_id: step.job_id,
    error_type: step.error_type,
    exit_reason: step.exit_reason,
    started_at: step.started_at,
    finished_at: step.finished_at
  }
end

# Then use it like:
|> Map.update!(:steps, fn steps -> Enum.map(steps, &format_step/1) end)
```

3. I notice we're fetching run data twice - once in `get_run_steps` and potentially again in `get_workorders_with_runs`. Is there an opportunity to optimize this, or is it necessary for the business logic?

4. Just a thought - since this is a private function, the name is fine, but would `get_run_details_and_workflow_history` be more descriptive of what it actually does?

I love the structure and the separation of concerns between run steps and history. What do you think about these suggestions?

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -2802,13 +2851,41 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+    workflow_id = socket.assigns.workflow.id
+
+    %{run_steps: run_steps} =
+      get_run_steps_and_history(workflow_id, run_id)
+
+    snapshot = snapshot_by_version(workflow_id, version_tag)
+
+    # snapshots currently don't have positions
+    # if we don't do this, the patch wouldn't replace the existing position in params
+    updated_snapshot = Map.put(snapshot, :positions, nil)
```
elias-ba: Should we add `positions` as a field to the Snapshot schema instead of doing this manual `Map.put`? Doing this manual manipulation feels fragile and could break if the Snapshot structure changes. Having `positions` as an actual field with a default of `nil` would make the data structure more explicit and eliminate the need for this workaround.

Something like:

```elixir
schema "workflow_snapshots" do
  # ... existing fields ...
  field :positions, :map, default: nil
  # ... rest of schema ...
end
```

What do you think?

elias-ba (changes requested):
Hey @doc-han, this is really huge work! I loved reviewing it, nice work 👏🏽.

I left a few suggestions/questions throughout, nothing really blocking, as this seems to work well. Some of my questions are more related to maintainability, clarity, and performance, but they may be out of scope for this PR. In those cases, we can just open issues to track them separately rather than blocking this work.

Great job overall! Let me know if you'd like to discuss any of the points I raised.

lib/lightning/invocation.ex
doc-han: This function is dead code. to be removed.

lib/lightning/work_orders.ex
midigofrank: @doc-han can we move the mapping logic over to the Liveview? This file contains logic that can be reused/called out by other modules. 

Actually, I don't think we need this function here at all. I would recommend that you do this in the liveview.
You can get the run by calling `Lightning.Runs.get(run_id, include: [:created_by, :steps]` and then doing the mapping logic in the liveview

lib/lightning/work_orders.ex
```diff
@@ -484,6 +484,89 @@ defmodule Lightning.WorkOrders do
...
+          where: r.id == ^run_id and wo.workflow_id == ^workflow_id,
+          preload: [:snapshot, runs: [:steps]],
+          union: ^workorder_with_run
+        )
+      else
+        workorder_with_run
+      end
+
+    final_query
+    |> order_by(desc: fragment("last_activity"))
```
midigofrank: I think you can simplify the query using dynamic queries https://hexdocs.pm/ecto/dynamic-queries.html#building-dynamic-queries:

```elixir
include_selected_run = 
     if run_id do
        dynamic([wo, r], r.id == ^run_id)
     else
       true
      end


query =
      from(wo in WorkOrder,
        join: r in assoc(wo, :runs),
        where: wo.workflow_id == ^workflow_id,
        or_where: ^include_selected_run,
        preload: [:snapshot, runs: [:steps]],
        distinct: wo.id,
        order_by: [desc: wo.last_activity],
        limit: 20
      )
```

lib/lightning/work_orders.ex
```diff
@@ -484,6 +484,89 @@ defmodule Lightning.WorkOrders do
...
+    |> Repo.all()
+    |> Enum.map(fn worder ->
```
midigofrank: Can we move the mapping logic to the liveview?

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -2765,6 +2810,10 @@ defmodule LightningWeb.WorkflowLive.Edit do
       %{"method" => method, "m" => nil, "s" => nil, "a" => nil} ->
         handle_method_assignment(socket, method)
 
+      %{"m" => "history", "a" => run_id, "v" => version_tag, "s" => selected_id}
+      when not is_nil(version_tag) ->
```
midigofrank:
```suggestion
      when is_binary(version_tag) ->
```

taylordowns2000 (dismissed):
Looks great @doc-han ! It feels so stable and simple. (And I know it's _anything but simple_, so you should feel very proud that it now _feels_ simple.)

A few things:
1. please rebase on/merge `main` and resolve conflicts
2. the link to "go to the history page, filtered by this workflow" doesn't work and is not _quite_ vertically aligned (seems low compared to the text) as far as i can tell (image 1)
3. when there aren't many runs for a workflow, the table is still too large. it should have a max height, but not a min. (image 2)
4. when there is no activity, it should say "No related activity" rather than "No recent activity" - because "recent" makes it sound like there might be older activity. (image 3)

-------- new issues to create --------

5. please use the `.datetime` "relative" components (or use the same hook in react) for datetimes in the mini history table so we keep the standardization now on `main`?
6. which run does it show when there are 2? highest exit code? (can we show some sort of _count_ of how many runs there are in that step? image 4)
7. indicate when the loaded run doesn't match the nodes on `latest` and can't be fully visualized
8. use standard state `pills` for list, use standard step icons as `step_icons` on canvas for step exit codes
9. when a run is loaded, we should have the option to retry or create a new work order, exactly like the button on the inspector (image 5)

## image 1
Image

## image 2

Image

## image 3

Image

## image 4

Image

## image 5

Image

lib/lightning/invocation.ex
doc-han: dead code removed.

lib/lightning_web/live/workflow_live/edit.ex
doc-han: Yeah, Issue created. #3493

lib/lightning_web/live/workflow_live/edit.ex
doc-han: updated 👍🏽

lib/lightning_web/live/workflow_live/edit.ex
doc-han: This is cool because we'll obviously have to support node positions in snapshots too. 
with this I run a migration after right?

doc-han: resolved. added a migration file for snapshot that adds positions.

assets/js/workflow-diagram/MiniHistory.tsx
doc-han: ts compiler isn't smart enough to know that when length is 1 then index 0 should be accessible.

assets/js/workflow-diagram/nodes/Node.tsx
doc-han: Yeah makes sense. I'll try making this better. I think the general Node component has to expose props for adding these tiny icons to a node. so that either Trigger Node or Step Node could make use of that.

assets/js/workflow-editor/WorkflowEditor.tsx
doc-han: I'll remove the comment. This is a higher level enough for now. this makes the tooltip available in the entire canvas as soon as you provide props to any html, then it has tooltips. 

Example
```html
      <span
        data-tooltip={tooltip}
        data-tooltip-placement="top"
        className={`${STATE_ICONS[type]} w-full h-full`}
      ></span>
```

You can ignore the placement attribute and It'll auto place to the right.

assets/js/workflow-store/WorkflowStore.tsx
doc-han: yup

lib/lightning_web/live/workflow_live/edit.ex
doc-han: 1. Yeah we can add all the expected keys. [**done**]
2. Yeah that's possible. I don't see this being reused anywhere hence the reason I left it there but yes can be extracted. [**done**]
3. They're two difference cases. `get_run_steps` gets the steps of a run whilst `get_workorder_with_runs` gets several workorders and their corresponding runs.
- `get_workorder_with_runs` is for getting the workorder and runs to be shows as the history of the workflow
- `get_run_steps` gets the steps involved in a run when a user select a run to view.

4. `get_workorders_with_runs` (getting workorders with their runs) vs `get_run_details_and_workflow_history`. I'll for the first one. what do you say?

lib/lightning/work_orders.ex
doc-han: logic moved out into liveview and logic simplified. Thanks.

lib/lightning/work_orders.ex
doc-han: logic moved to liveview

lib/lightning/work_orders.ex
doc-han: tested. didn't seem to work. 

I've simplified the query. 
1. get recent 20 workorders
2. get workorder that contains the `run_id` when `run_id` isn't `nil`.
3. merge them with a union.

lib/lightning/work_orders.ex
doc-han: 1. query has been simplified now. gets last 20 recent workorders. also gets the workorder with the run if `run_id` isn't `nil`. merges them using union with a distinct workorder id. 
2. Nice catch. we actually don't use the steps in this way. we wait for a user to select a run before getting steps for that run. [**resolved**] Thanks	
3. wasn't trivial for me making this change. can be dealt with later. 
4. formatting extracted into liveview. [**resolved**]
5....8. Didn't want to have multiple trips to DB. single trip (**note: logic has been simplified now**)

assets/js/utils/timeSpent.ts
doc-han: I doubt. I recently had a discussion with @taylordowns2000 around this and I'm sure we'll have a look at these once again in the entire application

doc-han: function name updated from `timeSpent` to `duration`. 👍🏽

assets/js/utils/timeSpent.ts
doc-han: Accurate. the duration from liveview uses `ms/s/m` as mentioned. Updated.

assets/js/workflow-diagram/MiniHistory.tsx
doc-han: I'll just remove the comments and not extract into a component for now.

assets/js/workflow-diagram/nodes/Node.tsx
josephjclark: I don't think we need to do this now: a comment in the code reminding us to refactor later is fine :) More important to get the work merged

doc-han:
> Looks great @doc-han ! It feels so stable and simple. (And I know it's _anything but simple_, so you should feel very proud that it now _feels_ simple.)
> 
> A few things:
> 
> 1. please rebase on/merge `main` and resolve conflicts
> 2. the link to "go to the history page, filtered by this workflow" doesn't work and is not _quite_ vertically aligned (seems low compared to the text) as far as i can tell (image 1)
> 3. when there aren't many runs for a workflow, the table is still too large. it should have a max height, but not a min. (image 2)
> 4. when there is no activity, it should say "No related activity" rather than "No recent activity" - because "recent" makes it sound like there might be older activity. (image 3)
> 5. please use the `.datetime` "relative" components (or use the same hook in react) for datetimes in the mini history table so we keep the standardization now on `main`?
> 


1. synced with main [**resolved**]
2. go to history vertically aligned & implemented to work [**resolved**]
3. height of mini-history made to fit [**resolved**]
4. wording updated to "No related activity. [**resolved**]
5. Not clear on that yet

doc-han:
Issue created for updating datetime format on the history panel. 
#3499

josephjclark (approved):

priv/repo/migrations/20250806154047_add_positions_to_snapshots.exs
doc-han, line 1: @elias-ba 
Can you double check this migration file? adds a positions field to the "workflow_snapshots" schema with default being nil. 

I also added a corresponding field in the schema file.

priv/repo/migrations/20250806154047_add_positions_to_snapshots.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Lightning.Repo.Migrations.AddPositionsToSnapshots do
+  use Ecto.Migration
+
+  def change do
+    alter table(:workflow_snapshots) do
+      add :positions, :map, null: true
+    end
+  end
+end
```
elias-ba, line 9: Lovely ❤️

elias-ba:
Hey @doc-han, my commit ([6b9aec3](https://github.com/OpenFn/lightning/pull/3382/commits/6b9aec3770dfae533dff8c3bee3f4cd052f3279d) here does two things:

### 1. Add positions to snapshot creation

I added `positions` to the cast and build functions of the snapshot data model, this allows workflow positions to be  preserved when creating snapshots

### 2. Fix flaky ordering in `get_workorders_with_runs/2`

While adding tests for the workorders query functions, I discovered the test `"orders workorders by last_activity descending"` was flaky - it would pass/fail randomly.

**The issue:** The original function used `order_by(desc: fragment("last_activity"))` which caused non-deterministic ordering due to:
- Ambiguous column references when joining `work_orders` with `runs`
- Conflicts between `DISTINCT` and `ORDER BY` clauses
- Complex UNION queries when `run_id` is provided

**The fix:** I rewrote the function to:
- Eliminate UNION queries and fragments
- Use explicit column references
- Handle "with run_id" and "without run_id" cases separately
- Ensure deterministic ordering

This test proves the fix - it fails with the original implementation but passes consistently with the rewrite.

https://github.com/OpenFn/lightning/blob/6b9aec3770dfae533dff8c3bee3f4cd052f3279d/test/lightning/work_orders_test.exs#L2766

**Please verify:** The UI still works correctly with these changes. I did some manual testing and everything looked good, but would appreciate if you could double-check the workorders display still behaves as expected.

Feel free to review the code in the commit and let me know if you have any concerns about the approach.

elias-ba:
In this commit ([6d0e031](https://github.com/OpenFn/lightning/pull/3382/commits/6d0e031f60196f62b38d60a1509f68758b47c935)), I am adding few liveview tests to increase coverage in the Workflow Edit module.

elias-ba (approved):
Great work @doc-han, this feature is a banger

elias-ba:
You need to add a Changelog line before merging this

midigofrank (changes requested):
Nicely done @doc-han , this now look ready to go. I just have one last request, can we get rid of `WorkOrders.get_run_steps/1`:
```elixir

  def get_run_steps(run_id) do
    Run
    |> where([r], r.id == ^run_id)
    |> preload([:created_by, :steps])
    |> Repo.one()
  end
 ```
 
Given that this function fetches a run, we can utilize the `Runs.get/2`:
```elixir
Runs.get(run_id, include: [:created_by, :steps])
```

doc-han:
> Nicely done @doc-han , this now look ready to go. I just have one last request, can we get rid of `WorkOrders.get_run_steps/1`:
> 
> ```elixir
>   def get_run_steps(run_id) do
>     Run
>     |> where([r], r.id == ^run_id)
>     |> preload([:created_by, :steps])
>     |> Repo.one()
>   end
> ```
> 
> Given that this function fetches a run, we can utilize the `Runs.get/2`:
> 
> ```elixir
> Runs.get(run_id, include: [:created_by, :steps])
> ```

Resolved. I've removed the function and used `Runs.get` directly as suggested. 
[ba14ab0](https://github.com/OpenFn/lightning/pull/3382/commits/ba14ab0c602028b737099daddcc53240aceb3c34)
Thanks man.

midigofrank (approved):
Nicely done

doc-han:
> Hey @doc-han, my commit ([6b9aec3](https://github.com/OpenFn/lightning/pull/3382/commits/6b9aec3770dfae533dff8c3bee3f4cd052f3279d) here does two things:
> 
> ### 1. Add positions to snapshot creation
> I added `positions` to the cast and build functions of the snapshot data model, this allows workflow positions to be preserved when creating snapshots
> 
> ### 2. Fix flaky ordering in `get_workorders_with_runs/2`
> While adding tests for the workorders query functions, I discovered the test `"orders workorders by last_activity descending"` was flaky - it would pass/fail randomly.
> 
> **The issue:** The original function used `order_by(desc: fragment("last_activity"))` which caused non-deterministic ordering due to:
> 
> * Ambiguous column references when joining `work_orders` with `runs`
> * Conflicts between `DISTINCT` and `ORDER BY` clauses
> * Complex UNION queries when `run_id` is provided
> 
> **The fix:** I rewrote the function to:
> 
> * Eliminate UNION queries and fragments
> * Use explicit column references
> * Handle "with run_id" and "without run_id" cases separately
> * Ensure deterministic ordering
> 
> This test proves the fix - it fails with the original implementation but passes consistently with the rewrite.
> 
> https://github.com/OpenFn/lightning/blob/6b9aec3770dfae533dff8c3bee3f4cd052f3279d/test/lightning/work_orders_test.exs#L2766
> 
> **Please verify:** The UI still works correctly with these changes. I did some manual testing and everything looked good, but would appreciate if you could double-check the workorders display still behaves as expected.
> 
> Feel free to review the code in the commit and let me know if you have any concerns about the approach.

I've tested several times and everything seems to work just fine. Thanks for this @elias-ba 
really appreciated. ❤️

doc-han:
This PR actually closes #3387 which is a sub-issue of #3103

PR #3384, opened 2025-07-10
add keystrokes to run from canvas

### LINKED ISSUE #3320: Play Button: ctrl + enter should open run dialog
This is a ticket for 8 hours of refactoring proposed on #3384 

... and pressing again should trigger the run

Probably need a tooltip somewhere with the shortcut. What do we do in the inspector?

https://docs.openfn.org/documentation/keyboard-shortcuts

taylordowns2000:
## Description

This PR closes #3320 .

Merge docs when this is merged: https://github.com/OpenFn/docs/pull/683

## Validation steps

1. Go to the canvas, press control-enter (or command-enter on mac) to open the "Run" dialog.
2. Press it again to run with the selected input
3. Go back to the canvas, select a step and do it again, this time noting that you're running from a specific step rather than from the trigger.
4. Check to see that all other keybindings still work: https://docs.openfn.org/documentation/keyboard-shortcuts

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/hooks/KeyHandlers.ts
+++ b/assets/js/hooks/KeyHandlers.ts
@@ -160,7 +160,7 @@ function createKeyCombinationHook(
 
         const maxPriority = Math.max(...matchingHandlers.map(h => h.priority));
         const topPriorityHandlers = matchingHandlers.filter(
-          h => h.priority === maxPriority
+          h => h.priority === maxPriority as PriorityLevel
         );
 
         // Take the last handler if there are more than one with the same priority.
@@ -370,7 +370,7 @@ export const CloseNodePanelViaEscape = createKeyCombinationHook(
 );
 
 /**
- * Handles Ctrl+Enter to directly click the appropriate run button based on URL state.
+ * Handles Ctrl+Enter to trigger run actions directly based on current state.
  *
  * BEHAVIOR (based purely on URL state):
  * 1. If in inspector (URL contains 'm=expand'):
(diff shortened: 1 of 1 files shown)


taylordowns2000:
@theroinaochieng , can i have a UX review from you _before_ I ask for a code review?

theroinaochieng:
Sure thing @taylordowns2000 taking a look in about 20 minutes.

assets/js/hooks/KeyHandlers.ts
```diff
@@ -366,3 +368,84 @@ export const CloseNodePanelViaEscape = createKeyCombinationHook(
...
+ * 1. If in inspector (URL contains 'm=expand'):
+ *    → Click #save-and-run button to execute the workflow
+ * 2. If in run panel (URL contains 'm=workflow_input'):
+ *    → Click #run-from-input-selector button to execute the workflow
+ * 3. If step selected but not in inspector (URL contains 's=' but no 'm=expand'):
+ *    → Click the appropriate run button (#run-from-step or #run-from-trigger)
+ * 4. If no step selected and no panel:
+ *    → Click #run-from-top button to run from trigger
+ */
+const openRunPanelAction = () => {
```
doc-han, line 385: To run the workflow from the run-panel. we don't need to search for the run button using selectors which will be problematic with time. We seem to already have a function in the js world that does that. It'll be best if we can register our key events close to this function and then just call it directly instead of depending on searching for a button.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -213,8 +213,12 @@ defmodule LightningWeb.WorkflowLive.Edit do
           class="transition-all duration-300 ease-in-out"
         />
         <div
-          class={"relative h-full flex grow transition-all duration-300 ease-in-out #{if @show_new_workflow_panel, do: "w-2/3", else: ""}"}
+          class={"relative h-full flex grow transition-all duration-300 ease-in-out #{if @show_new_workflow_panel, do: "w-2/3", else: ""} focus:outline-none"}
           id={"workflow-edit-#{@workflow.id}"}
+          phx-hook="OpenRunPanelViaCtrlEnter"
```
doc-han, line 218: This keyhandler registration doesn't seem to get unregistered.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -213,8 +213,12 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
+          data-keybinding-scope="workflow-editor"
+          tabindex="0"
+          phx-mounted={JS.focus()}
```
doc-han, line 221: wondering why we need to focus here.

lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -225,13 +229,17 @@ defmodule LightningWeb.WorkflowLive.Edit do
...
                 "fixed left-0 top-0 right-0 bottom-0 flex-wrap",
                 "hidden opacity-0",
-                "bg-white inset-0 z-30 overflow-hidden drop-shadow-[0_35px_35px_rgba(0,0,0,0.25)]"
+                "bg-white inset-0 z-30 overflow-hidden drop-shadow-[0_35px_35px_rgba(0,0,0,0.25)] focus:outline-none"
               ]}
               phx-mounted={fade_in()}
               phx-remove={fade_out()}
+              data-keybinding-scope="workflow-editor"
+              tabindex="0"
+              phx-hook="AutoFocusOnMount"
```
doc-han, line 242: another focusing of the inspector. wondering once again.

assets/js/hooks/KeyHandlers.ts
```diff
@@ -366,3 +368,84 @@ export const CloseNodePanelViaEscape = createKeyCombinationHook(
...
+      if (runFromTriggerButton instanceof HTMLElement) {
+        runFromTriggerButton.click();
+      }
+    }
+  } else {
+    // No step selected - trigger the same navigation as the run button
+    const runButton = document.querySelector('#run-from-top:not([disabled])');
+    if (runButton instanceof HTMLElement) {
+      runButton.click();
+    }
```
doc-han, line 427: Using ids here is brittle. we might want to actually call events directly when they exist.

assets/js/hooks/KeyHandlers.ts
```diff
@@ -366,3 +368,84 @@ export const CloseNodePanelViaEscape = createKeyCombinationHook(
...
+  
+  // Only work on workflow pages
+  if (!url.includes('/w/')) {
+    return;
+  }
+  
+  // Parse URL state
+  const hasStepSelected = url.includes('s=');
+  const isInInspector = url.includes('m=expand');
+  const isInRunPanel = url.includes('m=workflow_input');
```
doc-han, line 396: I don't think we've ever made keyHandlers depend on url params. It's mostly dependent on element that are available when we're on the screen we want the key events to be active. 

eg. Undo/Redo Key handlers are registered when the workflowDiagram is mounted and unregistered when it's unmounted. Hence we wouldn't have to depend on query params.

doc-han (changes requested):
This doesn't feel like the right approach. 
I think
- First ctrl + Enter to open the runPanel is fine living in the elixir world. 
- The second ctrl + Enter to click the run button seems too off since we have to search for the button using selectors. I guess this is because the approach we have in `KeyHandlers.ts` is for registering handlers in elixir but we now have part of our code living in the JS world. Hence, let's register events for them in the JS way. in that case we wouldn't have to search for a button to click but instead call the event/function performing the action of the button directly.

josephjclark:
To resolve the issue of Liveview and React both handling keyboard events (and different events at different times), we may need to build out a dedicated headless controller in React. Probably it catches the events first and decides whether to handle or delegate.

Not convinced we need to do that NOW, but maybe?

Let's have a design session and work out the best way forward

taylordowns2000:
> To resolve the issue of Liveview and React both handling keyboard events (and different events at different times), we may need to build out a dedicated headless controller in React. Probably it catches the events first and decides whether to handle or delegate.
> 
> Not convinced we need to do that NOW, but maybe?
> 
> Let's have a design session and work out the best way forward

Hello there @josephjclark !

**_Please note_** (pulling this in from Slack in case you're taking a peek) that the current implementation (here in this PR) works 95% the way we want but the 5% that I don't like makes me feel like something is wrong.

Specifically, it all works perfectly in behavior, but in _style_ the node (styled by `nodeIconStyles`) loses its `EDGE_COLOR_SELECTED` when you pick a node and press ctrl-enter. You get to the right screen, all the logic works, but it's as if the `nodeIconStyles` doesn't know that you're still looking at the same node and it loses its purple borders.

This _doesn't_ happen when you actually click the "Run" button after selecting a step. Only when you press Ctrl-Enter after selecting a step.

So.... it's spooky and I don't like it. Other than that disappearing purple border on the selected node, I'm happy to merge. (But the way it disappears makes me think there's some deeper problem.)

josephjclark:
@taylordowns2000 Yes, there is some weird stuff in the diff around focus which may account for that.

I'm 100% behind Farhan here and I _wouldn't_ want to merge as is: that key handler makes a lot of assumptions about the shape and structure of the app, and I don't want to have to maintain it every time we add new pages, or rebuild existing ones. I don't think there's any urgency to merge bad code.

I'll look at this with the team this week and we'll work out a more sustainable solution. I don't want it to be super expensive - it just needs a bit more thought

doc-han:
**Note:** When the AI chat panel is open and the textarea is active, **CTRL + ENTER** is for submitting the content in the textarea. Let's not forge to test that.

PR #3409, opened 2025-07-15
Name a dataclip

### LINKED ISSUE #311: Name a global dataclip
### User story

As an admin that wants to run a job with an arbitrary dataclip (#238) I want to be able to select that input dataclip from a list of saved and _named_ `globals`.

Consider the two types of globals:

**Type 1:** There are inputs I want to save to test my jobs. These, when saved, form a “test suite” of cases that we’d like to validate for a particular job in a flow. They are results from previous runs (save as a test input for job3 in the flow) and saved webhook requests.

**Type 2:** There are constants, or blobs of JSON data, that non-technical users want to build via a simple UI and, most commonly, store “mapping” data that links, e.g. a clinic code to it’s underlying clinic name.

### Details

Currently, there is a picklist which allows the user to change the type of a dataclip. This picklist should dissappear. Instead, they should have a button which says "save as global" and if they do so, they should be prompted to give that dataclip a name which follows the same naming pattern as projects (i.e., it's a slug.) and be able to edit the body of the dataclip.

Once it's saved as a global, they can **_not_** convert it back to a `run_result` or a `http_request`.

Once it's saved as a global, they **_can_** set the name to null, effectively removing it from the "named globals" list.

### Implementation notes

- [ ] add a nullable column on the dataclips table called `name`
- [ ] add this nullable name to the dataclip model
- [ ] disable the body editor when type is `run_result` or `http_request`
- [ ] remove the type picklist
- [ ] add a "Save as Global" button which, when clicked makes the `body` editable and displays the `name` field... from this altered view of the form the user may still "Cancel" or "Save".

### QA

- [ ] This has been QA’d by Amber, branch can be merged after review

midigofrank:
## Description

Closes #311 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


midigofrank:
@taylordowns2000 as always, your feature breaking skills are always appreciated

midigofrank:
@doc-han how do you rate my gpt-assisted React skills?

taylordowns2000 (changes requested):
UI looks great!

1. please add an audit trail event for the creation and removal of a label, then i think we're good
2. please add a flash message when a label is added "Label created. Dataclip will be saved permanently."
3. please add a flash message when a label is removed "Label removed. Dataclip will be removed when past your retention policy limit."

midigofrank:
@taylordowns2000 I've added the audits and flash messages

taylordowns2000 (changes requested):
This is looking great, but it seems as though the search doesn't return a named dataclip when that named dataclip is the _next_ cron run. In this image, I'd expect "long" to match and it does. I'd also expect "here" to match, but it returns no results.

image

This isn't likely a big problem, as the next cron run will almost never be named/saved. But it's still strange enough to warrant some investigation. Please see if you can get it to return when searching for the name of the next cron run.

midigofrank:
@taylordowns2000 this is likely the culprit https://github.com/OpenFn/lightning/pull/3409/files#diff-b400713bb29056407bd66ecb6e7a8663bde8d8adbf8a5a8f68f6bf3ffb8366eaR879

midigofrank:
I have run into a situation where if the query is of base 16, like `abc`, the filter will only search for `uuid`s. Because of this, I'm changing the `id_prefix` filter to be `name_or_id_part`.

midigofrank:
Good catch @taylordowns2000 , the error was being caused by trying to match against the next-cron dataclip that doesn't have a name.

I have fixed the bug and added a test

elias-ba (approved):
Yo dude! this is a lot of JS/TS lol I love this. I clicked tested a bit and look at the code also. Everything is looking good to me. Nice implementation btw 👏🏽 ❤️

taylordowns2000 (approved):
thanks frank!

PR #3439, opened 2025-07-18
partial string (not prefix) search for dataclip name

taylordowns2000:
hey @midigofrank , the name search shouldn't just be prefix, that's only for performance on the UUIDs.

this is the concept i'm looking for, but the implementation isn't quite right. can you check it out? (or re-do these changes on your own branch to allow partial string search on dataclip names?)


midigofrank (approved):
Good catch

PR #3451, opened 2025-07-25
Keychain Credentials

### LINKED ISSUE #3359: Bring KeyChain credentials to v2
https://github.com/OpenFn/lightning/blob/1035cb16181fbdcdf00a013884374e168468d71f/lib/lightning_web/channels/run_channel.ex#L88-L89

- we have the run
- we have the credential uuid

when calling `Runs.get_credential(run, id)` we have access to the credential ID and the run (with its input dataclip)

if the credential is a keychain credential, we _evaluate_ and then return the Credential.body for whatever it points to.

stuartc:
## Description

Keychain Credentials enable dynamic credential selection for jobs based on runtime data.
Instead of being locked to a single credential, jobs can automatically choose the
appropriate credential using JSONPath expressions that evaluate incoming data.

Database Layer:
- KeychainCredential schema with name, path (JSONPath), and optional default_credential_id
- Job schema updated with keychain_credential_id field (mutually exclusive with
project_credential_id)
- Database constraints ensuring project scope and single credential type per job

Backend:
- Lightning.Credentials context functions for CRUD operations
- JSONPath validation ensuring expressions start with $
- Permission policies restricting access to admin/owner roles

Frontend:
- KeychainCredentialFormComponent for create/edit modals with JSONPath input
- CredentialIndexComponent updated to display keychain credentials table
- Project settings integration showing keychain credentials alongside regular credentials
- New credential dropdown menu with "Keychain Credential" option

Using it:
1. Admin/editor/support user creates keychain credential with JSONPath expression (e.g., $.user_id)
2. Job configured to use keychain instead of static credential
3. At runtime, JSONPath extracts value from input dataclip to select credential
4. Falls back to default credential if no match found

Closes #3359 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


mix.exs
```diff
@@ -99,6 +99,7 @@ defmodule Lightning.MixProject do
       {:jsonpatch, "~> 2.2"},
       {:junit_formatter, "~> 3.0", only: [:test]},
       {:libcluster, "~> 3.3"},
+      {:live_debugger, "~> 0.3.0", only: :dev},
```
midigofrank, line 102: 🤩

mix.exs
```diff
@@ -130,6 +131,7 @@ defmodule Lightning.MixProject do
       {:telemetry_metrics, "~> 1.0"},
       {:telemetry_poller, "~> 1.0"},
       {:tesla, "~> 1.13"},
+      {:tidewave, "~> 0.1.0", only: :dev},
```
midigofrank, line 134: 🤩

lib/lightning_web/live/credential_live/credential_index_component.ex
```diff
@@ -98,34 +158,40 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
...
-    with :ok <- can_edit_credential(socket, client) do
+    if can_edit_credential(socket.assigns.current_user, client) do
       {:noreply,
        assign(socket,
          active_modal: :edit_oauth_client,
          credential: nil,
          oauth_client: client
        )}
+    else
+      not_authorized(socket)
```
midigofrank, line 178: 🤩

midigofrank (approved):
This looks great @stuartc, can't wait to try out `Tidewave` and the `LiveDebugger`.

I really love the use of `not_authorized(socket)` function here, reminds me of this article I read of having `noreply and ok` helpers: https://www.elixirstreams.com/tips/awesome-ok-noreply-liveview-helpers

I think we should add them

lib/lightning_web/live/credential_live/credential_index_component.ex
stuartc: Ha! Glad you like it, it actually comes from your work, I cut down the logic and renamed it... so you started it. <3

stuartc: And yes to adding common reply patterns, when doing controller responses, there are ton of wrappers for conn we use without thinking, we should at least have those two from that post, and a not_authorized one - but yeah I know why we haven't tried very hard on this so far; it's easy to find exceptions to the rule, but yeah paying attention to the exceptions and trying to mold them back into common patterns should be the goal instead.

stuartc:
@taylordowns2000 I think I've sorted this out.

Was pretty tricky at first, but I observed the credential being "changed" magically - I believe it had to do with the new credential dropdown having stale element references and so it was hanging onto the first element that arrived.

I've fixed up the CredentialSelector hook and its references and events seem to be stable now.

I also thought the credential resolution may be sensitive to integer values in the data (external ids are strings), but turns out my query accommodated for that already; but I added a test for it anyway.

I've fixed the "current credential" block in the JobEditView, it was only referencing project credentials, so needed a case statement.

And lastly, I've found a bug on the MetadataService, it was throwing errors when trying to get metadata when a Keychain Credential is being used - so I've added a fall through clause because right now we can't use keychains with the MetadataService (we don't have a data clip context easily available), so for now it's not supported but it won't crash.

I did also notice that metadata doesn't use Oauth credentials properly, I've made a bug over [here](https://github.com/OpenFn/lightning/issues/3466).

taylordowns2000 (approved):
looks great. thanks, @stuartc

PR #3485, opened 2025-08-05
show alert box before removing a project in credential form

### LINKED ISSUE #537: Alert user when they are removing project access for an actively used credential
### User story
As a user that is removing project access for a credential, I would like to be alerted if that project is using the credential in a workflow, so that I don't accidentally break a workflow / so that I can go and change the credential in that workflow if need be. 

### Details
In credentials > edit > project access, when a user clicks 'remove', check if the credential is being used in any workflows for that project. If so, give them an alert that says 'Are you sure you want to remove access? This credential is currently being used in <workflow_name>'. 
If the user clicks 'Yes', remove access. If the user clicks 'Cancel', do nothing. 

### Release notes
- Alert user if a credential is being used before they remove access to a project.

midigofrank:
## Description

This PR shows an alert box if the project being removed from the credential is using the credential in one of it's workflows

Closes #537 

## Validation steps

1. Ensure you have 2 projects, each having workflows
2. Create a credential. Link it to both projects
3. In one of the 2 projects, associate the credential with a job in the project workflows.
4. Go back to edit the credential
5. Try removing the project referenced in step 3 above
6. A browser alert box should pop-up saying the credential is being used in some workflows. It should separate the workflow names using comma if they're more than one.
7. Now try removing the other project (the one that didn't associate any job to the credential)
8. The alert box should not pop up

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


lib/lightning/workflows.ex
```diff
@@ -57,6 +57,27 @@ defmodule Lightning.Workflows do
...
   end
 
+  @spec workflow_names_using_project_credentials_per_project([
+          Ecto.UUID.t(),
+          ...
+        ]) ::
+          %{
+            optional(project :: Ecto.UUID.t()) => [workflow :: binary(), ...]
+          }
+  def workflow_names_using_project_credentials_per_project(project_credentials) do
```
elias-ba: Can we consider shortening a little bit this function name to `workflows_by_project_for_credentials(credential_ids)`? The current name feels a bit long (not bad tho, just trying to see if we can do better), and `credential_ids` might be clearer than `project_credentials` since we're passing IDs rather than full credential records.

lib/lightning/workflows.ex
```diff
@@ -57,6 +57,27 @@ defmodule Lightning.Workflows do
...
+      from w in Workflow,
+        join: j in assoc(w, :jobs),
+        where: j.project_credential_id in ^project_credentials,
+        select: %{name: w.name, project_id: w.project_id},
+        distinct: true
+
+    query
+    |> Repo.all()
+    |> Enum.group_by(& &1.project_id)
+    |> Map.new(fn {key, val} -> {key, Enum.map(val, & &1.name)} end)
```
elias-ba: Can we simplify this to just `Enum.group_by(& &1.project_id, & &1.name)`? The 3-arity version of group_by already gives us the exact structure we need here.

test/lightning_web/live/credential_live_test.exs
```diff
@@ -1109,6 +1109,75 @@ defmodule LightningWeb.CredentialLiveTest do
...
+      assert button_html =~
+               "data-confirm=\"This credential is in use by workflows #{workflow_with_credential_1.name}, #{workflow_with_credential_2.name}. Revoking access to #{project.name} will likely break those existing workflows"
+
+      assert button_html =~ workflow_with_credential_1.name
+      assert button_html =~ workflow_with_credential_2.name
+      refute button_html =~ workflow_without_credential.name
+
+      refute delete_credential_button(view, another_project.id) |> render() =~
+               "data-confirm"
+    end
```
elias-ba: Lovely ❤️

lib/lightning/workflows.ex
midigofrank: Good, catch. Thanks

lib/lightning/workflows.ex
midigofrank: I didn't want to use `credentials` here because there is an entity called `Credential` and an entity called `ProjectCredential`. In this case I'm targeting the `ProjectCredential`

midigofrank: Okay, how about: `project_workflows_using_credentials`

midigofrank: I was trying to shorten the variable names as well, that's why I din't append the `_ids` there. But the typespecs show exactly what they are

elias-ba: Yeah this is great Frank. I like `project_workflows_using_credentials`.

PR #3489, opened 2025-08-05
Fix PostgreSQL UTF-8 error when inserting log lines with null bytes

### LINKED ISSUE #3090: Postgrex UTF-8 error when inserting a log line
Sentry Issue: [LIGHTNING-XA](https://openfn.sentry.io/issues/6509410920/?referrer=github_integration)

```
Postgrex.Error: ERROR 22021 (character_not_in_repertoire) invalid byte sequence for encoding "UTF8": 0x00
  File "lib/ecto/adapters/sql.ex", line 1096, in Ecto.Adapters.SQL.raise_sql_call_error/1
  File "lib/ecto/repo/schema.ex", line 837, in Ecto.Repo.Schema.apply/4
  File "lib/ecto/repo/schema.ex", line 416, in anonymous fn/15 in Ecto.Repo.Schema.do_insert/4
  File "lib/ecto/repo/schema.ex", line 1099, in anonymous fn/3 in Ecto.Repo.Schema.wrap_in_transaction/6
  File "lib/ecto/adapters/sql.ex", line 1400, in anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
...
```

elias-ba:
## Description

This PR fixes a PostgreSQL error that occurs when log messages contain null bytes or other control characters. 

When workers send log messages containing null bytes (`\x00`) or other control characters, PostgreSQL rejects them with error 22021: "invalid byte sequence for encoding UTF8". While these characters are technically valid UTF-8, PostgreSQL's text fields cannot store them due to internal C-style string handling.

The fix adds UTF-8 sanitization directly to the `LogMessage` Ecto type, ensuring all messages are cleaned before they reach the database. Invalid characters are replaced with the Unicode replacement character (�) to maintain visibility that sanitization occurred.

Closes #3090

## Validation steps

1. Run the new tests: `mix test test/lightning/invocation/log_line_test.exs`
2. Try inserting a log line with null bytes:
   ```elixir
   # In iex -S mix:
   run = Lightning.Repo.get(Lightning.Run, "some-run-id")
   attrs = %{
     message: "Test message with \x00 null byte",
     timestamp: DateTime.utc_now() |> DateTime.to_unix(:millisecond),
     level: :info
   }
   changeset = Lightning.Invocation.LogLine.new(run, attrs, nil)
   Lightning.Repo.insert!(changeset)
   # Should succeed and show "Test message with � null byte"
   ```
3. Verify JSON messages with control characters work:
   ```elixir
   attrs = %{
     message: %{"error" => "Failed\x00here", "code" => 500},
     timestamp: DateTime.utc_now() |> DateTime.to_unix(:millisecond),
     level: :error
   }
   changeset = Lightning.Invocation.LogLine.new(run, attrs, nil)
   Lightning.Repo.insert!(changeset)
   # Should succeed with JSON containing � instead of \u0000
   ```

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ecto_types.ex
+++ b/lib/lightning/ecto_types.ex
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
 
   In the case of JSON objects we serialize them to a string, and in the case of
   arrays we serialize them individually and join them with a space.
+
+  This type also sanitizes strings to remove characters that PostgreSQL text
+  fields cannot store (null bytes and certain control characters).
   """
   use Ecto.Type
+
+  # Regex to match PostgreSQL-incompatible characters
+  # Matches: NULL (0x00), control chars (0x01-0x08, 0x0B, 0x0C, 0x0E-0x1F), DEL (0x7F)
+  @invalid_chars_regex ~r/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/
+
+  @unicode_escape_regex ~r/\\u00[0-8B-CE-F][0-9A-F]|\\u007F/i
+
+  @replacement_char "�"
+
   def type, do: :string
 
-  def cast(d) when is_binary(d), do: Ecto.Type.cast(:string, d)
-
-  def cast(d) when is_integer(d),
-    do: Ecto.Type.cast(:string, d |> Integer.to_string())
-
-  def cast(d) when is_boolean(d),
-    do: Ecto.Type.cast(:string, d |> to_string())
-
-  def cast(d) when is_float(d),
-    do: Ecto.Type.cast(:string, d |> Float.to_string())
-
-  def cast(d) when is_list(d) do
-    {:ok,
-     d
-     |> Enum.map(&cast/1)
-     |> Enum.map(fn {:ok, v} -> v end)
-     |> Enum.intersperse(" ")
-     |> IO.iodata_to_binary()}
+  def cast(d) when is_binary(d) do
+    {:ok, sanitize_string(d)}
   end
 
-  def cast(d) when is_map(d) or is_list(d) or is_nil(d) do
-    Jason.encode(d)
+  def cast(d) when is_integer(d) do
+    d |> Integer.to_string() |> cast()
+  end
+
+  def cast(d) when is_boolean(d) do
+    d |> to_string() |> cast()
+  end
+
+  def cast(d) when is_float(d) do
+    d |> Float.to_string() |> cast()
+  end
+
+  def cast(d) when is_list(d) do
+    sanitized_parts =
+      d
+      |> Enum.map(fn item ->
+        case cast(item) do
+          {:ok, v} -> v
+          _ -> ""
+        end
+      end)
+      |> Enum.intersperse(" ")
+      |> IO.iodata_to_binary()
+
+    {:ok, sanitized_parts}
+  end
+
+  def cast(d) when is_map(d) or is_nil(d) do
+    case Jason.encode(d) do
+      {:ok, json_string} ->
+        {:ok, sanitize_json_string(json_string)}
+
+      {:error, _} = error ->
+        error
+    end
   end
 
   def load(d) do
     Ecto.Type.load(:string, d)
   end
 
+  def dump(d) when is_binary(d) do
+    {:ok, sanitize_string(d)}
+  end
+
   def dump(d) do
     Ecto.Type.dump(:string, d)
+  end
+
+  defp sanitize_string(string) when is_binary(string) do
+    string
+    |> String.replace(@invalid_chars_regex, @replacement_char)
+  end
+
+  defp sanitize_string(other), do: other
+
+  defp sanitize_json_string(json_string) when is_binary(json_string) do
+    json_string
+    |> sanitize_string()
+    |> String.replace(@unicode_escape_regex, @replacement_char)
   end
 end
--- a/test/lightning/ecto_types_test.exs
+++ b/test/lightning/ecto_types_test.exs
@@ -39,6 +39,75 @@ defmodule Lightning.EctoTypesTest do
       assert {:ok, ~s<false>} =
                LogMessage.cast(false)
     end
+
+    test "sanitizes null bytes in strings" do
+      assert {:ok, "Hello�World"} = LogMessage.cast("Hello\x00World")
+      assert {:ok, "���"} = LogMessage.cast("\x00\x00\x00")
+    end
+
+    test "sanitizes control characters" do
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x01here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x02here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x03here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x04here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x05here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x06here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x07here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x08here")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x0Bhere")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x0Chere")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x0Ehere")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x0Fhere")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x1Fhere")
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x7Fhere")
+    end
+
+    test "preserves valid control characters" do
+      assert {:ok, "Text\there"} = LogMessage.cast("Text\there")
+      assert {:ok, "Text\nhere"} = LogMessage.cast("Text\nhere")
+      assert {:ok, "Text\rhere"} = LogMessage.cast("Text\rhere")
+    end
+
+    test "sanitizes null bytes in lists" do
+      assert {:ok, "Hello �World Test"} =
+               LogMessage.cast(["Hello", "\x00World", "Test"])
+    end
+
+    test "sanitizes null bytes in JSON maps" do
+      input = %{"message" => "Error\x00here", "level" => "error"}
+      {:ok, result} = LogMessage.cast(input)
+
+      assert result =~ "Error�here"
+      assert result =~ "error"
+      assert result == ~s<{"level":"error","message":"Error�here"}>
+    end
+
+    test "sanitizes Unicode escape sequences in JSON" do
+      # When Jason encodes control characters, it uses \u00XX format
+      # This test ensures we handle those too
+      input = %{"text" => "Has\x01control\x00chars"}
+      {:ok, result} = LogMessage.cast(input)
+
+      # Should replace both \u0001 and \u0000 with �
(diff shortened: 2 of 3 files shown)


lib/lightning/ecto_types.ex
```diff
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
...
+  This type also sanitizes strings to remove characters that PostgreSQL text
+  fields cannot store (null bytes and certain control characters).
   """
   use Ecto.Type
+
+  # Regex to match PostgreSQL-incompatible characters
+  # Matches: NULL (0x00), control chars (0x01-0x08, 0x0B, 0x0C, 0x0E-0x1F), DEL (0x7F)
+  @invalid_chars_regex ~r/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/
+
+  @unicode_escape_regex ~r/\\u00[0-8B-CE-F][0-9A-F]|\\u007F/i
```
rorymckinley, line 89: I like the addition of the `i` flag, but I can see someone (me) accidentally leaving it off one day - there does not seem to be a  test covering this - is there a cheap way to generate mixed/upper/lower-case escape sequences?

lib/lightning/ecto_types.ex
```diff
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
...
   end
 
-  def cast(d) when is_map(d) or is_list(d) or is_nil(d) do
-    Jason.encode(d)
+  def cast(d) when is_map(d) or is_nil(d) do
+    case Jason.encode(d) do
+      {:ok, json_string} ->
+        {:ok, sanitize_json_string(json_string)}
+
+      {:error, _} = error ->
```
rorymckinley: Given that Jason encode only returns {:ok, string} or {:error, error}, would it be ok to just have `error => error` here?

lib/lightning/ecto_types.ex
```diff
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
...
+
   def load(d) do
     Ecto.Type.load(:string, d)
   end
 
+  def dump(d) when is_binary(d) do
+    {:ok, sanitize_string(d)}
+  end
+
   def dump(d) do
```
rorymckinley, line 146: I think I am just missing context here,  should the default `dump` not also be doing sanitisation? The Internet says it is is possible to get to `dump` without necessarily having called `cast`. For that matter - if the problem lies with Postgres would it not be cleaner to only do the sanitisation in the `dump` methods? This would defer the 'data loss' until it is absolutely unavoidable (database says "nyet!")?

lib/lightning/ecto_types.ex
```diff
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
...
+  @invalid_chars_regex ~r/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/
+
+  @unicode_escape_regex ~r/\\u00[0-8B-CE-F][0-9A-F]|\\u007F/i
+
+  @replacement_char "�"
+
   def type, do: :string
 
-  def cast(d) when is_binary(d), do: Ecto.Type.cast(:string, d)
+  def cast(d) when is_binary(d) do
```
rorymckinley, line 95: What if the string is serialised JSON - would this produce an undesirable result?

lib/lightning/ecto_types.ex
```diff
@@ -76,39 +76,85 @@ defmodule Lightning.LogMessage do
...
+
+    {:ok, sanitized_parts}
   end
 
-  def cast(d) when is_map(d) or is_list(d) or is_nil(d) do
-    Jason.encode(d)
+  def cast(d) when is_map(d) or is_nil(d) do
+    case Jason.encode(d) do
+      {:ok, json_string} ->
+        {:ok, sanitize_json_string(json_string)}
```
rorymckinley, line 129: If we first sanitised the values and then encoded the map, would we still need `sanitize_json_string`?

test/lightning/invocation/log_line_test.exs
```diff
@@ -5,51 +5,247 @@ defmodule Lightning.Invocation.LogLineTest do
...
+            level: :error
+          },
+          nil
+        )
+
+      assert {:ok, log_line} = Repo.insert(changeset)
+      assert log_line.message == "Error occurred: � null byte in log"
+    end
+
+    test "handles multiple control characters", %{run: run} do
```
rorymckinley: Nice test coverage - including those simulating the error condition! If you wanted , I think you could remove a few   - specifically those that `LogLine.new` has no control over (e.g. `handles multiple control characters` is no something `LogLine.new` can influence.

I find this minimises the blast radius when making changes but I also recognise that it is subjective and I would much rather have too many tests than too few!

test/lightning/ecto_types_test.exs
```diff
@@ -39,6 +39,89 @@ defmodule Lightning.EctoTypesTest do
...
                LogMessage.cast(false)
     end
+
+    test "sanitizes null bytes in strings" do
+      assert {:ok, "Hello�World"} = LogMessage.cast("Hello\x00World")
+      assert {:ok, "���"} = LogMessage.cast("\x00\x00\x00")
+    end
+
+    test "sanitizes control characters" do
+      assert {:ok, "Text�here"} = LogMessage.cast("Text\x01here")
```
rorymckinley, line 49: This is just a thought - I have not done any research into this and I am not sure if the scale of the problem even makes it worth the effort of researching it. If we replace all the 'verboten' characters we suffer some (most possibly trivial) data loss - i.e. we can never replace the control character once the string comes out of the DB.

Is there any sort of standard way to identify/escape these chars? Even it is not automagically reversable.

Like I said, thinking out loud - feel free to ignore :).

midigofrank (approved):
Nicely done

taylordowns2000:
@elias-ba , @midigofrank , what do you think of @rorymckinley 's comments and suggestions. Should they be implemented?

test/lightning/invocation/log_line_test.exs
elias-ba: Thanks man! I’ve trimmed those tests from LogLineTest so it now only covers the integration scenarios `LogLine.new/3` can actually influence. The detailed sanitization cases are already covered in `EctoTypesTest`, so we’re keeping full coverage without the extra blast radius you mentioned.

test/lightning/ecto_types_test.exs
elias-ba: There **is** intentional, minimal data loss: we replace only the Postgres-illegal, non-printable bytes with U+FFFD (`�`). Rationale:

* These bytes aren’t useful in human-readable logs and can’t be stored by Postgres anyway, so we prefer a clear placeholder over failing inserts.
* We preserve TAB/LF/CR, and leave pre-serialized JSON strings opaque; only raw disallowed bytes (or their `\u00xx` escapes when we’re given a map/list) get normalized.
* This keeps logs consistent and searchable across producers.

If we ever need **reversibility**, we can switch to a reversible strategy (e.g., replace with tokens like `<NUL>` or keep `\u0000` escapes) or add a mode/flag.

lib/lightning/ecto_types.ex
elias-ba: Good question. We sanitize in **both** places on purpose:

* **At `cast/1`**: fail-early and make the value consistent everywhere it flows (changesets, scrubber, telemetry, UI previews). What you see on the changeset is what will be stored. It also prevents invalid bytes from leaking into code that inspects the struct before persistence.
* **At `dump/1`**: safety net for any path that might bypass cast (e.g., manual struct assignment) so Postgres can’t reject the write.

The sanitization is **idempotent**, so doing it at cast and again at dump is harmless and avoids surprises. Deferring only to dump would defer “data loss,” but it would also mean the app sees one value pre-persist and a different one post-persist, which is more confusing in practice.

lib/lightning/ecto_types.ex
elias-ba: Great catch, I pushed a commit to fix this.

Jerejef sama xarit (this means, thanks a lot my friend) !

lib/lightning/ecto_types.ex
elias-ba: We keep `sanitize_json_string/1` because encoding first gives us a **canonical, ASCII-only representation** where any raw control bytes are turned into `\u00xx` escapes by Jason. Then a single pass can:

1. replace any **escaped** control chars (`\u0000`, `\u000B`, …), and
2. (via `sanitize_string/1`) catch any **raw** bytes that slipped through.

Pre-sanitizing would require a deep walk over all nested strings **and** still wouldn’t catch cases where producers already supplied the literal text `"\u0000"` (six ASCII chars). Post-encode sanitize is O(n), structure-agnostic, and avoids mutating non-string values or risking double-encoding.

lib/lightning/ecto_types.ex
elias-ba: Great question, Rory. We deliberately treat the input as an opaque string and only replace raw Postgres-invalid bytes (e.g., `0x00`, control chars). We **do not** parse or rewrite serialized JSON there, so `\u00xx` escape sequences remain untouched. If JSON sanitization is desired, callers should pass a map/list (that path encodes with Jason and normalizes `\u00xx`). Tests added to confirm both behaviors.

lib/lightning/ecto_types.ex
elias-ba: Thanks man! I’ve added a test that generates mixed-case `\u00xx` escapes to make sure the `i` flag behaviour is locked in. If the flag ever gets dropped, the test will fail.

elias-ba:
> @elias-ba Some cosmetic comments/questions - some of them just for my own education, most likely.
> 
> Code looks great - especially the extra tests! I would approve given that most of my comments are really silly questions, but I am uncertain whether we should be sanitising in the fallthrough 'dump' method, so will wait on your input re: that.

Hey Rory, thanks a lot for this review, it was really educational. I’ve provided responses to some of your questions and addressed other changes directly. Really appreciate the catches and the amazing questions that pushed me to rethink and challenge the implementation.

lib/lightning/ecto_types.ex
rorymckinley: Amul sóló @elias-ba (I have added Jerejef sama xarit to my file of Wolof phrases :) ).

lib/lightning/ecto_types.ex
rorymckinley: Jerejef, @elias-ba . Based on the safety net approach then, should we not be sanitising the string in this dump() variant then as well (i.e. the one that caters for arguments that do not pass `is_binary`)?

rorymckinley:
> > @elias-ba Some cosmetic comments/questions - some of them just for my own education, most likely.
> > Code looks great - especially the extra tests! I would approve given that most of my comments are really silly questions, but I am uncertain whether we should be sanitising in the fallthrough 'dump' method, so will wait on your input re: that.
> 
> Hey Rory, thanks a lot for this review, it was really educational. I’ve provided responses to some of your questions and addressed other changes directly. Really appreciate the catches and the amazing questions that pushed me to rethink and challenge the implementation.

Thanks for taking the time to respond, @elias-ba ! I found your answers educational! I have asked one refining question but I am going to pre-approve so that I do not block you if the answer is "nah, we don't have to bother".

rorymckinley (approved):

lib/lightning/ecto_types.ex
elias-ba: Great suggestion, I updated `dump/1` to cover non-binary inputs and added tests to lock it in. Jerejef, Rory!

elias-ba:
@rorymckinley Jerejef, I really loved this review and the chat here. I learnt a lot.

PR #3508, opened 2025-08-11
Enable cross-cluster node discovery

### LINKED ISSUE #3482: Use libcluster_postgres to address cluster migration concerns
As part of moving the web pods from one cluster to the other, we need to cater for the scenario where each cluster has a set of pods and both sets are sharing the same database. 

With this scenario, the current libcluster topology, which relies on Kubernetes DNS will not function. This means that each set of pods is only aware of other pods in the same cluster and this may lead to tears. 

[libcluster_postgres](https://hexdocs.pm/libcluster_postgres/readme.html) may provide a way to solve this problem without having to make any changes to the existing pod configuration.

rorymckinley:
## Description

This PR introduces optional plumbing to allow Lightning to use LibclusterPostgres as an additional libcluster strategy. I am not sure if we will keep this post-migration, so for now, I have biased towards a clunky implementation that will be easy to remove.

At the moment, the risks posed by the solution appears to be reasonable, given the primary advantage,  there is no need to mess with any existing GKE infrastructure.

Closes #3482

## Validation steps

I am not sure, tbh.  The least painful way is to test this is to deploy Lightning to web pods in multiple clusters. I am happy to deploy to staging and do a demo of it working.

At present my branch is on staging - so you can connect to a staging pod to check:

First, check that my branch is still there:

```
Application.get_env(:lightning, :k8s)

```

If the above returns `[enable_cross_cluster_discovery: true]`, then my branch is still present and you can continue with the testing:

```
Node.list()
```

The response should contain 3 nodes: 1 node from the same cluster as your test pod, and the other two from the other cluster (you can use the IP addresses as a guide).

*(Is there anything else the reviewer should know or look out for?)*

In 1973, American musician and actor Dean Reed settled in East Germany. As a result of his performances in this communist country , he would earn the nickname of 'Red Elvis'.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


rorymckinley:
Codecov is compaining about a file unrelated to my changes.

stuartc (changes requested):
Could we rather name it `LIBCLUSTER_POSTGRES=true` (optionally setting `LIBCLUSTER_POSTGRES_CHANNEL=<channel_name>` or something like that? Separating the concept of k8s from clustering here, since this could be used without k8s if someone wanted. 

I can see why you named it like this, and in hindsight the `K8S_HEADLESS_SERVICE` env variable really doesn't pass the sniff test now, since we can now technically opt into more than one strategy, the k8s one should be `LIBCLUSTER_K8S=true` and `LIBCLUSTER_K8S_SERVICE=<service>`. Or a shorthand of `LIBCLUSTER_<K8S|POSTGRES>=<service|channel_name>` to enable and set the most common config value.

midigofrank (commented):
LGTM. 
I agree with Stu's renaming change request here. I would be happy to approve once resolved

rorymckinley:
I renamed the ENV var to `ERLANG_NODE_DISCOVERY_VIA_POSTGRES_ENABLED` and added `ERLANG_NODE_DISCOVERY_VIA_POSTGRES_CHANNEL_NAME` - unfortunately the latter required a rearranging of the table to accommodate it and broke git diff.

The naming should hopefully alow us space for the naming of any future strategies.

midigofrank (approved):

stuartc (approved):

PR #3515, opened 2025-08-12
Fix diagram collapsing

### LINKED ISSUE #3506: Workflow diagram collapses or renders incorrectly during creation
### Description
When creating a new workflow using either the **Templates** feature or the **AI Assistant**, the workflow diagram sometimes collapses as if all edges were deleted, causing nodes to stack in the same area. In some cases, multiple triggers are rendered, but only one is connected to the rest of the diagram.

This behavior is unpredictable, but based on observation, it seems to occur **only during workflow creation** and not during editing.

Image
Image

#### Scenario 1 – Templates

1. Create a new workflow
2. Click rapidly on a template multiple times, or switch templates quickly

#### Scenario 2 – AI Assistant

1. Create a new workflow
2. Use the AI Assistant to generate a workflow
3. Press `CTRL+S` multiple times immediately after generation

### Expected Behavior
The workflow diagram should render consistently, with all nodes and edges in their correct positions, regardless of how quickly templates are switched or how many times `CTRL+S` is pressed.

### Suspected Cause
Likely related to a race condition or rendering conflict during the creation process, possibly triggered by multiple rapid state updates.

### Goal of This Issue
Identify the root cause of the diagram collapse and fix it so the diagram always renders correctly during workflow creation.

josephjclark:
This PR fixes an issue where when switching between templates in the new Workflow UI, sometimes the diagram can seem to collapse in on itself.

Fixes #3506

This behaviour was caused by a little bit of code which defaults the position of items in the workflow to 0,0

That code only triggers because of a second render after the template is set. This second render hasn't yet recorded issues (I think that's a timing thing) and so sets a bad default and overrides the layout.

I've removed the code. The solution needs testing carefully because presumably this just breaks something else. I'm investigating now.

### DIFF
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -85,7 +85,7 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
     updatePositions,
     updatePosition,
     undo,
-    redo
+    redo,
   } = useWorkflowStore();
   const isManualLayout = !!fixedPositions;
   // value of select in props seems same as select in store. one in props is always set on initial render. (helps with refresh)
@@ -111,8 +111,9 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
   // on option 2. chartCache isn't updated. Hence we call updateSelection to do that
   useEffect(() => {
     // we know selection from server has changed when it's not equal to the one on client
-    if (selection !== chartCache.current.lastSelection) updateSelection(selection);
-  }, [selection, updateSelection])
+    if (selection !== chartCache.current.lastSelection)
+      updateSelection(selection);
+  }, [selection, updateSelection]);
 
   const {
     placeholders,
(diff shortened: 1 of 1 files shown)


josephjclark:
Here is the commit which introduced the issue https://github.com/OpenFn/lightning/commit/936f0d206c3fcafa445ec9e62fc777feacf9f678#diff-b70025de8b89226d5416c22018346c8552e92efecb05e8c935b39da6a05d7b5c

It has this super unhelpful comment, so go me:
```
// If layout is id, ensure nodes have positions
// This is really only needed when there's a single trigger node
```

I'm not really sure why this is needed when there's a single trigger. I'm not even sure how to get a workflow with a single trigger! I suspect this is just very out of date with how the component runs now.

assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -209,29 +210,33 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
-        // This is really only needed when there's a single trigger node
+      } else if (isManualLayout) {
+        // if isManualLayout, then we use values from store instead
         newModel.nodes.forEach(n => {
-          // if isManualLayout, then we use values from store instead
-          if (isManualLayout && n.type !== 'placeholder')
+          if (isManualLayout && n.type !== 'placeholder') {
             n.position = fixedPositions[n.id];
-          if (!n.position) {
-            n.position = { x: 0, y: 0 };
```
elias-ba, line 220: @stuartc here for the collapsing diagram bug !

PR #1163, opened 2023-09-28
Webhook Security Verification

elias-ba:
## Notes for the reviewer

This Pull Request addresses several enhancements and fixes to our webhook authentication and authorization system. The intention behind these changes is to make the system more robust and handle different edge cases, ensuring that only authenticated and authorized requests can access the designated endpoints.

All the logic for verifying security on the webhook request is implemented in a plug that is branched in the router.

### Changes Include:

1. Refactoring WebhookAuth Module
    - Enhanced the `WebhookAuth` plug to handle different paths and improve the readability and maintainability of the code.
    - Refactored `check_auth/3` to precisely handle conditions where a request is authenticated but doesn’t match any valid key or user.

2. Improved Test Coverage
     - Added new tests to `WebhookAuthTest` to cover various scenarios and edge cases.
     - Ensured that all the new and existing tests are passing, providing better confidence in the reliability and stability of our system.

3. Code Clarity and Conciseness
     - Ensured that the code is concise, clear, and well-commented to aid any future developments and reviews.

4. Handling Edge Cases
     - Addressed the scenarios' where authenticated requests do not match any valid key or user, returning appropriate responses.

### Tests implemented:
- Given an HTTP request to a trigger that doesn't exist, show 404.
- Given an unauthenticated HTTP request to a trigger which requires an auth method, show 401.
- Given an authenticated HTTP request to a trigger which requires auth but DOESN’T match, show 404.
- Given an authenticated HTTP request to a trigger with a matching auth_method, show a 2XX.
- Given an HTTP request to a trigger with no auth method configured, show a 2XX.

### Additional Notes:
These changes are crucial for maintaining the security and reliability of our webhook system. It’s essential to review them thoroughly to ensure that all edge cases are handled appropriately, and the system behaves as expected in all scenarios.

## Related issue

Fixes #1150 

## Review checklist

- [x] I have performed a **self-review** of my code
- [x] I have verified that all appropriate **authorization policies** have been implemented and tested
- [x] If needed, I have updated the **changelog**
- [ ] Product has **QA'd** this feature

PR #3528, opened 2025-08-19
3097 retry webhook

### LINKED ISSUE #3097: Retry DB inserts on inbox endpoint
Adding basic retry-with-backoff to the `/i/` endpoint will take pressure off Lightning instance administrators to maintaining high-availability in their Postgres deployments.

We don't reply to the webhook request until it's in the DB, but nothing's stopping us from trying again if we get a `DBConnectionError` and delaying that response by up to 60s.

Needs more detail from @stuartc

elias-ba:
## Description

Adds resilient webhook processing: `WebhooksController#create` now retries transient database connection errors using a new `Lightning.Retry` helper with exponential backoff and optional jitter. Retry behavior is configurable via `Lightning.Config.webhook_retry` (optionally set by `WEBHOOK_RETRY_*` env vars). If retries are exhausted, the endpoint returns **503 Service Unavailable** with a **Retry-After** header based on the configured `timeout_ms`.

Closes #3097

## Validation steps

1. **Start the app** (dev is fine).

2. **Create a workflow** with a **webhook trigger** and copy the webhook url.

3. **Happy path:**

   ```bash
   curl -X POST <webhook_url> -H 'content-type: application/json' -d '{}'
   ```

   Expect **200** with `{ "work_order_id": "<uuid>" }`.

4. **Configure retry & restart the app:**

   ```bash
   export WEBHOOK_RETRY_MAX_ATTEMPTS=3
   export WEBHOOK_RETRY_INITIAL_DELAY_MS=100
   export WEBHOOK_RETRY_TIMEOUT_MS=5000
   ```

5. **Simulate DB outage**:

   * Docker: `docker stop <postgres-container>`
   * Homebrew: `brew services stop postgresql@15`
   * Generic: `pg_ctl -D <data_dir> stop`

6. **POST again** (same curl).
   Expect **503** with header `Retry-After: 5` and body:

   ```json
   {
     "error": "service_unavailable",
     "message": "Unable to process request due to temporary database issues. Please try again later.",
     "retry_after": 5
   }
   ```

7. **(Optional) Observe retries:** watch logs for retry messages (`retry_sleep`, `retry_exhausted`). If you want telemetry, in IEx:

   ```elixir
   :telemetry.attach_many("retry-debug",
     [
       [:lightning, :retry, :start],
       [:lightning, :retry, :attempt],
       [:lightning, :retry, :stop],
       [:lightning, :retry, :exhausted],
       [:lightning, :retry, :timeout]
     ],
     fn ev, meas, meta, _ -> IO.inspect({ev, meas, meta}, label: "retry") end,
     nil)
   )
   ```

8. **(Prod-only) idle timeout sanity check:** run with `MIX_ENV=prod` and `WEBHOOK_RETRY_TIMEOUT_MS=60000` (plus required prod env like `SECRET_KEY_BASE`, `DATABASE_URL`). In IEx:

   ```elixir
   Application.get_env(:lightning, LightningWeb.Endpoint)
   |> get_in([:http, :protocol_options, :idle_timeout])
   # => 75_000  (60_000 + 15_000 buffer)
   ```

9. **Cleanup:** bring Postgres back up, unset the `WEBHOOK_RETRY_*` envs if you changed them, and restart the app.

* **Idle timeout:** Default is now `max(60_000, retry_timeout_ms + 15_000)` to avoid the HTTP connection closing while webhook DB retries are in progress.
* **Error shape (limits):** Rate/usage limit responses now include an error code **and** message.

  * 402 → `{"error":"runs_hard_limit", "message":"Runs limit exceeded"}`
  * 429 → `{"error":"too_many_requests", "message":"Too many runs in the last minute"}`
    Tests updated accordingly.
* **Telemetry:** `Lightning.Retry` emits `:start`, `:attempt`, `:stop`, `:exhausted`, `:timeout` under the `[:lightning, :retry, ...]` namespace.
* **Docs & envs:** `DEPLOYMENT.md` and `.env.example` document `WEBHOOK_RETRY_*` variables and usage.
* **Backwards compatibility:** If no `WEBHOOK_RETRY_*` envs are set, sensible defaults apply; existing behavior remains unchanged.

## AI Usage

Please tick what applies for this PR:

* [x] Learning or fact checking
* [ ] Strategy / design
* [ ] Other
* [ ] I have not used AI

## Pre-submission checklist

* [x] I have implemented and tested all related **authorization policies** (N/A for this change; controller path already guarded).
* [x] I have ticked a box in "AI usage" in this PR.


lib/lightning/config/bootstrap.ex
```diff
@@ -50,6 +50,10 @@ defmodule Lightning.Config.Bootstrap do
       """
     end
 
+    # Load storage and webhook retry config early so endpoint can respect it.
+    setup_storage()
```
midigofrank, line 54: I'm not seeing any `--` change for this call. Is it still being called somewhere else?

lib/lightning_web/controllers/webhooks_controller.ex
```diff
@@ -36,30 +38,71 @@ defmodule LightningWeb.WebhooksController do
...
+          )
+        end,
+        retry_on: &Retry.retriable_error?/1,
+        context: %{trigger_id: trigger.id, workflow_id: trigger.workflow.id}
+      )
+      |> case do
+        {:ok, work_order} ->
+          json(conn, %{work_order_id: work_order.id})
+
+        {:error, %DBConnection.ConnectionError{} = error} ->
```
midigofrank, line 71: Will we ever know that there was a `DBConnection.ConnectionError` in our monitoring tools?

lib/lightning_web/controllers/webhooks_controller.ex
```diff
@@ -69,16 +112,17 @@ defmodule LightningWeb.WebhooksController do
...
         conn
         |> put_status(status)
-        |> json(%{"error" => message})
+        |> json(%{error: reason, message: message})
 
       nil ->
         conn
         |> put_status(:not_found)
-        |> json(%{"error" => "Webhook not found"})
+        |> json(%{error: :webhook_not_found})
```
midigofrank: Do we know if some people were matching on the returned response values? This might potentially break some workflows

lib/lightning_web/endpoint.ex
midigofrank, line 1: 🤔 wait, does this mean all along the `WebhookAuth` has been happening after the parsers? Whaat?

lib/lightning_web/plugs/webhook_auth.ex
```diff
@@ -5,76 +5,97 @@ defmodule LightningWeb.Plugs.WebhookAuth do
...
+
+            methods = (trigger && trigger.webhook_auth_methods) || []
+            {:ok, {trigger, methods}}
+          end,
+          retry_on: &Retry.retriable_error?/1,
+          context: %{op: :webhook_auth_lookup, webhook: webhook}
+        )
+        |> case do
+          {:ok, {trigger, methods}} ->
+            validate_auth(trigger, methods, conn)
```
midigofrank: Can't this just be part of the `fn` block? I think the success case should be inside the retry block and then  we only handle `error` cases

lib/lightning_web/plugs/webhook_auth.ex
```diff
@@ -5,76 +5,97 @@ defmodule LightningWeb.Plugs.WebhookAuth do
...
+            conn
+            |> put_resp_header("retry-after", Integer.to_string(retry_after))
+            |> put_status(:service_unavailable)
+            |> json(%{
+              error: :service_unavailable,
+              message:
+                "Temporary database issue during webhook lookup. Please retry in #{retry_after}s.",
+              retry_after: retry_after
+            })
+            |> halt()
```
midigofrank: I think we can have this in one place for use by both the plug and controller

midigofrank (changes requested):
Hey @elias-ba , great job, I mostly have questions here.

1. Did you consider that if `max_retries=5` then we could potentially retry `10 times` since we're retrying twice?
2. Now that we're capturing the `DBConnection` exception, will our monitoring services ever pick it up? (sentry, prometheus ..)
3. I'm surprised that the `WebhookAuth` plug was placed after `Plug.Parsers` all along. Great catch, I wonder if there was a reason it was moved later. Here is the [original commit](https://github.com/OpenFn/lightning/pull/1163/files#diff-ef9ce86f2dafe34a48de7bf062894f5f5a69d68847d01590b45741b747cd0ecaR45) that you did which placed it before `Plug.Parsers`. Could you please double check as to why this was changed?
4. I see you've updated the error response payload, I have a feeling this might break workflows else where. I'm okay with it but could you just check with the implementation team to see if they ever match on the error response?

lib/lightning/retry.ex
```diff
@@ -0,0 +1,227 @@
...
+
+    e ->
+      # NOT retryable; preserve original behavior
+      reraise e, __STACKTRACE__
+  end
+
+  defp build_config(opts) do
+    merged = Keyword.merge(@default_opts, opts)
+
+    %{
```
rorymckinley, line 215: Some of these transformations do not appear to have test coverage, e.g.

image

lib/lightning/retry.ex
```diff
@@ -0,0 +1,227 @@
...
+      backoff_factor: merged[:backoff_factor] |> to_float() |> max(1.0),
+      timeout_ms: merged[:timeout_ms] |> to_int() |> max(0),
+      retry_on: merged[:retry_on] || (&default_retry_check/1),
+      context: merged[:context] || %{},
+      jitter: !!merged[:jitter]
+    }
+  end
+
+  defp calculate_next_delay(base_delay, %{jitter: true}) when base_delay > 0 do
+    max_jitter = div(base_delay, 4)
```
rorymckinley, line 231: The "normalisation" of the jitter does not appear to have test coverage:

image

lib/lightning/retry.ex
```diff
@@ -0,0 +1,227 @@
...
+
+        {:retry, sleep_ms, next_base_delay(delay, config)}
+    end
+  end
+
+  defp next_base_delay(delay, config) do
+    delay
+    |> Kernel.*(config.backoff_factor)
+    |> trunc()
+    |> min(config.max_delay_ms)
```
rorymckinley, line 197: The `min` and application of `config.backoff_factor` do not appear to have test coverage.

image
image

DEPLOYMENT.md
```diff
@@ -363,3 +363,31 @@ Once the client has been created set the following environment variables:
...
+
+Lightning automatically retries webhook processing on **transient database
+connection errors** using exponential backoff. This helps prevent data loss
+during brief database outages.
+
+The following environment variables control webhook retry behavior:
+
+| **Variable**                     | **Description**                                                                                  | **Default** |
+| -------------------------------- | ------------------------------------------------------------------------------------------------ | ----------: |
+| `WEBHOOK_RETRY_MAX_ATTEMPTS`     | Maximum number of attempts (the first attempt runs immediately; backoffs occur between retries). |         `5` |
```
rorymckinley, line 377: For the first implementation of this, it feels like we have a lot of knobs that we can fiddle with - my bias is towards wondering if there is space for a simpler implementation that allows us to iterate towards the additional complexity as we see what is required?

lib/lightning/config/bootstrap.ex
```diff
@@ -487,16 +491,20 @@ defmodule Lightning.Config.Bootstrap do
 
       url_scheme = env!("URL_SCHEME", :string, "https")
 
+      retry_timeout_ms = Lightning.Config.webhook_retry(:timeout_ms)
```
rorymckinley, line 494: I am trying to decide why this feels weird. I am not sure if you are familiar with the concept of "layering" as suggested by Eric Evans but this feels like it might reversing a pattern, here. 

I.e. in my head, I think of the "layers" within how config gets accessed as follows:

Other application code
|_ Lightning.Config
|_ stuff in bootstrap

And by referring to Lightning.Config inside it bootstrap, it feels like we are reversing that 'layering' (quotes because it has been a long time since I read the DDD book so I may be mangling it :). Especially as  Lightning.Config overrides the 'default_webhook_retry` values unconditionally?

As a rule of thumb I try to avoid this kind of stuff as it can get me into trouble, but it is subjective.

lib/lightning/config.ex
```diff
@@ -350,6 +350,48 @@ defmodule Lightning.Config do
     def sentry do
       Application.get_env(:lightning, :sentry, Sentry)
     end
+
+    @impl true
+    def webhook_retry do
+      default_webhook_retry()
+      |> Keyword.merge(Application.get_env(:lightning, :webhook_retry, []))
```
rorymckinley, line 357: Given that this happens unconditionally, could we not just set defaults in bootstrap and do away with `default_webhook_retry` ? That seems to be a pattern that is quite common?

test/lightning_web/plugs/webhook_auth_test.exs
```diff
@@ -7,15 +7,89 @@ defmodule LightningWeb.Plugs.WebhookAuthTest do
...
+
+    refute conn.halted
+    assert conn.assigns[:trigger] == expected_trigger
+  end
+
+  test "returns 503 with Retry-After when DB lookup errors are exhausted" do
+    Mimic.copy(Lightning.Retry)
+
+    Mimic.expect(Lightning.Retry, :with_webhook_retry, fn _fun, _opts ->
+      {:error, %DBConnection.ConnectionError{message: "db down"}}
```
rorymckinley, line 68: Does Mimic allow matching of arguments passed to the method being mocked? If not, I imagine you could perhaps do pattern matching in the function that handles the response?

For me, when I am writing a test where the code is calling another function, the two questions I want answered are:

- Does the code I am testing call the method correctly?
- Does the code I am testing do the right thing with the method response?

I think we may be missing coverage of the former, so it would be great if we could use Mimic for that.

test/lightning_web/plugs/webhook_auth_test.exs
```diff
@@ -7,15 +7,89 @@ defmodule LightningWeb.Plugs.WebhookAuthTest do
...
+      |> put_req_header("x-api-key", "secret")
+      |> WebhookAuth.call([])
+
+    expected_trigger =
+      trigger
+      |> unload_relation(:workflow)
+      |> Repo.preload([:workflow, :edges, :webhook_auth_methods])
+
+    refute conn.halted
+    assert conn.assigns[:trigger] == expected_trigger
```
rorymckinley, line 61: In cases where I am testing code that is performing a lookup, I like to have "negative" examples if it is cheap to do so - e.g. at least one other instance of trigger (in this case) that I am not interested to try and offer some reassurance that my code will still work when there is more than one instance of the thing that I am looking for (i.e. the way things would be in a non-test env).

test/lightning/config/bootstrap_test.exs
```diff
@@ -30,7 +33,16 @@ defmodule Lightning.Config.BootstrapTest do
 
   describe "prod" do
     setup do
-      Process.put(@opts_key, {:prod, ""})
+      Process.put({Config, :opts}, {:prod, ""})
```
rorymckinley, line 36: Is there an advantage to using `{Config, opts}` instead of `@opts_key`? Or `@config_key`, `@import_key`?

test/lightning/config/bootstrap_test.exs
```diff
@@ -471,4 +530,18 @@ defmodule Lightning.Config.BootstrapTest do
       nil -> nil
     end
   end
+
+  defp reconfigure(envs) do
```
rorymckinley, line 534: Nice! Given that it these helpers re only used in a single block of tests, how would you feel about moving the method definitions to be closer to the tests that use them?

rorymckinley: Also, is it worth extending this method a little bit so that we can also use it in the `setup` block?

lib/lightning/config/bootstrap.ex
```diff
@@ -812,6 +819,23 @@ defmodule Lightning.Config.Bootstrap do
...
+    webhook_retry_config =
+      [
+        max_attempts: env!("WEBHOOK_RETRY_MAX_ATTEMPTS", :integer, nil),
+        initial_delay_ms: env!("WEBHOOK_RETRY_INITIAL_DELAY_MS", :integer, nil),
+        max_delay_ms: env!("WEBHOOK_RETRY_MAX_DELAY_MS", :integer, nil),
+        backoff_factor: env!("WEBHOOK_RETRY_BACKOFF_FACTOR", :float, nil),
+        timeout_ms: env!("WEBHOOK_RETRY_TIMEOUT_MS", :integer, nil),
+        jitter: env!("WEBHOOK_RETRY_JITTER", &Utils.ensure_boolean/1, nil)
+      ]
+      |> Enum.reject(fn {_, value} -> is_nil(value) end)
```
rorymckinley, line 828: Nice - although someone should add a `compact` method for Elixir! :P

Is there any readability value to code-golfing this to extend the pipeline into a `cond` to replace the `if`?

lib/lightning/config.ex
```diff
@@ -350,6 +350,48 @@ defmodule Lightning.Config do
...
+        max_attempts: 5,
+        initial_delay_ms: 100,
+        max_delay_ms: 10_000,
+        backoff_factor: 2.0,
+        timeout_ms: 60_000,
+        jitter: true
+      ]
+    end
+
+    defp normalize_retry(opts) do
```
rorymckinley, line 377: There seems to be a lot of overlap between this and `Retry.build_config`? Is there a case to be made for deferring the normalisation to `build_config`?

test/lightning/retry_test.exs
```diff
@@ -0,0 +1,345 @@
...
+  end
+
+  describe "option parsing fallbacks (:error branches)" do
+    test "string parse failures fall back to safe defaults" do
+      attempts = :counters.new(1, [])
+
+      result =
+        Retry.with_retry(
+          fn ->
+            :counters.add(attempts, 1, 1)
```
rorymckinley, line 30: TIL - thanks, very cool!

lib/lightning/retry.ex
```diff
@@ -0,0 +1,227 @@
...
+
+    Logger.debug(
+      "retry finished: result=#{status} duration_ms=#{duration} ctx=#{inspect(config.context)}"
+    )
+
+    result
+  end
+
+  @spec retriable_error?(term()) :: boolean()
+  def retriable_error?({:error, %DBConnection.ConnectionError{}}), do: true
```
rorymckinley, line 84: This appears to be a duplicate of `default_retry_check` - what if we made `retriable_error?` the default for `retriable_on`, then you would need one less argument in the two current calls?

rorymckinley (commented):
Hey Jambaar! Round 2 done - more silly questions. I just need to run the manual tests and then I am done :).

lib/lightning_web/endpoint.ex
midigofrank: Here is the PR that changed it: https://github.com/OpenFn/lightning/pull/1541

rorymckinley (approved):
Jerejef @elias-ba - manual tests done - all looks good. Great job - none of my comments are showstoppers, so feel free to implement any/none of the suggestions :) - I am going to approve in the meantime so that I do not block you, but happy to do follow-ups if you require!

rorymckinley:
@elias-ba Almost, forgot! This: `Telemetry: Lightning.Retry emits :start, :attempt, :stop, :exhausted, :timeout under [:lightning, :retry, ...].` is awesome! Could you point me to the code where there is happening, I would love to surface this in Grafana :)

lib/lightning_web/endpoint.ex
elias-ba: hey @frank, I am pretty sure `WebhookAuth` was moved after `Replug` accidentally in #1541. I tested this in Lightning and the billing app and everything behaves as expected, so we’re good and safe to move `WebhookAuth` back **before** Replug.

lib/lightning_web/controllers/webhooks_controller.ex
elias-ba: Nice catch man, I just restored it. We can standardize error formats later if we think it's important enough.

lib/lightning_web/plugs/webhook_auth.ex
elias-ba: Yes it's actually nicer to have this in the `fn` block, thanks for the suggestion.

lib/lightning_web/plugs/webhook_auth.ex
elias-ba: Thanks man for this catch. I created a helper function (`respond_service_unavailable`) under `LightningWeb.Utils`, I didn't know where to put the function exactly and I didn't wanna create a new module for it, so I think for now utils is fine.

elias-ba:
Hey @midigofrank I handled most of your change requests / questions. Thanks a lot for your eagle eyes on this. This is really great. Here are few responses to your general questions

1) Each call to `Retry.with_webhook_retry/2` has its **own** cap. We run *two separate retry loops* in a request: one for **auth lookup** in the plug, one for **create\_workorder** in the controller. So the same DB operation is **never** retried twice; we just have two different operations that can each retry up to `max_attempts` (default 5). Worst case per request you’ll see up to 5 failed auth lookups **and** up to 5 failed work-order creations, not 10 for the same step. The DB driver’s own reconnects don’t change that cap; they just influence how a single attempt behaves internally.

2) Thanks for the catch, you’re right, we weren’t surfacing those because Retry rescues DBConnection.ConnectionError, so PlugCapture never saw them. I’ve pushed a change: we now explicitly capture exhausted retries to Sentry in the 503 path, and emit telemetry for the retry lifecycle plus a webhook.db_unavailable event for PromEx. This restores monitoring without spamming (only on exhaustion) and keeps payloads minimal. cc @rorymckinley 

3) Did a quick check, and this has no implication. We can safely place `WebhookAuth` before `Replug`. The switch noted in #1541 was probably accidental.

4) Nice catch, I restored the "Webhook not found" error

5) Verified and all good on the billing app, as it should be. This has no implication in anything that would break the billing app

lib/lightning_web/controllers/webhooks_controller.ex
elias-ba: Yes, I just pushed a change to capture exception in Sentry and emit telemetry. Thanks for the catch

lib/lightning/config/bootstrap.ex
elias-ba: Great catch, actually I forgot to remove the call of `setup_storage()` at the end of that function. Done now

elias-ba:
> @elias-ba Almost, forgot! This: `Telemetry: Lightning.Retry emits :start, :attempt, :stop, :exhausted, :timeout under [:lightning, :retry, ...].` is awesome! Could you point me to the code where there is happening, I would love to surface this in Grafana :)

Hey @rorymckinley sorry that commit about telemetry never got into this PR, I am happy you asked. You can find all the telemetry events now [in this commit](https://github.com/OpenFn/lightning/pull/3528/commits/00289f359f4e6f182bcea72ce63c2289f41c804a)

midigofrank (approved):

PR #3531, opened 2025-08-19
feat: hide mini-history on new page

### LINKED ISSUE #3529: Regression: Overlapping elements on the canvas
**Describe the bug** We need to disable view history on the templates view because the elements are overlapping

1. On main, create a workflow using a template
2. View the top left of the canvas and note overlapping element

Image

doc-han:
## Description

When on the new page for creating a new workflow, you shouldn't see the mini history panel. 

Screenshot 2025-08-19 at 9 35 40 AM

Closes #3529 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/workflow-diagram/WorkflowDiagram.tsx
+++ b/assets/js/workflow-diagram/WorkflowDiagram.tsx
@@ -590,15 +590,18 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
         liveAction={props.liveAction}
         drawerWidth={drawerWidth}
       />
-      <MiniHistory
-        collapsed={!runSteps.start_from}
-        history={someHistory}
-        selectRunHandler={onRunChange}
-        onCollapseHistory={onCollapseHistory}
-        drawerWidth={drawerWidth}
-        hasSnapshotMismatch={hasSnapshotMismatch}
-        missingNodeCount={missingNodeCount}
-      />
+      {
+        props.liveAction === 'edit' ?
+          <MiniHistory
+            collapsed={!runSteps.start_from}
+            history={someHistory}
+            selectRunHandler={onRunChange}
+            onCollapseHistory={onCollapseHistory}
+            drawerWidth={drawerWidth}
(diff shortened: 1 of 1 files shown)


assets/js/workflow-diagram/WorkflowDiagram.tsx
```diff
@@ -590,15 +590,18 @@ export default function WorkflowDiagram(props: WorkflowDiagramProps) {
...
-        collapsed={!runSteps.start_from}
-        history={someHistory}
-        selectRunHandler={onRunChange}
-        onCollapseHistory={onCollapseHistory}
-        drawerWidth={drawerWidth}
-        hasSnapshotMismatch={hasSnapshotMismatch}
-        missingNodeCount={missingNodeCount}
-      />
+      {
+        props.liveAction === 'edit' ?
```
elias-ba, line 594: Just for curiosity sake and learning JS, is this better than `props.liveAction === 'edit' && ...` ?

doc-han: It's just a habit I have when doing react.
I always do a full ternary when writing React JSX. 
In React, a react node is either
1. A react element i.e an html element or another react component
2. string
3. number
4. booleans, null, undefined

React renders react nodes. but when the node it's supposed to render is in 4. it ignores them and doesn't render them. 

So, the reason I always do a full ternary instead of using `&&` is that. you might be rendering an undesirable node if you're not careful or when the first condition doesn't fall in 4. 

eg. 
```js
<div>
	{
		users.length && <div>...</div>
	}
</div>
```
(where users.length = 0)
What this code will do is to actually render the number 0 somewhere on your screen. 
Hence, to make this right you actually need to turn the first part of `&&` to be in scope of 4. so that it'll be ignored when the `&&` fails. 
So you do `!!users.length` instead. 

I just choose to always do a full ternary

theroinaochieng (approved):

PR #3537, opened 2025-08-19
Fix: generate workflow YAML on page load

### LINKED ISSUE #3536: Bug: empty workflow YAML on first load
### Describe the bug

Entering the workflow editor generates an **empty YAML** on first load:

```yaml
name: ""
jobs: {}
triggers: {}
edges: {}
```

This breaks “View workflow as code” and causes Apollo/AI to think the workflow is empty until the page is reloaded.

### Where it happens

* **Staging:** v2.14.2-pre (1.5.34)
* **Prod:** v2.14.2-pre (1.5.34)

### To Reproduce

1. Go to app.staging.openfn.org / app.openfn.org
2. Create a new workflow
3. Immediately open **View workflow as code** or invoke the AI assistant (which requests YAML).
4. Observe empty YAML or assistant reporting “empty workflow.”
5. Reloading the page shows correct YAML.

### Expected behavior

YAML should be generated correctly on initial load without requiring a reload.

### Actual behavior / logs

Browser console shows the first `get-current-state` roundtrip returning an object with empty `jobs/triggers/edges`, followed by empty YAML.

### Root cause

Regression: the LiveView stopped firing the `generate_workflow_code` event on initial assignment. The `WorkflowToYAML` hook only generates YAML after that event, so first load returned an empty spec.

### Screenshots

Image

elias-ba:
## Description

This PR fixes a regression where the YAML view/code panel did not generate on first load of the Workflow editor. We now push the `"generate_workflow_code"` event from `assign_workflow/3` so the client builds the YAML immediately for:

* Existing workflows (on first mount), and
* Newly created workflows (after the initial save on `/w/new`).

This was affecting prod and staging.

Closes #3536

## Validation steps

1. **Existing workflow**

   * Navigate to `/projects/:project_id/w/:workflow_id?v=:lock_version`.
   * Open “View as YAML”.
   * Correct YAML appears without any additional interaction.
   
2. **New workflow**

   * Go to `/projects/:project_id/w/new`.
   * Choose any template, click **Continue/Save** so a workflow is created.
   * Open “View as YAML”.
   * Correct YAML appears.

## AI Usage

* [ ] Strategy / design
* [ ] Translation / spellchecking / doc gen
* [ ] Other
* [x] I have not used AI

## Pre-submission checklist

* [x] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
* [ ] I have updated the **changelog**.
* [x] I have ticked a box in "AI usage" in this PR

### DIFF
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -236,7 +236,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
             ai_assistant_component_id={@workflow_ai_assistant_id}
             module={LightningWeb.WorkflowLive.WorkflowAiChatComponent}
             workflow={@workflow}
-            workflow_code={@workflow_code_with_ids}
+            workflow_code={@workflow_code_with_ids |> dbg()}
             project={@project}
             base_url={@base_url}
             query_params={@query_params}
@@ -3098,6 +3098,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     |> assign(show_workflow_ai_chat: show_workflow_ai_chat)
     |> assign_changeset(changeset)
     |> maybe_disable_canvas()
+    |> generate_workflow_code()
   end
 
   defp apply_query_params(socket, params) do
--- a/test/lightning_web/live/workflow_live/edit_test.exs
+++ b/test/lightning_web/live/workflow_live/edit_test.exs
@@ -26,6 +26,38 @@ defmodule LightningWeb.WorkflowLive.EditTest do
   setup :register_and_log_in_user
   setup :create_project_for_current_user
 
+  describe "initial YAML generation (regression)" do
+    setup :create_workflow
+
+    test "pushes generate_workflow_code on first mount", %{
+      conn: conn,
+      project: project,
+      workflow: workflow
+    } do
+      {:ok, view, _html} =
+        live(
+          conn,
+          ~p"/projects/#{project.id}/w/#{workflow.id}?#{[v: workflow.lock_version]}",
+          on_error: :raise
+        )
+
+      assert_push_event(view, "generate_workflow_code", %{})
+    end
+
+    test "fires after a new workflow is created on the canvas", %{
+      conn: conn,
+      project: project
+    } do
+      {:ok, view, _html} =
+        live(conn, ~p"/projects/#{project.id}/w/new", on_error: :raise)
+
+      select_template(view, "base-webhook-template")
+      render_click(view, "save")
+
+      assert_push_event(view, "generate_workflow_code", %{})
+    end
+  end
+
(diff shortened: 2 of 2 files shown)


lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -3098,6 +3098,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     |> assign(show_workflow_ai_chat: show_workflow_ai_chat)
     |> assign_changeset(changeset)
     |> maybe_disable_canvas()
+    |> generate_workflow_code()
```
doc-han, line 3101: I think we already have a call to `generate_workflow_code` that could deal with this issue. 

**Check below**
https://github.com/OpenFn/lightning/blob/b7522e28a606cb5a456a6619064d3ffeaa43522a/lib/lightning_web/live/workflow_live/edit.ex#L3284-L3294

In this code, we're restricting the function to only run when there's a diff between two states of the workflow. Hence for a new workflow, there will be no patches available and then the function wouldn't be called. Is there another reason we put the function in there? if not, I believe removing the existing call from that if-block should resolve this issue without introducing another call to the function.

doc-han (changes requested):

lib/lightning_web/live/workflow_live/edit.ex
elias-ba: Good catch! we do call `generate_workflow_code()` after patches, but that path only runs when `length(patches) > 0`. On a fresh mount and on `/w/new` there are no patches yet, so nothing fires and the client never generates YAML. That’s the gap this PR closes.

I added a push on initial assign/mount so the client can generate YAML immediately, and I kept the existing call inside the patches branch so edits still regenerate YAML. If we removed the in-patch call, YAML wouldn’t update after subsequent changes. The client hook is debounced, so if both events happen back-to-back we still only run once.

In another PR, we can follow up by emitting a generic ‘state\_changed’ event both on mount and after patches and have the hook listen to that. For this PR I kept the change minimal to fix the regression.

midigofrank: Hey @elias-ba, sorry I missed the notification for your response. 

Doing a follow-up PR is fine, but I was hoping to get clarity as why we can't move the `generate_workflow_code()` out of the `if do` block as Farhan suggested above. Calling `generate_workflow_code()` inside `assign_workflow` means it will be called twice if there is a follow-up call to `push_patches_applied/2`

midigofrank: Had a chat in slack, `generate_workflow_code` is harmless and so there is no problem calling it more than once

midigofrank (approved):

theroinaochieng (approved):

PR #3541, opened 2025-08-21
feat: clean UI for rendering errors & allow re-renders

### LINKED ISSUE #3239: Cleaner UI for React Error Reporting
**Update:** 

I've clarified the issue title - this issue is really about the cosmetics of how these errors are reported, not about the details of the error.

**Original Post:**

In the New React world, we've got some generic error handling:

Image

The error is rendered by the Fallback component. 

This particular message is very terse and has no context. I have no idea where it's come from.

But the point of this issue is to think about the rendering of this:

- Can we console.log the original error?
- Can we console.log any better context about the error? Which component it came from? Is there a staack trace?
- Is the message dismissable?
- Can we render the message like a toast?
- Should we have better text?
- How does this look in the Inspector?

doc-han:
## Description
Whenever we have an Error somewhere in react, we now show a beautiful UI and also allow the user to recover from the error.
Screenshot 2025-08-21 at 5 57 10 AM

Closes #3239 

## Validation steps
Currently on main, switching between templates pretty quick breaks the app. Goodluck on recovering from it.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/react/components/Fallback.tsx
+++ b/assets/js/react/components/Fallback.tsx
@@ -2,11 +2,37 @@ import type { FallbackProps } from 'react-error-boundary';
 
 export type { FallbackProps };
 
-export const Fallback = ({ error }: FallbackProps) => (
-  <div role="alert">
-    <p>Something went wrong:</p>
-    <pre style={{ color: 'red' }}>
-      {error instanceof Error ? error.message : String(error)}
-    </pre>
-  </div>
-);
+const renderStyle = "border border-gray-300 bg-[repeating-linear-gradient(45deg,theme(colors.gray.200)_0_12px,theme(colors.gray.100)_12px_24px)] dark:border-gray-600 dark:bg-[repeating-linear-gradient(45deg,theme(colors.gray.700)_0_12px,theme(colors.gray.800)_12px_24px)]"
+
+export const Fallback = ({ error, resetErrorBoundary }: FallbackProps) => {
+  return <>
+    <div className={'w-full h-full ' + renderStyle}>
+    </div>
+    <div role="alert" className='z-50 fixed top-0 right-0 bottom-0 left-0 flex items-center justify-center bg-red-100/50'>
+      <div className='bg-white rounded-lg p-4 w-[512px] h-auto shadow'>
+        <h2 className='text-xl mb-3 flex items-center gap-2'> <span className='hero-exclamation-circle text-yellow-500 size-6'></span> Rendering Error</h2>
+        <p>An error occurred while rendering this part of the application.</p>
+        <p>You can either re-render the content or reload the entire page. </p>
+        <pre className={'rounded p-1 text-red-400 my-4 py-4 text-sm overflow-auto font-mono font-semibold' + renderStyle}>
+          {error instanceof Error ? error.message : String(error)}
+        </pre>
+        <div className="mt-2 flex gap-2">
+          <button
+            type="button"
+            className="rounded-md text-sm font-semibold shadow-xs phx-submit-loading:opacity-75 bg-primary-600 hover:bg-primary-500 text-white disabled:bg-primary-300 focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-primary-600 px-2 py-1 flex items-center gap-1"
+            onClick={resetErrorBoundary}
+          >
+            <span className="hero-tv w-4 h-4"></span>  Re-render
+          </button>
+          <button
+            type="button"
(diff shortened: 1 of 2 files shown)


lib/lightning_web/live/workflow_live/edit.ex
```diff
@@ -1314,7 +1314,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
     <div
       id={"selected-template-label-#{@template.id}"}
       phx-mounted={fade_in()}
-      class="absolute z-[9999] m-4 opacity-75 bg-white/100 border border-gray-200 rounded-lg p-6"
+      class="absolute z-40 m-4 opacity-75 bg-white/100 border border-gray-200 rounded-lg p-6"
```
doc-han, line 1317: @elias-ba I updated this to z-40. is there a reason z-[9999] was used? if not then I think z-40 should be fine.

assets/js/react/components/Fallback.tsx
```diff
@@ -2,11 +2,37 @@ import type { FallbackProps } from 'react-error-boundary';
...
-);
+const renderStyle = "border border-gray-300 bg-[repeating-linear-gradient(45deg,theme(colors.gray.200)_0_12px,theme(colors.gray.100)_12px_24px)] dark:border-gray-600 dark:bg-[repeating-linear-gradient(45deg,theme(colors.gray.700)_0_12px,theme(colors.gray.800)_12px_24px)]"
+
+export const Fallback = ({ error, resetErrorBoundary }: FallbackProps) => {
+  return <>
+    <div className={'w-full h-full ' + renderStyle}>
+    </div>
+    <div role="alert" className='z-50 fixed top-0 right-0 bottom-0 left-0 flex items-center justify-center bg-red-100/50'>
+      <div className='bg-white rounded-lg p-4 w-[512px] h-auto shadow'>
+        <h2 className='text-xl mb-3 flex items-center gap-2'> <span className='hero-exclamation-circle text-yellow-500 size-6'></span> Rendering Error</h2>
```
josephjclark: This looks lovely but users don't know what "Rendering" is, so we can't give them the option.

I suppose we could say "refresh" or "continue", but given that we don't really know what the error is, and what the consequences of either is (will work better lost?) I don't think that's a great answer either.

I think all we can do is provide a button the close the dialog

assets/js/react/components/Fallback.tsx
```diff
@@ -2,11 +2,37 @@ import type { FallbackProps } from 'react-error-boundary';
...
 
-export const Fallback = ({ error }: FallbackProps) => (
-  <div role="alert">
-    <p>Something went wrong:</p>
-    <pre style={{ color: 'red' }}>
-      {error instanceof Error ? error.message : String(error)}
-    </pre>
-  </div>
-);
+const renderStyle = "border border-gray-300 bg-[repeating-linear-gradient(45deg,theme(colors.gray.200)_0_12px,theme(colors.gray.100)_12px_24px)] dark:border-gray-600 dark:bg-[repeating-linear-gradient(45deg,theme(colors.gray.700)_0_12px,theme(colors.gray.800)_12px_24px)]"
```
josephjclark, line 5: This striped background looks super cool! But I'm not sold on it to be honest. It doesn't look like the rest of the app, and it makes the error message harder to read.

It also looks in the screenshot like the canvas has the same effect? I think I'd rather just leave it blank, nice and simple

assets/js/react/components/Fallback.tsx
doc-han: Nice. 
I think we can change the wording to "Something went wrong"
And the actions will be "retry" and "reload"

I doubt we can use a close on the dialog. that'll mean closing the dialog and the app still being unusable.

assets/js/react/components/Fallback.tsx
doc-han: Removing from the error message

lib/lightning_web/live/workflow_live/edit.ex
elias-ba: Thanks @doc-han. The 999 was probably a debugging value that I forgot to remove or some copy-pasta thing

elias-ba (commented):
@doc-han this looks good to me. My only question is whether a pop-up is the right UI element for this kind of element. I often find pop-ups intrusive and disruptive. Wouldn’t it be better to render the component directly in the canvas area with action buttons instead?

doc-han:
> @doc-han this looks good to me. My only question is whether a pop-up is the right UI element for this kind of element. I often find pop-ups intrusive and disruptive. Wouldn’t it be better to render the component directly in the canvas area with action buttons instead?

Cool. 
1. since an error renders the application unusable, we need a popup with an overlay that doesn't allow you to perform any further actions
2. I don't think rendering in the space of the error would work. Some of the components where the error emanates from have very little or no UI at times. eg. we can't fit this entire UI and action buttons into a 100px button or div when it errors.

elias-ba (approved):

PR #3547, opened 2025-08-25
Sandboxes: parent/child projects + workflow version provenance

### LINKED ISSUE #3422: sandboxes: basic DB support
The aim of this issue is to setup the database and tests so that Sandbox projects can be added at a later date.

At the end of the this issue all the properties required by Sandboxes should be supported, but otherwise sandboxes are indistinct from other projects. A sandbox can be access by putting the UUID in the URL. It looks just like any other poject.

## Design

A Sandbox is just a Project with a parent:

Image

This is not a database schema or yaml spec, but from a high level, this is what we need to add to a Project to support sandboxes:

* **parent_id** - a reference to the primary project that this sandbox is associated with
* **color** - metadata to allow us to colour the lightning UI when a sandbox is being viewed
environment. This can be any hex string
* **version** a list of version hashes* 
* **env** An arbitrary string which sets the environment for this project. All credentials use the matching variant if it exists. See #3400 for details (maybe we want to add this when we do #3400? There's a quesdtion over whether this goes on the Project, Workflow or Step)

When viewing the yaml for a Sandbox, we should see these keys at the top level.

Make sure project names are unique per `name + parent_id` combo (I.e., EXCLUDING sandboxes (ie, a sandbox copies the project name).

As well as extending the Project table, we need to add a `workflow_versions` table. This needs:

- hash: a 12 char version hash
- source: a short string (3 chars) indicating who created the version (app or CLI)
- inserted_at: a date for when the version was created
- workflow: the associated workflow id

elias-ba:
## Description

This PR introduces sandboxed projects and workflow version provenance to make project organization and change tracking safer, faster, and more deterministic.

Sandboxed projects add a simple parent/child self-association on `projects`, with uniqueness scoped to a parent so siblings can’t collide while root names can repeat. The new queries (`list_sandboxes`, `create_sandbox`, `get_workspace_projects`) exist to make common workspace views cheap and explicit. The “why” here is clearer multi-env or per-user sandboxes without bolting on a second model, plus predictable name rules that mirror how teams think about workspaces.

For workflows, we add an append-only `version_history` array on the workflow row and a normalized `workflow_versions` table to record provenance (`workflow_id`, `hash`, `source`, `inserted_at`). The array gives us constant-time “latest head” lookups and deterministic comparisons without a join; the table preserves a full audit trail and supports backfills. Helpers in `WorkflowVersions` (`record_version(s)` with idempotency + `FOR UPDATE` to avoid dupes, `history_for`, `latest_hash`, `reconcile_history!`, and `classify*`) provide the primitives we need to compare histories (same/ahead/behind/diverged) and to rebuild or inspect state. The motivation is reproducibility and cheap diffing across UIs/CLIs while remaining concurrency-safe and replay-safe.

At the project level we expose `compute_project_head_hash/1` and small append helpers. This gives a deterministic 12-hex digest of the latest heads across a project, useful for caching, change detection, and “is anything different?” checks without pulling full trees.

Migrations add the parent relation, the workflow `version_history` column, and the `workflow_versions` table with a `(workflow_id, hash)` unique index. Validation of `version_history` element format is enforced in application code (12 lowercase hex) because Postgres cannot use a subquery in a CHECK to validate each array element; this keeps the guarantee while staying compatible with PG’s constraints model.

Finally, tests and factories were updated to preload associations and use stable timestamps so assertions are deterministic. The aim is coverage of the new rules and APIs without flakiness, ensuring these foundations are reliable for future features like branching, workspace views, and deploy pipelines.

Closes #3422

## Validation steps

1. **Migrate**

   ```bash
   mix ecto.migrate
   ```
2. **Run test suite**

   ```bash
   mix test
   ```
3. **Manual smoke (IEx)**

   ```elixir
   # Create parent + sandbox
   {:ok, parent} = Lightning.Projects.create_project(%{name: "parent", project_users: [%{user_id: insert(:user).id, role: :owner}]})
   {:ok, sandbox} = Lightning.Projects.create_sandbox(parent, %{name: "child-1", project_users: [%{user_id: insert(:user).id, role: :owner}]})
   Lightning.Projects.list_sandboxes(parent.id) |> Enum.map(& &1.name) # => ["child-1"]

   # Workflow version provenance
   wf = insert(:workflow, project: parent)
   {:ok, wf} = Lightning.WorkflowVersions.record_version(wf, "aaaaaaaaaaaa", "app")
   {:ok, wf, _count} = Lightning.WorkflowVersions.record_versions(wf, ["aaaaaaaaaaaa","bbbbbbbbbbbb"], "cli")
   Lightning.WorkflowVersions.history_for(wf)      # => ["aaaaaaaaaaaa","bbbbbbbbbbbb"]
   Lightning.WorkflowVersions.latest_hash(wf)      # => "bbbbbbbbbbbb"

   # Project head hash
   Lightning.Projects.compute_project_head_hash(parent.id) # => 12-hex digest
   ```

## AI Usage

Please disclose how you've used AI in this work (it’s cool, we just want to know!):

* [x] Code generation (copilot but not intellisense)
* [ ] Strategy / design
* [ ] Other
* [ ] I have not used AI

## Pre-submission checklist

* [x] I have implemented and tested all related **authorization policies** (e.g., `:owner`, `:admin`, `:editor`, `:viewer`).
* [x] I have ticked a box in **AI usage** in this PR.

### DIFF
--- a/lib/lightning/utils/hex.ex
+++ b/lib/lightning/utils/hex.ex
@@ -101,7 +101,7 @@ defmodule Lightning.Validators.Hex do
   @spec format(length_spec) :: Regex.t()
   @spec format(keyword) :: Regex.t()
   @spec format(length_spec, keyword) :: Regex.t()
-  def format(), do: build_regex(@default_len, [])
+  def format, do: build_regex(@default_len, [])
 
   def format(opts) when is_list(opts),
     do: build_regex(@default_len, opts)


josephjclark:
Thank you @elias-ba 

As mentioned on Slack I think I've worked out a way we can simplify workflow versions. I just need to work it through a few examples before I'm sure. We're probably better to wait a little bit before merging, but I don't want to hold it back too long.

I'll take a look over the PR when I get a chance

elias-ba:
Thanks a lot @josephjclark. That's super clear. I will branch off this PR for the next issue to not stay blocked. And also I tried to make the commits in this PR atomic, so reverting anything related to workflow versioning should be easy and isolated. Let me know how we should move forward when you have more clarity

elias-ba:
Per my discussion with @josephjclark in Slack and here on simplifying the versioning, we can safely merge this PR whenever it passes code review.

The sandbox DB work and the `workflow_versions` provenance table are self-contained, and nothing in here commits us to keeping `workflows.version_history` long term. Most read paths already use the provenance rows (with deterministic ordering), so there’s no lock-in.

If we later move to Joe's simpler approach (e.g., a `parent_version` pointer), the follow-up is straightforward:
1. remove the code that appends to `workflows.version_history`
2. run a small migration to drop that column
3. update the few tests that assert array appends
4. and keep everything else as-is.

`Projects.compute_project_head_hash/1` and the compare helpers already operate off the `workflow_versions` table, so they won’t need changes. No data migration is required, existing provenance rows remain the source of truth.

I’ll proceed with this PR as the base; once the new spec lands I’ll open a focused cleanup PR to drop the array and wire in the new approach (pointer).

lib/lightning/projects.ex
```diff
@@ -1214,4 +1215,167 @@ defmodule Lightning.Projects do
 
     query |> Repo.all()
   end
+
+  @doc """
+  Returns the *direct* sandboxes (children) of a parent project, ordered by `name` (ASC).
+
+  This is a flat view: only rows where `parent.id == child.parent_id` are returned.
+  If we later support arbitrarily deep nesting, switch this to a recursive CTE.
```
josephjclark, line 1224: I think we will want to support deep nesting. I haven't yet found a good reason to restrict forking and merging - although I've played with the idea for sure.

Let's stick with a rule off flat hierarchies now - no forks of sandboxes - and add support later when we understand a bit more about the behaviour of them

lib/lightning/projects.ex
```diff
@@ -1214,4 +1215,167 @@ defmodule Lightning.Projects do
...
+  * `{:error, %Ecto.Changeset{}}` on validation/unique errors
+  """
+  @spec create_sandbox(Project.t(), map(), boolean()) ::
+          {:ok, Project.t()} | {:error, Ecto.Changeset.t()}
+  def create_sandbox(%Project{id: parent_id}, attrs, schedule_email? \\ false) do
+    attrs |> Map.put(:parent_id, parent_id) |> create_project(schedule_email?)
+  end
+
+  @doc """
+  Returns a read-only “workspace” view for a parent project: the parent itself
```
josephjclark, line 1257: So the difference between this and `list_sandboxes` is just that this includes the parent as well?

josephjclark:
For the record @elias-ba and I decided yesterday that workflow history is definitely the right approach. The simplified `formed_from_version` (or whatever) is great for main-staging sandboxes, but doesn't really work if you have sandbox-sandbox merges.

So we're going to keep it. I recommend we merge this and move forward quickly.

priv/repo/migrations/20250824190735_add_sandbox_fields_to_projects.exs
```diff
@@ -0,0 +1,31 @@
...
+    create index(:projects, [:parent_id])
+
+    # Sandbox names must be unique per parent; top-level projects may share names
+    create unique_index(:projects, [:parent_id, :name],
+             where: "parent_id IS NOT NULL",
+             name: "projects_unique_child_name"
+           )
+
+    # Optional safety checks; validations will also exist in changesets
+    create constraint(:projects, :color_must_be_hex,
```
midigofrank: Err, I would rather have this purely at the application layer. These things can change and it's very painful to have to write another migration to change 😂

priv/repo/migrations/20250824190735_add_sandbox_fields_to_projects.exs
```diff
@@ -0,0 +1,31 @@
...
+             check: "color IS NULL OR color ~ '^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3}|[A-Fa-f0-9]{8})$'"
+           )
+
+    create constraint(:projects, :env_must_be_slug,
```
midigofrank: And this as well

priv/repo/migrations/20250824193557_create_workflow_versions.exs
```diff
@@ -0,0 +1,35 @@
...
+      timestamps(type: :utc_datetime_usec, updated_at: false, null: false)
+    end
+
+    create unique_index(:workflow_versions, [:workflow_id, :hash])
+
+    create index(:workflow_versions, [:workflow_id, :inserted_at, :id],
+             name: :workflow_versions_latest_per_wf
+           )
+
+    create constraint(
```
midigofrank: Let's have these validations at the application layer

priv/repo/migrations/20250824193557_create_workflow_versions.exs
```diff
@@ -0,0 +1,35 @@
...
+             :workflow_versions,
+             :hash_is_12_hex,
+             check: "hash ~ '^[a-f0-9]{12}$'"
+           )
+
+    create constraint(
```
midigofrank: Let's have this in the application layer

lib/lightning/projects/project.ex
```diff
@@ -80,8 +90,31 @@ defmodule Lightning.Projects.Project do
...
 
+  defp disallow_self_parent(%{data: %{id: nil}} = cs), do: cs
+
+  defp disallow_self_parent(cs) do
+    if get_field(cs, :parent_id) == cs.data.id,
+      do: add_error(cs, :parent_id, "cannot be self"),
+      else: cs
+  end
+
+  def sandbox?(%__MODULE__{parent_id: nil}), do: false
```
midigofrank: I always prefer to evaluate the truthy head first when dealing with conditionals. In this case, we know that what makes a project a sandbox is `if is_binary(parent_id)`. The `false` value should be on the `else` block

elias-ba:
Hey @midigofrank, thanks for the eagle eyes on the migration constraints. I’ve updated the PR to move those format checks entirely to the application layer. Mind giving it another quick pass/approval so we can merge ?

elias-ba:
Hey @midigofrank I removed the migration related to the DB constraints about the version_history. Now validations live in the application layer. Could you give me another review / approval ?

midigofrank (approved):

PR #3556, opened 2025-09-03
Provision sandboxes from parent projects

### LINKED ISSUE #3430: sandbox: backend API to create a new Sandbox
We need to add an internal API to allow a new Sandbox to be created. While the UI isn't built here, understanding what will come from the UI and what must be done in the backend is important. Here we go:

**Inputs from the user**
  - Enter the Name for your sandbox (must be a unique combination of `parent_id + name`)
  - Pick Color
  - Pick Credential ENV to use
  - Pick Collaborators (default to current user)
 
** In the backend:**
  - Check if user is allowed (admin or higher)
  - Clone project settings from the `parent_id`
  - Generate project.yaml from the source poject
  - Import the project.yaml as a new project
  - Copy credential UUIDs from original project (ie, new project references the same credentials as the old)
  - Duplicate version history for each workflow (TBD - I think we just take the most recent version of the base and create a duplicate with the NEW workflow id)
  - Accept a list of dataclip ids to import into the project (they'll optionally choose to clone **_named_** dataclips as their test suite)
  - Set new name
  - Set `parent_id` column
  - Set `color`
  - Ser `credential_env`
  - Create `project_user` for the creator (but accept a list, as this will later be multiple people)
  - Redirect to: /parent-project-uuid/new-name/w

Note: 
  - Credentials must be referenced, not cloned. So if Project A has access to Credential C, then the sandbox must also have access to credential C. 

So basically we create clone of an existing project and some subset of its related artefacts, attached to a billing account. There should be no runs associated with this new project.

This will later be called by the provisioner API and the Lightning app. You may want to hook one of these up temporarily - or just implement the API through scripts and unit tests. That's up to you.

elias-ba:
## Description

This change introduces end-to-end **sandbox provisioning** for projects. When an owner or admin provisions a sandbox, we create a child project that mirrors the parent’s core settings, **reuses the same credentials by reference** (no secrets are duplicated), and **clones the workflow graph** (workflows, jobs, triggers, edges). Any triggers in the sandbox start **disabled**. We also **remap node positions** to the new IDs and carry over **only the latest** workflow version per workflow so history begins in a sensible state.

If the parent’s jobs use **Keychain credentials**, we **clone just the keychain metadata** (`name`, `path`, `default_credential_id`) into the sandbox and **rewire the cloned jobs** to those keychains. Because project credentials are copied first, the keychain’s default credential remains valid. Optionally, the caller can request a subset of **named dataclips** (of allowed types) to be copied.

The entire operation runs in a single transaction. If project creation fails, we roll back and return the changeset (no partial state). This PR also adds docs for the `Projects.provision_sandbox/3` wrapper.

Closes #3430 

## Validation steps

1. **Run tests**

   * `mix test test/lightning/projects/sandboxes_test.exs`
   * `mix test test/lightning/projects_test.exs: "project head hash" & "sandboxes" sections`
   * Optional full run: `mix test`

2. **Happy path (using `iex`)**

   * Create a parent project with an owner and at least one credential, one workflow, and a job.
   * Call: `Projects.provision_sandbox(parent, owner, %{name: "sb-1"})`
   * Verify: sandbox has `parent_id`, owner exists, and triggers are `enabled: false`.

3. **Keychain rewiring (using `iex`)**

   * Add a keychain to the parent and a job that uses it.
   * Provision a sandbox; confirm the cloned job has `keychain_credential_id` (set) and `project_credential_id` (nil),
     while static-cred jobs keep mapped `project_credential_id`.

4. **Credentials & Workflow positions**

   * Compare `credential_id` sets in `project_credentials` between parent and sandbox → equal.
   * Open a cloned workflow and confirm `positions` contains only **new** node IDs.

5. **Error/authorization**

   * Expect `{:error, :unauthorized}` for a non-admin/non-owner actor.
   * Expect `{:error, %Ecto.Changeset{}}` when name is invalid (e.g., `""`), and verify no sandbox row was created.

## AI Usage

* [ ] Strategy / design
* [x] Optimisation / refactoring
* [ ] Other
* [ ] I have not used AI

## Pre-submission checklist

* [x] I have implemented and tested all related **authorization policies**.
* [x] I have ticked a box in "AI usage" in this PR.

### DIFF
--- a/test/lightning/sandboxes_test.exs
+++ b/test/lightning/sandboxes_test.exs
@@ -417,23 +417,21 @@ defmodule Lightning.Projects.SandboxesTest do
     end
 
     test "copies trigger webhook auth methods when present" do
-      if Code.ensure_loaded?(Lightning.Workflows.WebhookAuthMethod) do
-        %{actor: actor, parent: parent} = build_parent_fixture!(:admin)
+      %{actor: actor, parent: parent} = build_parent_fixture!(:admin)
 
-        {:ok, sandbox} = Sandboxes.provision(parent, actor, %{name: "sb-wam"})
+      {:ok, sandbox} = Sandboxes.provision(parent, actor, %{name: "sb-wam"})
 
-        cnt =
-          from(t in Trigger,
-            join: w in assoc(t, :workflow),
-            where: w.project_id == ^sandbox.id,
-            join: j in "trigger_webhook_auth_methods",
-            on: j.trigger_id == t.id,
-            select: count(j.webhook_auth_method_id)
-          )
-          |> Repo.one()
+      cnt =
+        from(t in Trigger,
+          join: w in assoc(t, :workflow),
+          where: w.project_id == ^sandbox.id,
+          join: j in "trigger_webhook_auth_methods",
+          on: j.trigger_id == t.id,
+          select: count(j.webhook_auth_method_id)
+        )
+        |> Repo.one()
 
-        assert cnt >= 1
-      end
+      assert cnt >= 1
     end
   end
 


lib/lightning/projects/sandboxes.ex
```diff
@@ -0,0 +1,478 @@
...
+  defp create_workflows!(parent, sandbox) do
+    Enum.reduce(parent.workflows, %{}, fn w, acc ->
+      {:ok, new_w} =
+        %Workflow{}
+        |> Workflow.changeset(%{
+          name: w.name,
+          project_id: sandbox.id,
+          concurrency: w.concurrency,
+          enable_job_logs: w.enable_job_logs,
+          positions: %{}
```
midigofrank, line 273: Why are we not copying the positions?

lib/lightning/projects/sandboxes.ex
```diff
@@ -0,0 +1,478 @@
...
+        })
```
midigofrank, line 274: Hmm, and I think we should have a `snapshot` here as well given these workflows are going to operate like actual workflows.

test/lightning/sandboxes_test.exs
```diff
@@ -0,0 +1,754 @@
...
+          select: {d.name, d.type}
+        )
+        |> Repo.all()
+        |> MapSet.new()
+
+      assert copied == MapSet.new([{"g1", :global}, {"req1", :http_request}])
+    end
+
+    test "copies trigger webhook auth methods when present" do
+      if Code.ensure_loaded?(Lightning.Workflows.WebhookAuthMethod) do
```
midigofrank: hmm, was the module not getting loaded?

lib/lightning/projects/sandboxes.ex
elias-ba: Hey @midigofrank we are copying positions [here](https://github.com/OpenFn/lightning/blob/07f89553d123b35d214834bb26c9ae859e11db33/lib/lightning/projects/sandboxes.ex#L148). Positions involve record ids, so we need to copy the records first, then we remap the old positions with the new records ids.

lib/lightning/projects/sandboxes.ex
elias-ba: Great catch, yes we do need a snapshot. I realized that when I was building the UI on top of this. I fixed it in that PR [here](https://github.com/OpenFn/lightning/blob/33d64c4ee51cb8d168de3de50109d07ed84e808d/lib/lightning/projects/sandboxes.ex#L152).

test/lightning/sandboxes_test.exs
elias-ba: Nice catch. Removed it. Thanks a lot

elias-ba:
> Hey @elias-ba , great job here. I've left some comments regarding workflow positions and snapshot.

Hey @midigofrank thanks a lot. I just responded to your questions. Could you give me another quick round of review ?

midigofrank (approved):
Nicely done

Issue #3431, opened 2025-07-17
sandboxes: backend API to remove a sandbox


josephjclark:

elias-ba:
This done in #3562

Issue #3474, opened 2025-08-04
sandboxes: update breadcrumbs UI to include sandbox name


josephjclark:
`Home > Projects > [project name] > [Sandbox name] > Workflows`

josephjclark:
Currently the breadcrumbs are wrong:

<img width="640" height="139" alt="Image" src="https://github.com/user-attachments/assets/f7698957-92c6-4eac-8ddc-742e44102e95" />

This needs to be `Home > Projects > [project name] > [Sandbox name] > Workflows`

Issue #3424, opened 2025-07-16
sandboxes: make sandboxes accessible/viewable in the app (update URLs)


josephjclark:
By the end of this issue, new Sandbox projects should be accessible in the app via direct URL with very very basic UI.
 
- Update URL structure to allow sandboxes to be accessed
- Ensure proper access control is in place (according to the sandbox's users)
- Hide Sandbox projects in the main Projects viwe
- Ensure that runs work as expected (history will probably be broken)

## URL structures

A sandbox name should be appended to ALL urls after `/project`. In other words, all project resources are only accessible _through_ a sandbox.

```
/projects/<project-id>/sandboxes/ - sandbox list (not implemented yet)
/projects/<project-id>/<sandbox-name>/w/ - workflow list
/projects/<project-id>/<sandbox-name>/w/xyz - workflow editor
/projects/<project-id>/<sandbox-name>/history/ - history (filtered by sandbox, but only implemented for main for now)
/projects/<project-id>/<sandbox-name>/settings/ -  settings
```

The primary Project should be accessible through the magic reserved string `main`. Or we can use a special character if prefferred, like `projects/_/` or  `projects/m/` or `projects/~/`. But I like words and believe a URL should always be human readable.

For backwards compatibility, if no sandbox name is provided, the URL should redirect to `/projects/main/<whatever>`

Routing includes permissons, so this may be related to https://github.com/OpenFn/thunderbolt/issues/562

Triggers will continue to work at /projects/<project-id> where project id is a sandbox UUID

## View Updates

Existing views History, Settings and Subscription, all need to be updated to use the active sandbox data, not the OG project data. So we need to override the key used for this stuff to not be the project key in the URL

josephjclark:
All project links in the app UI need to be made relative to the current sandbox.

So if the URL reads projects/abc/main, then all links on the left navbar go to the main sandbox. Ie, projects/abc/settings, projects/abc/history

I don't think we get this for free - I think we need to update how the links are generated in the page.

It's super important that when inside a sandbox, clicking a link like History takes me to the history of that sandbox only

elias-ba:
We’re pausing this issue for now and will revisit it later.

The main goal of this issue is to make app URLs reflect the user’s position based on the sandbox they’re viewing. This morning, both @josephjclark, @taylordowns2000, and I agreed that while this cosmetic change in the URLs is important for user experience, it’s not an immediate priority.

[...]

PR #3562, opened 2025-09-05
Basic CRUD UI for Sandboxes

### LINKED ISSUE #3412: sandboxes: Sandbox List UI
A minimal UI to list sandboxes.

Just looking for the sandbox name and colour. Clicking the list item should open the sandbox Workflows list.

The idea is just to get the layout in place

More details to follow

This should be behind a feature flag (disabled by default) so that we can merge it without exposing the unfinished feature to users

Image

elias-ba:
## Description

This PR adds a very basic UI for creating, updating, listing, and deleting sandboxes per project. It also introduces app theming depending on the Sandboxes colors.

Closes #3412, #3431 

## Validation steps

1. Login to the app
2. Pick a project
3. Identify the Sandboxes menu item in the left nav bar; click on it
4. You are now in the Sandboxes List UI
5. Create a new Sandbox by clicking on the 'Create Sandbox' button
6. Fill in the form in the modal
7. Click create to save your sandbox
8. When the modal closes, see your new Sandbox created and listed in the UI.
9. Repeat that process until you have as many sandboxes as you want
10. Click on a sandbox row to open it
11. Note how the app theme changes to match the Sandbox color you picked earlier
12. Note how workflows, are copied from the parent project to the sandbox
13. In the Sandboxes list UI, note the presence of the Edit and Delete button
14. Click on each of those to edit a sandbox or delete a sandbox

See these videos for quick demos:
- https://www.loom.com/share/f0768fb9320c4733ba81068db39d02d3?sid=f05ea8df-0502-4208-84c5-fbe926b678a3
- https://www.loom.com/share/17f27153f2434eabb0f9a54fd29d1d26?sid=57a33f6e-85df-4237-93cc-cecb88cdedac

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


josephjclark:
The Sandboxes menu needs to be removed before merging

Can we stick it behind an env var or feature flag? Not really sure how this works in Elixir but it can't be exposed to users yet.

image


We can still get to the sandboxes menu from the `/sandboxes` URL, so while it's nice to see it in the UI we really don't need it now :)

josephjclark:
Something is up with the Sandboxes list

If you're in a sandbox, it needs to display all the sandboxes for the parent (ie for the "workspace") 

if you're on the parent, it shows the whole workspace

Right now, if I go into a sandbox, the Sandboxes page is empty

image

josephjclark:
Rather than show an arbitrary colour picker, I want to show 8, 12 or even 16 pre-picked colours which we know will look good in the app. 

I'd like to be involved in this bit but  I don't have to be. Let's tart with 8 - pick good deep strong colours. Maybe use a palette generator and start with the default blue. We can tweak values and add more options later.

image


EDIT: @elias maybe you can pick out 8 colours from the Material UI range? https://materialui.co/colors#google_vignette Pick them all from the same row - maybe the 800 range? No brown or grey.

josephjclark:
As you were saying at standup @elias-ba , sandboxes must be excluded from this pick-list here: 

image

Technically this probably needs fixing against main. If you use backend tools to create a sandbox in main, it probably shows there right now? So I'd prefer a separate PR for the fix, but if you'd rather bundle it all here then that's fine

josephjclark:
Thinking about it, the Sandbox list should probably also include the main/OG project (call it main for now but we'll have a chat about it later). Because you need an easy way to get back to the OG project.

I suppose the big Sandbox boxout in the left hand side could also have a "switch to main" link inside it, so maybe that's a cleaner pattern

elias-ba:
> The Sandboxes menu needs to be removed before merging
> 
> Can we stick it behind an env var or feature flag? Not really sure how this works in Elixir but it can't be exposed to users yet.
> 
> image
> We can still get to the sandboxes menu from the `/sandboxes` URL, so while it's nice to see it in the UI we really don't need it now :)

Hey @josephjclark we could have some sort of env variable to put this behing a flag. But I think the easiest for now is to comment the Sandboxes menu item line of code and we good to go. That way when we need to bring it back we just uncomment that line. It's just one line of code, we don't wanna have a whole feature flag mechanism just to hide it temporarily. What do you think ?

elias-ba:
> Something is up with the Sandboxes list
> 
> If you're in a sandbox, it needs to display all the sandboxes for the parent (ie for the "workspace")
> 
> if you're on the parent, it shows the whole workspace
> 
> Right now, if I go into a sandbox, the Sandboxes page is empty
> 
> image

@josephjclark I agree and nice catch. Would we include the current Sandbox we're viewing in the list of sandboxes ? Or would be a list of the other sandboxes in that workspace excluding the current sandbox ?

elias-ba:
> Rather than show an arbitrary colour picker, I want to show 8, 12 or even 16 pre-picked colours which we know will look good in the app.
> 
> I'd like to be involved in this bit but I don't have to be. Let's tart with 8 - pick good deep strong colours. Maybe use a palette generator and start with the default blue. We can tweak values and add more options later.
> 
> image
> EDIT: @elias maybe you can pick out 8 colours from the Material UI range? https://materialui.co/colors#google_vignette Pick them all from the same row - maybe the 800 range? No brown or grey.

Hey @josephjclark, interesting. I do agree that controling this allows us to render only colors that look good in the app. Would the color picker then become a dropdown list ? or maybe some sort of color cards ?

elias-ba:
> As you were saying at standup @elias-ba , sandboxes must be excluded from this pick-list here:
> 
> image
> Technically this probably needs fixing against main. If you use backend tools to create a sandbox in main, it probably shows there right now? So I'd prefer a separate PR for the fix, but if you'd rather bundle it all here then that's fine

I open a separate nice little PR for this one so that we can merge it to main faster

elias-ba:
> Thinking about it, the Sandbox list should probably also include the main/OG project (call it main for now but we'll have a chat about it later). Because you need an easy way to get back to the OG project.
> 
> I suppose the big Sandbox boxout in the left hand side could also have a "switch to main" link inside it, so maybe that's a cleaner pattern

Mhmm @josephjclark are we keeping the big Sandbox boxout in the left hand side ?

josephjclark:
Sorry for a bit of a terse review Elias but this is looking great! I really like that the feature becomes a bit fun.

1. Feel free to comment out the lines, I don't mind how you want to ship this
1. Yes, let's include the current sandbox in the list. Maybe we need a litle chip which says `original` and `active` or something
1. I would just show a grid of colours. Here's the [material UI widget](https://mui.com/material-ui/customization/color/#playground) - just pop something like that open. Make it pretty but don't spend hours on it
image

4. for the big Sandbox thing, can you spin out a new branch and try something out? Let's just add a boxout with the sandbox and a link back to main for now. It might jump when going in and out of sandboxes, I'd be interested to see how big a problem that will be. I was also thinking that a Sandbox could have a colour AND an icon - that would be a great backup for colour

lib/lightning/projects/sandboxes.ex
```diff
@@ -475,4 +478,71 @@ defmodule Lightning.Projects.Sandboxes do
...
+
+  Authorization: `actor` must be `:owner` or `:admin` on the **parent** project.
+  The `sandbox` must belong to the given `parent`.
+
+  Only basic project attributes are expected (`:name`, `:color`, `:env`).
+  """
+  @spec update(Project.t(), User.t(), Project.t() | Ecto.UUID.t(), map()) ::
+          {:ok, Project.t()}
+          | {:error, :unauthorized | :not_found | Ecto.Changeset.t()}
+  def update(%Project{} = parent, %User{} = actor, %Project{} = sandbox, attrs)
```
midigofrank:
```suggestion
  def update_sandbox(%Project{} = parent, %User{} = actor, %Project{} = sandbox, attrs)
```

`update` is quite ambiguous

lib/lightning/projects/sandboxes.ex
```diff
@@ -475,4 +478,71 @@ defmodule Lightning.Projects.Sandboxes do
...
+
+  @doc """
+  Deletes a sandbox (child project) under `parent` on behalf of `actor`.
+
+  Authorization: `actor` must be `:owner` or `:admin` on the **parent** project.
+  The `sandbox` must belong to the given `parent`.
+  """
+  @spec delete(Project.t(), User.t(), Project.t() | Ecto.UUID.t()) ::
+          {:ok, Project.t()} | {:error, :unauthorized | :not_found | term()}
+  def delete(%Project{} = parent, %User{} = actor, %Project{} = sandbox) do
```
midigofrank:
```suggestion
  def delete_sandbox(%Project{} = parent, %User{} = actor, %Project{} = sandbox) do
```

lib/lightning_web/router.ex
```diff
@@ -194,6 +194,10 @@ defmodule LightningWeb.Router do
         live "/w", WorkflowLive.Index, :index
         live "/w/new", WorkflowLive.Edit, :new
         live "/w/:id", WorkflowLive.Edit, :edit
+
+        live "/sandboxes", SandboxLive.Index, :index
```
midigofrank, line 199: Hmm, @elias-ba did the spec change? I thought it was `/project/project-name/sandboxes`? Or is that on a different issue

lib/lightning_web/utils.ex
```diff
@@ -165,4 +165,43 @@ defmodule LightningWeb.Utils do
...
+  @doc """
+  Normalizes a color string to `#RRGGBB` uppercase hex.
+
+  Accepts:
+    * `#RGB`, `RGB`
+    * `#RRGGBB`, `RRGGBB`
+    * Any other value -> fallback #79B2D6
+  """
+  @spec normalize_hex(nil | binary() | any()) :: binary()
+  def normalize_hex(nil), do: @fallback
```
midigofrank: Can rename this to `nomalize_color_hex`? There is the `hex12` somewhere as well, maybe that will add more clarity

lib/lightning_web/utils.ex
```diff
@@ -165,4 +165,43 @@ defmodule LightningWeb.Utils do
...
+
+  def normalize_hex(val) when is_binary(val) do
```
midigofrank: Do we really need this guy? I think there is a lot of ceremony here. Assuming the user picks a color, why do we then need to normalize it again?

midigofrank (changes requested):
Hey @elias-ba , this looks great!! I have left some comments regarding naming

lib/lightning_web/utils.ex
elias-ba: The idea here is to allow people type color hex values. My rule is never trust what's coming from the user.

lib/lightning_web/utils.ex
elias-ba: Yes agreed, thanks for the catch

lib/lightning_web/router.ex
elias-ba: That will come in [this issue](https://github.com/OpenFn/lightning/issues/3424), for now I wanted to keep this simple.

lib/lightning_web/utils.ex
midigofrank: Then shouldn't this be a changeset validation instead?

lib/lightning/projects/sandboxes.ex
elias-ba: Mhmm ! `Sanboxes.update_sandbox/3` vs `Sandboxes.update/3` I guess yeah but this is so subjective. I will change it tho, even tho I usually prefer avoiding suffixes in names when no conflict is at risk. I guess that's because I come from Python.

midigofrank: Yeah, it's very subjective. But given this is the context, there is a possibility of adding another update function in there

elias-ba: Yeah ! I think you're right. Better be extra clear rather than less clear

lib/lightning_web/utils.ex
elias-ba: Mhmm yeah maybe but that would make this input less standalone. I am happy to remove this piece of code tho and until whenever we need some sort of normalization. For now colors are controlled by us anyway, I just wanted to be extra cautious.

lib/lightning/projects/sandboxes.ex
elias-ba: Done

lib/lightning/projects/sandboxes.ex
elias-ba: Done

lib/lightning_web/utils.ex
elias-ba: Finally removed color normalization. We already validate project colors in the changeset.

lib/lightning_web/utils.ex
elias-ba: Finally removed the normalization and only trust changeset validation. Thanks

elias-ba:
- [x] When creating sandbox mention the place where we are creating this sandbox from
- [x] After creating a sandbox, directly redirect to the sandbox page
- [x] Pick random color when creating a sandbox
- [x] Make the color picker wide (8x2)

lib/lightning_web/components/new_inputs.ex
```diff
@@ -424,6 +424,11 @@ defmodule LightningWeb.Components.NewInputs do
     doc:
       "indicates if the tag input operates independently of a form's validation flow"
 
+  attr :shape, :string, default: "rounded", values: ~w(square rounded circle)
+  attr :wrapper_class, :string, default: ""
```
midigofrank: Hmm, I don't see these being used anywhere

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -0,0 +1,316 @@
...
+  defp sandbox_card(assigns) do
+    ~H"""
+    <div
+      class="group block cursor-pointer rounded-xl border border-slate-200 bg-white hover:border-slate-300 hover:shadow-md transition-all duration-200 overflow-hidden"
+      phx-click={JS.navigate(~p"/projects/#{@sandbox.id}/w")}
+      role="button"
+      tabindex="0"
+    >
+      <div class="flex items-stretch">
+        <% hex = @sandbox.color || "#9ca3af" %>
```
midigofrank: This is an anti-pattern: https://hexdocs.pm/phoenix_live_view/assigns-eex.html#variables . 
> Due to the scope of variables, LiveView has to disable change tracking whenever variables are used in the template, with the exception of variables introduced by Elixir block constructs such as case, for, if, and others

You can add it to the assigns instead. Something like: `assigns = assign(assigns, :hex, assigns.sandbox.color || "#9ca3af")`

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -0,0 +1,316 @@
...
+            <button
+              id={"duplicate-sandbox-#{@sandbox.id}"}
+              type="button"
+              class="rounded-lg p-2 hover:bg-slate-100 transition-colors"
+              phx-click={JS.patch(~p"/projects/#{@project.id}/sandboxes")}
+              phx-stop-click
+              phx-hook="Tooltip"
+              aria-label="Duplicate is coming soon"
+            >
+              <Heroicons.clipboard_document class="h-4 w-4 text-slate-400" />
```
midigofrank: FYI, we're trying to move away from the `Heroicons` lib because it brings all those svgs in the app. You can read this thread in elixirforum to know more: https://elixirforum.com/t/heroicon-defined-in-core-components-not-working/61182/7?u=midigof. 

You should use the `<.icon name="hero-clipboard-document" class="h-4 w-4 text-slate-400" />` style

lib/lightning_web/live/sandbox_live/index.ex
midigofrank, line 1: - Instead of `sb`, can we just call it `sandbox`? 
- instead of `cs`, why not `changeset`? 
- `confirm_cs` should just be `confirm_changeset`

By the way, I always try to avoid shortened variable names as much as I can. I only use them in higher order functions, i.e. `Enum.map(sandboxes, fn sb -> end)`

lib/lightning_web/components/new_inputs.ex
elias-ba: Nice catch, just removed them. Thanks

lib/lightning_web/live/sandbox_live/components.ex
elias-ba: This is removed in my last commit. But thanks for the catch and the link.

lib/lightning_web/live/sandbox_live/components.ex
elias-ba: This is also fixed in my last commit. Thanks

lib/lightning_web/live/sandbox_live/index.ex
elias-ba: Agreed, thanks for the catch. Fixed it

josephjclark:
@elias-ba I know you're on top of all this but just for the record, here's what needs doing:
* [x] Show the main project in the Sandboxes list (pinned to top)
* [x] Use a feature flag to disable the Sandboxes menu item (then I can test with it enabled, neat!)
* [x] CSS doesn't live update when Editing a sandbox, only for new
* [x] Looks like the whole Workspace still isn't visible under the sandbox list?
* [x] When I enter a sandbox, then Edit it, change the colour and hit save... I seem to navigate back to the top list? App goes back to blue

elias-ba:
To access the `Sandboxes` menu item in the left sidebar, go to the User Profile Management page, find the *Experimental Features* section, and enable it by toggling the switch. cc @josephjclark 

Screenshot 2025-09-15 at 17 11 15

lib/lightning_web/live/helpers/project_theme.ex
```diff
@@ -0,0 +1,176 @@
...
+    --primary-bg-dark: var(--color-primary-900);
+    --primary-text-light: var(--color-primary-300);
+    --primary-text-lighter: var(--color-primary-200);
+    --primary-ring: var(--color-gray-300);
+    --primary-ring-focus: var(--color-primary-600);
+    """
+    |> String.replace("\n", " ")
+  end
+
+  defp build_scale(hex) do
```
josephjclark, line 57: @elias-ba do we really need all this complexity here? Is this AI generated?

If we need colour utils like this, surely there's a standard library we can use?

josephjclark:
Looking absolutely great @elias-ba :tada: 

## Colours

Raised a PR with an alternative palette suggestion

I note that buttons are a different hue to the left border. I'm not sure that's right? I think they should match. Also the button colours, particularly in the weaker shades, are too light/bright, and don't contrast well against the white background.

I guess this is why you have all that colour converting code, to generate 3 hues based on one starting point?

Also, two UI elements are not dynamically coloured:

image

The link text should probably match the theme, and the radio button certainly should

## Sandbox Editing

It's a little weird to edit a sandbox, change the colour, watch the background change, and then when you hit "save" it reverts back to your current sandbox. I think the behaviour is correct - just visually weird. It's also annoying that the modal greys out the background as you click, because you're not seeing the true colour of the app.

I don't think we should do anything about these today, but I think we'll get feedback on them later.

If we have three buttons: `Cancel`, `save` and `SAVE & ACTIVATE`, it might be a little clearer what the behaviour is

## Edit Form

I can't work out what it is, but the padding on the sandbox name is a bit off:
image

Looks like there's padding or extra spaces in there - but looking in the inspector I can't actually work out what is is :thinking:  Would it work better to just use a `<code>` element?

Later, the environment form should auto-complete to use environment names already defined for this project. We'll have to think more about how that works.

## Sandbox List

Apart from the pinned OG< can we sort sandboxes by name?

The padding between the main/pinned projects and the others is bigger:

image

I suspect this is deliberate, to separate the sandboxes, and it _almost_ works. But if we're going to seperate, let's do it properly with a subtitle or at least a divider.

I think a label on the card does everything we need to there.

So for now, for this PR at least, can we just make the padding consistent? Then we'll swing back when everything is done and decide whether we need to make the OG clearer.

## Card Affordance

It's not clear what happens when you click on a sandbox card, or even whether it's clickable.

There's a very subtle shift in the border colour, and a change in the pointer to be fair.

On the Workflows page, the workflow list item goes grey when you hover, which helps me understand that the whole element is interactive. Can we quickly add the same styles here?

## Workflows

When a sandbox is active and I click into Workflows, it takes me to the Workflows of main.

I presume this will be fixed in the URL work?

lib/lightning_web/live/helpers/project_theme.ex
elias-ba: Good question @josephjclark. You're absolutely right to call out the complexity here, this is definitely some custom color math for what might seem like a straightforward feature.

I was probably a bit too comfortable writing this because of my computer vision background where I was used to wrestling with color spaces and doing RGB-to-HSL conversions by hand. It's like when you have a hammer, everything starts looking like a nail.

But you make a valid point about libraries. The core math here is really just standard color space conversions, and I'm sure there are Elixir packages that could handle the RGB to HSL dance without us having to manually implement the "is this hue in the red sector or the green sector" logic. Libraries like [chameleon](https://hexdocs.pm/chameleon/readme.html) might save us from writing our own `remf/2` function.

The palette generation logic with the saturation adjustments might still need to be custom since we have specific design requirements, but the underlying color conversions could definitely be outsourced to someone who's already solved the "wait, why is my hue calculation off by 60 degrees" problem.

Let me take a look at Chameleon and I'll refactor the color math to use it while keeping our custom logic focused on just the palette generation part.

josephjclark: You wrote this by hand?? That's incredible!

I wouldn't block the PR on this, I just think it's worth a conversation. If it takes longer to port to chameleon maybe it's not worth the effort Might be worth a third opinion from Stu or Frank on the best way forward

elias-ba: This addressed. We are now using Chameleon ! 🎉 

Reduced about 80 lines of custom color space math and replaced them with simple library calls. Chameleon handles all the RGB - HSL conversions that were causing the "hue calculation off by 60 degrees" headaches, while we kept the important design-specific logic (the saturation adjustments and lightness targets) that makes the color palettes actually look good.

The key insight was using `Chameleon.convert()` directly rather than the two-step struct creation process, which eliminated the hex prefix issues we were seeing. Now it cleanly handles colors from the database (with #) and generates the same quality color scales as before.

Also removed custom floating-point modulo functions, sector-based hue calculations, and all the RGB/HSL conversion edge cases, while maintaining the same visual output. The palette generation logic with the careful saturation adjustments stayed exactly the same since that's the domain-specific value we were adding.

elias-ba:
Hey @josephjclark I addressed most of the the UI change requests you proposed. Here is a list of what is changed and is not.

- [ ] Button hues - Let's do this separately as it may be too much back and forth
- [x] Link and toggle buttons to match sandbox theme
- [x] Right padding on sandbox name in creation or edition form
- [x] Bottom padding between main project and sandboxes
- [x] Match sandbox card hover effect to workflows list item hover effect
- [x] ~When a sandbox is active and I click into workflows, it takes me to the workflows of main - this will be done in the URLs PR~ When a sandbox is active and I click into workflows, it takes me to the workflows of that sandbox. **A sandbox is just another project**

josephjclark:
Colours still aren't right (that's on me) - tomato, crimson and vermillion are all way too close. But that shouldn't block this PR. I've opened that out as https://github.com/OpenFn/lightning/issues/3587

elias-ba:
Thanks @josephjclark for the colors. I agree the palette I have chosen is not great. I have made it such that it's really easy to change the palettes. Just change here: https://github.com/OpenFn/lightning/blob/c633c035c2538472c44343bd85e57dc9c6b528ab/lib/lightning_web/live/sandbox_live/components.ex#L7

josephjclark:
@elias-ba the only thing maybe missing is the active sandbox indicator

I like the new `main` label:

image

But still think the Active indicator is unclear and a bit inappropriate:

image

Options:
1. Create an issue to set a better active label later
1. Make a change now and use the `main` styles for the active sandbox (maybe use the theme colour but I think even the grey is fine)
1. Ignore this issue and move on

I'd love to do (2) before we merge this!

elias-ba:
Hey @josephjclark I finally went for option 2. Just a gray badge similar to the environments badges but the label is `active`. You can do one last click testing / review. But I think this is now ready for merge. It also has the changes we discussed with @taylordowns2000 this morning. Thanks Taylor for your eagle eyes and for raising the wonderful discussion we had.

Screenshot 2025-09-18 at 21 11 36

josephjclark:
Breadcrumbs aren't right but this will be fixed in #3474

elias-ba:
Yeah about breadcrumbs, I was thinking that it's gonna be weird to have them now without having URLs matching them.

josephjclark:
@elias-ba deleting a sandbox fails right now - is this expected? I'm not surprised by it. Maybe just to keep this PR tidy we should disable the delete button?

image

josephjclark:
@elias-ba not sure about the breadcrumb/urls thing - I think the breadcrumb is super important right now. But let's take that discussion out of this PR :)

elias-ba:
Hey @josephjclark two things:

1. I think we should keep breadcrumbs out of this PR. I created a placeholder issue [here](https://github.com/OpenFn/lightning/issues/3591) so we can continue that discussion separately.

2. About sandbox permissions, you actually caught a bug in deletion (thanks 🙏). I fixed it and also took the opportunity to introduce a **Permission manager module** for sandboxes, following the same pattern we use in other parts of the app. This centralizes the rules so that when we need to refine who can do what, it’s just one place to change.

The current rules are:

* **Provisioning**: You can create a new sandbox if you’re an **owner/admin of the parent project**, or a superuser.
* **Updating**: You can update a sandbox if you’re a **superuser**, an **owner/admin of that sandbox**, or an **owner/admin of the root project**.
* **Deleting**: Same as update, but with two extra behaviors:

  * Deleting a sandbox will **also delete all its descendants** (prevents orphans being promoted into root projects).
  * If you delete the **currently active sandbox**, Lightning automatically redirects you back to the root project to avoid broken navigation.

So we now have a consistent and extensible permissions framework for all sandbox operations.

elias-ba:
Hey @josephjclark since this is isolated under experimental features, I already did a round of basic QA to confirm nothing breaks, and we’ll still have proper QA in staging anyway. Plus, this code has been sitting here for a while. I’d suggest we go ahead and merge, then catch anything in staging if needed (I’m pretty confident it’s stable). What do you think?

Issue #3478, opened 2025-08-04
sandboxes: version pruning


josephjclark:
Every time a project is merged or deleted (or both!), we need to prune the version histories of all workflows.

josephjclark:
@midigofrank  here is the squashing issue but I need to add notes to it

josephjclark:
90% sure squashing is as simple as: if the source of the latest version is the same as new source, replace the latest version. Otherwise, push a new version

josephjclark:
Squashing is done - pruning still needs design. It can wait.

PR #3563, opened 2025-09-08
Generate hash on workflow save

### LINKED ISSUE #3452: sandbox: generate version hash
A version hash is a character string that uniquely identifies a particular version of a Workflow. It is not the same as, or related to, a snapshot.

A hash is not reversible/decodable: you cannot re-build a project or workflow from a hash alone.
Versions can be generated in the app or in the CLI. So kit and lightning will need to duplicate the hashing algorithm.

The hash includes a source string, which must be passed into the function (you can default it to, or just start with, `app:`. The source string identifies why the version was generated (this is important later, for squashing versions)

The algorithm is not complex:

	- Create a list
	- Add the workflow name to the start of the list
	- For each node (trigger, step and edge) in a consistent order
		- Get a sorted list of keys
		- Filter out ignored keys
		- Add each key and value to the list
	- Join the list into a string, no separator
	- Hash the list with SHA 256
	- Truncate the list to 12 (?)
	- Prepend the source string to the hash, like `<source>:<hash>`

I don't really know how long the hash needs to be - let's start at 12 chars.

The nodes must be consistently ordered. How do we ensure this? We could generate an id for each and sort by id (can't use UUID, it'll change arbitrarily across sandboxes). So let's do `trigger-jobs-edges`, sorted by name, where a name is generated for edges (source-target) and triggers (type).

The ignore list starts with `uuid`. What else do we include?
	
Notes: 
- Changes to UUIDs in lightning should not trigger version diff
- trivial changes, like workflow or step name change, should trigger version
	- name changes will change the ordering of the things in the hash. I think this is fine
- node positional change probably should trigger a version

midigofrank:
## Description

This PR creates a new workflow version when saving. 

- A new version is created only if the tracked changes have changed
- If the current latest version has the same source as the incoming version, then we squash i.e. delete the old version

Closes #3452 

## Validation steps

Can be tested via UI if you can get access to a db client. You make a change via UI then check for records in the db. 
I'd advise looking at the test cases in the changes

1. I have removed the unique constraint on workflow versions for `[:workflow_id, :hash]`. It is possible to have a combination of `a (app) -> b (cli) -> a (app)`
2. Squashing is not implemented for bulk insertion. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning/workflow_versions_test.exs
+++ b/test/lightning/workflow_versions_test.exs
@@ -585,7 +585,7 @@ defmodule Lightning.WorkflowVersionsTest do
     end
 
     test "generates the same hash for this simple workflow" do
-      predetermined_hash = "173cfef43fe8"
+      predetermined_hash = "a3605e1f6463"
 
       simple_workflow = %{
         name: "Simple Workflow",
@@ -612,6 +612,7 @@ defmodule Lightning.WorkflowVersionsTest do
           %{
             id: "trigger1",
             type: :webhook,
+            cron_expression: nil,
             enabled: true
           }
         ],
@@ -619,13 +620,19 @@ defmodule Lightning.WorkflowVersionsTest do
           %{
             source_trigger_id: "trigger1",
             target_job_id: "job1",
+            source_job_id: nil,
             condition_type: :always,
+            condition_label: nil,
+            condition_expression: nil,
             enabled: true
           },
           %{
+            source_trigger_id: nil,
             source_job_id: "job1",
             target_job_id: "job2",
             condition_type: :on_job_success,
+            condition_label: nil,
+            condition_expression: nil,
             enabled: true
           }
         ]


midigofrank:
Hey @elias-ba @josephjclark , kindly go through the additional notes in the description and let me know. 
tldr; 
I have implemented squashing but only when inserting a single hash. For the bulk insertion, it's a little tricky, should probably be handled in https://github.com/OpenFn/lightning/issues/3478

test/lightning/workflow_versions_test.exs
```diff
@@ -46,10 +46,115 @@ defmodule Lightning.WorkflowVersionsTest do
       assert {:error, :invalid_input} =
                WorkflowVersions.record_version(wf, @a, "web")
     end
+
+    test "does not insert duplicate when hash is same as latest" do
```
josephjclark: Hmm, not totally sure on these behaviours.

I think the correct behaviour is:

* If the source is the same, and the hash is the same, then it's basically a no-op. No new row inserted, and I guess we don't update the timestamp.
* If the source is the different but the hash is the same, then we must push a new row. So I think the assertion around line 63 is wrong
* If the source is the same and the hash is different, we squash (so still no new row)

I suppose the version history we're comparing doesn't include the source - so maybe I'm just getting misled by the test here?

josephjclark: Thinking about this I really think you need to include the source in the version history - so `['cli:aaaaa', 'app:bbbb']`

test/lightning/workflow_versions_test.exs
```diff
@@ -46,10 +46,115 @@ defmodule Lightning.WorkflowVersionsTest do
...
+      assert count_rows(wf.id) == 1
+
+      # Insert second version from cli
+      assert {:ok, wf2} = WorkflowVersions.record_version(wf1, @b, "cli")
+      assert wf2.version_history == [@a, @b]
+      assert count_rows(wf.id) == 2
+
+      # Insert third version from cli (same source as latest) - should squash
+      assert {:ok, wf3} = WorkflowVersions.record_version(wf2, @c, "cli")
+      # @b should be replaced by @c
```
josephjclark: This way of articulating the tests is absolutely brilliant - thank you! :trophy:

test/lightning/workflow_versions_test.exs
```diff
@@ -46,10 +46,115 @@ defmodule Lightning.WorkflowVersionsTest do
...
+      assert count_rows(wf.id) == 4
+
+      # All hashes should exist in database
+      assert Repo.get_by(WorkflowVersion, workflow_id: wf.id, hash: @a) != nil
+      assert Repo.get_by(WorkflowVersion, workflow_id: wf.id, hash: @b) != nil
+      assert Repo.get_by(WorkflowVersion, workflow_id: wf.id, hash: @c) != nil
+      assert Repo.get_by(WorkflowVersion, workflow_id: wf.id, hash: @d) != nil
+    end
+
+    test "handles edge case: first version (no latest to compare)" do
```
josephjclark, line 152: :clap:

josephjclark:
@midigofrank tell me more about the bulk versioning API. What is it and why do we need it?

test/lightning/workflow_versions_test.exs
```diff
@@ -180,6 +285,328 @@ defmodule Lightning.WorkflowVersionsTest do
     end
   end
 
+  describe "generate_hash/1" do
+    test "generates consistent hash for same workflow structure" do
+      project = insert(:project)
+
+      workflow =
```
josephjclark, line 270: I would love to see this represented here as a JSON or YAML object, with simple uuids like `1`, `2`, `3`, and to see actual hash values asserted in the tests.

This would help us compare versions against the CLI

midigofrank: Yeah, good idea. I'll add it

midigofrank:
> @midigofrank tell me more about the bulk versioning API. What is it and why do we need it?

Currently you can create versions in 2 ways:

1. Single hash: `record_version(%Workflow{}, "hash", "app")
2. Multiple hashes: `record_versions(%Workflow{}, ["hash1", "hash2", "hash3"], "app")

I think the multiple hashes here was an anticipation for the cli, kind of like git, where the user might commit multiple times and then finally push. I don't think we need it anymore because of [squashing](https://github.com/OpenFn/lightning/issues/3478#issuecomment-3252736340).

test/lightning/workflow_versions_test.exs
midigofrank: Aaah, that's good to know. Let me make the change

josephjclark:
@midigofrank ah right, yeah CLI should only ever send you one version, which it generates on-the-fly just before deploying. If we even bother - I mean it makes just as much sense to just send the state and let lightning generate the version hash.

CLI still needs to compare versions locally though. And maybe it's good for referential integrity to include the CLI version hash with the state file, like a checksum.

Or we could use it as a pre-validation: send the version hash, ask lightning if it matches, and if there's a diff, _then_ we send the full file.

So all sorts of uses.

Anyway I'm distracting myself: none of that requires a bulk API. I'd be inclined to drop it

midigofrank:
Agreed

midigofrank:
Sorry @josephjclark , I forgot to include the json structure in the tests. Don't review just yet, I'm adding it

elias-ba (approved):
@midigofrank this looks good to me. Great work, man 👏🏽

test/lightning/workflow_versions_test.exs
midigofrank: @josephjclark I have added a test with a predetermined hash here: https://github.com/OpenFn/lightning/pull/3563/files#diff-b61eca0a74785a9231c824e1d6f672b69ffe229789077d2afcb7c6a145122ce2R587

midigofrank:
Aaah, good point @josephjclark . I don't think we should hide it behind a feature flag, let's leave it and see how the versions are building up

PR #3566, opened 2025-09-08
Rollback "claimed" runs to "available" if worker doesn't join run channel

### LINKED ISSUE #3565: Runs are marked as lost when they should be re-enqueued
What's happening now can be reproduced on `main` by adding `Process.sleep(10_000)` right after line 39 of `worker_channel.ex` then running any run. Here's how the bug works:

1. The `ws-worker` joins the worker channel and requests work via the "claim" API.
2. Every once in a while, the DB poops itself and the query takes a very long time. (This is what we replicate with the `Process.sleep`)
3. The `ws-worker` goes away, after telling us [`✘ TIMEOUT on claim. Runs may be lost.`](https://github.com/OpenFn/kit/blob/main/packages/ws-worker/src/api/claim.ts#L119-L122)
4. The DB query _then_ finishes, setting the run to `:claimed` and replying to **_Nobody At All_** that it's got a run ready to be executed.
5. **_Nobody At All_** does anything (the worker is long gone at this point) and the run sits there as `:claimed` for N minutes—controlled by the Janitor's grace period.
6. The Janitor (after N minutes) comes by and marks this run as `:lost`, secretly giving it the status "LostAfterClaim".
7. We tell the user that we don't know what happened, apologizing profusely for losing the run and saying, "Sorry, we have no more info and there's nothing you can do."

Really, it seems like there _is_ something (or there are a couple of somethings) we can do here:

1. In the `worker_channel`, Make the socket timeout GREATER OR EQUAL to the query timeout so that we don't shut the door before the positive reply comes.
2. In the `worker_channel` Check if the connection to the worker is still alive before we `{:reply, {:ok, %{runs: runs}}, socket}`. If it's not still alive, then roll back the transaction and/or (more simply) set the run state back to `:available` so another worker can come along and claim it.
3. In the `janitor.ex`, don't set runs to `:lost` (with `LostAfterClaim`), set them back to `:available` so they get picked up. If we're **_certain_** that the run hasn't been started (`:claimed` is a different state than `:started`) then simply put it back in the queue... it's not lost, it's still waiting for some worker to work on it.

@stuartc , @josephjclark , @rorymckinley , a penny for your thoughts here? Of the last 230 lost runs we know of, 212 were `LostAfterClaim` and it seems that implementing any/all of these three enhancements might put a real dent in that.

I've opened an illustrative fix for concept 2 here: #3566 - if it has legs I'll write tests. so far just testing manually and it works a treat.

taylordowns2000:
Does two things:

1. Fixes #3565 
2. Renames some variables to be less ambiguous - it was confusing for me not knowing if they were seconds or milliseconds

**Question:** Should we add some sort of state to the run_channel to make sure that it cannot be joined by more than one worker? (Imagine we roll back after 30 seconds, and right at that moment worker A joins the run channel. Worker B will then join also once it gets its own claim. It would be awesome to either invalidate the first key or block joining the run channel once we decide to roll back. Is that possible?)

## Validation steps
1. set `RUN_CHANNEL_JOIN_TIMEOUT_SECONDS=15`
3. add process.sleep(11_000) as described in the issue (line 64/65 in worker_channel.ex in this PR)
4. run a run and see what happens!
5. it should be claimed, the worker should timeout, and then after 30 seconds it should be put back into available state.
6. if you remove the timeout, everything should work as usual

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI
- [x] Description of changes

#### Summary

This PR implements a timeout mechanism for the `WorkerChannel` to prevent runs from being stuck in `:claimed` state when workers disconnect or fail to join run channels after claiming runs. The mechanism works across clustered Lightning nodes using Phoenix.PubSub for cross-node communication.

#### Problem

Previously, when a worker claimed runs and then disconnected before joining the corresponding run channels, those runs would remain in `:claimed` state indefinitely, preventing other workers from claiming them. This created a resource leak where runs became unavailable.

#### Solution

- **Timeout Mechanism**: After a worker claims runs, a configurable timeout (default 30 seconds) is started
- **Automatic Rollback**: If the client doesn't join run channels within the timeout period, claimed runs are automatically rolled back to `:available` state
- **Cross-Node Communication**: Uses Phoenix.PubSub to work across clustered Lightning nodes
- **Optimization**: Only starts timeout when there are actually runs to track (no timeout for 0 runs)

##### Core Implementation
- **`lib/lightning_web/channels/worker_channel.ex`**: Added timeout logic, PubSub subscription, and rollback handling
- **`lib/lightning_web/channels/run_channel.ex`**: Added notification broadcast when run channels are joined
- **`lib/lightning/runs.ex`**: Added `rollback_claimed_runs/1` function for database rollback

##### Configuration
- **`lib/lightning/config/bootstrap.ex`**: Added `RUN_CHANNEL_JOIN_TIMEOUT_SECONDS` environment variable
- **`lib/lightning/config.ex`**: Added configuration callback and renamed existing callbacks for consistency

##### Testing
- **`test/lightning_web/channels/worker_channel_test.exs`**: Added comprehensive tests for timeout and cancellation scenarios

##### 🔍 **Critical Review Areas**

1. **Cross-Node Communication Logic** (`worker_channel.ex:39-44, 168-179`)
   - Verify PubSub subscription/unsubscription is handled correctly
   - Check that worker_id filtering prevents message cross-contamination
   - Ensure proper cleanup in `terminate/2`

2. **Timeout and Rollback Logic** (`worker_channel.ex:83-102, 119-135`)
   - Verify timeout only starts when `original_runs` count > 0
   - Check that rollback correctly resets run state to `:available`
   - Ensure timeout cancellation works when all runs are joined

3. **Database Rollback Function** (`runs.ex:350-374`)
   - Verify the `rollback_claimed_runs/1` function correctly updates run state
   - Check that it handles empty lists gracefully
   - Ensure proper logging of rollback operations

4. **Configuration Changes** (`config.ex:327-330, 453, 693-695`)
   - Verify new configuration follows existing patterns
   - Check that callback renaming doesn't break existing functionality
   - Ensure proper default values

##### 🧪 **Test Coverage**

5. **Timeout Test** (`worker_channel_test.exs:139-170`)
   - Verify test correctly simulates timeout scenario
   - Check that manual timeout triggering works in test environment
   - Ensure rollback verification is comprehensive

6. **Cancellation Test** (`worker_channel_test.exs:177-210`)
   - Verify PubSub message format matches implementation
   - Check that timeout cancellation prevents rollback
   - Ensure test cleanup removes configuration

##### ⚠️ **Potential Issues to Watch For**

7. **Memory Leaks**: Ensure PubSub subscriptions are properly cleaned up
8. **Race Conditions**: Verify timeout cancellation is atomic
9. **Configuration**: Check that environment variable naming is consistent
10. **Error Handling**: Ensure graceful handling of edge cases (empty runs, nil worker_id)

#### Configuration

The timeout can be configured via the `RUN_CHANNEL_JOIN_TIMEOUT_SECONDS` environment variable (default: 30 seconds).

#### Testing

All existing tests pass, and new tests cover:
- Timeout rollback when client doesn't join run channels
- Timeout cancellation when client successfully joins run channels
- Cross-node communication via PubSub

#### Backward Compatibility

This change is fully backward compatible. Existing functionality remains unchanged, and the new mechanism only activates when runs are claimed.

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/channels/worker_channel.ex
+++ b/lib/lightning_web/channels/worker_channel.ex
@@ -235,13 +235,9 @@ defmodule LightningWeb.WorkerChannel do
         )
       end)
 
-      # Roll back all pending runs at once
       run_ids = Map.keys(pending_run_timeouts)
 
       if not Enum.empty?(run_ids) do
-        # We need to get the actual run structs to roll them back
-        # For now, we'll log this - in practice, the runs will be rolled back
-        # by other mechanisms when the worker disconnects
         Logger.warning(
           "Worker channel terminating with pending run timeouts for runs: #{Enum.join(run_ids, ", ")}"
         )


lib/lightning_web/channels/worker_channel.ex
```diff
@@ -50,7 +51,19 @@ defmodule LightningWeb.WorkerChannel do
             }
           end)
 
-        {:reply, {:ok, %{runs: runs}}, socket}
+        # Check if the socket is still alive AND the client is still waiting for a reply
+        if Process.alive?(socket.transport_pid) do
```
josephjclark, line 88: Quick note ( I'm not at my desk yet but don't want to forget this thought): is this the right test?

If a claim event times out, I believe the socket is still connected. So I _think_ the socket is still alive and this will always be true.

> AND the client is still waiting for a reply

This bit.is interesting - is this implemented here?

I've always had this feeling that if Lightning replies after the worker has times out, that somehow the Lighting socket should know that. There should be some read receipt down at the socket level. And THAT would inform our rollback.

But maybe there's no such thing?

Maybe we DO need to implement an extra message - but that just feels clunky and brittle to me

taylordowns2000:
@rorymckinley , @josephjclark , @stuartc , i feel pretty good about this with one straggling question... reposting it from the PR itself down here:

> **Question:** Should we add some sort of state to the run_channel to make sure that it cannot be joined by more than one worker? (Imagine we roll back after 30 seconds, and right at that moment worker A joins the run channel. Worker B will then join also once it gets its own claim. It would be awesome to either invalidate the first key or block joining the run channel once we decide to roll back. Is that possible?)

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -107,4 +257,9 @@ defmodule LightningWeb.WorkerChannel do
     |> Runs.RunOptions.new()
     |> Ecto.Changeset.apply_changes()
   end
+
+  defp run_channel_join_timeout_ms do
+    # Default to 30 seconds if not configured
+    Application.get_env(:lightning, :run_channel_join_timeout_seconds, 30) * 1000
```
rorymckinley: @taylordowns2000 We could probably replace this with `Lightning.Config.run_channel_join_timeout_seconds`?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -31,6 +31,20 @@ defmodule LightningWeb.RunChannel do
            Runs.get_project_id_for_run(run) do
       Sentry.Context.set_extra_context(%{run_id: id})
 
+      # Notify the worker channel that this run channel has been joined
+      # Use PubSub for cross-node communication in clustered environments
+      case socket.assigns[:worker_id] do
```
rorymckinley, line 36: @taylordowns2000 `run_channel_test.exs` does not seem to have been updated to cover this additional functionality?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -50,7 +75,43 @@ defmodule LightningWeb.WorkerChannel do
...
+          Logger.error("Sending #{Enum.count(response_runs)} runs back!")
+
+          {:reply, {:ok, %{runs: response_runs}}, socket}
+        else
+          # Socket is no longer alive, roll back the transaction by setting runs back to :available
+          Logger.warning(
+            "Worker socket disconnected before claim reply, rolling back transaction for runs: #{Enum.map_join(original_runs, ", ", & &1.id)}"
+          )
+
+          Runs.rollback_claimed_runs(original_runs)
```
rorymckinley, line 123: @taylordowns2000 There does not appear to be test coverage here - is there a cheap way to test that the `Process.alive?` check failed?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -50,7 +75,43 @@ defmodule LightningWeb.WorkerChannel do
...
+              timeout_ref =
+                :timer.send_after(
+                  timeout_ms,
+                  self(),
+                  {:claim_timeout, original_runs}
+                )
+
+              # Store the timeout reference and original runs in socket assigns
+              assign(socket,
+                claim_timeout_ref: timeout_ref,
```
rorymckinley: @taylordowns2000 If I am reading this correctly, the reference to the timeout is set for the socket - does that make us vulnerable to the following:

Worker 1 sends a claim request for 1 run (R1)
Lightning sends back run details and sets timeout T1
Before Worker 1 joins the run channel and within the join timeout, Worker 1 sends through a claim for another run.
Lightning sends back run details (R2) and sets timeout T2.
Now, T1 has been replaced with T2, i.e. have we lost timeout coverage of R1?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -71,6 +132,64 @@ defmodule LightningWeb.WorkerChannel do
...
+    # Timeout occurred - client didn't join run channels in time
+    Logger.warning(
+      "Claim timeout reached, rolling back transaction for runs: #{Enum.map_join(runs, ", ", & &1.id)}"
+    )
+
+    {:ok, count} = Runs.rollback_claimed_runs(runs)
+    Logger.info("Successfully rolled back #{count} runs")
+
+    # Clear the timeout reference and pending runs from socket assigns
+    socket = assign(socket, claim_timeout_ref: nil, pending_runs: nil)
```
rorymckinley: @taylordowns2000 This line does not appear to have test coverage?

test/lightning_web/channels/worker_channel_test.exs
```diff
@@ -135,5 +135,86 @@ defmodule LightningWeb.WorkerChannelTest do
...
+
+      # Verify the run is claimed
+      run = Repo.reload!(run)
+      assert run.state == :claimed
+
+      # Manually trigger the timeout by sending the message directly
+      send(socket.channel_pid, {:claim_timeout, [run]})
+
+      # Wait a bit for the message to be processed
+      Process.sleep(100)
```
rorymckinley, line 165: @taylordowns2000 This feels like it could result in a flickering test? Could we poll with a timeout rather?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -71,6 +132,64 @@ defmodule LightningWeb.WorkerChannel do
...
+            nil ->
+              {:noreply, socket}
+
+            timeout_ref ->
+              :timer.cancel(timeout_ref)
+              socket = assign(socket, claim_timeout_ref: nil, pending_runs: nil)
+              {:noreply, socket}
+          end
+        else
+          # Still have pending runs, update the list
```
rorymckinley: @taylordowns2000 There does not appear to be test coverage for the case where there are runs remaining (and the timeout does not get cancelled)?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -71,6 +132,64 @@ defmodule LightningWeb.WorkerChannel do
...
+
+  def handle_info({:run_channel_joined, run_id, worker_id}, socket) do
+    # Handle PubSub message for run channel join notification
+    # Only process if this message is for this worker
+    case socket.assigns[:worker_id] do
+      ^worker_id ->
+        handle_info({:run_channel_joined, run_id}, socket)
+
+      _ ->
+        # Message not for this worker, ignore
```
rorymckinley, line 190: @taylordowns2000 Do we have a test to cover the case where the message is not for this worker?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -91,6 +210,37 @@ defmodule LightningWeb.WorkerChannel do
...
+        case socket.assigns[:pending_runs] do
+          nil ->
+            :ok
+
+          pending_runs ->
+            Logger.warning(
+              "Worker channel terminating with pending runs, rolling back: #{Enum.map_join(pending_runs, ", ", & &1.id)}"
+            )
+
+            Runs.rollback_claimed_runs(pending_runs)
```
rorymckinley: @taylordowns2000 This line does not appear to have test coverage?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -31,6 +31,20 @@ defmodule LightningWeb.RunChannel do
...
+        nil ->
+          # No worker ID available, continue normally
+          :ok
+
+        worker_id ->
```
rorymckinley, line 41: @taylordowns2000 Are we vulnerable to a race condition here> E.g, we retrieve the run, broadcast a message that the timer can be reset and then send data back to the worker? In the gap between the broadcast and something acting on it, the timeout could kick in, unclaiming the run and potentially allowing  situation where two workers are running it?

This feels like the 'ack-ack' scenario (although we can use the run channel response as the 'ack-ack' - and I think that addresses your question about the Worker A, Worker B scenario.

When a worker joins the run channel, I think that Lightning should confirm that the run is still claimed, lock the run and apply a change that makes rollback impossible (it feels like the responsible way to do this would be to introduce a new state), and then only send the :ok back to the worker.

If none of these conditions are true, the Worker gets an error.

lib/lightning/runs.ex
```diff
@@ -345,4 +345,29 @@ defmodule Lightning.Runs do
...
+  so they can be claimed by another worker.
+  """
+  @spec rollback_claimed_runs([Run.t()]) :: {:ok, non_neg_integer()}
+  def rollback_claimed_runs(runs) do
+    # Set the runs back to :available state so they can be claimed by another worker
+    run_ids = Enum.map(runs, & &1.id)
+
+    {count, _} =
+      from(r in Run, where: r.id in ^run_ids)
+      |> Repo.update_all(
```
rorymckinley, line 363: @taylordowns2000 To avoid race conditions, it may be worth locking each record, checking that the record should be rolled back and then rolling it back.

lib/lightning_web/channels/worker_channel.ex
taylordowns2000: great catch. investigating now

taylordowns2000: @rorymckinley , i think you were right... does this do the trick? https://github.com/OpenFn/lightning/pull/3566/commits/0a874d9323de08d2175bd08f6f5536dc679b480a

lib/lightning_web/channels/run_channel.ex
taylordowns2000: @rorymckinley , would this be handled better by ensuring that a run channel can only be joined once? or by a single worker? (this was a thought I had in the initial PR.)

taylordowns2000: Eish, or would that prevent workers from re-connecting to the run channel if they got temporarily disconnected?

rorymckinley: @taylordowns2000 Even if a single-connect is an option I do not think it would solve the problem entirely and may introduce a new problem:

 Showing my workings, apologies for the repetition of what I said before.

I can't see any obvious place where we are checking the state of a run before we send stuff back to the worker (probably never needed it before?) - so race 1 would be that the run has already been rolled back (essentially Worker A sans Worker B). A single connect does not seem to solve this.

Race 2 would be that the run has not been rolled back but in between the broadcast and the action the run gets rolled back. Using the Highlander approach you suggest would prevent another worker picking up the run once it has rolled back but......

- The first worker would think that it is ok to work on this, but Lightning thinks it is available. What happens when the worker sends through 'started' - would Lightning need a moment to compose itself?
- If we only allow one connection to the run, would that block legitimate cases where we do want another Worker to connect. So, let us say that Worker A does connect too late, the run has been rolled back but not claimed yet. In this case, we would tell Worker A to fly a kite and we would want Worker B to be able to connect, but the Highlander principle would block it?

rorymckinley: And, oops: I forgot to add that I do not know enough about workers to know know if it will reconnect mid-run to the run channel - @josephjclark ?

taylordowns2000: Eish. Yeah, it does sound like adding another state (`:locked`?) might be useful. Let's see what @stuartc and @josephjclark think here before I invest more time in this.

rorymckinley: `:confirmed` ?

josephjclark: I believe the worker will just reconnect when the connection is lost. The socket should handle this.  I'll test locally later and see if I can confirm.

lib/lightning_web/channels/worker_channel.ex
rorymckinley: @taylordowns2000 It looks like it - I have a question re: this [comment](https://github.com/OpenFn/lightning/commit/0a874d9323de08d2175bd08f6f5536dc679b480a#diff-7d6fd8b04470e59981c7ff0ec1272156079a92abf72571a760967d4f12f58611R239).

Is that functionality still coming?

taylordowns2000: No, we shouldn't roll them back. We should let the Janitor have its way with them if the worker channel closes

taylordowns2000: I've removed the comment. Thanks for catching it!

rorymckinley: Thanks @taylordowns2000 - will you be adding tests later?

taylordowns2000: Only if Joe/stu say yes, they want this approach in the app.

josephjclark:
@taylordowns2000 

> Should we add some sort of state to the run_channel to make sure that it cannot be joined by more than one worker?

Not sure, I keep changing my mind.

It definitely feels like there are a few teeny tiny gaps where a claim might resolve just after the timeout, or two workers might join, or hhe rollback might occur just after a worker starts, or maybe the wrong worker joins. These are really tiny gaps but possible. They might just be impossible to close?

If a worker is disconnected mid-run, it should reconnect and  continue sending events. In fact I've just tested this and it works great (although some events may time out if the connection is down for long enough)

So you could add validation that only one worker can be in the channel at the time, or after a worker joins, only another worker with the same id can rejoin. But none of those are ironclad.

I would leave it: this PR probably fixes 99% of the LostBeforeStart cases. Let's put that into the wild and come back and thing about the remaining 1% if it remains a problem.

stuartc:
One comment here, will digest the "ack" aspect tomorrow.

The idea of checking the transport pid is alive feels odd to me, the socket and channel process (afaik) are tied to the transport process so they shouldn't be running, ie that check will always be true. I haven't seen logs indicating a message failed because the transport was dead.

The original conversation was about a theory I had that transactions were continuing after socket had closed, since the claim handler is synchronous right now I find it hard to believe the client is still listening to replies after 60s. We observed multiple claim transactions on the db still going after 60s. So I'm interested in trying to replicate this. The client should have disconnected (and the channel process dies and takes the transaction with it) or the worker should still be waiting for a response. Really want to determine if the worker is still connected at least via a socket and isn't going to consume the response.

So the logic seems ok as a safe guard for a worst case scenario, since the reply isn't ack'd in our current design.  But it shouldn't replace understanding what's happening (and I'm thinking it might get the way of debugging the scenario above).

lib/lightning_web/components/layout_components.ex
```diff
@@ -22,7 +22,7 @@ defmodule LightningWeb.LayoutComponents do
       )}
     <% else %>
       <Common.combobox
-        items={assigns[:projects] || []}
+        items={(assigns[:projects] || []) |> Enum.sort_by(& &1.name)}
```
josephjclark, line 25: Cheeky. I think Elias has fixed this on main now?

lib/lightning_web/channels/worker_channel.ex
```diff
@@ -50,7 +84,46 @@ defmodule LightningWeb.WorkerChannel do
             }
           end)
 
-        {:reply, {:ok, %{runs: runs}}, socket}
+        # Check if the socket is still alive
+        if Process.alive?(socket.transport_pid) do
```
josephjclark, line 88: yeah I'm with @stuartc here - checking the socket is alive isn't the right thing. 

Is this a hangover from the original PR that needs removing? Or is there still some value?

taylordowns2000:
closing as nobody really likes this 🤣  and we don't think lost runs are coming from this source anymore!

@josephjclark , @rorymckinley , @stuartc , feel free to open it up again if it's helpful

PR #3582, opened 2025-09-17
sandbox: project merge

### LINKED ISSUE #3432: sandboxes: backend API to merge a sandbox

midigofrank:
## Description

Closes #3432

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/projects/sandboxes.ex
+++ b/lib/lightning/projects/sandboxes.ex
@@ -475,4 +475,451 @@ defmodule Lightning.Projects.Sandboxes do
       m -> m
     end
   end
+
+  @doc """
+  Merges a sandbox workflow back onto its parent workflow.
+
+  This function matches nodes between source and target workflows using:
+  1. Structural position in the DAG (primary criterion)
+  2. Name and/or adaptor for disambiguation when structure is ambiguous
+
+  Returns params compatible with `Provisioner.import_document`.
+
+  ## Parameters
+    * `source_workflow` - The sandbox workflow with modifications
+    * `target_workflow` - The parent workflow to merge changes onto
+
+  ## Returns
+    A map with the merged workflow structure ready for import:
+    * Matched nodes keep target UUIDs with source properties
+    * Unmatched source nodes get new UUIDs
+    * Unmatched target nodes are marked for deletion
+    * Edges are remapped to final node IDs
+  """
+  @spec merge_workflow(Workflow.t(), Workflow.t()) :: map()
+  def merge_workflow(
+        %Workflow{} = source_workflow,
+        %Workflow{} = target_workflow
+      ) do
+    source_workflow = preload_workflow(source_workflow)
+    target_workflow = preload_workflow(target_workflow)
+
+    # Build DAG structures for both workflows
+    source_dag = build_dag_structure(source_workflow)
+    target_dag = build_dag_structure(target_workflow)
+
+    # Match nodes between source and target
+    {job_matches, trigger_matches} = match_nodes(source_dag, target_dag)
+
+    # Build merged jobs
+    merged_jobs =
+      build_merged_jobs(
+        source_workflow.jobs,
+        target_workflow.jobs,
+        job_matches
+      )
+
+    # Build merged triggers (webhook and cron only)
+    merged_triggers =
+      build_merged_triggers(
+        source_workflow.triggers,
+        target_workflow.triggers,
+        trigger_matches
+      )
+
+    # Build ID mapping for edge remapping
+    id_map = build_id_map(job_matches, trigger_matches)
+
+    # Remap edges using the ID mapping
+    merged_edges = build_merged_edges(source_workflow.edges, id_map)
+
+    %{
+      "id" => target_workflow.id,
+      "name" => source_workflow.name,
+      "jobs" => merged_jobs,
+      "triggers" => merged_triggers,
+      "edges" => merged_edges
+    }
+  end
+
+  defp preload_workflow(workflow) do
+    Repo.preload(workflow, [:jobs, :triggers, :edges], force: true)
+  end
+
+  defp build_dag_structure(workflow) do
+    # Build adjacency lists for structural analysis
+    edges_by_source =
+      workflow.edges
+      |> Enum.group_by(fn edge ->
+        cond do
+          edge.source_trigger_id -> {:trigger, edge.source_trigger_id}
+          edge.source_job_id -> {:job, edge.source_job_id}
+          true -> nil
+        end
+      end)
+      |> Map.reject(fn {k, _} -> is_nil(k) end)
+
+    edges_by_target =
+      workflow.edges
+      |> Enum.group_by(fn edge ->
+        if edge.target_job_id, do: {:job, edge.target_job_id}, else: nil
+      end)
+      |> Map.reject(fn {k, _} -> is_nil(k) end)
+
+    jobs_map = Map.new(workflow.jobs, &{&1.id, &1})
+    triggers_map = Map.new(workflow.triggers, &{&1.id, &1})
+
+    %{
+      jobs: jobs_map,
+      triggers: triggers_map,
+      edges_by_source: edges_by_source,
+      edges_by_target: edges_by_target,
+      edges: workflow.edges
+    }
+  end
+
+  defp match_nodes(source_dag, target_dag) do
+    # Match triggers first (simpler: only webhook and cron)
+    trigger_matches = match_triggers(source_dag.triggers, target_dag.triggers)
+
+    # Match jobs using structural position and disambiguation
+    job_matches =
+      match_jobs_structurally(source_dag, target_dag, trigger_matches)
+
+    {job_matches, trigger_matches}
+  end
+
+  defp match_triggers(source_triggers, target_triggers) do
+    # Only handle webhook and cron triggers
+    source_list =
+      source_triggers
+      |> Map.values()
+      |> Enum.filter(&(&1.type in [:webhook, :cron]))
+
+    target_list =
+      target_triggers
+      |> Map.values()
+      |> Enum.filter(&(&1.type in [:webhook, :cron]))
+
+    # Try to match by type and structural position
+    Enum.reduce(source_list, %{}, fn source_trigger, acc ->
+      # Find candidates with same type
+      candidates = Enum.filter(target_list, &(&1.type == source_trigger.type))
+
+      matched_target =
+        case candidates do
+          [single] ->
+            single
+
+          multiple when length(multiple) > 1 ->
+            # Try to disambiguate using cron_expression for cron triggers
+            if source_trigger.type == :cron do
+              Enum.find(
+                multiple,
+                &(&1.cron_expression == source_trigger.cron_expression)
+              )
+            else
+              nil
+            end
+
+          _ ->
+            nil
+        end
+
+      if matched_target do
+        Map.put(acc, source_trigger.id, matched_target.id)
+      else
+        # New trigger - generate new ID
+        Map.put(acc, source_trigger.id, Ecto.UUID.generate())
+      end
+    end)
+  end
+
+  defp match_jobs_structurally(source_dag, target_dag, trigger_matches) do
+    source_jobs = Map.values(source_dag.jobs)
+    target_jobs = Map.values(target_dag.jobs)
+
+    # Build structural signatures for all jobs
+    source_signatures =
+      build_job_signatures(source_jobs, source_dag, trigger_matches)
+
+    target_signatures = build_job_signatures(target_jobs, target_dag, %{})
+
+    # Match jobs based on structural position
+    Enum.reduce(source_jobs, %{}, fn source_job, acc ->
+      source_sig = Map.get(source_signatures, source_job.id)
+
+      # Find target jobs with matching structural signature
+      candidates =
+        target_jobs
+        |> Enum.filter(fn target_job ->
+          target_sig = Map.get(target_signatures, target_job.id)
+          signatures_match?(source_sig, target_sig, trigger_matches)
+        end)
+        |> Enum.reject(fn target_job ->
+          # Don't match already matched targets
+          Map.values(acc) |> Enum.member?(target_job.id)
+        end)
+
+      matched_target =
+        case candidates do
+          [] ->
+            nil
+
+          [single] ->
+            single
+
+          multiple ->
+            # Disambiguate using name and/or adaptor
+            disambiguate_jobs(source_job, multiple)
+        end
+
+      if matched_target do
+        Map.put(acc, source_job.id, matched_target.id)
+      else
+        # New job - generate new ID
+        Map.put(acc, source_job.id, Ecto.UUID.generate())
+      end
+    end)
+  end
+
+  defp build_job_signatures(jobs, dag, id_map) do
+    Map.new(jobs, fn job ->
+      parents = get_parent_nodes(job.id, dag)
+      children = get_child_nodes(job.id, dag)
+
+      # Map parent/child IDs if needed
+      mapped_parents = map_node_refs(parents, id_map)
+      mapped_children = map_node_refs(children, id_map)
+
+      signature = %{
+        parents: mapped_parents |> Enum.sort(),
+        children: mapped_children |> Enum.sort(),
+        parent_count: length(mapped_parents),
+        child_count: length(mapped_children)
+      }
+
+      {job.id, signature}
+    end)
+  end
+
+  defp get_parent_nodes(job_id, dag) do
+    dag.edges_by_target
+    |> Map.get({:job, job_id}, [])
+    |> Enum.map(fn edge ->
+      cond do
+        edge.source_trigger_id -> {:trigger, edge.source_trigger_id}
+        edge.source_job_id -> {:job, edge.source_job_id}
+        true -> nil
+      end
+    end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp get_child_nodes(job_id, dag) do
+    dag.edges_by_source
+    |> Map.get({:job, job_id}, [])
+    |> Enum.map(fn edge ->
+      if edge.target_job_id, do: {:job, edge.target_job_id}, else: nil
+    end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp map_node_refs(node_refs, id_map) do
+    Enum.map(node_refs, fn {type, id} ->
+      mapped_id = Map.get(id_map, id, id)
+      {type, mapped_id}
+    end)
+  end
+
+  defp signatures_match?(source_sig, target_sig, trigger_matches) do
+    # Check if structural positions match
+    # Parent and child counts must match
+    # Check if parent sets match (accounting for trigger mapping)
+    # Children matching is more complex - we check count for now
+    source_sig.parent_count == target_sig.parent_count &&
+      source_sig.child_count == target_sig.child_count &&
+      parents_match?(source_sig.parents, target_sig.parents, trigger_matches) &&
+      length(source_sig.children) == length(target_sig.children)
+  end
+
+  defp parents_match?(source_parents, target_parents, trigger_matches) do
+    # Map source trigger parents using trigger_matches
+    mapped_source_parents =
+      Enum.map(source_parents, fn
+        {:trigger, id} -> {:trigger, Map.get(trigger_matches, id, id)}
+        other -> other
+      end)
+
+    # For now, check if counts match - more sophisticated matching could be added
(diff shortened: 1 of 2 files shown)


lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+  alias Lightning.Projects.Project
+  alias Lightning.Repo
+  alias Lightning.Workflows.Workflow
+
+  @doc """
+  Merges a source project onto a target project using workflow name matching.
+
+  Maps workflows from source to target project based on exact name matching.
+  Uses the existing merge_workflow/2 logic for individual workflow merging.
+  Workflows that don't match are marked for deletion (target) or creation (source).
```
josephjclark, line 18: We'll probably need more control over this

Users will soon be given a UI which says "which workflows do you want to merge?", and they'll tick the boxes.

These partial merges should not destroy unmapped workflows in the target.

We also need allow users to map/rename workflows in this UI.

In Kit we've done the following (not linked to the CLI yet, so it's just internal functions)
- merge takes an options object
- There's an option to map source -> target workflows explicitly
- If no mappings are passed, by default we map everything by name association
- There's an option, default false, to delete any unmapped workflows in the target. This is for wholesale "replace this project with this project", rather than a "merge these workflows into this project"

For the record, I think it's rare we'll delete a workflow  from a sandbox. And if that's needed you can just go int other project and manually delete after merging. I suggest we release with that and find a better solution if and when users ask for it

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+
+  ## Parameters
+    * `source_project` - The project with modifications to merge
```
josephjclark, line 21: We find it helpful in kit to use `main` and `staging` to clarify the flow, I'd do that like this here:
```
* `source_project` - The project with modifications to merge (ie, staging)
```

lib/lightning/projects/merge_projects.ex
josephjclark: Happy to defer the mapping logic btw - I know I only asked for basic mapping here. But the behaviour of marking for deletion doesn't feel quite right

josephjclark:
This is going to be a beast to review :sweat_smile: 

I'll spend an hour or two with it tomorrow to get feel for the merge logic. Mostly though I want to implement https://github.com/OpenFn/kit/issues/1010  and test this stuff at the yaml level

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+  end
+
+  defp build_merged_edges(source_edges, target_edges, node_mappings) do
+    merged_from_source =
+      Enum.map(source_edges, fn source_edge ->
+        from_id =
+          Map.get(source_edge, :source_trigger_id) ||
+            Map.get(source_edge, :source_job_id)
+
+        mapped_from_id = Map.fetch!(node_mappings, from_id)
```
elias-ba, line 564: What will happen when the job matching algorithm fails to map all nodes due to complex dependencies or structural differences between workflows?

I think this will crash with a `KeyError` since some edges might reference unmapped job/trigger IDs. I am not sure, but could we try a lesser strict (non bang version) fetch call and handle potential failures more gracefully ?

```elixir
case Map.fetch(node_mappings, from_id) do
  {:ok, mapped_id} -> mapped_id
  :error -> 
    Logger.warning("Error message")
    nil  # or handle gracefully
end
```

Situations I am thinking could make this fail are these. Let me know if you think these are improbable edge cases and we're safe from them. From my understanding of this algorithm, these situations can still happen.

1. **Complex dependency chains**: source workflow: `JobA → JobB → JobC → JobD` but target workflow: `JobX → JobY` (shorter chain). The structural matching might only map JobA→JobX and JobB→JobY, leaving JobC and JobD unmapped.

2. **Renamed workflows with new jobs**: source has 5 jobs, target has 3 jobs with completely different names. Name matching fails, structural matching only maps 2-3 jobs, but edges still reference all 5 source job IDs.

3. **Partially deleted workflows**: target workflow had some jobs deleted since the sandbox was created. Source still has edges pointing to those deleted jobs, but they can't be mapped to anything in the target.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+
+  # Get mapped children IDs for a source job
+  defp get_mapped_children(job_id, adjacency_map, node_mappings) do
+    get_children(job_id, adjacency_map)
+    |> Enum.map(fn child_id -> Map.get(node_mappings, child_id) end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  # Get parent IDs for a job
+  defp get_parents(job_id, adjacency_map) do
```
elias-ba, line 378: I think this is O(n²) because during structural matching, we have nested loops where the outer loop processes each unmapped job (up to n jobs) and the inner operations call `get_parents` which scans all n entries in the adjacency map. Specifically:

1. `attempt_structural_matching` loops through unmapped jobs => O(n)
2. For each job, `get_mapped_parents` calls `get_parents` => O(n) scan
3. Same job also calls `get_mapped_children` which may trigger more `get_parents` calls for target candidates => O(n) scan  
4. This happens across multiple iterations of `iterate_structure_matching` => multiply by iteration count

So we get: n jobs × n adjacency entries × iterations = O(n²) minimum, potentially O(n³) with iterations.

Maybe we could try pre-building reverse lookups:

```elixir
defp build_adjacency_maps(workflow) do
  parent_to_children = build_workflow_parent_children_map(workflow)
  
  child_to_parents = 
    Enum.reduce(parent_to_children, %{}, fn {parent, children}, acc ->
      Enum.reduce(children, acc, fn child, acc ->
        Map.update(acc, child, [parent], &[parent | &1])
      end)
    end)
    
  {parent_to_children, child_to_parents}
end
```

This makes all parent/child lookups O(1) instead of O(n) scans.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+
+        multiple_candidates when length(multiple_candidates) > 1 ->
+          # Multiple candidates - use job body matching
+          body_matches = filter_matching_by_body(source_job, multiple_candidates)
+
+          case body_matches do
+            [single_match] ->
+              Map.put(acc, source_job.id, single_match.id)
+
+            [position_match | _rest] when last_iteration? ->
```
elias-ba, line 273: What does "position_match" actually represent here? I think this variable name is misleading since the candidates are sorted by connectivity (children/parents count) not position. Could you try renaming it to most_connected_candidate or best_connected_match to clarify what the fallback logic is doing?

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+  end
+
+  # Find match among candidates using job body comparison
+  defp filter_matching_by_body(source_job, candidates) do
+    source_body = source_job.body || ""
+
+    candidates
+    |> Enum.filter(fn candidate ->
+      candidate_body = candidate.body || ""
+      String.trim(source_body) == String.trim(candidate_body)
```
elias-ba: What's the intended use case for body matching? detecting identical jobs or finding similar jobs that users might have modified? I think this exact string comparison is very strict and would miss legitimate matches if users made minor edits like adding comments, reformatting, or small code changes.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+    source_workflow = Repo.preload(source_workflow, [:jobs, :triggers, :edges])
+    target_workflow = Repo.preload(target_workflow, [:jobs, :triggers, :edges])
+
+    merge_workflow(
+      Map.from_struct(source_workflow),
+      Map.from_struct(target_workflow)
+    )
+  end
+
+  def merge_workflow(source_workflow, target_workflow) do
```
elias-ba, line 91: Did you consider concurrent modifications to the target workflow during merge operations? I think this could cause data loss because the merge function works with potentially stale workflow structs without checking if the target was modified by another user. We use optimistic locking (`lock_version`) but this merge doesn't validate the target is still current. Could we try wrapping the existing logic in a transaction with a freshness check:

```elixir
def merge_workflow(source_workflow, target_workflow) do
  Repo.transaction(fn ->
    # Re-fetch target to ensure it's still current
    current_target = 
      from(w in Workflow, where: w.id == ^target_workflow.id, lock: "FOR UPDATE")
      |> Repo.one!()
    
    if current_target.lock_version != target_workflow.lock_version do
      Repo.rollback(:target_workflow_modified)
    end
    
    # Proceed with existing merge logic using fresh target
    node_mappings = map_workflow_node_ids(source_workflow, current_target)
    build_merged_workflow(source_workflow, current_target, node_mappings)
  end)
end
```

What do you think ?

lib/lightning/projects/merge_projects.ex
midigofrank: The logic at the top ensures that we're only dealing with already mapped nodes, so we are expecting the mapping to exist. I deliberately used the `fetch!` because if it doesn't, then the logic is broken

lib/lightning/projects/merge_projects.ex
midigofrank: Yeah, great catch

lib/lightning/projects/merge_projects.ex
midigofrank: `position_match` literally means, it's the first match based on how the nodes were positioned. It is not the "best" candidate here, all of them are because they all matched the body. Remember we are matching the body one to one and not based on similarity. 

I used `position_match` because ideally, we should be mapping through the nodes from `top to bottom, then left to right`.  Maybe I should use `first_match` here instead

lib/lightning/projects/merge_projects.ex
midigofrank: The cli merge spec uses strict body matching, I was just matching it here: https://github.com/OpenFn/kit/pull/965/files#diff-2e5bbe1186b26a054ca6c644c357cf81e943dcf015a6d81a6bc079162e55230cR309

lib/lightning/projects/merge_projects.ex
midigofrank: Great idea, but I think that check should happen during the action. We can even go around it by only supplying the `workflow_id` and then selecting the latest for both.

FYI, this function does no harm at the moment, all it does is return params that can be used to do the actual update. And I think we should keep it that way to have it easily testable.  We can add the checks to the caller

lib/lightning/projects/merge_projects.ex
elias-ba: Nice! You're totally right. Thanks for clarifying the design intention behind the `fetch!`. I also agree that we should fail when logic is broken.

lib/lightning/projects/merge_projects.ex
elias-ba: That makes perfect sense, thanks for explaining the workflow positioning logic! `first_match` would definitely be clearer.

lib/lightning/projects/merge_projects.ex
elias-ba: Got it. Makes sense to stay consistent with the CLI spec!

lib/lightning/projects/merge_projects.ex
elias-ba: That's a good point about keeping the function pure and testable. Moving the concurrency check to the caller makes more sense architecturally.

elias-ba:
Thanks a lot @midigofrank for the detailed responses. That gives me better context on some of the choices made here. I think now if you address the quadratic (maybe even cubic) complexity we discussed, we should be good to merge this.

lib/lightning/projects/merge_projects.ex
josephjclark: @elias-ba just for exact matches!

I'd like to explore Best Match or likeness a bit later on to try and catch trivial edits like comments. But not today!

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+      matched_target =
+        find_matching_trigger(source_trigger, target_workflow.triggers, acc)
+
+      if matched_target do
+        Map.put(acc, source_trigger.id, matched_target.id)
+      else
+        acc
+      end
+    end)
+  end
```
doc-han, line 118: If I'm getting it right, this is to say that workflows can have multiple triggers.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+
+  defp find_matching_trigger(source_trigger, target_triggers, node_mappings) do
+    target_triggers
+    |> Enum.reject(fn target ->
+      target.id in Map.values(node_mappings)
+    end)
+    |> Enum.find(fn target ->
+      target.type == source_trigger.type
```
doc-han, line 126: I assume there's a high probability code execution wouldn't frequently reach here. From what I understand in the code so far, code execution reaches here when there are more than one triggers(which I doubt there's a case for).

But if we reach here, do we just want to do a type check?

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+  end
+
+  defp match_jobs_by_name(node_mappings, source_workflow, target_workflow) do
+    Enum.reduce(source_workflow.jobs, node_mappings, fn source_job, acc ->
+      if Map.has_key?(acc, source_job.id) do
+        acc
+      else
+        target_job =
+          Enum.find(target_workflow.jobs, fn target_job ->
+            target_job.name == source_job.name and
```
doc-han, line 171: In CLI we compare the generated slug id (eg. `some-node-name`) instead of the name directly, I think one advantage is simply character casing.
hence, all characters are lowercased for comparison

lib/lightning/projects/merge_projects.ex
elias-ba: Thanks for the context, @josephjclark. Yeah, this makes total sense.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+    unmapped_target_jobs =
+      Enum.reject(target_workflow.jobs, fn job ->
+        job.id in mapped_target_ids
+      end)
+
+    # Special case: exactly 1 unmapped source and 1 unmapped target
+    case {unmapped_source_jobs, unmapped_target_jobs} do
+      {[single_source], [single_target]} ->
+        # Map the single remaining source to single remaining target
+        Map.put(node_mappings, single_source.id, single_target.id)
```
doc-han: I don't agree to the fact that if there's a single unmapped node on both sides we should just map them.

lib/lightning/projects/merge_projects.ex
```diff
@@ -0,0 +1,715 @@
...
+        # Sort by number of children (desc), then by number of parents (desc) for tie-breaking
+        children_count =
+          target.id |> get_children(target_adjacency_map) |> Enum.count()
+
+        parents_count =
+          target.id |> get_parents(target_adjacency_map) |> Enum.count()
+
+        {children_count, parents_count}
+      end,
+      :desc
```
doc-han, line 359: I think there's a little misunderstanding somewhere here. in kit we don't do the ranking here or this way. we don't count how many nodes matched by children and how many matched by parents. 

We do ranking inside the `get_parents` function. this is because you we can have 3 nodes being returned as a parent to the provided children. we rank them according to which parent matches most of the children. 

eg. 
on the target tree, there a node x with children a,b,c then we need to find what x maps to in the source tree using the children. but then in the source tree we have nodes, m and n where m has children a,b and n has children a,b,c. in this case we want to rank the returned parent nodes according to which matched the children the most. hence in the order of which matched the most, the returned candidates will be [n, m] competing for the position of x.
n has more priority over m.

test/lightning/projects/merge_projects_test.exs
```diff
@@ -0,0 +1,2443 @@
...
+      assert find_edge_by_names(result, "a", "b")
+      # b->c
+      assert find_edge_by_names(result, "b", "c")
+      # b->d
+      assert find_edge_by_names(result, "b", "d")
+
+      for result_edge <- result["edges"] do
+        refute result_edge["delete"]
+      end
+    end
```
doc-han, line 1001: I think we need to assert also the UUIDs of the edges being maintained instead of just asserting whether they exist. 

eg. edge `y-z` becomes `b-c` hence we want to assert that `b-c` maps to the uuid of `y-z`

test/lightning/projects/merge_projects_test.exs
```diff
@@ -0,0 +1,2443 @@
...
+        build(:workflow)
+        |> with_trigger(source_trigger)
+        |> with_job(source_job_a)
+        |> with_job(source_job_b)
+        |> with_job(source_job_c)
+        |> with_job(source_job_d)
+        |> with_job(source_job_e)
+        |> with_job(source_job_f)
+        |> with_job(source_job_g)
+        # trigger->a
```
doc-han: I think the util for creating a workflow could be simplified to look something like what we have in kit. makes it quite easy to follow the tests

test/lightning/projects/merge_projects_test.exs
```diff
@@ -0,0 +1,2443 @@
...
+      # b->d
+      assert find_edge_by_names(result, "b", "d")
+      # c->e
+      assert find_edge_by_names(result, "c", "e")
+      # d->f
+      assert find_edge_by_names(result, "d", "f")
+      # e->g
+      assert find_edge_by_names(result, "e", "g")
+      # f->g
+      assert find_edge_by_names(result, "f", "g")
```
doc-han: Once again, we need to assert what the edges map to. we need to reserver the edge uuids also

lib/lightning/projects/merge_projects.ex
midigofrank: Yes, technically it's possible to have multiple triggers.

lib/lightning/projects/merge_projects.ex
midigofrank: Yes, if there are more than one triggers, we should use the `type` to disambiguate. The type is like the name of the node

lib/lightning/projects/merge_projects.ex
midigofrank: Great catch. I'll do a down case for comparison

lib/lightning/projects/merge_projects.ex
midigofrank: Good catch, the cli doesn't match last unmapped. But I think we should for the sake of preserving original uuids.

lib/lightning/projects/merge_projects.ex
midigofrank: Hey Farhan, I think this is doing the exact same thing. The reason why I sort by children first is because we're iterating top-> down, meaning the children would ideally be matched last

midigofrank:
Hey @doc-han , I have implemented the `generateWorkflow` helper function and it has unearthed a potential flakiness with the algorithm. I have this failing test:
```
 # Source: trigger->a, a->b, b->c, b->d
 # Target: trigger->x, x->y, y->z, y->q
 # Special case: node 'b' has both parent and children changed
 # Should map: a->x, b->y, c->z, d->q
```
After all the passes, you end up with conflicting choices for `c and d`. They can all match `z and d` and so the only way to break the tie is picking the first one in the array. Question is, how was the order in the array determined in the first place? Was it timestamps? 
I think it should be Left-to-Right, based on the `x` position on the `workflow.positions` but this would deviate from the cli implementation. Also, there is no guarantee that the the workflow has `positions`

Let me know your thoughts, cc: @josephjclark

doc-han:
> After all the passes, you end up with conflicting choices for `c and d`. They can all match `z and d` and so the only way to break the tie is picking the first one in the array. Question is, how was the order in the array determined in the first place? Was it timestamps? 

If we can write the test to ensure that `c` maps to either `z` or `q` and then `d` maps to what remains of the two. It's acceptable

lib/lightning/projects/merge_projects.ex
midigofrank: Noted. Let's have that implemented in a follow up issue. For this issue I was targeting the current provisioner logic. Actually what this does is return params that can be later used by the provisioner. It's more like a Step 1

josephjclark:
@midigofrank @doc-han sorry chaps I missed that discussion. It's a great case and I think we'll need to find a fix later on.

I would suggest we sort by name or UUID?

There was a plan in the original algorithm to build a sorted pool of all the nodes and then process that. That would guarantee you a fixed order in both implementations.

PR #3623, opened 2025-09-30
Fix bootstrap script

elias-ba:
## Description
This PR fixes the bootstrap script to support both older Bash versions (3.1+) and Intel Macs, addressing compatibility issues that were preventing team members from setting up the development environment.

The original script used Bash 4.2+ features (`declare -g`, `declare -A`) and had hardcoded ARM64-specific paths, causing failures on:
- macOS with default Bash 3.2
- Intel Macs (x86_64 architecture)
- Systems with C++ header/Xcode CLT issues

Changes made:
- Removed Bash 4.2+ specific syntax (`declare -g` and associative arrays)
- Added dynamic Homebrew path detection for Intel vs ARM Macs
- Added Xcode Command Line Tools validation and repair instructions
- Extended platform-specific compilation to both architectures
- Improved error messages and diagnostic output

## Validation steps
1. Test on Intel Mac:
   - Run `./bin/bootstrap` on an Intel Mac (x86_64)
   - Verify Homebrew packages install to `/usr/local`
   - Confirm all dependencies install successfully

2. Test on Apple Silicon Mac:
   - Run `./bin/bootstrap` on M1/M2/M3 Mac
   - Verify Homebrew packages install to `/opt/homebrew`
   - Confirm all dependencies install successfully

3. Test with older Bash:
   - Run `bash --version` to confirm version 3.2.x
   - Execute script and verify no syntax errors

4. Test C++ header detection:
   - If headers are missing, verify helpful error messages appear
   - Follow the suggested fix commands and re-run

1. The script now requires Bash 3.1+ (2005) instead of 4.2+ (2011), making it compatible with macOS's default Bash 3.2
2. The `HOMEBREW_PREFIX` variable dynamically adjusts based on architecture
3. The C++ header check now includes instructions for completely removing and reinstalling Xcode CLT when needed
4. All array operations have been converted from associative to regular arrays

## AI Usage
- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist
- [x] I have implemented and/or tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -5,12 +5,15 @@ set -euo pipefail
 
 trap 'echo "❌ Script interrupted at line $LINENO"; exit 1' ERR INT
 
-# Global environment facts
-declare -g OS ARCH
-declare -g HAS_RUST HAS_HOMEBREW
-declare -g HOMEBREW_PACKAGES_INSTALLED
-declare -A MISSING_SYSTEM_DEPS
-declare -A MISSING_BREW_PACKAGES
+# Global environment facts (no declare -g needed at global scope)
+OS=""
+ARCH=""
+HAS_RUST=""
+HAS_HOMEBREW=""
+HOMEBREW_PACKAGES_INSTALLED=""
+# Use regular arrays instead of associative arrays
+MISSING_SYSTEM_DEPS=()
+MISSING_BREW_PACKAGES=()
 
 # Function to gather environment facts
 gather_environment_facts() {
@@ -25,7 +28,7 @@ gather_environment_facts() {
   if command -v node &>/dev/null; then
     echo "✅ Node.js: $(node --version)"
   else
-    MISSING_SYSTEM_DEPS[node]="Node.js"
+    MISSING_SYSTEM_DEPS+=("Node.js")
   fi
 
   if command -v elixir &>/dev/null; then
@@ -33,7 +36,7 @@ gather_environment_facts() {
     elixir_version="$(elixir --version 2>/dev/null | head -1 || echo "version unknown")"
     echo "✅ Elixir: $elixir_version"
   else
-    MISSING_SYSTEM_DEPS[elixir]="Elixir"
+    MISSING_SYSTEM_DEPS+=("Elixir")
   fi
 
   if command -v rustc &>/dev/null; then
@@ -41,7 +44,7 @@ gather_environment_facts() {
     echo "✅ Rust: $(rustc --version)"
   else
     HAS_RUST=false
-    MISSING_BREW_PACKAGES[rust]="Rust"
+    MISSING_BREW_PACKAGES+=("rust")
   fi
 
   # Check Homebrew availability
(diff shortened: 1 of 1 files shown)


bin/bootstrap
```diff
@@ -5,115 +5,147 @@ set -euo pipefail
...
   ARCH="$(uname -m)"
   echo "Platform: $OS $ARCH"
 
-  # Check system commands
+  if [[ "$OS" == "Darwin" ]]; then
+    if [[ "$ARCH" == "arm64" ]]; then
+      HOMEBREW_PREFIX="/opt/homebrew"
+    else
+      HOMEBREW_PREFIX="/usr/local"
+    fi
```
stuartc: Is this right? Are we saying that arm64 and x86 have different homebrew prefixes?

I think we should use `brew config` and pluck out the HOMEBREW_PREFIX from that.

I have a hunch that this _may_ be because of either an older version of macos, or that it is an older version of homebrew that have been carried across over a year or three?

bin/bootstrap
```diff
@@ -143,19 +172,26 @@ install_missing_homebrew_packages() {
...
-    # Check if libsodium is available (either already installed or just installed)
+  if [[ "$OS" == "Darwin" ]]; then
     if echo "$HOMEBREW_PACKAGES_INSTALLED" | grep -q "^libsodium$" || brew list libsodium &>/dev/null; then
-      export CPATH=/opt/homebrew/include
-      export LIBRARY_PATH=/opt/homebrew/lib
-      echo "🔧 Set environment variables for libsodium"
+      export CPATH="$HOMEBREW_PREFIX/include"
+      export LIBRARY_PATH="$HOMEBREW_PREFIX/lib"
+      echo "🔧 Set environment variables for libsodium (using $HOMEBREW_PREFIX)"
+    fi
```
stuartc, line 175: Nice, I like that we are separating the homebrew prefix, with the change requested above this will be perfect.

bin/bootstrap
```diff
@@ -143,19 +172,26 @@ install_missing_homebrew_packages() {
...
+
+    if command -v xcrun &>/dev/null; then
+      local sdk_path
+      sdk_path=$(xcrun --show-sdk-path 2>/dev/null || echo "")
+      if [[ -n "$sdk_path" ]]; then
+        export SDKROOT="$sdk_path"
+        export CPATH="$CPATH:$SDKROOT/usr/include"
+        echo "🔧 Set SDKROOT to $SDKROOT"
+      fi
```
stuartc, line 184: Whats this about? Why is this required?

elias-ba: Hey Stu! So this is for helping the compiler find C++ headers when things go sideways on macOS. 

Basically, Hanna was getting that `'cstddef' file not found` error when compiling crc32cer, and this was one of the things we tried. It sets up the SDK paths explicitly so the compiler knows where to look for system headers.

Turns out the real fix was just reinstalling Command Line Tools with `xcode-select --install` - that's what actually solved it for her. So honestly, we might not even need this bit!

But I figured it's harmless to leave it in as a belt-and-suspenders thing, you know, just in case someone's setup is wonky after a macOS upgrade or they have multiple Xcode versions or whatever. It won't break anything if CLT is properly installed, just sets some paths that are probably already set.

Happy to yank it if you think it's overkill though! What do you think?

bin/bootstrap
elias-ba: Yeah nice catch! I think we can just use `brew --prefix` like this: `HOMEBREW_PREFIX="$(brew --prefix)"`. Much cleaner than hardcoding the paths!

elias-ba: This is done, thanks

stuartc (approved):

PR #3651, opened 2025-09-30
string, not json from postgres

taylordowns2000:


taylordowns2000:
still need to figure out how to pretty print it

lib/lightning_web/controllers/dataclip_controller.ex
```diff
@@ -40,18 +41,43 @@ defmodule LightningWeb.DataclipController do
   end
 
   defp respond_with_body(conn, dataclip_id) do
-    dataclip = Invocation.get_dataclip_details!(dataclip_id)
+    import Ecto.Query
```
taylordowns2000: This should replace get_dataclipdetails

stuartc (approved):

PR #3672, opened 2025-10-02
use started_at, fallback to claimed_at

taylordowns2000:
## Description

Closes #__

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/runs/query.ex
+++ b/lib/lightning/runs/query.ex
@@ -40,12 +40,20 @@ defmodule Lightning.Runs.Query do
       where: r.state not in ^final_states,
       where:
         fragment(
-          "? + ((? ->> 'run_timeout_ms')::int + ?) * '1 millisecond'::interval < ?",
+          "COALESCE(?, ?) + ((? ->> 'run_timeout_ms')::int + ?) * '1 millisecond'::interval < ?",
+          r.started_at,
           r.claimed_at,
           r.options,
           ^grace_period_ms,
           ^now
-        ) or (is_nil(r.options) and r.claimed_at < ^fallback_oldest_claim)
+        ) or
+          (is_nil(r.options) and
+             fragment(
+               "COALESCE(?, ?) < ?",
+               r.started_at,
+               r.claimed_at,
+               ^fallback_oldest_claim
+             ))
     )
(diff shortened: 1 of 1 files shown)


lib/lightning/runs/query.ex
```diff
@@ -40,23 +40,37 @@ defmodule Lightning.Runs.Query do
...
+
+    grace_cutoff =
+      Lightning.current_time() |> DateTime.add(-grace_period_seconds, :second)
 
     from s in Step,
       join: r in assoc(s, :runs),
       on: r.state in ^final_states,
-      where: is_nil(s.exit_reason) and is_nil(s.finished_at)
+      where: is_nil(s.exit_reason) and is_nil(s.finished_at),
+      where: r.finished_at < ^grace_cutoff
```
stuartc, line 73: I'm not fully understanding this, why do we need a grace period here? Do we have evidence of a situation where a run is complete and it didn't mark the steps as complete AND they were marked as lost prematurely?

test/lightning/janitor_test.exs
```diff
@@ -128,6 +128,7 @@ defmodule Lightning.JanitorTest do
             starting_trigger: trigger,
             dataclip: dataclip,
             state: state,
+            finished_at: DateTime.utc_now() |> DateTime.add(-3600),
```
stuartc: Can we bind this to use the value from Config, lest we change anything and the test fails.

lib/lightning/runs/query.ex
taylordowns2000: We don't have evidence, but since step:complete and run:complete messages are sent separately, it's possible that one would arrive and the other would not. If that's the case, we don't want to call the run lost, but we want to call the step lost. I'd like a grace period here because (for network reasons?) the step complete could have failed but be later retried, and there's no great harm in waiting a few more seconds in case that step:complete message comes in from the cold.

test/lightning/janitor_test.exs
taylordowns2000: done @stuartc !

PR #3673, opened 2025-10-02
don't use Logger and Shell IO in tests

taylordowns2000:
This PR makes your tests look like this! 🤩  No extra logs, but you still see the errors/failed tests. 👍 

image

It also makes one small change to the DocumentSupervisor:

#### Problem
  During shutdown, `GenServer.stop/3` can raise various exit signals depending on race conditions:
  - `{:noproc, _}` - process already gone
  - `{:normal, _}` - process exiting while we stop it
  - `{:shutdown, _}` - process shutting down
  - Other exit reasons during concurrent termination

  Previously we only caught `{:noproc, _}`, allowing other exit signals to escape and log errors despite `capture_log: true`.

  #### Solution
  Changed from `catch :exit, {:noproc, _} -> :ok` to `catch :exit, _ -> :ok` for both child processes.

  #### Why this is safe

  1. **Narrow scope**: The try/catch wraps only the `GenServer.stop/3` call, not arbitrary code
  2. **Shutdown context**: We're in `terminate/2` - the supervisor is already terminating
  3. **Best-effort cleanup**: During shutdown, we attempt graceful stops but don't care if processes are already gone
  4. **Race conditions expected**: Child processes may exit between our monitor check and stop call
  5. **No error masking**: Any legitimate errors would have already been surfaced by the monitor's `:DOWN` message in
  `handle_info/2`

  This is a cleanup operation where "process already stopped" (regardless of how) is the desired outcome.

## Validation steps

1. Run the tests! Notice how you only see things you _want_ to see.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -16,7 +16,10 @@ Mimic.copy(Mix.Tasks.Lightning.InstallSchemas)
 # Other ExUnit configuration can be found in `config/runtime.exs`,
 # for example to change the `assert_receive` timeout, configure it using the
 # `ASSERT_RECEIVE_TIMEOUT` environment variable.
-ExUnit.configure(formatters: [JUnitFormatter, ExUnit.CLIFormatter])
+ExUnit.configure(
+  formatters: [JUnitFormatter, ExUnit.CLIFormatter],
+  capture_log: true
+)
 
 Mox.defmock(Lightning.Extensions.MockRateLimiter,
   for: Lightning.Extensions.RateLimiting
--- a/lib/lightning/collaboration/document_supervisor.ex
+++ b/lib/lightning/collaboration/document_supervisor.ex
@@ -102,7 +102,7 @@ defmodule Lightning.Collaboration.DocumentSupervisor do
       try do
         GenServer.stop(state.shared_doc_pid, :normal, 5000)
       catch
-        :exit, {:noproc, _} -> :ok
+        :exit, _ -> :ok
       end
     end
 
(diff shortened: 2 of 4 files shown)


test/test_helper.exs
```diff
@@ -16,7 +16,10 @@ Mimic.copy(Mix.Tasks.Lightning.InstallSchemas)
 # Other ExUnit configuration can be found in `config/runtime.exs`,
 # for example to change the `assert_receive` timeout, configure it using the
 # `ASSERT_RECEIVE_TIMEOUT` environment variable.
-ExUnit.configure(formatters: [JUnitFormatter, ExUnit.CLIFormatter])
+ExUnit.configure(
+  formatters: [JUnitFormatter, ExUnit.CLIFormatter],
+  capture_log: true
```
stuartc, line 21: My one big hesistation for this is that it hides real OTP smells, when working with processes and supervisors, if your cleanup isn't right - in test environments you all sorts of incorrectly tested or side-effects of process failures. I'm not 100% sure this specific setting swallows the kinds of errors I'm talking about.

stuartc: ~~I'm going to leave it in, but have it be false if LOG_LEVEL is debug.~~

Nah thats waaay too noisy. Leaving it set to true.

stuartc (approved):

midigofrank (approved):
Nicely done

PR #3759, opened 2025-10-15
3758 prevent totp reuse

### LINKED ISSUE #3758: Prevent reuse of TOTP code
Implement the `since` option on NimbleTOTP to prevent reuse of TOTP codes.

rorymckinley:
## Description

Adds a field to track when last a TOTP code was successfully validated, and implements NimbleTOTP's `since` functionality to prevent reuse of TOTP codes.

if the last_totp_at field is nil, the code is accepted, so this will allow us to grandfather existing users in.

Closes #3758 

## Validation steps

- **Before switching to this branch, setup MFA for a user - this will represent the case of an existing user when this change goes live**
- Switch to the feature branch
- Log in using MFA
- View the user - `last_totp_at` should be set  

- Add MFA to a different user
- Confirm that you can log in as the user with MFA
- Log out and open up 2 log in session in different browsers / browser sessions, stopping at the point where you have to provide the OTP
- Use the OTP in one session and before it expires, try to use the MFA in your other session - it should be rejected.
- Once you have logged in, confirm that you can delete MFA - the user should have MFA disabled and the last_totp_at should be nil.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/accounts.ex
+++ b/lib/lightning/accounts.ex
@@ -216,10 +216,7 @@ defmodule Lightning.Accounts do
     |> Multi.update(:user, fn %{totp: totp} ->
       totp = Repo.preload(totp, [:user])
 
-      Ecto.Changeset.change(
-        totp.user,
-        %{mfa_enabled: true, last_totp_at: Lightning.current_time()}
-      )
+      Ecto.Changeset.change(totp.user, %{mfa_enabled: true})
     end)
     |> Multi.run(:backup_codes, fn _repo, %{user: user} ->
       user = Repo.preload(user, [:backup_codes])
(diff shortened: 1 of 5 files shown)


rorymckinley:
> @rorymckinley great job here. I think we should move the `last_totp_at` to the `Lightning.Accounts.UserTOTP` schema instead of the `User`. This way we have everything related to the TOTP in one place

@midigofrank A valid argument :) - my counter-argument would be that the totp_usage is related to user behaviour rather than being linked to a specific instance of UserTOTP. I will concede that my argument is slightly weakened by the fact that I clear the `last_totp_at` when MFA is deleted :), but that may not always be the desired behaviour. In addition, User is already home to `mfa_enabled`.

midigofrank:
@rorymckinley fair point, but we don't use `mfa_enabled` for any TOTP verification. `mfa_enabled` can be used to denote other forms of authentication like SMS/Email codes. But for this case, we are using a particular `secret` code which is in the `UserTOTP` to verify.

rorymckinley:
> @rorymckinley fair point, but we don't use `mfa_enabled` for any TOTP verification. `mfa_enabled` can be used to denote other forms of authentication like SMS/Email codes. But for this case, we are using a particular `secret` code which is in the `UserTOTP` to verify.

And a point to you good sir :)  - I still think that the UserTOTP is the wrong place for `last_totp_at` but in the interests of getting the fix merged in, I will move it.

midigofrank (approved):

Issue #3102, opened 2025-04-10
Configurable webhook responses (in sync-mode)


taylordowns2000:
Both the **GovStack** and **OpenHIE** communities want/require to be able to build "client applications" that make `GET` requests to OpenFn endpoints (think: webhook urls) with designated access tokens (think: webhook auth methods) and get back some meaningful data in the reply.

This could be as simple as "reverse-proxying" a GET to `/i/abc123?patientId=007` via a single step workflow, as in "please do a get from OpenMRS `/api/patients?patientId=007` and forward the reply".

Or it could be as complex as "please run a whole workflow but don't reply to me until you're done".

Currently, we send a 201 with the created workorder and run ID so that the user can check the status of this async workflow later.

Here, we'd enable two additional modes for a workflow:
1. Async -> how OpenFn always worked in the past. (✅ - done)
2. Sync -> standardized response with a 201 and final `state` after the run finishes. (✅ - done in #3785 and #4323 )
3. **Custom** -> the user determines the response step, status code, and contents. 🎯

Note that this may mean we hold HTTP requests open for a _long time_ but that's OK... we envision future enhancements where users could designate "dedicated workers" for a high-traffic "channel" like this that does a simple reverse-proxy. (Edit, this other feature is tracked here: #4322 )

🔥 **Here's a [concept Figma](https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App--new-?node-id=5264-2051&p=f).** 🔥

<img width="1177" alt="Image" src="https://github.com/user-attachments/assets/2cd91fb0-ca8a-4d55-999b-e368368c71dd" />

<img width="1001" alt="Image" src="https://github.com/user-attachments/assets/4c410379-7c4a-49e4-8695-6e07cb71ddf1" />

<img width="1192" alt="Image" src="https://github.com/user-attachments/assets/0d74e9b5-8972-41b5-b124-6f8b6b02afb6" />

brandonjackson:
The underlying mental model here is that there are two common use cases and a long tail of possible options, which we shouldn't build ourselves, but that will have positive ROI for users. What would be helpful for me is to see an enumeration of what that long tail actually contains? 

1. Async
2. Sync 
3. Perform step 1 (which validates data format) then respond
4. Perform step 1 and 2 (which checks an external system to see if data is valid) then responds
5. IDK
6. ?
7. ?
8. ?
9. ?
10. ?

brandonjackson:
[...]

PR #3785, opened 2025-10-18
Handle "delayed response" (sync-mode) for webhook triggers

taylordowns2000:
This PR adds support for delayed webhook responses, allowing workflows to return their final state as the HTTP response to the original webhook request.

_**This is step 1, in service of https://github.com/OpenFn/lightning/issues/3102**_

A new webhook_reply enum field is added to triggers with three values:
1. `:before_start` (default) - Returns work order ID immediately (current behavior)
2. `:after_completion` - Waits up to 30s for workflow completion, then returns final state
3. `:custom` - Reserved for future use (we'll add a special step to workflows that is handled by lightning (not the worker) and triggers the "reply" event

When `:after_completion` is enabled, the webhook controller subscribes to a PubSub topic and blocks until the RunChannel broadcasts the final run state after execution completes. The response includes the worker's final state output and an appropriate HTTP status code based on the run result.

## Validation steps

1. Prerequisites: Connect a ws-worker running on the send_final_state branch (the official release is not yet available; tests will pass once the new worker version is released and package.json is updated)
2. Create a workflow with a webhook trigger
3. Update the trigger's webhook_reply field to `:after_completion`:
  ```ex
  trigger |> Ecto.Changeset.change(webhook_reply: :after_completion)
  |> Repo.update!()
  ```
4. POST to the webhook endpoint and observe:
  - Request blocks (does not return immediately)
  - After workflow completes, response contains the final state from the last job
  - Status code reflects run outcome (200 for success, 500 for failure/errors, 504 for timeout)
5. Verify default behavior still works:
  - Triggers with webhook_reply: :before_start return work order ID immediately

⭐️ The **TODO** ‼️ on lib/lightning_web/channels/run_channel.ex:200-201: This is the critical design decision about the "contract" with external systems. The current mapping of run states to HTTP status codes and response bodies should be carefully reviewed:
```ex
    case state do
      :success -> 200
      :failed -> 422
      :crashed -> 500
      :exception -> 500
      :killed -> 410
      :cancelled -> 499
      _other -> 500
    end
```

This mapping defines the public API contract and should be discussed before merging.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/controllers/webhooks_controller.ex
+++ b/lib/lightning_web/controllers/webhooks_controller.ex
@@ -141,6 +141,10 @@ defmodule LightningWeb.WebhooksController do
         |> json(%{error: error})
     after
       Lightning.Config.webhook_response_timeout_ms() ->
+        Logger.warning(
+          "Webhook response timeout for work_order: #{inspect(work_order.id)}"
+        )
+
         conn
         |> put_status(:gateway_timeout)
         |> json(%{


lib/lightning_web/controllers/webhooks_controller.ex
```diff
@@ -121,6 +125,32 @@ defmodule LightningWeb.WebhooksController do
...
+        conn
+        |> put_status(status_code)
+        |> json(body)
+
+      {:webhook_error, status_code, error} ->
+        conn
+        |> put_status(status_code)
+        |> json(%{error: error})
+    after
+      Lightning.Config.webhook_response_timeout_ms() ->
```
elias-ba, line 143: Could we add log statement when timeout occurs for observability ? Something like
```elixir
Logger.warning("Webhook response timeout for work_order: #{work_order.id}")
```

test/integration/web_and_worker_test.exs
```diff
@@ -266,7 +267,7 @@ defmodule Lightning.WebAndWorkerTest do
 
       version_logs = pick_out_version_logs(run)
       assert version_logs["@openfn/language-http"] =~ "3.1.12"
-      assert version_logs["worker"] =~ "1.15"
+      assert version_logs["worker"] =~ "1.17"
       assert version_logs["node.js"] =~ "22.12"
       assert version_logs["@openfn/language-common"] == "3.0.2"
```
elias-ba, line 272: Are these assertions super important ? I am wondering what happens when any of these versions is bumped ? Would we need to update the test anytime ?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -183,6 +189,52 @@ defmodule LightningWeb.RunChannel do
...
+        topic,
+        {:webhook_response, status_code, body}
+      )
+    end
+  end
+
+  defp determine_status_code(state) do
+    case state do
+      :success -> 200
+      :failed -> 422
```
elias-ba: Interesting. This typically means "Unprocessable Entity" which a validation error. If the workflow executed but produced a failure state, example: a job intentionally returned failure, this would be semantically incorrect, right ? Could we consider 200 with state: "failed" in metadata, or 400 if it's more like a client error ?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -183,6 +189,52 @@ defmodule LightningWeb.RunChannel do
...
+      :crashed -> 500
+      :exception -> 500
+      :killed -> 410
```
elias-ba: 410 means "Gone" (resource permanently deleted). This seems odd for a killed workflow. Could we consider 503 (Service Unavailable) or 500 ?

lib/lightning_web/channels/run_channel.ex
```diff
@@ -183,6 +189,52 @@ defmodule LightningWeb.RunChannel do
...
+          work_order_id: work_order.id,
+          run_id: run.id,
+          state: run.state,
+          error_type: run.error_type,
+          inserted_at: run.inserted_at,
+          started_at: run.started_at,
+          claimed_at: run.claimed_at,
+          finished_at: run.finished_at
+        }
+      }
```
elias-ba, line 216: This is nice. Since this defines a public contract, should we consider adding a version field in meta: `%{api_version: "1.0", ...}` ?

taylordowns2000:
note that @stuartc 's vote is a `200` for all run states - don't provide info on run state in the HTTP status

PR #3788, opened 2025-10-20
fix: credential handling in collaborative editor jobs

### LINKED ISSUE #3760: Collaborative editor crashes when saving job without credential
### User story

When I create a job in the collaborative editor and don't select a credential, the workflow fails to save with a validation error.

### Details

The job form sends `undefined` for empty credential fields. The backend expects `nil` or a valid credential ID, so the save fails.

This blocks users from creating jobs that don't need credentials (like simple data transformations with the common adaptor).

The job form also lacks test coverage in both unit and E2E tests. The LiveView workflow editor has tests in `workflow-steps-creation.spec.ts`, but the React collaborative editor doesn't.

### Implementation notes

1. Add unit tests for the JobInspector component
2. Add E2E tests that cover the credential selection scenarios
3. Fix the bug so jobs can be saved without credentials

### Release notes

Jobs in the collaborative editor can now be saved without selecting a credential.

### User acceptance criteria

- [ ] Jobs can be saved without selecting a credential
- [ ] Jobs can be saved with a project credential
- [ ] Jobs can be saved with a keychain credential
- [ ] Credentials can be cleared after being selected
- [ ] Unit tests cover the JobInspector form behavior
- [ ] E2E tests verify the credential scenarios work end-to-end

elias-ba:
## Description

This PR fixes a bug where the collaborative editor crashes when saving a job without selecting a credential. The root cause was that the `addJob()` function left credential fields as `undefined` instead of explicitly setting them to `null`, causing backend validation failures.

Closes #3760

## Validation steps

1. **Create job without credential:**
   - Create a new workflow in the collaborative editor
   - Add a job without selecting any credential
   - Click "Save"
   - Verify workflow saves successfully with success toast
   - Reload page and verify job appears with no credential selected

2. **Test credential selection:**
   - Select a project credential and save
   - Verify it persists after reload
   - Clear the credential and save
   - Verify it's cleared after reload

3. **Test both editors:**
   - Repeat above steps in the LiveView editor
   - Verify consistent behavior

1. **Root Cause:** The bug was in `createWorkflowStore.ts:addJob()` which only set `id`, `name`, and `body` fields. Credential fields were never initialized, leaving them as `undefined` in Y.Doc. When workflows were saved before user interaction, `undefined` was sent to the backend, causing validation failures.

2. **Minimal Fix:** The fix adds just 2 lines to explicitly initialize credentials:
   ```typescript
   jobMap.set("project_credential_id", job.project_credential_id || null);
   jobMap.set("keychain_credential_id", job.keychain_credential_id || null);
   ```

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have implemented and tested all related **authorization policies**. (N/A - no authorization changes)


assets/js/collaborative-editor/stores/createWorkflowStore.ts
```diff
@@ -474,7 +474,12 @@ export const createWorkflowStore = () => {
     // Observer handles the rest: Y.Doc → immer → notify
   };
 
-  const addJob = (job: Partial<Session.Job>) => {
+  const addJob = (
+    job: Partial<Session.Job> & {
+      project_credential_id?: string | null;
+      keychain_credential_id?: string | null;
+    }
+  ) => {
```
stuartc: This isn't right, these attributes should be on Session.Job; i.e. they are missing from the type. Not sure why they were unioned on the arguments.

Also they shouldn't be optional.

assets/js/collaborative-editor/components/inspector/JobInspector.tsx
```diff
@@ -345,6 +345,11 @@ export function JobInspector({ job, renderFooter }: JobInspectorProps) {
                   "keychain_credential_id",
                   resolved.keychain_credential_id
                 );
+                // Manually trigger updateJob to persist credentials to Y.Doc
+                updateJob(job.id, {
+                  project_credential_id: resolved.project_credential_id,
+                  keychain_credential_id: resolved.keychain_credential_id,
+                });
```
stuartc, line 352: Do we really have to do this? I thought that by setting the field value on the form would trigger the onChange at the top? Hmm, guess it makes sense. Just want to be sure.

Also there is a type warning here, Session.Job is missing project_credential_id, weird I'm not getting it for keychain_credential_id as well. But yeah we need to add the properties to the type.

stuartc (approved):
I've addressed my comments 👍

PR #3791, opened 2025-10-21
Workflow Editor Header: Adds read only indicator to collaborative editor

### LINKED ISSUE #3627: Workflow Editor Header: Read-only warning indicator
### User story

As a user, I need a clear indication of when I'm not able to edit a workflow because either (a) I'm viewing a previous snapshot or (b) I don't have sufficient privileges (via my `project_user.role`)

### Details

Display a read-only warning when viewing historical workflow snapshots or viewing as `project_user.role == :viewer` .

Currently implemented at edit.ex:134-150 but _only_ includes the snapshot check, the role "viewer" check is done at edit.ex:171-175.

**Behavior:**

- Shows when conditions are met:
  - Viewing a snapshot (`snapshot_version_tag != "latest"`) ~but not in creation screen (!show_new_workflow_panel)~
  - User does not have edit permissions (`!can_edit_workflow`)
 
- Shows information icon (hero-information-circle-solid, primary-600, 50% opacity)
- Text displays "Read-only"
- Tooltip: Either:
  - "You cannot edit or run an old snapshot of a workflow"
  - "You do not have permission to edit this workflow"
- Uses bottom placement for tooltip

### Implementation notes

The new UI only cares about one Boolean (`isReadOnly`) and the tooltip itself will need more logic to determine which kind of read-only it is. We hold all the state (like the lock versions, permissions etc) in the stores.

- Conditional rendering based on three socket assigns
- Component ID is "edit-disabled-warning"
- Icon is h-4 w-4 size
- Uses `phx-hook="Tooltip"` with `data-placement="bottom"`
- Depends on #3624

### Release notes

N/A - Internal refactor maintaining existing functionality

### User acceptance criteria

- [ ] Warning appears when viewing old snapshots
- [ ] Warning hidden when viewing latest version
- [ ] Warning appears for users without edit permissions
- [ ] Information icon and "Read-only" text display correctly
- [ ] Tooltip explains different restrictions on hover

lmac-1:
## Description

This PR implements a read-only indicator in the collaborative editor header that shows when a workflow can't be edited. 

The indicator appears as a small info icon with "Read-only" text in the header (right after the breadcrumbs). When you hover over it, a tooltip explains why editing is disabled - either because you don't have permission, you're viewing an old snapshot or the the workflow has been deleted. 

I created a `useWorkflowReadOnly` hook that checks these three conditions in priority order (deleted > permissions > snapshot). The hook is used by the new `ReadOnlyWarning` component in the header, and also passed to the workflow diagram and Monaco editors so they know when to disable editing. 

The implementation follows the same pattern as the existing `useCanSave` hook.

Closes #3627

## Validation steps

1. Test permission restrictions: Log in as a viewer on a project (someone with read-only access). Open any workflow in the collaborative editor - you should see the read-only indicator in the header with the tooltip "You do not have permission to edit this workflow"
2. Test snapshot viewing: Add another collaborator to your project as an editor. Open the collaborative editor for both users in the workflow (you will need to log in in a separate incognito tab). Make changes to the workflow and save for user A. User B should now see the read-only indicator "You cannot edit or run an old snapshot of a workflow" (Note: Claude suggested that there might be a more simple way to test this with one single user by adding something like `?v=` to the url but I couldn't get this to work - let me know if you know how)
3. Test new workflow creation: Go to create a new workflow and open collaborative . Even if you're a viewer, the indicator should NOT appear during new work
4. Check for flickering: Refresh the page while viewing an editable workflow, or observe when opening the collaborative editor for an editable workflow. The read-only indicator should not briefly flash on screen during loading.

1. While the disabled state is correctly passed to the `WorkflowDiagram` component and Monaco editor components, users can still click on nodes and type in the editor when read-only. The save button is correctly disabled, so changes can't be persisted, but the UI doesn't prevent the interactions. This is a pre-existing issue with how these components handle the `disabled` prop - they need additional work to fully disable interactions. This should be addressed in a separate ticket.
2. The component uses `edit-disabled-warning` as the Phoenix LiveView template already uses this ID. Both implementations coexist without conflicts.
3. Claude added about 35 tests covering all scenarios - hook logic, component rendering, header integration, and edge cases like null permissions. I'm not too familiar with your testing strategy so not sure if this is too much or too little. Please let me know. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,3 @@
 assets/vendor
+CLAUDE.md
+.claude
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,7 +87,6 @@ iex -S mix phx.server
 ```
 
 **Available stores:**
-
 - WorkflowStore - Workflow data, jobs, triggers, edges
 - SessionContextStore - User, project, config, permissions
 - SessionStore - Connection and sync state
@@ -96,13 +95,11 @@ iex -S mix phx.server
 - CredentialStore - Project and keychain credentials
 
 **Features:**
-
 - View current state of any store
 - See action history with timestamps
 - Export/import state for bug reproduction
 
-**Note:** DevTools is disabled in production builds. For detailed store
-architecture and usage guidelines, see `.claude/guidelines/store-structure.md`.
+**Note:** DevTools is disabled in production builds. For detailed store architecture and usage guidelines, see `.claude/guidelines/store-structure.md`.
 
 ### Docker Development
 
@@ -202,14 +199,11 @@ components:
 - Key dependencies: @xyflow/react (workflow visualization), Yjs (collaborative
   editing), Monaco Editor (code editing)
 - React components integrated into Phoenix app via LiveView mounting
-- **Icons**: Use heroicons via Tailwind classes (e.g.,
-  `className="hero-check-micro h-4 w-4"`). Never create custom SVG icons.
+- **Icons**: Use heroicons via Tailwind classes (e.g., `className="hero-check-micro h-4 w-4"`). Never create custom SVG icons.
 
 #### Toast Notifications
 
-For toast notifications in the collaborative editor, see
-`.claude/guidelines/toast-notifications.md`:
-
+For toast notifications in the collaborative editor, see `.claude/guidelines/toast-notifications.md`:
 - Usage patterns for info, alert, success, and warning toasts
 - Integration with workflow operations
 - Styling conventions matching Lightning's design system
@@ -217,11 +211,8 @@ For toast notifications in the collaborative editor, see
 
 #### Collaborative Editor Store Architecture
 
-When working with the collaborative editor stores (creating, modifying, or
-debugging), see `.claude/guidelines/store-structure.md`:
-
-- Store hierarchy and responsibilities (SessionStore, WorkflowStore,
-  AwarenessStore, etc.)
+When working with the collaborative editor stores (creating, modifying, or debugging), see `.claude/guidelines/store-structure.md`:
+- Store hierarchy and responsibilities (SessionStore, WorkflowStore, AwarenessStore, etc.)
 - Decision tree for "where should this state go?"
 - Store update patterns (Y.Doc, Phoenix Channel, local state)
 - When to create new stores vs extending existing ones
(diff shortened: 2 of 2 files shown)


lmac-1:
@stuartc I merged main into this branch to fix conflicts, but looks like some files have formatting changes from that merge, despite not touching them in this branch. Perhaps I did something wrong.. 

Should I be running something before committing?

lmac-1:
Good point @taylordowns2000, will double check this today and report back. @stuartc hold off on reviewing for now until I have checked this.

lmac-1:
@taylordowns2000 re your comment: It seems that this is the current behaviour by design, although I agree that it's a bit confusing, so it might need some reworking/rethinking in a separate issue. 

Lightning uses optimistic locking with `lock_version` to create immutable workflow snapshots. 

So when User A saves: 
1. A new snapshot is created with `lock_version` = `N + 1`
2. User B is still working on `lock_version` = `N` (the old snapshot)
3. User B's editor detects that they are viewing an "old snapshot" and becomes read only

I think the relevant code is in [useWorkflow.ts:549-552](https://github.com/OpenFn/lightning/blob/main/assets/js/collaborative-editor/hooks/useWorkflow.ts#L549-L552) - it checks if the current workflow's `lock_version` differs from `latestSnapshotLockVersion` received from the server.  The system has two collaboration mechanisms that don't fully integrate: 
- Yjs handles real-time sync between users (works great before anyone saves)
- Snapshot versioning creates immutable checkpoints (causes the disconnect on save)

This means that "collaboration" currently only works until someone saves, then other users get locked out until they refresh. You could auto-reload User B's editor when User A saves, but you'd need to carefully merge any unsaved changes User B has made that differ from User A's saved snapshot. Otherwise User B might lose their work on reload. This feels like a separate discussion. @stuartc any thoughts? Should we open an issue to track this, or will the team discuss in an upcoming meeting?

TLDR; I don't _think_ this affects this PR and should maybe be discussed elsewhere

stuartc:
@lmac-1 excellent summary, I believe this should be fixed in main as of yesterday. I was reviewing @taylordowns2000 changes yesterday (the ones that likely fixed this) and found some other smells around syncing the lock_version; again the scenario you described _should_ be fixed; could you verify this after I merge this?

I'm tracking the smells over [here](https://github.com/OpenFn/lightning/issues/3863), please let me know if you run into "This means that "collaboration" currently only works until someone saves" or anything similar.

PR #3826, opened 2025-10-22
Manual run from new canvas

taylordowns2000:
@stuartc , this mostly works but the existing dataclip search doesn't appear to be right. can you review the structure and then either (a) fix/implement the dataclip search so it matches the functionality of the liveview dataclip search or (b) give me a thumbs up to continue barking up the same tree?

### DIFF
--- a/assets/js/collaborative-editor/components/ManualRunPanel.tsx
+++ b/assets/js/collaborative-editor/components/ManualRunPanel.tsx
@@ -272,33 +272,13 @@ export function ManualRunPanel({
     (selectedTab === "existing" && selectedDataclip) ||
     selectedTab === "custom";
 
-  // Use HotkeysContext to manage runPanel scope
-  const { enableScope, disableScope, activeScopes } = useHotkeysContext();
-
-  logger.debug({ activeScopes });
-
-  // Enable runPanel scope when this component mounts
-  // Parent (WorkflowEditor) manages the "panel" scope to prevent conflicts
-  useEffect(() => {
-    enableScope("runPanel");
-
-    return () => {
-      disableScope("runPanel");
-    };
-  }, [enableScope, disableScope]);
-
   // Handle Escape key to close the run panel
   useHotkeys(
     "escape",
     () => {
-      logger.debug("Triggered Esc");
       onClose();
     },
-    {
-      enabled: true,
-      scopes: ["runPanel"],
-      enableOnFormTags: true,
-    },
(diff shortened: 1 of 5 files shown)


stuartc:
I've just pushed a bunch of changes.

Cleanup and refactoring:

- Removed unused useOutsideClick hook
- Refactored ManualRunPanel to use the InspectorFooter pattern for consistency
- Replaced console.log calls with proper logger namespaces in FullScreenIDE and workflow hooks

Architectural changes: Introduced UIStore to manage run panel state

- No props drilling between Header, WorkflowEditor, and ManualRunPanel
- Removed the InnerEditor wrapper component (it only existed to lift state, UIStore takes care of this now)
- Fixed hotkey scope management so Escape works correctly when switching between inspector and run panel

Some notes (non-exhaustive):

1. I'm not sure if it existed before my changes, but if you open the manual run panel from a trigger/job and hit escape it also dismisses the selected node as well. While we have separate hotkey scopes for this, something is causing the inspector to be closed as well. I would assume it would be layered as the enable/disableScope functions are stacked. I didn't get to the bottom of it, but we do need to pay attention to this, whether we are ok with this behaviour or not, we need to be aware of why dismissing the run panel also causes the job/trigger to be deselected as well.
2. No proper error ahndling on the dataclipApi or it's consumers. We should address this and potentially building an API client wrapper to separate calling a specific API and our API's common specifics like headers, csrf etc.
3. Need to make sure we reflect in an issue that comment on Header.tsx saying that can_run_workflow isn't available, that would need to be retrieved when we fetch the session context (i.e. session + current user specific data).
4. Interesting (intentional) side effect of the ManualRunPanel and the useEffect around line 98, if users click different nodes it throws out their search state, not sure that was the original behaviour. 

_Edit:_ Ok so for 4, when you change nodes it doesn't update the run panel - so that useEffect doesn't really do anything. Since we don't change the runContext when clicking different nodes those props never change; so now it feels unnecessary or defensive. But it's kind odd to be able to click different nodes and not actually see any change other than on the diagram.

stuartc (approved):

taylordowns2000:
@stuartc , so we don't lose this in Slack:
1. need to restore the run panel context updating - if you click another node, your runpanel must still udpate to let you run from that step/trigger
2. need to restore the disableScope behaviour - ok if we don't do it that way, but must make sure that the behaviour is preserved: an `Esc` press on ManualRunPanel shouldn't _also_ close JobInspector/TriggerInspector panels

PR #3872, opened 2025-10-27
fix formatter

taylordowns2000:
formatting only.

@stuartc , i'm going to wait until we're at an ebb in open PRs related to the re-write to merge this.

### DIFF
--- a/assets/dev-server.js
+++ b/assets/dev-server.js
@@ -1,4 +1,5 @@
 import path from 'node:path';
+
 import esbuild from 'esbuild';
 import { copy } from 'esbuild-plugin-copy';
 import postcss from 'esbuild-postcss';
--- a/assets/js/adaptor-docs/Docs.tsx
+++ b/assets/js/adaptor-docs/Docs.tsx
@@ -1,9 +1,10 @@
 import { useCallback } from 'react';
+
 import DocsPanel from './components/DocsPanel';
 
 type DocsProps = {
   adaptor: string; // name of the adaptor to load. aka specfier.
-}
+};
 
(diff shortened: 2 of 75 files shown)


bin/format
```diff
@@ -26,7 +26,7 @@ npm exec prettier -- \
   --config "$PROJECT_ROOT/.prettierrc" \
   --ignore-path "$PROJECT_ROOT/.prettierignore" \
   --write \
-  js/collaborative-editor test | grep -v "unchanged" || {
+  "**/*.{js,jsx,ts,tsx}" | grep -v "unchanged" || {
```
stuartc: I think this should just be `.` or `js test`. Prettier has it's own config about what files to handle, so we could even run the from the root.

config/dev.exs
```diff
@@ -145,7 +145,7 @@ config :git_hooks,
         {:cmd,
          "FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -E '\\.(js|ts|tsx|jsx|md)$' || true); " <>
            "if [ -n \"$FILES\" ]; then " <>
-           "npx --prefix assets prettier --write $FILES && git add $FILES; " <>
+           "npx --prefix assets prettier --config .prettierrc --ignore-path .prettierignore --write $FILES && git add $FILES; " <>
```
stuartc, line 156: Kinda surprised we had to do this, did that fix the changelog issue for you? I'm surprised because those config files are defaults for prettier _and_ we're cd'd into the same folder as them. 🤷 But maybe it's the `--prefix` thats changing the cwd.

Issue #3940, opened 2025-11-07
Bugs with history canvas viewing


taylordowns2000:
1. Clicking "view all history for this workflow" button yields a 500

<img width="224" alt="Image" src="https://github.com/user-attachments/assets/ec1a2899-cc32-4b10-82ec-5dd7aa37e0f5" />

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/fd7ee6e8-0d8c-4961-8a25-bf2c38b517ed" />

2. Monaco editor for custom input not displayed properly

<img width="1087" height="949" alt="Image" src="https://github.com/user-attachments/assets/f98d75e0-4660-444f-a325-0eb12f642e0e" />

lmac-1:
A few questions to understand better:

1. @taylordowns2000 which button is "View all history" button? When I click on this uuid for the workflows I have I don't have a 500 for the workflows I have:

<img width="3022" height="1340" alt="Image" src="https://github.com/user-attachments/assets/49e6e428-0111-4b98-bab7-3b68aed63bbe" />

Can you give me more details? or steps to replicate?


2. What should it look like? Do you mean it should look like this with the code editor (pasted from LiveView). Is anything else broken?

<img width="2968" height="1550" alt="Image" src="https://github.com/user-attachments/assets/617d3394-a0b3-4265-9ea8-2d5293ea5ab9" />

lmac-1:
Ah I got it for 1 - it's the little icon! On it!

lmac-1:
Fixed in #3944 and #3941

PR #3941, opened 2025-11-07
Fix History Panel bugs and add version mismatch warning in collaborative editor

### LINKED ISSUE #3814: Canvas: View a Run on the Canvas
1. Entrypoints

- [ ] Via url (must be able to refresh the page and have ?run={id} in url)
- [ ] Via MiniHistory.tsx (see above - 8c?)

2. The visuals look exactly like they do now for a completed run

Image

4. New: A pending run (updates automatically - subscribed to changes in the run channel)
5. New: An in-progress run (updates automatically - subscribe to changes in the run channel)
6. If an editor is viewing a completed run for an old snapshot, but switches to “latest” so they can make edits, display a warning banner saying: “Note that the steps in this latest version of the workflow may be different from those used when this run was executed.”
7. When an editor opens the IDE for a step, they do not lose the run that is loaded (i.e., they are taken to the IDE for that step/run/input exactly as it works now on production.)

lmac-1:
## Description

This PR improves the history panel experience in the collaborative editor by fixing critical bugs that prevented navigation to the "view all history" page and adding a version mismatch warning to prevent user confusion.

Closes #3814 and point 1 of #3940

### History page 500 error when navigating from collaborative editor
image

- **Frontend**: Fix workflow ID extraction in MiniHistory - was taking "collaborate" instead of the actual UUID from /w/:id/collaborate URLs
- **Backend**: Handle nil case in humanize_workflow when workflow not found
- **Impact**: "View History" button now works from collaborative editor

### Feature: Version mismatch warning banner

Shows a warning when viewing latest workflow but selected run was executed on an older version. Prevents confusion when workflow structure has changed since the run executed. Appears when: 
- Run selected AND viewing "latest" AND versions differ

Example message:

image

⚠️ This run was executed on version 66, but you're viewing version 67. Steps shown may not match the current workflow structure.

### Testing
- All existing tests passing
- 8 new unit tests for version mismatch detection hook

## Validation steps

**Version mismatch warning:**
You have two choices 
1. Pick an existing workflow where the structure has changed over time in the last 20 runs
2. Make an update to an existing workflow and click Save

Once you have a workflow where the latest version is different to previous version: 
1. Select an old run while viewing "latest" -> Banner appears
2. Switch to matching snapshot version -> Banner disappears
3. Deselect run -> Banner disappears

**Bug fix:**
1. Open workflow in collaborative editor -> Click the little icon next to "View full history" (marked in screenshot above)
2. Verify history page loads AND that it has the correct content (it previously crashed)

I tried to make the banner look similar to existing banner in LiveView using Tailwind colours.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/collaborative-editor/components/diagram/MiniHistory.tsx
+++ b/assets/js/collaborative-editor/components/diagram/MiniHistory.tsx
@@ -29,7 +29,14 @@ import React, { useState } from 'react';
 import { relativeLocale } from '../../../hooks';
 import { duration } from '../../../utils/duration';
 import truncateUid from '../../../utils/truncateUID';
+import { useProject } from '../../hooks/useSessionContext';
+import { useWorkflowState } from '../../hooks/useWorkflow';
 import type { Run, WorkOrder } from '../../types/history';
+import {
+  navigateToRun,
+  navigateToWorkOrderHistory,
+  navigateToWorkflowHistory,
+} from '../../utils/navigation';
 
 // Extended types with selection state for UI
 type RunWithSelection = Run & { selected?: boolean };
@@ -102,6 +109,10 @@ export default function MiniHistory({
   const [expandedWorder, setExpandedWorder] = useState('');
   const now = new Date();
 
+  // Get project and workflow IDs from state for navigation
+  const project = useProject();
+  const workflow = useWorkflowState(state => state.workflow);
+
   // Clear expanded work order when panel collapses
   React.useEffect(() => {
     if (collapsed) {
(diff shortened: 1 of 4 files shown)


lmac-1:
This one is ready for review @stuartc @taylordowns2000 @theroinaochieng ☺️. Probably one for Monday now.

lmac-1:
@stuartc I decided to put the version mismatch stuff in a hook to not overcomplicate components. There's also a test to test the logic of the hook.

elias-ba (approved):
Hey @lmac-1 this is looking great to me from a code point of view. I alkso clicked tested and it looks great. I will leave the UI approvals to Taylor. @taylordowns2000 just wanna flag that the nodes collapsing issue might be out of scope of this work. It has became rare now but it's still happening from time to time in the app on main. I think we may need to have a proper issue for it and do deep diagnostic for it. I do think that it's deeply related to nodes positionning and layout algorithm.

lmac-1:
Hey @taylordowns2000 thanks for the Loom, yes that makes sense, will take a look! I didn't realise that the version should update at the top.

And for the nodes collapsing, I have also been seeing it a bit in other bits of dev work that I'm working on in other branches / main, but not sure what are the exact steps to replicate. Seems to be a bit random. As @elias-ba suggested, it might be worth diving deeper into that in a separate issue. Will try to keep an eye out for it though to see what is causing it.

lmac-1:
I think I've fixed the node position bug... but would appreciate some help in understanding how 😅

I was able to consistently reproduce the collapsing nodes issue across multiple branches using these steps:
1. Select a run from the history panel with a "Failed" status.
2. Refresh the page.
3. The nodes will appear collapsed at position (0,0).
4. If you then select a different run and return to the failed run, the nodes display correctly.

I looked into this with Claude, and it suggested the fix I implemented here in this branch in [adc80d9](https://github.com/OpenFn/lightning/pull/3941/commits/adc80d91c6ce039bfa37b8e693689679c3b993ba). The fix works, but I don't fully understand why: The change suppresses the "could not auto-calculate position" warnings when the positions cache is empty on initial render (which is expected before layout runs). But removing these `console.warn()` calls also fixes the visual issue where nodes appear collapsed - not just the console spam. My theory is that the console operations during the render cycle were causing some kind of timing/performance issue that prevented the layout system from completing before React painted the nodes. But I'm not confident in this explanation. @elias-ba - do you have insight into why removing these warnings would fix the visual positioning issue? Is there something about how ReactFlow or the layout system interacts with synchronous operations during render that I'm missing? 

@taylordowns2000 can you retest where you saw the error? It should happen when you have a run selected with a non-successful status and then refresh the page.

lmac-1:
Okay @taylordowns2000, I think I have fixed everything you mentioned:
1. Clicking on runs now automatically updates the version in the header
2. If you then manually select "Latest", the warning appears as expected

I also improved the warning logic so it appears whenever you're viewing a different version than the run's version, not just when viewing "latest". For example:
1. You select a run that was executed on version 4
2. The header automatically switches to v4 (no warning - versions match)
3. You manually change to "Latest" (version 5) → Warning appears: "This run was executed on version 4, but you're viewing version 5..."
4.You manually change to version 3 → Warning still appears: "This run was executed on version 4, but you're viewing version 3..."

This gives you flexibility to compare runs against any workflow version while staying aware of mismatches.

PR #3951, opened 2025-11-10
fix: flickering of active collaborator icons between states

### LINKED ISSUE #3931: Inactive Collaborator Avator sdissaperas on teh Editor insead of greying out
**Description**  
The gray border indicator for inactive collaborators (inactive for more than 2 minutes) is not working as expected.  
Instead of switching to a gray border, the avatar either disappears entirely or remains displayed as active on the navigation bar.

Image

**Expected Behavior**  
- Avatars for inactive collaborators (no activity for > 2 minutes) should remain visible with a **gray border** to indicate inactivity.  
- Avatars should only disappear when the collaborator disconnects (e.g., closes the browser tab).

**Actual Behavior**  
- After 2–4 minutes of inactivity, the avatar disappears even though the user’s tab remains open and connected.  
- Sometimes, the avatar remains active with a green border beyond the 2-minute inactivity window.

**Steps to Reproduce**  
1. Open the same workflow in two different browser sessions (as different users)/ Have anotehr.  
2. Keep one session inactive for more than 2 minutes (without mouse or keyboard input).  
3. Observe whether the avatar in the other session changes to gray or disappears.

doc-han:
## Description

Implements active, inactive and unavailable states for active collaborators on a workflow. 

**active** - a user is active when a workflow is open and the user is actively on the tab.
**inactive** - a user is inactive after 12s if they leave the tab where the workflow is open.
**unavailable** - a user is removed from active collaborators after approx. 1min of closing the tab where the workflow is open. 

### Main issue
**[a] Throttling of browser timers is our biggest villain in this PR.** 
What problems does it cause? 
- Users are supposed to keep updating their awareness in ydoc by updating their state else they get removed from awareness. there's a 30secs window for this by ydoc else user is marked offline
- We currently perform this update every 10secs to have 3 chances of updating the ydoc state.
- Browser throttling reduced the 3 chances to be unpredictable(mostly 0 when the user leaves the tab).
- Hence, when a user leaves the tab. there's a high chance they'll be removed from ydoc awareness because throttling would've increased our 10secs interval to 30+secs. 
Hence, when a user leaves a tab, they will be marked as offline when throttling kicks in, and when after being throttled, their event finally goes through, they then get marked back as online. making them flicker

**[b] Another thing to be aware of 10secs interval**
Every 10secs we actually update the user's `lastSeen` to `Date.now()` but then when the user leaves the tab, we don't update this `lastSeen` value. And because there's no state change/diff happening when the user leaves the tab. No event is actually triggered to keep ydoc live.

### Approach in this PR
1. When user leaves the tab. for every 10secs, we increment lastSeen by 1ms which will trigger an event to keep ydoc live. This is to battle [b]
2. We have a small cache for activeCollaborators that lives for 60secs. when a user shows up as an active-collaborator we keep him for 60secs max. this 60secs is renewed when they show up again in the new set of active collaborators we receive. This resolves [a]

1. and 2. work hand in hand to smoothen the active collaborators regardless of the ydoc and browser throttling limits. 

### Side Effect
When a user closes a tab or a workflow. They disappear from other users active-collaborators after 60secs. but users would know inactive collaborators real quick(12secs)

Closes #3931 

## Validation steps

Note: Some values got changed. 
eg. 
1. inactivity time changed from 2min to 12secs.
3. it also takes approx. 1min(60secs) to remove a user from the active collaborators when they actually leave.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/collaborative-editor/components/ActiveCollaborators.tsx
+++ b/assets/js/collaborative-editor/components/ActiveCollaborators.tsx
@@ -1,5 +1,5 @@
 import { cn } from '../../utils/cn';
-import { useRemoteUsers } from '../hooks/useAwareness';
+import { useAwareness } from '../hooks/useAwareness';
 import { getAvatarInitials } from '../utils/avatar';
 
 import { Tooltip } from './Tooltip';
@@ -15,7 +15,7 @@ interface ActiveCollaboratorsProps {
 }
 
 export function ActiveCollaborators({ className }: ActiveCollaboratorsProps) {
-  const remoteUsers = useRemoteUsers();
+  const remoteUsers = useAwareness({ cached: true });
 
   if (remoteUsers.length === 0) {
     return null;
--- a/assets/js/collaborative-editor/components/CollaborationWidget.tsx
+++ b/assets/js/collaborative-editor/components/CollaborationWidget.tsx
@@ -5,14 +5,15 @@
 
 import { useSocket } from '../../react/contexts/SocketProvider';
 import { cn } from '../../utils/cn';
-import { useAwarenessUsers } from '../hooks/useAwareness';
+import { useAwareness } from '../hooks/useAwareness';
 import { useSession } from '../hooks/useSession';
 
 export function CollaborationWidget() {
   const { isConnected: socketConnected, connectionError } = useSocket();
   const { isConnected: yjsConnected, isSynced } = useSession();
 
-  const users = useAwarenessUsers();
+  // Get remote users only (local user is always excluded)
+  const remoteUsers = useAwareness({ cached: true });
 
   const getStatusColor = () => {
     if (socketConnected && yjsConnected && isSynced) return 'bg-green-500';
@@ -40,19 +41,20 @@ export function CollaborationWidget() {
         </div>
 
         {/* Separator */}
-        {users.length > 0 && <div className="w-px h-3 bg-gray-300" />}
+        {remoteUsers.length > 0 && <div className="w-px h-3 bg-gray-300" />}
 
         {/* Online users */}
-        {users.length > 0 && (
+        {remoteUsers.length > 0 && (
           <div className="flex items-center gap-1">
             <span className="text-gray-500">
-              {users.length} user{users.length !== 1 ? 's' : ''}:
+              You + {remoteUsers.length} other
+              {remoteUsers.length !== 1 ? 's' : ''}:
             </span>
             <div className="flex gap-1">
-              {users.slice(0, 3).map(user => (
+              {remoteUsers.slice(0, 3).map(user => (
                 <div
                   key={user.clientId}
-                  className="flex items-center gap-1 px-2 py-0.5 
+                  className="flex items-center gap-1 px-2 py-0.5
                              bg-gray-50 rounded-full"
                   title={`${user.user.name} (Client ${user.clientId})`}
                 >
@@ -65,9 +67,9 @@ export function CollaborationWidget() {
                   </span>
                 </div>
               ))}
-              {users.length > 3 && (
+              {remoteUsers.length > 3 && (
                 <span className="text-gray-400 px-1">
-                  +{users.length - 3} more
+                  +{remoteUsers.length - 3} more
                 </span>
               )}
             </div>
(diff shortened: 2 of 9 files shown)


doc-han:
suggestion from @stuartc, "use phoenix channel callback instead of timeouts to keep awareness active".

doc-han:
**update**
1. caching awareness users for 1min to make user availability smooth to others.
2. reducing online idle time from 2min to 12sec. hence, if you leave the tab, you'll be set to inactive after 12secs instead of the initial 2mins.
3. when users are active on the tab, they update their lastSeen every 30secs with their current time.
4. when users leave the tab, we update their lastSeen with their previous lastSeen + 1ms. approx. nothing. just to update the store to keep awareness active.

doc-han:
@stuartc I think I understand clearly what you mean.
So if I'm getting it right. we have to do. 
1. `lastSeen` - for keeping ydoc awareness alive
2. `lastState` - `active`|`away`|`idle` to show the state of the user. based on window visibility API

- with the separation of these. we stand the chance to be able to show user state in realtime. hence, 2. does a great job
- with 1. lastSeen will then be updated at our normal 10s. (which might be throttled, causing a flicker). 

don't we still need the caching to smoothen out when `lastSeen` gets throttled? In this case our `lastState` has been decoupled from it and will be realtime. 

thinking...
to make it better, we only cache when `lastState` moves to `away`|`idle`. once out of those states, we clear from cache. I think that will smoothen it out without any effects.

stuartc:
> @stuartc I think I understand clearly what you mean. So if I'm getting it right. we have to do.
> 
> 1. `lastSeen` - for keeping ydoc awareness alive
> 2. `lastState` - `active`|`away`|`idle` to show the state of the user. based on window visibility API
> 
> * with the separation of these. we stand the chance to be able to show user state in realtime. hence, 2. does a great job
> * with 1. lastSeen will then be updated at our normal 10s. (which might be throttled, causing a flicker).
> 
> don't we still need the caching to smoothen out when `lastSeen` gets throttled? In this case our `lastState` has been decoupled from it and will be realtime.

> thinking... to make it better, we only cache when `lastState` moves to `away`|`idle`. once out of those states, we clear from cache. I think that will smoothen it out without any effects.

Yeah I agree, we still need the smoothing. But we can use logic like `lastActivity > lastSeen` i.e. we can derive that they aren't engaged but the window is still open from that.
We can use it to unify multiple tabs for the same user, and pick the one with newest lastActivity. 

I think `lastState` (like an enum) shouldn't be set in the awareness, since we might lose the ability to change it in throttling situations. I see lastState as derived. But having `lastMouseMove`, `lastKeyDown`, `lastUnfocus` (when visibility changes); would all be timestamps that would be updated with a debounce. _I'm not saying we want ALL off those right now_, but I guess my point is, by having different values for each signal we can derive the exact state we want to show other users, down to the second if we wanted and the smoothing comes just before the presentation layer and the updating of the awareness object happens whenever we need it to without any coupling to how we show the state of a user.

doc-han:
@stuartc I've opened a new PR around this. separates the log for keeping ydoc alive and the one for updating the last activity. Works very fine so far. with smoothing also. 

here #4053 
currently handles two activity states. `active`|`away`.

doc-han:
For the consolidation of multiple tabs open by a user. I already have that in place but I've changed how the priority is picked in the other PR. `lastActivity` > `lastSeen` just as you mentioned.

doc-han:
with this said. on #4053 
1. user last activity is tracked realtime `active` | `away`.
2. we keep our 10s `lastSeen` updates to ydoc.
3. to battle possible throttling. when a users activity state changes to `away`, we cache that user for 60s, we only make use of this cache when we can't find the user in our awareness while cache TTL isn't elapsed.

This provides a better user experience than before. 

thinking...
@stuartc does it make sense to put in a measure before activating this caching mechanism? eg. checking the time difference between our 10s ydoc keep-alive trigger. if we start noticing near 30s then we activate the caching mechanism.

What do you think about this and the whole caching stuff in general.

doc-han:
Closing this PR soon in favour of #4053

PR #3956, opened 2025-11-10
fix: use provisioner to validate inputs of merge_projects mix task

elias-ba:
## Description

This PR fixes a security vulnerability in the `mix lightning.merge_projects` task by using Provisioner validation instead of direct atom key conversion.

This PR also updates the `CLAUDE.md` prompt file to add a rule that forces Claude to not include _Coauthored by_ lines as it usually does when it commits. This is quite subjectives, I hate seeing those Coauthored with Claude lines in every single commit and Claude always commits too much for what can be just one line commit messages. cc @rorymckinley @stuartc @midigofrank 

## Validation steps

1. Run Mix task tests:
   ```bash
   mix test test/mix/tasks/merge_projects_test.exs
   ```

2. Run MergeProjects module tests:
   ```bash
   mix test test/lightning/projects/merge_projects_test.exs
   ```

3. Verify security tests catch malicious input:
   ```bash
   mix test test/mix/tasks/merge_projects_test.exs:463
   ```

**Security Issue**: The Mix task previously used `Jason.decode(content, keys: :atoms)` which could create unlimited atoms from malicious JSON input. Thanks @rorymckinley for the catch.

**Solution**: Uses `Provisioner.parse_document()` which leverages Ecto's `cast()` to safely convert string keys to atoms only for known schema fields, while enforcing all validation rules. This allows to no modify the `Lightning.Projects.MergeProjects` module at all. Thanks to @stuartc and @midigofrank for stopping me from going that route. As they said, the MergeProjects module is expecting Project structs, we should give it Project structs not modify it to adapt to the mix task's constraints. This is where using the Provisionner is important and helpful. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have performed a self-review of my code.
- [x] I have implemented and tested all related authorization policies. (N/A - no auth changes)
- [ ] I have updated the changelog.
```

### DIFF
--- a/lib/mix/tasks/merge_projects.ex
+++ b/lib/mix/tasks/merge_projects.ex
@@ -89,29 +89,44 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
     write_output(output, output_path)
   end
 
-  defp perform_merge(source_project, target_project) do
+  defp perform_merge(source_data, target_data) do
+    # Convert string-keyed maps to atom-keyed maps using only existing atoms
+    # This prevents atom exhaustion attacks while allowing the merge algorithm
+    # to work with its expected data structure
+    source_project = atomize_keys(source_data)
+    target_project = atomize_keys(target_data)
+
     MergeProjects.merge_project(source_project, target_project)
   rescue
-    e in KeyError ->
+    ArgumentError ->
       Mix.raise("""
-      Failed to merge projects - missing required field: #{inspect(e.key)}
+      Failed to merge projects - encountered unknown field in JSON
 
-      #{Exception.message(e)}
-
-      This may indicate incompatible or corrupted project state files.
-      Please verify both files are valid Lightning project exports.
-      """)
-
-    e ->
-      Mix.raise("""
-      Failed to merge projects
-
-      #{Exception.message(e)}
-
-      This may indicate incompatible project structures or corrupted data.
-      Please verify both files are valid Lightning project exports.
+      This may indicate the JSON contains invalid or unexpected fields.
+      Please ensure both files are valid Lightning project exports.
       """)
   end
+
+  defp atomize_keys(data) when is_map(data) do
+    Map.new(data, fn {key, value} ->
+      atom_key =
+        if is_binary(key) do
+          # Only convert to atoms that already exist
+          # This prevents creating new atoms from malicious input
+          String.to_existing_atom(key)
+        else
+          key
+        end
+
+      {atom_key, atomize_keys(value)}
+    end)
+  end
+
+  defp atomize_keys(data) when is_list(data) do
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: CLAUDE.md [hunk 1]
```diff
@@ -279,6 +279,14 @@ When working with the collaborative editor stores (creating, modifying, or debug
 - Follow security best practices (authentication, authorization, input
   validation)
 
+### Git Commit Guidelines
+
+- **NEVER** add Claude Code attribution or Co-Authored-By: Claude lines to commit messages
```
rorymckinley: @elias-ba I am curious as to why we would not want to be transparent about this? Or is the assumption that all code will have Claude's fingerprints over it given enough time ? :)

lib/lightning/utils/maps.ex
```diff
@@ -13,4 +26,44 @@ defmodule Lightning.Utils.Maps do
...
+      iex> deep_stringify_keys(%{user: %{name: "John", roles: [:admin]}})
+      %{"user" => %{"name" => "John", "roles" => [:admin]}}
+
+      iex> deep_stringify_keys(%{items: [%{id: 1}, %{id: 2}]})
+      %{"items" => [%{"id" => 1}, %{"id" => 2}]}
+
+      iex> deep_stringify_keys(%User{name: "John", email: "john@example.com"})
+      %{"name" => "John", "email" => "john@example.com"}
+  """
+  def deep_stringify_keys(%_{} = struct) do
```
rorymckinley, line 51: @elias-ba It feels like we could write some nice clean tests to cover this functionality?

rorymckinley (approved):
Nice work, Jambaar!

A couple of small comments/questions. It would be nice if we could get some unit-level tests of the new stringify method.

lib/lightning/utils/maps.ex
elias-ba: I removed this func. It's no longer needed. Changed my approach

CLAUDE.md
elias-ba: Well yeah and also we have our AI disclaimer in PRs. And also it's super annoying to always have those Coauthored lines in every single commit. But this is subjective hey ! what do you think ?

elias-ba: I am wondering what @josephjclark would think about this !?

josephjclark: We should absolutely have claude attribution in a claude co-authored commit

elias-ba:
Closing this PR in favor of #3973. This current PR changed a lot and became too noisy. This fix should be straightforward and minimal. Also I had a conversation with Joe [here](https://openfn.slack.com/archives/C05KZNPEJFN/p1762960810097039) that led to a simpler fix.

PR #3965, opened 2025-11-11
In generator project mix task, pass a UUID map

### LINKED ISSUE #3964: Sandboxes: when merging, how can we make UUID generation predictable?
When writing tests against sandbox merging, sometime a merge creates a node or edge with a new UUID.

It can be hard to write tests against this because I can't predict the UUID. I can't write a test which says: "merge node X into Workflow W and ensure that X has THIS UUID".

This must be a problem inside Lightning tests too?

This affects me most in the CLI mix task.

The best way I can think to solve this is to pass in a preordained list of new UUIDs against node **names**. Before creating a new UUID, Lightning should always check against this list and use the mapped values. I use a similar pattern in CLI testing and it works well

I think CLI has the same problem btw

josephjclark:
I am writing unit tests against the `mix generate_workflow` command. I am taking two projects, merging them, and comparing the results.

I've hit a fun problem on new nodes created in the source sandbox (ie, staging). Lightning generates new UUIDs for them in the target project. But I can't predict those UUIDs, so it's very hard to write compelling tests.

This PR allows me to explicitly pass a UUID mapping into the command. So if there's a new node called `abc` in the source, I want to assign it a UUID of `xyz` in the target:
```
mix lightning.generate_workflow staging.json main.json --uuid abc:xyz
```

Note that in tests I very rarely use actually UUIDs, so the command allows me to pass in fake UUIDs as strings or numbers.

I have deliberately not added a node to the changelog. I'm not sure it's relevant?

Closes #3964

## AI Usage

Claude generated all the code over a series of commits, with some refactoring and design from myself.

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/lib/mix/tasks/merge_projects.ex
+++ b/lib/mix/tasks/merge_projects.ex
@@ -129,18 +129,10 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
     |> Enum.reduce(%{}, fn mapping_str, acc ->
       case String.split(mapping_str, ":") do
         [source_id, target_id] ->
-          source_id = String.trim(source_id)
-          target_id = String.trim(target_id)
+          source_id = cast_int_or_string(source_id)
+          target_id = cast_int_or_string(target_id)
 
-          # Support both string and integer keys by adding both forms
-          acc
-          |> Map.put(source_id, target_id)
-          |> then(fn map ->
-            case Integer.parse(source_id) do
-              {int_id, ""} -> Map.put(map, int_id, target_id)
-              _ -> map
-            end
-          end)
+          Map.put(acc, source_id, target_id)
 
         _other ->
           Mix.raise("""
@@ -151,6 +143,13 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
           """)
       end
     end)
+  end
+
+  defp cast_int_or_string(value) do
+    case Integer.parse(value) do
+      {int, ""} -> int
+      _ -> String.trim(value)
+    end
   end
 
   defp encode_json(project) do


josephjclark:
Thanks @midigofrank. What are you envisioning by "make it priority"? What sort  of changes would you imagine?

I think this is only needed when new nodes and edges are created in the source. Which is a common use-case but also not totally core.

If we want to broaden this up, we could explore a flag that enables seeded "uuids" which are just incrementing numbers. I use this approach all the time in the CLI, and it a least means the ids are predictable.

But tbh this approach here works great for me today and I wouldn't want to spend a lot of time anything more substantial. Not yet anyway :)

midigofrank:
Aaah I see. I meant having it as the accumulator during the first step of mapping in the merge algorithm. Think of it as them being already mapped nodes

josephjclark:
Maybe. I think I'd rather just merge it as-is. Given that it's a test util I think I prefer the idea that the implementation is out on the fringes. I also don't want to take a lot of resources on it, given that it works great from my point of view

lib/mix/tasks/merge_projects.ex
```diff
@@ -113,6 +123,36 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
...
+  defp parse_uuid_mappings(opts) do
+    opts
+    |> Keyword.get_values(:uuid)
+    |> Enum.reduce(%{}, fn mapping_str, acc ->
+      case String.split(mapping_str, ":") do
+        [source_id, target_id] ->
+          source_id = String.trim(source_id)
+          target_id = String.trim(target_id)
+
+          # Support both string and integer keys by adding both forms
```
midigofrank: @josephjclark by the way why do we need to add both keys in here? And why are we only parsing the `source` and not the `target`?

midigofrank: Hmm, it's actually this part that's failing in the CI. I'll have a look tomorrow

josephjclark:
Thank you so much @midigofrank :pray:

lib/mix/tasks/merge_projects.ex
josephjclark: This is probably AI silliness that has not been properly validated by this silly human

I do need string and integer types, although maybe there's a better way I can handle this (like maybe my test code should be casting numbers to strings)

PR #3973, opened 2025-11-12
fix: prevent atom exhaustion in `merge_projects` mix task

elias-ba:
## Description

This PR fixes a security vulnerability in the `mix lightning.merge_projects` task where malicious JSON input with arbitrary keys could cause atom exhaustion and crash the VM.

The fix uses `String.to_existing_atom/1` to safely convert JSON keys to atoms, only allowing keys that already exist in the system. This prevents creation of unlimited atoms from malicious input while maintaining compatibility with the merge algorithm.

Closes #3956

## Validation steps

1. **Test basic merge functionality:**
   ```bash
   mix test test/mix/tasks/merge_projects_test.exs
   ```
   All tests should pass.

2. **Test with non-UUID IDs (Joe's requirement):**
   - Run merge with projects using simple IDs like "1", "2", "test-source-1"
   - Verify merge works without requiring valid UUIDs

3. **Test security:**
   - Try merging a project with unknown JSON keys
   - Should raise clear error about unknown fields

4. **Test offline operation:**
   - Run mix task without database connection
   - Should work without any database access

1. **Implementation approach:** Uses `atomize_keys/1` function that recursively converts only map keys to atoms (not values). UUIDs and other string values remain as strings.

2. **No Provisioner coupling:** Intentionally does not use `Provisioner.parse_document/1` to avoid coupling to validation rules and usage limiting checks (per Joe's feedback).

3. **Lets merge fail naturally:** If project structure is truly invalid, the merge algorithm itself will fail with appropriate errors rather than blocking upfront.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/mix/tasks/merge_projects_test.exs
+++ b/test/mix/tasks/merge_projects_test.exs
@@ -705,7 +705,7 @@ defmodule Mix.Tasks.Lightning.MergeProjectsTest do
       assert result["name"] == "target-project"
     end
 
-    test "works offline without database access", %{tmp_dir: tmp_dir} do
+    test "runs standalone without database fixtures", %{tmp_dir: tmp_dir} do
       source_state =
         build_project_state(
           id: "offline-source",


elias-ba:
Hey @rorymckinley and @josephjclark 👋

This is a new PR that supersedes #3956. Based on Joe's feedback about not coupling the merge task to Provisioner validation or database requirements, I've implemented a much simpler approach.

### What changed from #3956

The new implementation:
- No Provisioner dependency (avoids coupling to validation rules and usage limits)
- No database access (works completely offline as Joe needed)
- Simple `String.to_existing_atom/1` for security (prevents atom exhaustion)
- Supports non-UUID IDs like "1", "test-source-1" (Joe's requirement for testing)
- Lets the merge algorithm handle validation naturally

### How it works

1. Takes JSON files as input
2. Converts string keys to existing atoms using `String.to_existing_atom/1`
3. Passes atom-keyed maps to `MergeProjects.merge_project/2`
4. Returns the merged result

This approach keeps the Mix task decoupled from Lightning's validation constraints while still preventing the atom exhaustion security vulnerability.

@josephjclark - As discussed on Slack, this should satisfy your testing needs. When you have time, please test this branch thoroughly and let me know if you see any issues. I've added you as a reviewer.

@rorymckinley - When you get a chance, would you be able to help review this and get it merged? I believe this approach is cleaner and aligns better with the Mix task's purpose of being a simple, offline utility.

Thanks both! 🙏

josephjclark (approved):
Not going to pretend to 100% understand what's happening here - but this passes against my test suite, so I'm all for it!

lib/mix/tasks/merge_projects.ex
```diff
@@ -89,30 +89,34 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
...
+    Map.new(data, fn {key, value} ->
+      atom_key = if is_binary(key), do: String.to_existing_atom(key), else: key
+      {atom_key, atomize_keys(value)}
+    end)
+  end
 
-      This may indicate incompatible project structures or corrupted data.
-      Please verify both files are valid Lightning project exports.
-      """)
+  defp atomize_keys(data) when is_list(data) do
```
rorymckinley: @elias-ba If I comment out this method, no tests fail - is it still required?

lib/mix/tasks/merge_projects.ex
```diff
@@ -89,30 +89,34 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
...
 
-    e ->
-      Mix.raise("""
-      Failed to merge projects
-
-      #{Exception.message(e)}
+  defp atomize_keys(data) when is_map(data) do
+    Map.new(data, fn {key, value} ->
+      atom_key = if is_binary(key), do: String.to_existing_atom(key), else: key
+      {atom_key, atomize_keys(value)}
```
rorymckinley: @elias-ba If you are going to be atomizing things that are not keys - maybe this should be called just `atomize`, or `maybe_atomize`?

lib/mix/tasks/merge_projects.ex
```diff
@@ -89,30 +89,34 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
...
 
-  defp perform_merge(source_project, target_project) do
+  defp perform_merge(source_data, target_data) do
+    source_project = atomize_keys(source_data)
+    target_project = atomize_keys(target_data)
+
     MergeProjects.merge_project(source_project, target_project)
   rescue
-    e in KeyError ->
+    ArgumentError ->
```
rorymckinley, line 130: @elias-ba This does not appear to have any test coverage?

lib/mix/tasks/merge_projects.ex
```diff
@@ -89,30 +89,34 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
     write_output(output, output_path)
   end
 
-  defp perform_merge(source_project, target_project) do
+  defp perform_merge(source_data, target_data) do
```
rorymckinley: @elias-ba I can unfortunately not comment on the line I want to as it is not part of the diff, but:

`read_state_file` still contains `Jason.decode(content, keys: :atoms)` and `read_state_file` provides the input to `perform_merge` so it gets executed _before_ we cll `atomize_keys` so are we not still vulnerable to a DoS via JSON parsing?

lib/mix/tasks/merge_projects.ex
```diff
@@ -89,30 +89,34 @@ defmodule Mix.Tasks.Lightning.MergeProjects do
...
+  end
 
-    e ->
-      Mix.raise("""
-      Failed to merge projects
-
-      #{Exception.message(e)}
+  defp atomize_keys(data) when is_map(data) do
+    Map.new(data, fn {key, value} ->
+      atom_key = if is_binary(key), do: String.to_existing_atom(key), else: key
```
rorymckinley, line 141: @elias-ba Are we sure that all the string keys will already be present in memory as items at the time this task runs? I can imagine that it will be quite a frustrating user experience if this is not the case.

test/mix/tasks/merge_projects_test.exs
```diff
@@ -378,9 +400,157 @@ defmodule Mix.Tasks.Lightning.MergeProjectsTest do
...
+      # enumerate workflows (which is a string instead of a list)
+      # This is expected behavior - we let the merge algorithm fail naturally
+      assert_raise Protocol.UndefinedError, fn ->
         Mix.Tasks.Lightning.MergeProjects.run([source_file, target_file])
       end
     end
   end
+
+  describe "run/1 - flexibility for testing (Joe's requirements)" do
+    test "allows non-UUID IDs for testing purposes", %{tmp_dir: tmp_dir} do
```
rorymckinley, line 640: @elias-ba Would passing in non-UUID IDs be a valid option for users other than @josephjclark ?

If not, I would suggest that we perhaps disable the UUID -checking via an ENV variable to give Joe the flexibility he requires, but keep it enabled otherwise, which would allow other users tighter validation?

rorymckinley: @elias-ba This test looks as if it may be a duplicate of the test on line 9?

test/mix/tasks/merge_projects_test.exs
```diff
@@ -378,9 +400,157 @@ defmodule Mix.Tasks.Lightning.MergeProjectsTest do
...
+          Mix.Tasks.Lightning.MergeProjects.run([source_file, target_file])
+        end)
+
+      {:ok, result} = Jason.decode(output)
+
+      assert result["id"] == "2"
+      assert result["name"] == "target-project"
+    end
+
+    test "successfully merges projects with deeply nested structures", %{
```
rorymckinley: @elias-ba This test also looks as if it may be a duplicate of the test on line 9?

test/mix/tasks/merge_projects_test.exs
```diff
@@ -378,9 +400,157 @@ defmodule Mix.Tasks.Lightning.MergeProjectsTest do
...
+
+      workflow = hd(result["workflows"])
+      assert workflow["name"] == "Workflow 1"
+
+      job = hd(workflow["jobs"])
+      assert job["body"] == "console.log('updated')"
+      assert job["name"] == "Job 1"
+    end
+
+    test "works offline without database access", %{tmp_dir: tmp_dir} do
```
rorymckinley: @elias-ba For this test are we simulating 'without database access' by just not inserting any fixtures into the database?

rorymckinley:
@elias-ba Ok, done. If my assessment re: the continued presence of `Jason.decode` with atoms is correct, that is the ony blocker - the rest are nice-to-haves (depending on how you view test coverage :) ).

test/mix/tasks/merge_projects_test.exs
josephjclark: Happy with a strict mode in prod and chill mode in dev! I don't need this stuff to run on the platform

elias-ba:
@rorymckinley Thanks for the thorough review Rory! You're absolutely right about the blocker - that was embarrassing. I could swear I had removed the `keys: :atoms` from Jason.decode, but it must have come back through merge conflicts or something. Either way, I'm really glad you caught it because that was the whole point of this PR. I've now removed the `keys: :atoms` option entirely, so we parse with string keys and then explicitly atomize using `String.to_existing_atom/1`. This actually closes the DoS vulnerability.

For the atom availability concern, great catch. I've added an `ensure_schemas_loaded()` function that calls `Code.ensure_loaded/1` on all the schema modules we need (Project, Workflow, Job, Trigger, Edge). This ensures their field atoms are in memory before we attempt atomization. Used a module attribute `@required_schemas` at the top of the file so it's easy to add new schemas as the export format evolves.

For the function naming, you're right that `atomize_keys` was misleading since it recursively atomizes maps and lists. Renamed it to just `atomize` throughout.

Added test coverage for the ArgumentError rescue block - it creates JSON with a dynamically generated field name that won't exist as an atom, verifying the protection mechanism actually works when unknown fields are present.

For the UUID validation discussion, since this doesn't involve any database queries we're probably running no risk by allowing non-UUID IDs. The flexibility is useful for Joe's testing scenarios without compromising anything.

On test duplication, you were spot on. The "deeply nested structures" test was essentially identical to the basic merge test, just with different ID formats. Removed it and enhanced the remaining test with clearer comments.

I also renamed the "works offline without database access" test to "runs standalone without database fixtures" which better describes what it's actually testing.

Please could you do another round of review and let me know if you see any other issues! Thanks a lot man for your help

taylordowns2000:
yeah @rorymckinley , we were supposed to be done with the quote wars (`"` to `'`) but they keep popping up

Issue #4024, opened 2025-11-19
Saving in collaborative editor should update the version number in VersionDropdown without refreshing


lmac-1:
**Describe the bug** If you save a workflow in the collaborative editor and it updates the version of the workflow, the version number in the version dropdown in the header does not update without refreshing the page.

**Version number** v2.14.15-pre (main)

I have reproduced this locally on main:

- [x] Yes
- [ ] No

**To Reproduce** Steps to reproduce the behavior:

1. Open collaborative editor and make a note of the current version of the workflow by opening the VersionDropdown (e.g. latest = version 90)
2. Make a change to the workflow
3. Click Save
4. Check the VersionDropdown version

**Expected behavior** The version number should have increased by one (i.e. latest is now 91)

**Actual behavior** The version number remains unchanged. If I refresh the page, the version number correctly increases by one.

**Screenshots** 

<img width="1363" height="329" alt="Image" src="https://github.com/user-attachments/assets/c0e21e65-5bf8-40e0-b0ca-94ffe1306329" />

lmac-1:
Assigning to myself as I think this is connected to other bugs I'm looking at

lmac-1:
Closed in #4062

PR #4029, opened 2025-11-19
Fix phantom snapshot creation from virtual trigger field

### LINKED ISSUE #4021: Does a new version get created on every manual run?
Possible bug here. Does a new workflow version get created on every manual run in the new collaborative editor?

**🎯 Goal:** It shouldn't! A new version should only be created if the workflow changed.

**Placeholder**

@lmac-1 and I have just detected something where the version number increases on every run (even if no changes are made) 

**Lucy** [Today at 2:46 PM] I'm wondering if it's related to my PR from before which was touching all of that (clicking "latest" at the top and the warning message) https://github.com/OpenFn/lightning/pull/3941

**Lucy** [1 hour ago](https://openfn.slack.com/archives/C05KZNPEJFN/p1763569188565829?thread_ts=1763559651.370189&cid=C05KZNPEJFN) I just had a little look .. I don't think that the PR I linked is what broke things, but it does draw attention to the underlying bug.

When you run a workflow without making any changes, the version number still increments. Then, when you view that run, the UI shows the “old version” warning because the run’s version no longer matches the version in the header. That’s the unexpected behaviour.

**Stu Corbishley** [44 minutes ago](https://openfn.slack.com/archives/C05KZNPEJFN/p1763569306864729?thread_ts=1763559651.370189&cid=C05KZNPEJFN) The strange thing is, that only the server increments the number. The manual run function/behaviour on the backend is the same as it always was afaik

**Lucy** Yeah and I don't remember seeing version numbers going so high before.. strange

**Lucy** Extra context: The version number doesn't appear to increase when it runs automatically. Seems to only happen when manually kicking off a run from collab editor. For example - if the run runs via cron job, that doesn't increase version number, but if i manually click run from collab editor it does.

**Taylor Downs** Kicking off a run definitely saves and runs the workflow. The question is if the save event is sent, and there are no changes, does it still increment the version?

### LINKED ISSUE #4025: Saving a workflow in collaborative editor without making any changes creates a new version
**Describe the bug** If you open a workflow in collaborative editor and click Save without making any new changes, it updates the version number.

**Version number** v2.14.15-pre

- [x] Yes
- [ ] No

1. Open collaborative editor and make a note of the current version (e.g. latest = version 90)
2. Click Save 
3. Refresh the page to update version dropdown (due to #4024 - UI version number without refreshing the page)

NOTE: If you aren't sure if your workflow has unsaved changes - you can repeat the steps - as the second time you save, you can be sure that you don't have any new changes

**Expected behavior** Latest should not have changed (e.g. should still be latest = version 90)
**Actual behavior** Latest has increased by 1 (e.g. latest = version 91)

lmac-1:
_FYI - this fix was applied by Claude after investigating and debugging. I would like a careful review / opinion from someone with more knowledge to make sure that this doesn't break something else down the line.

**This PR is still in draft as some Elixir tests are breaking but would appreciate some feedback as to whether this is the right approach. Thanks! **
_

## Description

Snapshots were being created on every save, even when no changes were made to the workflow and when kicking off a manual run. This caused version numbers to increment unnecessarily and created confusion about what actually changed.

Root cause: The Trigger schema has a virtual field `has_auth_method` that is computed at query time from the `webhook_auth_methods` association. This field was being serialized to Y.Doc and sent back on save, causing Ecto to detect it as a change even though it's a read-only computed value.

Virtual fields should never be serialized/deserialized as they are always recomputed by the backend from their source data. The frontend manages this field locally for UI display (lock icon on triggers) via `webhook_auth_methods_updated` broadcasts.

Changes:
- Remove `has_auth_method` from trigger serialization (line 199)
- Remove `has_auth_method` from trigger deserialization (line 247)
- Add comprehensive tests verifying:
  * Saves without changes do not create snapshots
  * Saves with changes do create snapshots
  * Round-trip serialization produces no phantom changes

This fixes: 
1. Creating a new snapshot when there are no new changes
2. Creating a new snapshot when you manually kick off a run

2 for the price of one!

Closes #4021 and closes #4025

## Validation steps

1. Open collaborative editor for a workflow
2. Click Save to ensure starting from a clean state (in case there are lingering unsaved changes)
3. Refresh the page to see the updated version number (due to #4024)
4. Make a note of the current version number int he version dropdown (e.g. latest = version 90)
5. Manually kick off a run
6. Return to collaborative editor and check the version number
7. Version should still be the same (e.g. latest = version 90) and manually running shouldn't create a snapshot
8. Click Save without making any changes to the workflow
9. Refresh the page to see the updated version number
10. Check the version number in the dropdown - it should still be the same (e.g. latest = version 90) 

Expected result: version number stays the same after manually kicking off a run and saving the workflow without any new changes. 

I debugged this with Claude and still don't have a big picture of `has_auth_method` and whether this might break something further down the line. Claude says it's fine but I don't 100% trust it. I'd appreciate the input of someone with more knowledge of how everything else is working end-to-end 💗

<details>

<summary>More notes from Claude: </summary>

Why Removing It Doesn't Break Anything
The backend never reads has_auth_method from the frontend. Here's the data flow: Frontend → Backend (Workflow Save):
- ❌ Before fix: Included has_auth_method: true → Backend saw virtual field "change" → Created snapshot
- ✅ After fix: Omits has_auth_method → Backend doesn't see phantom change → No snapshot

Backend → Frontend (Page Load):
- Backend loads trigger from database
- Computes has_auth_method fresh from the webhook_auth_methods association
- Frontend receives this value and can display lock icons correctly

Frontend Internal (Real-time Updates):
- When auth methods change, Phoenix Channel broadcasts webhook_auth_methods_updated
- Frontend updates has_auth_method in Y.Doc directly (line 552 of createWorkflowStore.ts)
- Lock icon updates in real-time
- This local update never needs to go back to the server

</details>

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning/collaboration/session_test.exs
+++ b/test/lightning/collaboration/session_test.exs
@@ -220,7 +220,7 @@ defmodule Lightning.SessionTest do
         assert doc_trigger = find_in_ydoc_array(triggers_array, trigger.id),
                "Trigger #{trigger.id} not found in doc"
 
-        ~w(cron_expression enabled has_auth_method id type)
+        ~w(cron_expression enabled id type)
         |> Enum.each(fn key ->
           doc_value = Yex.Map.fetch!(doc_trigger, key)
 
--- a/test/lightning/collaboration/workflow_serializer_test.exs
+++ b/test/lightning/collaboration/workflow_serializer_test.exs
@@ -179,7 +179,6 @@ defmodule Lightning.Collaboration.WorkflowSerializerTest do
       assert trigger["id"] == workflow_trigger.id
       assert trigger["type"] == "webhook"
       assert trigger["enabled"] == true
-      assert trigger["has_auth_method"] == false
       assert is_nil(trigger["cron_expression"])
     end
 
@@ -759,7 +758,6 @@ defmodule Lightning.Collaboration.WorkflowSerializerTest do
                "id" => trigger.id,
                "type" => "webhook",
                "enabled" => true,
-               "has_auth_method" => trigger.has_auth_method,
                "cron_expression" => nil
              } == extracted_trigger
 
@@ -978,8 +976,7 @@ defmodule Lightning.Collaboration.WorkflowSerializerTest do
                "id" => original_trigger.id,
                "type" => to_string(original_trigger.type),
                "enabled" => original_trigger.enabled,
-               "cron_expression" => original_trigger.cron_expression,
-               "has_auth_method" => original_trigger.has_auth_method
+               "cron_expression" => original_trigger.cron_expression
              } == extracted_trigger
 
       # Positions
@@ -1259,7 +1256,7 @@ defmodule Lightning.Collaboration.WorkflowSerializerTest do
              } == edge
     end
 
-    test "preserves all 5 trigger fields" do
+    test "preserves all trigger fields" do
       workflow =
         build(:workflow)
         |> with_trigger(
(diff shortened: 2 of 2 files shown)


lmac-1:
Sorry @taylordowns2000 @stuartc @elias-ba I have removed the request to review as this is still a draft PR. Would appreciate some feedback/discussion though!

PR #4042, opened 2025-11-20
feat: ai assistant in collaborative editor for workflows and jobs

### LINKED ISSUE #3816: Canvas & IDE: AI Assistant Chat
As a user, I would like to be able to access the AI Assistant on the canvas and the IDE.

https://www.figma.com/design/4hGWdO4sajnjcOwE81j8BK/App--new-?node-id=5590-325&t=d1Eipx5hrUWWC8wz-1

### LINKED ISSUE #3891: Keystrokes: Ctrl+Return to Send AI Chat
### User story
As a user, while actively on a chat in the AI assistant, I would like to be able to user the keyboard shortcut Ctrl+Enter to send my AI chat.

### Details

1. There should be text in the chat on the assistant
2. User presses Ctrl+Return/Enter on their keyboard to send the chat

elias-ba:
## Description

This PR brings the AI Assistant for workflows and jobs into the new collaborative editor. It also has massive and important new UI designs. We now make the AI Assistant as a standalone view that slides in from the right on user click to assist them with their workflow creation or job writing. This is a very important UI / UX decision because it makes us closer to a globally available assistant across the whole App.

Closes #3816 
Closes #3891 

## Validation steps

1. Create a new workflow (make sure you are in the collab editor)
2. See the chat icon at the top right of the header (after the Save button) ? click on it and see the AI Assistant slide in from the right
3. Send a message to start a new session or open an existing session
4. Chat with the AI to assist you in generating your workflow
5. See the workflow automatically rendered when it's generated
6. See the workflow code rendered in the chat as an artifact
7. See the Apply and Copy buttons
8. Click on them to force or reapply or copy the workflow code
9. Click on Save when you happy with the workflow; note how you transition from create mode to edit mode without the AI Assistant closing
10. Keep using your chat session to bring in more modifications (or open a new chat session, it's up to you)
11. Click on a job node to open the IDE
12. See the AI Assistant still there but now loads different chat sessions (job chat sessions and workflow chat sessions are separated)
13. Open a job chat session or type a message to create a new one
14. Ask the assistant about your adaptor and the context it knows about your job
15. Change the adaptor on the fly and ask about it again; note how it's aware of the change
16. Chose to send or not your job expression or your run logs
17. Send a message and see how the AI receives or not your job expression and or your run logs
18. Generate a job expression using the AI Assistant
19. Note the Add and Copy buttons in the code artifact
20. Click on Add to add the job expression directly in the Monaco job editor
21. Use the step picker in the IDE header to change your step
22. See how the AI Assistant syncs and load the sessions for that step
23. Close the IDE and go back to your workflow canvas
24. See the AI Assistant load appropriate sessions
25. Close the AI Assistant; see it slide out to the right
26. Click around, test more, try different combinations, try to break it

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/test/collaborative-editor/__helpers__/index.ts
+++ b/assets/test/collaborative-editor/__helpers__/index.ts
@@ -171,6 +171,8 @@ export {
 export {
   createMockSessionContextStore,
   createMockHistoryStore,
+  createMockUIStore,
   createMockStoreContextValue,
   defaultSessionContextState,
+  defaultUIState,
 } from './storeMocks';
--- a/assets/test/collaborative-editor/__helpers__/storeMocks.ts
+++ b/assets/test/collaborative-editor/__helpers__/storeMocks.ts
@@ -37,10 +37,17 @@
 import { vi } from 'vitest';
 
 import type { SessionContextStore } from '../../../js/collaborative-editor/stores/createSessionContextStore';
-import type { HistoryStore } from '../../../js/collaborative-editor/types/history';
+import type {
+  HistoryStore,
+  HistoryState,
+  RunDetail,
+} from '../../../js/collaborative-editor/types/history';
 import type { SessionContextState } from '../../../js/collaborative-editor/types/sessionContext';
 import type { StoreContextValue } from '../../../js/collaborative-editor/contexts/StoreProvider';
-import type { RunDetail } from '../../../js/collaborative-editor/types/history';
+import type {
+  UIStore,
+  UIState,
+} from '../../../js/collaborative-editor/types/ui';
 
 // =============================================================================
 // Default State Objects
@@ -68,6 +75,30 @@ export const defaultSessionContextState: SessionContextState = {
   isLoading: false,
   error: null,
   lastUpdated: null,
+};
+
+/**
+ * Default empty UI state
+ * Used as the baseline for all UI store mocks
+ */
+export const defaultUIState: UIState = {
+  runPanelOpen: false,
+  runPanelContext: null,
+  githubSyncModalOpen: false,
+  aiAssistantPanelOpen: false,
+  aiAssistantInitialMessage: null,
+  createWorkflowPanelCollapsed: true,
+  templatePanel: {
+    templates: [],
+    loading: false,
+    error: null,
+    searchQuery: '',
+    selectedTemplate: null,
+  },
+  importPanel: {
+    yamlContent: '',
+    importState: 'initial',
+  },
 };
 
 // =============================================================================
@@ -110,7 +141,7 @@ export function createMockSessionContextStore(
   const defaultStore: SessionContextStore = {
     // Queries
     getSnapshot: vi.fn(() => defaultSessionContextState),
-    subscribe: vi.fn((listener: () => void) => {
+    subscribe: vi.fn((_listener: () => void) => {
       // Return unsubscribe function
       return () => {};
     }),
@@ -198,11 +229,11 @@ export function createMockHistoryStore(
   const defaultStore: HistoryStore = {
     // Queries
     getSnapshot: vi.fn(() => defaultState),
-    subscribe: vi.fn((listener: () => void) => {
+    subscribe: vi.fn((_listener: () => void) => {
       return () => {};
     }),
-    withSelector: vi.fn(<T>(selector: (state: any) => T) => {
-      return () => selector(defaultState);
+    withSelector: vi.fn(<T>(selector: (state: HistoryState) => T) => {
+      return () => selector(defaultState as HistoryState);
     }),
     getRunSteps: vi.fn(() => null),
     getActiveRun: vi.fn(() => activeRun),
@@ -240,6 +271,102 @@ export function createMockHistoryStore(
 }
 
 // =============================================================================
+// UIStore Mock Factory
+// =============================================================================
+
+/**
+ * Creates a mock UIStore with all required methods
+ *
+ * All methods are vi.fn() mocks that can be inspected and configured.
+ * State queries return sensible defaults that can be overridden.
+ *
+ * **State Tracking:**
+ * By default, the mock tracks state changes made via setImportState,
+ * setImportYamlContent, and clearImportPanel. This allows tests to
+ * verify component behavior that depends on state updates.
+ *
+ * @param overrides - Partial UIStore to override defaults
+ * @param initialState - Optional initial state to start with
+ * @returns Complete UIStore mock
+ *
+ * @example
+ * // Basic usage with default state
+ * const store = createMockUIStore();
+ * expect(store.getSnapshot().runPanelOpen).toBe(false);
+ *
+ * @example
+ * // Override state
+ * const store = createMockUIStore({}, {
+ *   ...defaultUIState,
+ *   runPanelOpen: true,
+ * });
+ */
+export function createMockUIStore(
+  overrides: Partial<UIStore> = {},
+  initialState: UIState = defaultUIState
+): UIStore {
+  // Mutable state that tracks changes
+  let currentState = { ...initialState };
+
+  const defaultStore: UIStore = {
+    // Queries
+    getSnapshot: vi.fn(() => currentState),
+    subscribe: vi.fn((_listener: () => void) => {
+      return () => {};
+    }),
+    withSelector: vi.fn(<T>(selector: (state: UIState) => T) => {
+      return () => selector(currentState);
+    }),
+
+    // Commands
+    openRunPanel: vi.fn(),
+    closeRunPanel: vi.fn(),
+    openGitHubSyncModal: vi.fn(),
+    closeGitHubSyncModal: vi.fn(),
+    openAIAssistantPanel: vi.fn(),
+    closeAIAssistantPanel: vi.fn(),
+    toggleAIAssistantPanel: vi.fn(),
+    collapseCreateWorkflowPanel: vi.fn(),
+    expandCreateWorkflowPanel: vi.fn(),
+    toggleCreateWorkflowPanel: vi.fn(),
+    setTemplates: vi.fn(),
+    setTemplatesLoading: vi.fn(),
+    setTemplatesError: vi.fn(),
+    setTemplateSearchQuery: vi.fn(),
+    selectTemplate: vi.fn(),
+    clearTemplatePanel: vi.fn(),
+    // Import panel commands that track state
+    setImportYamlContent: vi.fn((content: string) => {
+      currentState = {
+        ...currentState,
+        importPanel: { ...currentState.importPanel, yamlContent: content },
+      };
+    }),
+    setImportState: vi.fn(
+      (
+        importState: 'initial' | 'parsing' | 'valid' | 'invalid' | 'importing'
+      ) => {
+        currentState = {
+          ...currentState,
+          importPanel: { ...currentState.importPanel, importState },
+        };
+      }
+    ),
+    clearImportPanel: vi.fn(() => {
+      currentState = {
+        ...currentState,
+        importPanel: { yamlContent: '', importState: 'initial' },
+      };
+    }),
+  };
+
+  return {
+    ...defaultStore,
+    ...overrides,
+  };
+}
+
+// =============================================================================
 // Full StoreContextValue Mock Factory
 // =============================================================================
 
@@ -248,8 +375,8 @@ export function createMockHistoryStore(
  *
  * Provides a full StoreContextValue suitable for StoreProvider tests.
  * By default, most stores are minimal `{} as any` placeholders except
- * for sessionContextStore and historyStore which use their dedicated
- * mock factories.
+ * for sessionContextStore, historyStore, and uiStore which use their
+ * dedicated mock factories.
  *
  * @param overrides - Partial StoreContextValue to override defaults
  * @returns Complete StoreContextValue mock
@@ -286,7 +413,7 @@ export function createMockStoreContextValue(
     workflowStore: {} as any,
     sessionContextStore: createMockSessionContextStore(),
     historyStore: createMockHistoryStore(),
-    uiStore: {} as any,
+    uiStore: createMockUIStore(),
     editorPreferencesStore: {} as any,
     aiAssistantStore: {} as any,
   };


taylordowns2000 (changes requested):
thanks elias! noting things from our call:

- [x] timestamps missing
- [x] retry button missing
- [x] code isn't applied (must manually click apply)
- [x] esc shouldn't close AI assistant

maybe more, we can add as we go!

taylordowns2000 (changes requested):
awesome. few more odds and ends

1. bottom border too big on collapsed artifact card
2. new chats kick themselves out of their own sessions?! https://www.loom.com/share/eb590d9cfcce481d9ef23872a3f6d00b
3. i get kicked out of chat sessions randomly: https://www.loom.com/share/8b9af7cc4006428eaf691a05b0a4c1a3

taylordowns2000 (approved):
it's brilliant. thanks @elias-ba !

stuartc (changes requested):
Lets a bag of changes requests with you on Slack, but overall this is good.

I did get spooked by:

`lib/lightning_web/channels/workflow_channel.ex:168`

```elixir
# CRITICAL: Always fetch the actual latest workflow from DB
# socket.assigns.workflow could be an old snapshot (e.g., v22)
# but we need to send the actual latest lock_version (e.g., v28)
fresh_workflow = Lightning.Workflows.get_workflow(workflow.id)
```

**Changed to:**

```elixir
fresh_workflow = Lightning.Workflows.get_workflow(workflow.id) || workflow
```

I need to know why this was changed and what issues we might be exposing ourselves to. It feels very disconnected to what it's solving, and `get_context` is also starting to feeling like it's perhaps doing too much. Are we calling it multiple times? If a workflow is unsaved, then would it not be more appropriate to return something like `nil`? Should the frontend maybe be smart enough to know the difference?

elias-ba:
> Lets a bag of changes requests with you on Slack, but overall this is good.
> 
> I did get spooked by:
> 
> `lib/lightning_web/channels/workflow_channel.ex:168`
> 
> ```elixir
> # CRITICAL: Always fetch the actual latest workflow from DB
> # socket.assigns.workflow could be an old snapshot (e.g., v22)
> # but we need to send the actual latest lock_version (e.g., v28)
> fresh_workflow = Lightning.Workflows.get_workflow(workflow.id)
> ```
> 
> **Changed to:**
> 
> ```elixir
> fresh_workflow = Lightning.Workflows.get_workflow(workflow.id) || workflow
> ```
> 
> I need to know why this was changed and what issues we might be exposing ourselves to. It feels very disconnected to what it's solving, and `get_context` is also starting to feeling like it's perhaps doing too much. Are we calling it multiple times? If a workflow is unsaved, then would it not be more appropriate to return something like `nil`? Should the frontend maybe be smart enough to know the difference?

Hey @stuartc, good question and thanks for the attention. The `|| workflow` fallback was causing issues because it masked the real problem. I've updated it to explicitly check `lock_version`:
```
  fresh_workflow =
    if is_nil(workflow.lock_version) do
      workflow
    else
      Lightning.Workflows.get_workflow(workflow.id)
    end
```
The issue is when a user creates a new workflow and opens the AI assistant before saving, workflow.lock_version is nil because it hasn't been persisted yet. Calling `get_workflow(workflow.id)` returns nil for unsaved workflows, which crashed downstream code.

If `lock_version` is `nil`, we know it's an unsaved workflow, so we use the in-memory workflow directly. If it has a `lock_version`, we fetch fresh from DB to get the latest version.

This is more explicit than `|| workflow` which would silently fall back even if the DB fetch failed for other reasons.

elias-ba:
Hey @stuartc, the registry pattern (`useAIChannelRegistry`) centralizes all Phoenix channel management for the AI assistant. Instead of each component managing its own connect/disconnect logic with `useEffect` chains and `setTimeout` workarounds, components just declare "I need this session" and the registry handles the rest. This follows our discussion of last time and it works super well currently.

Now the components call `registry.subscribe(topic, subscriberId, context)` and the registry:
  - Tracks subscribers per topic
  - Reuses existing connections when possible
  - Handles cleanup with proper timing
  - Prevents duplicate joins
  - Guards against stale async callbacks

assets/js/collaborative-editor/components/MessageList.tsx
```diff
@@ -0,0 +1,553 @@
...
+      preElement.appendChild(buttonContainer);
+    });
+  }, [html, showAddButtons]);
+
+  return (
+    <div
+      ref={containerRef}
+      className={className}
+      dangerouslySetInnerHTML={{ __html: html }}
+    />
```
stuartc: This isn't safe, we're passing unsantized html from an llm into the dom. Are we sanitising it on the server? In liveview we santised it at render time iirc.

If not, we should be using `dompurify`, something like: 

```typescript
import DOMPurify from 'dompurify';

const html = useMemo(() => {
  const result = marked.parse(content, { async: false });
  return DOMPurify.sanitize(
    typeof result === 'string' ? result : content,
    {
      ALLOWED_TAGS: ['p', 'br', 'strong', 'em', 'code', 'pre',
                     'ul', 'ol', 'li', 'h1', 'h2', 'h3', 'a', 'blockquote'],
      ALLOWED_ATTR: ['href', 'class'],
    }
  );
}, [content]);
```

lib/lightning_web/channels/ai_assistant_channel.ex
```diff
@@ -0,0 +1,775 @@
...
+          job_name: unsaved_job["name"],
+          workflow_name: workflow.name,
+          is_unsaved: true
+        })
+
+      session.job_id ->
+        case Jobs.get_job(session.job_id) do
+          {:ok, job} ->
+            workflow = Workflows.get_workflow(job.workflow_id)
+
```
stuartc: Looks like we are introducing and N+1 issue here. When you list say 20 sessions, thats 40+ extra queries. Should we not push the preloading into `list_sessions/3` and make this function purely view/rendering?

assets/js/collaborative-editor/lib/AIChannelRegistry.ts
```diff
@@ -0,0 +1,686 @@
...
+      const typedPayload = payload as {
+        message_id: string;
+        status: MessageStatus;
+      };
+      this.store._updateMessageStatus(
+        typedPayload.message_id,
+        typedPayload.status
+      );
+    });
+  }
```
stuartc: Looks like a memory leak waiting to happen, I don't see any explicit deregistration for these listeners/callbacks.

assets/js/collaborative-editor/stores/createAIAssistantStore.ts
```diff
@@ -0,0 +1,581 @@
...
+
+  /**
+   * Update message status
+   * @internal Called by useAIAssistantChannel hook
+   */
+  const _updateMessageStatus = (messageId: string, status: string) => {
+    state = produce(state, draft => {
+      const message = draft.messages.find(m => m.id === messageId);
+      if (message) {
+        message.status = status as Message['status'];
```
stuartc: Use MessageStatus directly instead.

assets/js/collaborative-editor/components/MessageList.tsx
```diff
@@ -0,0 +1,553 @@
...
+        };
+
+        buttonContainer.appendChild(addButton);
+      }
+
+      preElement.style.position = 'relative';
+      preElement.classList.add('group');
+      preElement.appendChild(buttonContainer);
+    });
+  }, [html, showAddButtons]);
```
stuartc: Whats going on here? Why not proper components? React can't track or selectively rerender these.

assets/js/collaborative-editor/components/AIAssistantPanel.tsx
```diff
@@ -0,0 +1,601 @@
...
+      className={cn(
+        'h-full flex flex-col overflow-hidden bg-white relative',
+        !isResizable && [
+          'flex-none border-l border-gray-200',
+          'transition-[width,border] duration-300 ease-in-out',
+          isOpen ? 'w-[400px]' : 'w-0 border-l-0',
+        ]
+      )}
+      role="dialog"
+      aria-modal="false"
```
stuartc: This isn't a modal, we're setting it to "dialog" and then turning off focus trapping with `aria-modal="false"`, essentially undoing the point of a dialog.

The main panel should use role="complementary" (or just be an `<aside>`) since it's a persistent side panel that coexists with main content.

And there are two internal overlays that might need to be `aria-modal`s. The about modal and the disclaimer screen, not deal breakers since their focus trapping impact is limited. 

The way to test the UX for this is by pressing enter or tab after you open and close things, it shouldn't be tabbing to different inputs behind things you can't see (this is what `role=dialog` and `aria-modal` essentially controls)

assets/js/collaborative-editor/hooks/useAISession.ts
```diff
@@ -0,0 +1,281 @@
...
+          // Registry handles connection and sets session when join succeeds
+          registry.subscribe(topic, subscriberId, finalContext);
+        } catch (err) {
+          logger.error('Failed to serialize workflow:', err);
+          // On error, still subscribe but without workflow YAML
+          if (currentSubscriptionRef.current === topic) {
+            registry.subscribe(topic, subscriberId, context);
+          }
+        }
+      })();
```
stuartc: The dynamic import makes this code path async, which exposes it to race
conditions when switching sessions quickly. The stale-callback guard on line 214
tries to catch this, but it can't distinguish between "topic A from click 1" and
"topic A from click 3" if the user navigates A → B → A.

serializeWorkflowToYAML is synchronous and the module is small. A static import
at the top of the file would make this entire block synchronous and immune to
race conditions:

```typescript
import { serializeWorkflowToYAML } from '../utils/workflowSerialization';

// In the effect - no async wrapper, no guard needed:
const workflowForSerialization = { ... };
const workflowYAML = serializeWorkflowToYAML(workflowForSerialization);
const finalContext = { ...context, code: workflowYAML };
registry.subscribe(topic, subscriberId, finalContext);
```

assets/js/collaborative-editor/stores/createAIAssistantStore.ts
```diff
@@ -0,0 +1,581 @@
...
+    });
+
+    // Save session to localStorage with mode-specific key
+    const storageKey = getStorageKeyFromContext();
+    if (storageKey) {
+      try {
+        localStorage.setItem(storageKey, session.id);
+      } catch (error) {
+        logger.error('Failed to save session to localStorage', error);
+      }
```
stuartc: I can't see any hook called useAIAssistantChannel, also we seem to be setting keys for sessions like `ai-job-<jobId>`, but I can't see it being read anywhere? Where are these being read from?

I see that ChatInput uses a `<storageKey>:attach-code` and logs key, both read and write. 

I would suggest dramatically cutting down/simplifying the localStorage stuff for the time being, keeping it just for the ChatInput keys.

There is a 'low-impact' issue here in that, 1. these keys are never cleared out on log out, 2. they aren't scoped by the user so we would be keeping data between logins. Again low risk but might be a bug waiting to happen _if_ we were to be actually using the `ai-job` etc keys. So if we were to properly implement the reading of those keys we would need to filter out those keys on store mount and on logout remove them as well.

stuartc (changes requested):

assets/js/collaborative-editor/components/MessageList.tsx
elias-ba: Fixed! Added DOMPurify with a strict whitelist - tags and attributes are defined in a shared `sanitizeHtml` utility so it's easy to maintain. Also spotted the same issue in the adaptor docs (`Function.tsx`) so fixed that too while I was at it. Both usages are now covered by tests - there's a dedicated test suite for the utility plus the integration tests in MessageList.

lib/lightning_web/channels/ai_assistant_channel.ex
elias-ba: Fixed! Moved the preloading into `list_sessions` - now preloads `:workflow` and `job: :workflow` so the formatting functions just use the preloaded data.

For unsaved jobs, I also set `workflow_id` directly on the session when creating it (instead of only storing it in meta), so it can be preloaded like any other session. No extra queries at all now.

assets/js/collaborative-editor/lib/AIChannelRegistry.ts
elias-ba: Fixed by storing handler references in the channel entry and calling `channel.off()` before `channel.leave()` in both `cleanup()` and `destroy()`. Same pattern used elsewhere in the codebase (e.g., `createSessionContextStore.ts`).

assets/js/collaborative-editor/stores/createAIAssistantStore.ts
elias-ba: Done. Thanks

assets/js/collaborative-editor/components/MessageList.tsx
elias-ba: Refactored to use `react-markdown` with a custom `CodeBlock` component for the COPY/ADD buttons. No more DOM manipulation - everything is proper React components now. Also updated `Function.tsx` (adaptor docs) to use `react-markdown`. I removed `marked`, `dompurify`, and the `sanitizeHtml` utility entirely.

assets/js/collaborative-editor/components/AIAssistantPanel.tsx
elias-ba: Changed the main panel to use a semantic <aside> element which has an implicit role="complementary" - much better suited for a persistent side panel that doesn't trap focus. Also added proper role="dialog" aria-modal="true" to the internal overlays (About modal and Disclaimer screen) that actually do behave as modal dialogs with trapped focus.

assets/js/collaborative-editor/hooks/useAISession.ts
elias-ba: Thanks for this. Fixed

assets/js/collaborative-editor/stores/createAIAssistantStore.ts
elias-ba: You're right - the session ID localStorage writes (`ai-job-*, ai-workflow-*, ai-project-*`) were dead code. They were being written but never read anywhere. Removed `getStorageKeyFromContext` and the `localStorage.setItem` call entirely. The ChatInput localStorage usage for `attach-code` and `attach-logs` checkboxes remains - that's functional and properly scoped via `useAIStorageKey`. Those keys are actively read/written by ChatInput to persist the checkbox states.

PR #4053, opened 2025-11-21
fix: flickering of inactive collaborator

### LINKED ISSUE #3931: Inactive Collaborator Avator sdissaperas on teh Editor insead of greying out

doc-han:
## Description

Better implements
1. activity state for connected collaborators on a workflow showing their current state being `active`|`away`. also deals with when they switch from the tab and back.
2. maintains our 10s ydoc keep-alive trigger updates using `lastSeen`.

#### What was causing the flickering?
- ydoc requires us to update a user's state within a 30s interval to keep them considered in awareness(being online)
- due to this, we have a `setInterval` which updates the user's `lastSeen` every 10s, just so we are in the 30s interval
- we try to do this all the time but some browsers have a smart way of saving resources which is to throttle JS execution on tabs that aren't active(user has switched away from)
- when this happens, JS timers like `setInterval` start to miss their interval calls. eg. a 10s interval call might be actually triggered after 60s.
- hence, we start missing the 30s requirement by ydoc and then ydoc removes all users who have started missing the time.
- but if a user missed the 30s threshold and got called after 60s due to throttling. it shows a flicker on all other active collaborator. because the user was removed when they missed 30s but got back in awareness when their throttled trigger came in after 60s. 

#### Approach taken
- we keep our 10s interval trigger to ydoc
- we have another field `lastState` controlled by the browser visibility API to track whether a user is `active` or `away`

**How do we smoothing the presence of a user to avoid the flickering?**
- Now, we only cache a user in awareness for 60s when they are `away`. (throttling likely happens to away users)
- If they're away and keep showing up in awareness, we give that precedence over what's in the cache. (probably they're not throttled)
- As soon as they come back to active, we dont wait for TTL of cache but just remove them from cache. (throttling never happens to active users)

Closes #3931 

## Validation steps

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/collaborative-editor/hooks/useAwareness.ts
+++ b/assets/js/collaborative-editor/hooks/useAwareness.ts
@@ -96,8 +96,9 @@ export const useAwarenessUsers = (): AwarenessUser[] => {
 /**
  * Hook to get only remote users (excluding the local user)
  * Useful for cursor rendering where you don't want to show your own cursor
- * Deduplicates users by keeping the one with the latest lastSeen timestamp
- * and adds connectionCount to show how many connections they have
+ * Deduplicates users by keeping the one with the highest priority state
+ * (active > away > idle), then by latest lastSeen timestamp
+ * Adds connectionCount to show how many connections they have
  */
 export const useRemoteUsers = (): AwarenessUser[] => {
   const awarenessStore = useAwarenessStore();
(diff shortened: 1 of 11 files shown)


doc-han:
Previous conversations at #3951

assets/js/collaborative-editor/hooks/useAwareness.ts
```diff
@@ -93,50 +93,84 @@ export const useAwarenessUsers = (): AwarenessUser[] => {
   return useSyncExternalStore(awarenessStore.subscribe, selectUsers);
 };
 
+const awayUserCache = new Map<
+  string,
+  { user: AwarenessUser; expiresAt: number }
+>();
+
+const AWAY_CACHE_DURATION = 60000; // 60 seconds
```
elias-ba, line 101: I am wondering if we shouldn't move this into `AwarenessStore` so that we can properly scope it and clean it up to avoid cache pollutions and memory leaks when users open workflows in multiple tabs. Wouldn't the fact that it's module-level Map make it shared across all component instances and persists across React re-renders ?

assets/js/collaborative-editor/stores/createAwarenessStore.ts
```diff
@@ -342,6 +355,30 @@ export const createAwarenessStore = (): AwarenessStore => {
...
+    if (visibilityProps) {
+      const handler = activityStateChangeHandler.bind(undefined, setState);
+      // initial call
+      handler();
+      document.addEventListener(visibilityProps.visibilityChange, handler);
+
+      // return cleanup function
+      return () => {
+        document.removeEventListener(visibilityProps.visibilityChange, handler);
+      };
```
elias-ba, line 378: What happens when `visibilityProps` is `null` ? Wouldn't the function return `undefined` instead of a no-op cleanup function ? Wouldn't this cause issues in `StoreProvider.tsx:160` where the cleanup is called unconditionally ?

assets/js/collaborative-editor/hooks/useAwareness.ts
elias-ba, line 101: The 60-second duration looks arbitrary. If that's the case, shouldn't we consider documenting why this specific value was chosen. Is it related to typical browser throttling behavior? Should it be configurable?

assets/js/collaborative-editor/hooks/useAwareness.ts
```diff
@@ -93,50 +93,84 @@ export const useAwarenessUsers = (): AwarenessUser[] => {
...
-        userMap.set(userId, user);
-      } else {
-        // Keep the user with the latest lastSeen timestamp
-        const existingLastSeen = existingUser.lastSeen || 0;
-        const currentLastSeen = user.lastSeen || 0;
-        if (currentLastSeen > existingLastSeen) {
-          userMap.set(userId, user);
-        }
+      connectionCounts.set(userId, (connectionCounts.get(userId) || 0) + 1);
+      (userMap.get(userId) || userMap.set(userId, []).get(userId)!).push(user);
```
elias-ba, line 129: This looks dense. No way to simplify ?

elias-ba (changes requested):
Hey @doc-han this is a well-thought-out solution. What do you think about my questions / suggestions down below ?

doc-han:
Moving this PR to draft since its priority at the moment has been changed to low

josephjclark:
@doc-han to work out and propose next steps for this

midigofrank:
@doc-han is this still a thing?

doc-han:
> @doc-han is this still a thing?

Nope. this PR got over-run by another PR by @stuartc that got merged in
https://github.com/OpenFn/lightning/pull/4068

PR #4067, opened 2025-11-26
fix: logs panel not allowing scroll after being resized to minimum height (collab-editor)

### LINKED ISSUE #4034: Logs panel doesn't scroll in new collab editor on resize
When resizing the meta/steps panel, the logs panel isn't properly responsive to scrolling

1. run job with lots of logs
2. resize steps/meta panel above logs panel
3. see that logs panel doesn't know it's been resized and wont scroll
4. change tabs (to output or logs)
5. change back to logs
6. see that logs panel NOW scrolls

doc-han:
## Description

This PR resolves an issue where the logs panel loses its scroll after a resize. 
**The cause of this issue is that**
1. The logs panel has enough space to show more logs
2. When you resize it from large to small, the editor doesn't get resized but actually takes more space below and off the screen making us unable to scroll to reach there. 

Note: At times the scroll might be available but you might not be able to reach the bottom of your logs. 

**Solution**
use a grid where the first item in the row's height matches its content(`auto`) and then the second item in row takes up the remaining space.

Closes #4034 

## Validation steps

1. Run a workflow with a lot of logs (alternatively, just switch the log type to debug logs)
2. After your run and the logs for the run are showing. use the resize handle to resize the logs panel to it's minimum viewable height. 
3. Are you able to scroll all the way down to the last log? If yes, we're good. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/test/collaborative-editor/components/run-viewer/LogTabPanel.test.tsx
+++ b/assets/test/collaborative-editor/components/run-viewer/LogTabPanel.test.tsx
@@ -295,7 +295,11 @@ describe('LogTabPanel', () => {
       // Check for log viewer container
       const logViewerContainer = container.querySelector('.bg-slate-700');
       expect(logViewerContainer).toBeInTheDocument();
-      expect(logViewerContainer).toHaveClass('flex', 'h-full', 'flex-col');
+      expect(logViewerContainer).toHaveClass(
+        'grid',
+        'h-full',
+        'grid-rows-[auto_1fr]'
+      );
 
       // Check for log level filter header
       const filterHeader = container.querySelector(


doc-han:
> What about if we used CSS Grid on `LogTabPanel` so we don't have any magic numbers for header?
> ...

Oh thanks @lmac-1. updated the PR to use the grid suggestion. 👍🏽

taylordowns2000:
> @doc-han , this seems ok to me but I can't figure out where the calc function is from your PR "Solution" section:
> 
> > use css calc function to calculate the height of container around the editor by subtracting the height of the above mini-header.
> 
> is this the `grid-rows-[auto_1fr]` bit? that seems like a different approach (using a grid?) than the one explained in the PR (using css calc?)

Edit: I now see the other comments and realize that this must have simply been left over from the original approach. All good from my end. (But maybe update the PR description so when your reviewers read it they don't get confused?)

doc-han:
> @doc-han , this seems ok to me but I can't figure out where the calc function is from your PR "Solution" section:
> 
> > use css calc function to calculate the height of container around the editor by subtracting the height of the above mini-header.
> 
> is this the `grid-rows-[auto_1fr]` bit? that seems like a different approach (using a grid?) than the one explained in the PR (using css calc?)

Yeah. changed to grid. updating the PR description

theroinaochieng (approved):

PR #4065, opened 2025-11-25
Fix: Improves panel management in workflow editor

lmac-1:
## Description

This PR includes two related improvements to panel management in collaborative editor: 

1. **Fix settings panel not opening:** When the Run Panel is open and user clicks the settings icon, the settings panel now opens correctly instead of immediately being overridden back to the Run Panel

2. **Improve Run Panel context switching:** When the Run Panel is open and user clicks a different node, the panel now stays open and updates to show the new node's context, instead of closing entirely.

NOTE: 2 was something I noticed and fixed after fixing 1. I added the `key` prop to `ManualRunPanel` meaning that when you change node, any custom text input will be wiped as we remount the component. This means that any unsaved custom input, tab selection, or dataclip selection will be lost. This trade-off was necessary to ensure the panel properly resets to the correct default state for each node type and prevents stale state. However, this may go against the expected behaviour and need to change.

Each change is in its own commit. 

Closes #4008

## Validation steps

Change 1
1. Open collaborative editor
2. Click on a node on the diagram
3. Click "Run" button in the job inspector footer to open the Run Panel
4. Click the settings icon in the top header
5. Verify that settings correctly opens and it doesn't rerun 

This is the screenshot of what it should look like before clicking settings: 
image

Change 2: 
1. Open collaborative editor
2. Click on a node
3. Click "Run" button in the job inspector footer to open the Run Panel
4. Click on another node
5. It should reload correctly with that node in context, as if you have clicked on that node for the first time

## Additional notes for the reviewer

Feel free to play round with clicking on different nodes from different contexts and/or settings to make sure it loads properly

## AI Usage

Please disclose how you've used AI in this work (it's cool, we just want to know!):

- [x] Code generation (copilot but not intellisense)
- [x] Learning or fact checking
- [x] Strategy / design
- [ ] Optimisation / refactoring
- [ ] Translation / spellchecking / doc gen
- [ ] Other
- [ ] I have not used AI

You can read more details in our [Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Pre-submission checklist

- [x] I have performed a **self-review** of my code.
- [ ] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [x] I have updated the **changelog**.
- [x] I have ticked a box in "AI usage" in this PR

PR #4082, opened 2025-11-28
Fix: Improves panel management in workflow editor (take 2)

### LINKED ISSUE #4008: Settings panel doesn't open when JobInspector is visible (Collaborative Editor)
**Describe the bug** When the JobInspector is open (showing job details in the right panel), clicking the settings icon in the header doesn't open the Workflow Settings panel. Instead, the JobInspector appears to reload.

**Version number**  Production (not sure of version number)

- [x] Yes
- [ ] No

1. Open a workflow in the collaborative editor
2. Click on a job node
3. Click "Run" to open JobInspector
4. Click the settings icon in the header
5. Notice that JobInspector reloads rather than opening Workflow settings

**Expected behavior** The Workflow settings panel should open at step 5 instead of reloading JobInspector

Image

Image

Expected behaviour:
Image

**Additional context** If you click the settings icon at step 2 (after clicking on a job node but _before_ clicking "Run"), you will notice that the settings panel opens correctly. 
It appears to also be "broken" on LiveView - but the broken behaviour at step 5 is that the workflow settings panel briefly appears before fading out and disappearing.
Also, sorry if my vocabulary isn't completely right about the different parts of the project.

**Technical context**: The Inspector component lives at assets/js/collaborative-editor/components/inspector/index.tsx

lmac-1:
## Important context

This was previously raised in #4065 but we had to revert it due to a regression. @taylordowns2000 noted that when you clicked on FullScreenIDE steps, it would close the IDE rather than show the correct part of the logs: 

image

This PR includes the same commits from before + fixes the above bug.

## Description

This PR includes two related improvements to panel management in collaborative editor: 

1. **Fix settings panel not opening:** When the Run Panel is open and user clicks the settings icon, the settings panel now opens correctly instead of immediately being overridden back to the Run Panel

2. **Improve Run Panel context switching:** When the Run Panel is open and user clicks a different node, the panel now stays open and updates to show the new node's context, instead of closing entirely.

NOTE: 2 was something I noticed and fixed after fixing 1. I added the `key` prop to `ManualRunPanel` meaning that when you change node, any custom text input will be wiped as we remount the component. This means that any unsaved custom input, tab selection, or dataclip selection will be lost. This trade-off was necessary to ensure the panel properly resets to the correct default state for each node type and prevents stale state. However, this may go against the expected behaviour and need to change.

Each change is in its own commit. 

Closes #4008

## Validation steps

Change 1
1. Open collaborative editor
2. Click on a node on the diagram
3. Click "Run" button in the job inspector footer to open the Run Panel
4. Click the settings icon in the top header
5. Verify that settings correctly opens and it doesn't rerun 

This is the screenshot of what it should look like before clicking settings: 
image

Change 2: 
1. Open collaborative editor
2. Click on a node
3. Click "Run" button in the job inspector footer to open the Run Panel
4. Click on another node
5. It should reload correctly with that node in context, as if you have clicked on that node for the first time
**6. Open FullScreenIDE and click through run steps. Make sure it doesn't close the IDE **
7. Click around the FullScreenIDE and do other actions and make sure nothing breaks.

Feel free to play round with clicking on different nodes from different contexts and/or settings to make sure it loads properly - also when FullScreenIDE is open.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


assets/js/collaborative-editor/components/run-viewer/StepItem.tsx
```diff
@@ -31,15 +31,23 @@ export function StepItem({ step, selected, runInsertedAt }: StepItemProps) {
...
       job: step.job_id,
       run: currentRunId, // Preserve run context
       step: step.id, // Update to this step
-    });
+    };
+
+    // Only include panel in updates if it exists (avoids deleting it)
+    if (currentPanel) {
+      updates.panel = currentPanel;
+    }
```
stuartc, line 48: This feels a bit off, or rather that we have some responsibility to address somewhere else. But this is defensive at worst, so gonna leave it in.

assets/js/collaborative-editor/hooks/useWorkflow.ts
```diff
@@ -300,15 +302,26 @@ export const useNodeSelection = () => {
       const foundTrigger = state.triggers.find(trigger => trigger.id === id);
       const foundEdge = state.edges.find(edge => edge.id === id);
 
+      // Preserve special panels (run, editor, settings, code); otherwise clear panel to show node inspector
+      const specialPanels = ['run', 'editor', 'settings', 'code'];
```
stuartc, line 306: Also not a dealbreaker, but this special panel stuff feels like we could find a more 'pointed' way of dealing with this logic. It's good that it's in a hook though. I have a branch in the wings that addresses some of the difficulties around getting and replacing the url state, will take some of this logic into consideration before I finish.

PR #4084, opened 2025-11-28
fix: save&run disabled after new workflow save

### LINKED ISSUE #4071: Save and Run Button Disabled for Users with Access
1. Click on create new workflow button and switch to collaborative editor
2. Click on the import button
3. Paste your YAML or upload a yaml file and click on “create” button
4. The left panel goes away but the save and run buttons at the top are disabled with the tooltip message “You do not have permission to edit this workflow”
5. The URL also does not update to the created workflow. It stays at `/new/collaborate` instead of `/workflow_id/collaborate`

doc-han:
## Description

This PR resolves an issue where save&run buttons were disabled after creating a new workflow

Closes #4071 

## Validation steps

1. Go to create a new workflow
2. Click the experimental icon(beaker) in the header to switch to the new editor
3. You'll notice we don't support templates there yet. Now use the import and then import some workflow via yaml. (example yaml provided below)
4. After clicking create, are your run & save buttons enabled? Are you able to perform any action as usual? If yes then we're good

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Pre-submission checklist


test/lightning_web/channels/workflow_channel_test.exs
```diff
@@ -334,6 +334,37 @@ defmodule LightningWeb.WorkflowChannelTest do
...
+
+      ref = push(new_socket, "get_context", %{})
+
+      assert_reply ref, :ok, response
+
+      # CRITICAL: For new unsaved workflows, fresh_workflow will be nil
+      # so latest_snapshot_lock_version should fall back to workflow.lock_version
+      assert %{latest_snapshot_lock_version: latest_lock_version} = response
+      # New workflows start at lock_version 0 (or nil, which should be handled)
+      assert latest_lock_version == 0 || is_nil(latest_lock_version)
```
elias-ba: Shouldn't this assert be `assert latest_lock_version == 0` to make the test more precise and catch any unexpected nil values. Based on the code logic, we should always get 0 for new workflows since we create a workflow struct with implicit lock_version: 0 (Ecto schema default) and we fall back to workflow.lock_version when fresh_workflow is nil. Right ?

test/lightning_web/channels/workflow_channel_test.exs
```diff
@@ -334,6 +334,37 @@ defmodule LightningWeb.WorkflowChannelTest do
...
+      # Verify we're in a new workflow context
+      assert new_socket.assigns.workflow.name == "Untitled workflow"
```
elias-ba, line 367: Could we assert new_socket.assigns.workflow.lock_version == 0 too ? This was we verify the in-memory workflow actually has lock_version: 0

test/lightning_web/channels/workflow_channel_test.exs
doc-han: resolved

test/lightning_web/channels/workflow_channel_test.exs
doc-han: resolved

doc-han:
> Hey @doc-han nice one. The fix is great. I could actually approve but I would love to draw your attention to the comments I left in my review. Also maybe confirm with @taylordowns2000 if we want to enable the 'Run' button during the creation mode. I could run my workflow before even saving it. The Run saved the workflow then ran but as a regular user that surprised me because I was expecting the button to just run. And after the run, I was directly in edit mode.
> 
> Image

Resolved this too. 
Made the buttons on the right of the header not accessible at this point in time

taylordowns2000:
@theroinaochieng , is this still relevant after https://github.com/OpenFn/lightning/pull/4087 ?

stuartc:
@taylordowns2000 you might be right, I'm going to review the other PR and circle back.

assets/js/collaborative-editor/components/left-panel/index.tsx
```diff
@@ -40,7 +40,7 @@ export function LeftPanel({
 
   return (
     <div
-      className={`absolute inset-y-0 left-0 w-1/3 transition-transform duration-300 ease-in-out z-10 ${
```
stuartc, line 43: 😘

stuartc (approved):

PR #4130, opened 2025-12-04
Add promotional banner for collab editor in legacy

### LINKED ISSUE #4091: Prepare collab-editor for mainstream adoption
To facilitate release of the new editor, we need to make some modifications to the experimental features flag around the new collaborative editor:

- [x] Maintain the users choice to use new collaborative editor in user preferences (so that any time you access the editor, from any link in the app, you're using the new collab editor) once it's been toggled "on" via the beaker. (e.g., `{ ...stuff, userPrefs.useNewEditor: true }`)
- [x] Add a dismissible banner to the legacy editor asking users to switch to the new editor
- [x] Make collab editor available to all users, not just experimental feature users

## These follow-up issues are handled in later sprints:
- [ ] #4132 
- [ ] #4111
- [ ] #4112

midigofrank:
## Description

Implements a dismissible promotional banner that encourages users to try the collaborative editor when viewing workflows in the legacy editor.

Features:
- Displays on all workflow versions (including snapshots)
- Self-contained component that extracts projectId and workflowId from URL
- Cookie-based dismissal persisting for 90 days
- Preserves 'a' (run) and 'v' (version) query params when navigating
- Blue-themed info banner positioned at top-center of canvas

The banner uses window.location to extract necessary IDs, avoiding prop drilling through multiple component layers and maintaining consistency with existing legacy editor patterns.

Closes #4091

## Validation steps

1. Open legacy editor
2. You should see a banner appear on top of the workflow

Screenshot 2025-12-04 at 17 01 22

1. This appears regardless of whether you have experimental features flag
2. Only `v` and `a` query params are passed on to the collab editor url. This means if you have a job selected and click the try now link, you find the job open in the new editor
3. The `projectID` and `workflowID` are parsed from the URL

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/workflow-diagram/CollaborativeEditorPromoBanner.tsx
+++ b/assets/js/workflow-diagram/CollaborativeEditorPromoBanner.tsx
@@ -2,11 +2,10 @@
  * CollaborativeEditorPromoBanner - Promotional banner encouraging users to try the collaborative editor
  *
  * Displays when:
- * - User has experimental features enabled
- * - Banner hasn't been previously dismissed
+ * - Banner hasn't been previously dismissed (checked via cookie)
  *
  * Features:
- * - Fixed position at bottom-center of viewport
+ * - Absolute position at bottom-center of workflow canvas
  * - Dark themed design matching Tailwind sticky banner pattern
  * - Dismissible via X button
  * - Persists dismissal in cookies for 90 days
--- a/assets/test/workflow-diagram/CollaborativeEditorPromoBanner.test.tsx
+++ b/assets/test/workflow-diagram/CollaborativeEditorPromoBanner.test.tsx
@@ -1,0 +1,232 @@
+/**
+ * CollaborativeEditorPromoBanner Tests
+ *
+ * Verifies banner behavior:
+ * - Renders when not dismissed
+ * - Hides when dismissed (via cookie)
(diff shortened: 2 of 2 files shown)


taylordowns2000 (approved):
great stuff. removed dependency on experimental, updated CL, good to go. subsequent work can be tracked here but is lower priority: #4132

elias-ba:
Doesn't this create consistency problem on how display banners in the app ? Some banners are occupying the whole width of the page (see the yellow banner) and now some are just floating at top middle of the page (see the blue banner). Why do we do that ? If the aim is to change how we display banners then why not change them all ? The blue banner is hiding the yellow banner's content anyway. Maybe the blue banner should be pop-up at the bottom right corner of the banner and be sticky. Same as we do with the "Accept cookies" pop-up in the billing app. Or could we try something simpler and more discrete like Github does, just a pulse animation on the beaker icon with a tooltip (maybe let's change the beaker icon to another one) that would be enough to trigger the user's attention. I really don't think stacking banners like this is great, let me know what you think guys. I know this is super hard because, we have a real problem of real state in the workflow editor.

Screenshot 2025-12-05 at 15 22 22

elias-ba (approved):

PR #4166, opened 2025-12-10
Improve version pinning behavior in collaborative editor

### LINKED ISSUE #4121: Regression: Version dropdown gets stuck
**Describe the bug** Version dropdown is getting stuck in different scenarios. This has regressed and is quite difficult to replicate.

**Version number** v2.15.0-pre1 `44fa909`

- [x] Yes
- [ ] No

1. Go to collab editor 
2. Make a note of the current latest version and click through runs in history to see versions of previous runs (e.g v167)
3. Make sure the most recent run in history is selected in history panel
4. Make an update to the run and click Save
5. Change the version dropdown to different versions 

Expected behaviour:
- You can click through the versions 

Actual behaviour: 
- You get stuck. When you click on the version that is `latest - 1` and try to go back to latest, it doesn't go to latest. It stays on `latest-1`

**Additional context** It seems that you must have the previous run selected when making the update to the workflow and saving. That's when the bug happens. It's difficult to replicate, but if you see the version dropdown getting stuck, you're on the right track. It seems to be related to clicking between `latest` and the version before latest. As soon as you click on a run with a previous version to `latest - 1`, things seem to reset as expected. It's a weird one.

### LINKED ISSUE #4149: Unable to go to latest
**Describe the bug** As a user for a workflow I own, I'm unable to go to the latest version of the workflow. This also affects my ability to enable and disable my workflow.

https://www.loom.com/share/ea803bd37aca48d597b3558acc416c9e

**Version number** v2.15.0-pre4 (1.5.52-pre4)

- [ ] Yes
- [x] No

1. Go to a project then a workflow
2. Click on the version picker (notice that the version is by default on a older snapshot)
3. Try to select an earlier snapshot 
4. Try to move to latest
5. Notice the weird behavior

midigofrank:
## Description
Changes the version dropdown and read-only logic to use URL parameter presence instead of comparing lock versions. When viewing a run from history, only pins the version if it differs from latest, preventing unnecessary read-only mode when viewing latest version runs.

The version dropdown now shows both "latest" and version numbers as separate options, making it clearer when editing the latest version versus viewing a pinned snapshot.

Closes #4121
Closes #4149 

### 1. Version Dropdown Behavior
- Open a workflow in the collaborative editor
- Click the version dropdown in the header
- Verify it shows "latest" as the first option
- Verify it also shows individual version numbers (e.g., v19, v18, v17)
- Click "latest" and verify the URL has no `?v=` parameter
- Click a specific version (e.g., v17) and verify the URL shows `?v=17`

### 2. MiniHistory and Read-Only Mode
- Open a workflow and make several edits to create multiple versions
- Open the MiniHistory panel (run history)
- Click on a run from an older version
- Verify the URL includes `?v=X` parameter
- Verify the editor shows a read-only warning
- Verify you cannot edit the workflow

### 3. Latest Version with Pinned Parameter
- Get the latest version number from the dropdown (e.g., v19)
- Manually add `?v=19` to the URL (pinning to latest version)
- Verify the editor enters read-only mode 
- Verify you cannot edit the workflow
- Click "latest" in the version dropdown
- Verify the `?v` parameter is removed from the URL
- Verify the editor becomes editable again

### 4. Clicking Latest Version Run in MiniHistory
- Create a new run (this will be at the latest version)
- Click on this run in the MiniHistory panel
- Verify the URL does NOT include a `?v=` parameter
- Verify the editor remains editable (no read-only warning)

The key behavioral change is that we now use URL parameter presence (`?v=X`) to determine read-only state, rather than comparing workflow lock versions. This makes the behavior more predictable and prevents unnecessary read-only mode when viewing runs at the latest version.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning_web/live/run_live/show_test.exs
+++ b/test/lightning_web/live/run_live/show_test.exs
@@ -232,6 +232,70 @@ defmodule LightningWeb.RunLive.ShowTest do
     |> render_async()
   end
 
+  describe "workflow link" do
+    setup :register_and_log_in_user
+    setup :create_project_for_current_user
+
+    test "excludes version param when run snapshot matches current workflow", %{
+      conn: conn,
+      project: project
+    } do
+      workflow = insert(:simple_workflow, project: project) |> with_snapshot()
+      %{triggers: [%{id: webhook_trigger_id}]} = workflow
+
+      # Post to webhook to create a run
+      assert %{"work_order_id" => wo_id} =
+               post(conn, "/i/#{webhook_trigger_id}", %{"x" => 1})
+               |> json_response(200)
+
+      %{runs: [%{id: run_id}]} = WorkOrders.get(wo_id, include: [:runs])
+
+      {:ok, view, _html} = live(conn, ~p"/projects/#{project.id}/runs/#{run_id}")
+
+      html = view |> element("#run-detail-#{run_id}") |> render_async()
+
+      # Find the workflow link - should NOT include version param
+      assert html =~
+               ~r/href="\/projects\/#{project.id}\/w\/#{workflow.id}\?a=#{run_id}"/
+
+      refute html =~ ~r/&v=/
+    end
+
+    test "includes version param when run snapshot differs from current workflow",
+         %{
+           conn: conn,
+           project: project
+         } do
+      # Create workflow with initial snapshot
+      workflow =
+        insert(:simple_workflow, project: project, lock_version: 1)
+        |> with_snapshot()
+
+      %{triggers: [%{id: webhook_trigger_id}]} = workflow
+
+      # Post to webhook to create a run on v1
+      assert %{"work_order_id" => wo_id} =
+               post(conn, "/i/#{webhook_trigger_id}", %{"x" => 1})
+               |> json_response(200)
+
+      %{runs: [%{id: run_id}], snapshot: snapshot} =
+        WorkOrders.get(wo_id, include: [:runs, :snapshot])
+
+      # Update workflow to v2 (simulating a change after the run was created)
+      workflow =
+        Lightning.Repo.update!(Ecto.Changeset.change(workflow, lock_version: 2))
+
+      {:ok, view, _html} = live(conn, ~p"/projects/#{project.id}/runs/#{run_id}")
+
+      html = view |> element("#run-detail-#{run_id}") |> render_async()
+
+      # Find the workflow link - should include version param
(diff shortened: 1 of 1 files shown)


midigofrank:
@stuartc @taylordowns2000 while working on this, I learnt that opening the collab editor with  `?v=10` where 10 is the latest version would start a different Y.Doc instance from when you opening without the `v` param. That made me so happy because it means this solution fits right in 👑 

@taylordowns2000 I have updated the read-only tooltips to  "You are viewing a pinned version of this workflow" instead of "You cannot edit because this an old snapshot.."

taylordowns2000 (approved):
awesome. thank you @midigofrank

lmac-1:
I really love how this has been simplified! The logic using URL parameters makes so much more sense. Great work, @midigofrank ☺️

One thing for future consideration (definitely not for this PR and I may have missed context from earlier conversations): the dropdown UX where both "latest" and "v19" appear when v19 is the latest version. Since they represent the same content but have different behaviours (editable vs read-only), it might be a bit confusing to users. 

Maybe "latest" could be a separate button outside of version dropdown, or the dropdown could use visual indicators to show which options are editable vs read-only (although that might be a bit cluttered). Not sure what the best solution would be yet... one to discuss and flag for a rainy day perhaps @taylordowns2000 @theroinaochieng .

lmac-1:
@midigofrank I also noticed this, but I think this should be a follow up issue https://www.loom.com/share/09bbafb4587d4aa386f9d13d766c2a51

There's some fishy behaviour after refreshing if two runs within the same workorder have different versions. It doesn't seem to be showing the correct version number. Stu was noting some observations about this on [this issue](https://github.com/OpenFn/lightning/issues/4121)

lmac-1:
So to be clear, I think we only need action on the stuff mentioned in "Some feedback" before merging:
1. ~Updating the tooltip message~ 
2. Confirming expected behaviour for when you do a run and come back to the canvas with latest and fixing if necessary

Follow up issues: 
1. Handling invalid version number in url like v=9999 (low priority, won't happen often)
3. Handling multiple runs in the same work order with different versions correctly (higher priority but I do think it should be separate from this PR)

cc @theroinaochieng @taylordowns2000

taylordowns2000:
@lmac-1 , confirmed that theres one missing link - the run detail page

image

lmac-1 (approved):

PR #4174, opened 2025-12-10
Support batched logs from the worker

### LINKED ISSUE #4123: Support batching of worker log events
Right now the worker sends log events one at a time - which can result in very high volume of traffic during intensive logging. And this can block the websocket and introduce latency across all events.

One solution we want to explore is allowing multiple log events to be sent in a single `worker:log` message. So you get fewer messages but with larger payloads (there may be some problems with this)

Spec to follow. It should be pretty clean from the Lightning side - the complexity is in the worker

See the main issue in kit https://github.com/OpenFn/kit/issues/790

josephjclark:
## Background

The production worker now supports batch logging (optionally).

What this means in short that instead of sending each log line straight to the `run:socket` channel, it'll save up a bunch of them and send a batch of logs in one go.

This should reduce latency on log processing in the worker, and may eliminate even more lost runs.

## Overview

This is a very simple, minimal PR which supports log events being set from the worker in batches.

It is designed to be backward compatible with batched and un-batched logs.

So basically the behaviour here is is simply:
* Always handle log events as an array of logs (under the `logs` key)
* If receiving a "legacy" event, without a `logs`, key, massage the payload into an array of 1 log
* Process the logs one at a time

This PR does not do any batch uploading in the database. It probably makes sense to do this? But it feels beyond the scope of what I ought to be doing (with or without Claude's help)

I have not touched tests, but we should introduce a few tests against batched logs.

Closes #4123

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI


josephjclark:
Hey @stuartc - this is all I've done on the lightning so far for batch logging. Just enough for local testing really.

There's a lot of work we _could_ do here to improve. I don't know how much appetite we have?

josephjclark:
Hi @midigofrank - as Stu is going to be tied up until the end of the year, can you help me with this? CC @theroinaochieng

midigofrank:
Hey @josephjclark , I'll have a look

midigofrank:
@josephjclark I have named the event `run:batch_logs`. Let me know your thoughts

josephjclark:
Ok @midigofrank I've just updated this on the Worker side - works great with the new event

I think we can merge this guy down now.

Issue #4154, opened 2025-12-08
Job updates not persisted after saving and reopening workflow


martalovescoffee:
**When editing an existing job in a workflow, changes are not persisted after saving and reopening the workflow. The initial creation of the job (with multiple steps) saves correctly and the job can be run successfully. However, after adding an additional step to the same job and clicking Save, reopening the workflow shows the very first version of the job without the newly added step. This gives the impression that the job was saved, but the latest changes are silently discarded.** A clear and concise description of what the bug is.

**v2.15.0-pre4** The version of Lightning you have seen the bug on.

I have reproduced this locally on main:

- [x] Yes
- [ ] No

**To Reproduce** Steps to reproduce the behavior:

Here’s a clean version you can copy into the ticket:

Reproduction steps:

1. Create a new workflow.
2. Inside that workflow, create a new job.
3. Add multiple steps to the job (a “huge” job). **And rename the new job**
5. Click **Save** on the job.
6. Run the job and wait for the run to complete.
7. Go back to the workflow view.
8. Confirm that the job appears in the workflow.
9. Open the same job again and add one more step.
10. Click **Save** on the job.
11. Go back to the workflow list.
12. Open the same workflow again.Expected behavior:
The job (with the newly added step) is saved and shown in the workflow.

Actual behavior:
The latest changes to the job (the additional step) are not saved; the workflow shows the very first version of the job.

Loom: https://www.loom.com/share/d36a6f0bc0f94f16b68939ff73c654e6

doc-han:
Found a more reliable way to reproduce this issue. 
1. After creating a new node, rename that same node before hitting save. 
2. Now go back to the workflows page and back into the workflow

doc-han:
The issue here is very flaky. 

### Current finding
When a user visits a workflow, the state of the workflow(`w1`) in jobs DB and ydoc document states (which is derived from the DB initially) end up being the same. 

**After some changes.**
jobs DB = `w1`
ydoc = `w1` + `some changes`

The changes in ydocs have two different places they could  be, 
- In memory: new changes are kept in memory with a debounce time to gather more changes before performing a transaction to persist them to the db.
- In the DB: changes that have been persisted from In memory end up in the DB after some time of inactivity(no new changes coming in)

**On Save**
- Because our changes go to ydoc and not the db, ydoc is up to date during save. Just that most of the changes might still be in-memory and not persisted to the document states DB yet
[...]

PR #4175, opened 2025-12-10
fix: infinite loading when reopening workflow in collaborative editor

### LINKED ISSUE #4164: Infinite loading when opening workflow "Any workflow 1"
**Describe the bug** When opening workflows "Any workflow 1" under the QA project with fast workers, the workflow editor gets stuck in an infinite loading state.

**Version number** v2.15.0-pre4

- [ ] Yes
- [ x] No

1. Go to  project '[qa-project-with-fast-workers](https://app.staging.openfn.org/projects/7ca57ff5-df60-4106-bc7c-ae5ea866c028/w)'
2. Click on 'Any workflow 1'
3. See the error; 

Additionally in the video I have created a new workflow from the same template which I did before. Not sure it's the template problem tho. 

**Expected behavior** Workflow should open normally

Loom: https://www.loom.com/share/4751e050c4de45ea8413dfec691cf843

elias-ba:
## Description

This PR **fixes** the infinite loading issue when reopening a workflow in the collaborative editor.

**1. `extract_lock_version` not handling `nil` values (`persistence.ex`)**

When a workflow is opened before first save, the Y.Doc gets `lock_version: nil`. The `extract_lock_version/1` function didn't handle the `{:ok, nil}` case from `Yex.Map.fetch/2`, causing it to fall through incorrectly. The `stale?/2` function was also refactored for clarity.

**2. `merge_updates` applying deltas to empty doc (`persistence_writer.ex`)**

This was the root cause of the infinite loading. When the `PersistenceWriter` batches multiple Yjs updates together, `merge_updates/1` was applying them to an empty `Yex.Doc.new()`. Since Yjs deltas only contain changes (not full state), applying them to an empty doc produced corrupted 2-byte outputs, losing all workflow data.

**Evidence from reproduction:**
- Before fix: merged updates = 2 bytes (corrupted/empty)
- After fix: merged updates = 969-1742 bytes (correct full state)

The fix loads the existing persisted state (checkpoint + updates) from the database before applying new deltas on top.

### How this caused the `nil` lock_version crash

1. **User creates new workflow** → Y.Doc initialized with `lock_version: nil`
2. **After 2s debounce** → First batch saved with full state (including `lock_version: nil`) ~1000 bytes ✓
3. **User saves workflow** → DB gets real `lock_version`, Y.Doc updated via `merge_saved_workflow_into_ydoc`
4. **This generates a delta update** → Only contains the `lock_version` change
5. **After 2s debounce** → `merge_updates` merges pending deltas
6. **BUG**: Deltas applied to empty doc → Produces 2-byte corrupted output, **lock_version update lost**
7. **User reopens workflow** → Loads first update (`lock_version: nil`) + corrupted update (2 bytes)
8. **Result**: Y.Doc has `lock_version: nil`, DB has real value → stale check fails → crash

Closes #4164 #4154

## Validation steps

1. Create a new workflow in collab editor and save it
2. Add a job node (but don't save)
3. Navigate back to workflows list
4. Open the workflow again
5. Verify workflow loads correctly (no infinite loading)

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to
 - Fix incorrect field access and version type in checkpoint creation that would
   crash after 500 document updates
   [#4176](https://github.com/OpenFn/lightning/issues/4176)
+- Fix issue where users were not able to go to the latest versin of a workflow
+  [#4149](https://github.com/OpenFn/lightning/issues/4149)
 - Fix AI Assistant disclaimer not persisting after acceptance
   [#4158](https://github.com/OpenFn/lightning/issues/4158)
 - Credential form now shows inline validation errors for JSON schema and raw


lib/lightning/collaboration/persistence_writer.ex
```diff
@@ -355,13 +357,19 @@ defmodule Lightning.Collaboration.PersistenceWriter do
       {:error, exception}
   end
 
-  defp merge_updates([update]), do: update
+  defp merge_updates([update], _document_name), do: update
```
doc-han, line 360: The fix for my issue #4154  overlaps with what you're doing here. I want to get rid of this function all together. If there's a single update **doesn't mean that**

1. we don't want to generate a delta between the current state of the document(`_document_name`)
2. also, properly encode the changes using `Yex.encode_state_as_update!`

with 2. I don't think that just returning the single update is the same as passing it through a new doc and encoding.

doc-han: On my end, I've consistently reproduced the error whenever there's a single update coming through ydoc which leads to this function. checking whether I can reproduce on this branch.

doc-han: Awesome. I think this PR fixes #4154 also. just revalidated whether a single update coming through causes an issue and it actually doesn't. Hence the issue with #4154 also had to do with the corrupted bytes. Took me a long route to get to `merge_updates` function.

taylordowns2000 (commented):
@elias-ba , sadly, i'm not able to reproduce the bug on `main` here so i cant approve this one either. when you're online this morning maybe you can help?

PR #4177, opened 2025-12-10
Fix keychain credential `:external_id` bug

### LINKED ISSUE #4170: Credentials can have the same external id
External ID is unique per project - it's how keychain credentials decide how credentials are selected. It looks like this is broken and you can have two creds with the same external id in the same project.

Here's how it works in v1:
```ex
  @spec validate_external_id(Ecto.Changeset.t(), String.t()) :: Ecto.Changeset.t()
  def validate_external_id(changeset = %Ecto.Changeset{}, slug) do
    staged_map = %{
      slug: slug,
      external_id: get_field(changeset, :external_id),
      project_ids:
        get_field(changeset, :project_credentials)
        |> Enum.map(fn pc -> pc.project_id end)
    }

    if CredentialService.valid_external_id(staged_map),
      do: changeset,
      else:
        add_error(
          changeset,
          :external_id,
          "Invalid external id. There's already a credential with the same external ID."
        )
  end
```

taylordowns2000:
This PR fixes #4170 by adding validation on credential create or update (including project_credential addition) that prevents two different credentials that are both shared in the same project from having the same external ID.

The implementation was based on the Keychain Credential feature from OpenFn v1.

## Validation steps

1. Attempt to create a credential with the same external ID as another in the same project
2. Attempt to update a credential to use the same external ID as another in the same project
3. Attempt to add a credential with the same external ID to a project in which there is another credential that already has that external ID

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/credentials.ex
+++ b/lib/lightning/credentials.ex
@@ -209,7 +209,7 @@ defmodule Lightning.Credentials do
       build_create_multi(changeset, credential_bodies)
       |> derive_events(changeset)
       |> Repo.transaction()
-      |> handle_transaction_result()
+      |> handle_transaction_result(changeset)
     end
   end
 
@@ -274,7 +274,7 @@ defmodule Lightning.Credentials do
       build_update_multi(credential, changeset, credential_bodies)
       |> derive_events(changeset)
       |> Repo.transaction()
-      |> handle_transaction_result()
+      |> handle_transaction_result(changeset)
     end
   end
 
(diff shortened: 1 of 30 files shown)


lib/lightning/credentials/credential.ex
```diff
@@ -50,6 +50,9 @@ defmodule Lightning.Credentials.Credential do
     |> unique_constraint([:name, :user_id],
       message: "you have another credential with the same name"
     )
+    |> unique_constraint([:external_id, :user_id],
+      message: "you have another credential with the external ID"
+    )
```
elias-ba, line 55: This constraint enforces uniqueness per-user, but the bug we're fixing requires uniqueness per-project, right ? Also, this needs a corresponding database index/migration to actually work. I scanned a little bit the migrations and the changes in this PR but I haven't seen one. Without a db migration for the constraint, this probably won't work.

lib/lightning/credentials.ex
```diff
@@ -203,7 +203,9 @@ defmodule Lightning.Credentials do
     credential_bodies = get_credential_bodies(attrs)
 
     with :ok <- validate_credential_bodies(credential_bodies, attrs),
-         changeset <- change_credential(%Credential{}, attrs) do
+         changeset <- change_credential(%Credential{}, attrs),
+         changeset <- validate_external_id_uniqueness(changeset),
```
elias-ba, line 207: This validation here runs outside the multi and transaction. This means there are high chances of a race condition. In fact, two concurrent requests could both pass validation before either inserts. This flow for example:

```
1. Request A: checks DB → no conflict
2. Request B: checks DB → no conflict  
3. Request A: inserts credential with external_id "foo"
4. Request B: inserts credential with external_id "foo" → duplicate!
```

Now this may be very rare situation but it can happen, and when it happens it will break things.

We should move this validation inside the Multi so the check and insert happen in the same transaction.

test/lightning/credentials/keychain_credential_test.exs
```diff
@@ -411,4 +411,72 @@ defmodule Lightning.Credentials.KeychainCredentialTest do
       assert loaded_keychain.default_credential.id == credential.id
     end
   end
+
+  describe "external_id uniqueness validation" do
```
elias-ba, line 415: Nice test documenting the failure mode for legacy data! But this makes me think, users in prod might already have duplicates in their database, right? They'd hit this `Ecto.MultipleResultsError` at runtime. Maybe worth a follow-up issue to add a migration or mix task that detects and reports existing duplicates so admins know to clean them up?

taylordowns2000: yeah, good call. i was planning to update known instances manually but we should include a migration here. ill do so

taylordowns2000:
this was done in #4422 , right @elias-ba ?

Issue #4139, opened 2025-12-05
Use the new channel registry for runs


taylordowns2000:

taylordowns2000:
@elias-ba and @midigofrank looked at this, but because all of the related bugs have been fixed (and this would be a very costly, destabilizing rewrite) they've decided _not_ to move forward with it right now. moving to "tech backlog"

elias-ba:
## Related: Run selection not persisting on page reload (#4188, PR #4201)

Found another race condition that the channel registry would fix properly.

**The bug:** When reloading a page with `?run=xxx` in the URL, the run visualization sometimes doesn't appear on the diagram. It's flaky - sometimes works, sometimes doesn't.

**Root cause:** `LoadingBoundary` waits for Y.Doc sync and workflow data, but NOT for history channel connection. So the diagram can render while history channel is still connecting:

1. Diagram renders → `useRunSteps(runId)` called
2. Effect has guard: `if (!isHistoryChannelConnected) return`
3. Guard bails → no subscribe, no fetch
4. Selector reads empty cache → returns null
5. Diagram renders without run data
6. Channel connects later → effect re-runs → fetch → re-render

**Current fix (PR #4201):** Server pre-populates run data via `data-initial-run-data` attribute so cache has data before React mounts. Works, but leans back toward LiveView patterns.

**How channel registry fixes this properly:** If the registry ensures all channels (including history) reach `active` state before `LoadingBoundary` resolves, the race disappears. Components can assume channels are ready when they render.

This fits under the "Pluggable settling" concept in this issue - history channel would need its own settling logic (maybe just successful join + initial data received).

PR #4201, opened 2025-12-15
fix: persist run selection across page reload in collaborative editor

### LINKED ISSUE #4188: Runs sometimes show up on the canvas, and sometimes don't
On `main` as of Friday, December 12th at 7:58am UTC (c88757946325eabf81307b865adf43c4bb08c24f)

https://www.loom.com/share/16f7a3db5d3f4f2fa8beb83f614c01df

Is this related to #4139 ?

elias-ba:
## Description

This PR **fixes** a flaky bug where selecting a run to view in the workflow diagram didn't persist across page reloads in the collaborative editor. Sometimes the diagram would show the run selection after reload, sometimes it wouldn't.

This was caused by a race condition where React renders before the Phoenix channel connects, causing a brief flash of empty state. In development, this is worse due to React StrictMode's double-mount cycle (mount → unmount → mount) which also clears the run steps cache between mounts.

To fix this, we now pre-populate run data from server in the `collaborate.ex` liveview specifically in the `handle_params/3` (making data available before React mounts) and we also now never clear cache on unsubscribe (this prevents the StrictMode race condition in dev).

Closes #4188

## Validation steps

1. Open a workflow in the collaborative editor
2. Select a run from the history panel to view it on the diagram
3. Refresh the page (Cmd+R)
4. Verify the run selection persists and the diagram shows the run visualization immediately without flashing
5. Repeat 3 and 4 very frequently to try to break it

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


elias-ba:
Hey @midigofrank, so I dug into both things you mentioned:

**On duplicate fetching** - there's none. The `subscribeToRunSteps` function checks `!runStepsCache[runId]` before triggering a fetch. If the server already pre-populated the cache, that check returns false and `requestRunSteps` never gets called. Pretty straightforward deduplication.

**On the actual race condition** - it's not a ReactFlow thing at all. It's a timing issue between when React renders and when the history channel connects. Here's what happens: `LoadingBoundary` waits for Y.Doc to sync and for the workflow to be non-null, then it lets children render. But it doesn't wait for `isHistoryChannelConnected`. So the diagram can render while the history channel is still connecting. Inside `useRunSteps`, there's a guard:
```typescript
if (!runId || !isHistoryChannelConnected) {
  return; // early exit, no subscribe, no fetch
}
```
So the diagram renders, `useRunSteps` fires, but the guard bails early because channel isn't connected yet. The selector reads from an empty cache, returns null, and the diagram renders without run data.

The server pre-population fixes this by putting data in the cache *before* React even mounts. So when that selector runs, there's already data there, no waiting for channel.

There's also a secondary issue with React StrictMode in dev doing its mount-unmount-mount dance. The old code cleared the cache on unsubscribe, so even if data was fetched on first mount, it got wiped before second mount. Now we just leave the cache alone on unsubscribe.

**Re: Channel Registry** - yeah, if it ensures the history channel is connected before `LoadingBoundary` resolves, that would fix the root cause. The race exists because LoadingBoundary doesn't gate on history channel readiness.

midigofrank:
Aaah, great sleuthing @elias-ba . Makes a lot of sense

midigofrank (approved):

doc-han:
> I'm a little worried with the approach here it's taking us away from the stores. Is this the only way out of this?


I think I share the same sentiment as this line from @midigofrank. we're going away from our new way of getting data through the setup channels back to liveview.

Initially, liveview was both rendering content and also piping the data from elixir to it. 
currently, we're trying to decouple liveview so that react takes charge of the rending and then we depend on websockets(phoenix-channels) for the data.

This change takes us back to the liveview ways. 

Do we want to use this approach and attempt it once again later?
cc: @elias-ba

doc-han:
Also, this works just fine. It guards from runs not being available on initial paint of the canvas. 
There're scenarios where the runs get missing when switching  between two runs instead.

elias-ba:
Hey @doc-han, you're right, this does lean back toward the LiveView pattern for this specific case. It's a pragmatic fix for now because the race condition exists at a fundamental level: `LoadingBoundary` doesn't gate on history channel readiness, only on Y.Doc sync.

The "proper" fix would be the Channel Registry ensuring all channels (including history) are connected before we let React render. Once that's in place, we can remove this server pre-population entirely since the data would flow through channels like everything else. So yeah, think of this as a stopgap.

And thanks for testing! Good to know it fixes the reload case. For the switching between runs issue, can you give me steps to reproduce? I'd like to see if it's related or a separate thing entirely.

doc-han:
> Hey @doc-han, you're right, this does lean back toward the LiveView pattern for this specific case. It's a pragmatic fix for now because the race condition exists at a fundamental level: `LoadingBoundary` doesn't gate on history channel readiness, only on Y.Doc sync.
> 
> The "proper" fix would be the Channel Registry ensuring all channels (including history) are connected before we let React render. Once that's in place, we can remove this server pre-population entirely since the data would flow through channels like everything else. So yeah, think of this as a stopgap.
> 
> And thanks for testing! Good to know it fixes the reload case. For the switching between runs issue, can you give me steps to reproduce? I'd like to see if it's related or a separate thing entirely.

Ohkay, so maybe we open an issue to be sure the registry stuff considers it.

The switching between runs is quite flaky. best way is to switch between runs with different statuses. eg. from success to crash and back to success.

**I'll approve this then**. but I think `LoadingBoundary` shouldn't wait for the channels to connect either. The channel after connecting and having the required data should be able to cause a re-render with runs. I've not looked into the cause but initially I was thinking it was the other way round. where history channel gets its data before render and canvas paints workflows without the runs.

elias-ba:
Thanks @doc-han ! Could you open an issue for the switching runs bug since you've got the repro steps? I think it's a separate thing. And it probably happens when runs are on different workflow versions, the version change triggers session recreation and there's probably a window where data disappears. Just a theory !

elias-ba:
About the re-render, it does happen. The issue is there's still that window while the fetch is in progress where the selector returns null. The server pre-population just eliminates that window entirely by having data ready from the start.

elias-ba:
@doc-han @taylordowns2000 @midigofrank I added a comment in the channel registry issue that should help do the necessary cleaning when being implemented. https://github.com/OpenFn/lightning/issues/4139#issuecomment-3656614044

PR #4219, opened 2025-12-16
fix: make template preview read-only in collaborative editor

### LINKED ISSUE #4203: Regression: Templates are editable
**Describe the bug** Users can now add nodes, edit job names and edit templates before creating a workflow

- [ ] Yes
- [ ] No

happen.

elias-ba:
## Description

This PR **fixes** the template preview read-only behavior in the collaborative editor. Users viewing a template or AI-generated workflow preview can no longer edit until they click "Create". This prevents confusing bugs where edits would be lost.

Closes #4203

## Validation steps

1. Go to a project and click "New Workflow"
2. Select a template from the template browser
3. Verify the workflow renders on canvas with the details overlay
4. Verify you **cannot**:
   - Toggle the workflow enable/disable switch
   - Click the "Code" button in the job inspector
   - Click the "Run" button in the job or trigger inspector
   - Add new nodes to the canvas
5. Click "Create" to save the workflow
6. Verify all editing features are now enabled
7. Do the same for AI generated templates and for the YAML imported workflows too

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ and this project adheres to
   [#4069](https://github.com/OpenFn/lightning/issues/4069)
 - "Try the new editor" wasn't persisting properly to user preferences
   [#4199](https://github.com/OpenFn/lightning/issues/4199)
+- Make template preview read-only in collaborative editor until workflow is
+  created [#4203](https://github.com/OpenFn/lightning/issues/4203)
 
 ## [2.15.0-pre5] - 2025-12-13 🎂 ❤️ Happy Birthday, Mom!
 
--- a/assets/test/collaborative-editor/components/NewRunButton.test.tsx
+++ b/assets/test/collaborative-editor/components/NewRunButton.test.tsx
@@ -1,0 +1,100 @@
+/**
+ * NewRunButton Component Tests
+ *
+ * Tests for the NewRunButton component focusing on the disabled prop behavior.
+ * Verifies that the button correctly respects both the disabled prop and useCanRun hook.
+ */
+
+import { render, screen } from '@testing-library/react';
+import { describe, expect, test, vi } from 'vitest';
+
+import { NewRunButton } from '../../../js/collaborative-editor/components/NewRunButton';
+
+// Mock useCanRun hook
+vi.mock('../../../js/collaborative-editor/hooks/useWorkflow', () => ({
+  useCanRun: () => ({
+    canRun: true,
+    tooltipMessage: 'Run workflow',
+  }),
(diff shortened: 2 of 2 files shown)


assets/js/collaborative-editor/components/NewRunButton.tsx
```diff
@@ -29,8 +29,8 @@ export function NewRunButton({
 }: NewRunButtonProps) {
   const { canRun, tooltipMessage } = useCanRun();
 
-  // Allow parent to add additional disabled conditions
-  const isDisabled = disabledProp ?? !canRun;
+  // Disable if parent requests OR if canRun is false
+  const isDisabled = disabledProp || !canRun;
```
doc-han, line 33: I think `??` was used so that when `disabledProps` is `false`, that is treated as an expected value and we wouldn't fallback to `canRun`

doc-han (approved):

taylordowns2000 (approved):
fantastic. thanks @elias-ba

theroinaochieng:
Works great. Thanks @elias-ba

taylordowns2000 (changes requested):
@elias-ba , please update the changelog then merge

assets/js/collaborative-editor/components/NewRunButton.tsx
elias-ba: Hey dude, good catch on the intent there! You're right that `??` was probably deliberate. It lets `disabled={false}` explicitly override the `canRun` check.

The issue we ran into was with template preview mode. When a user is viewing a template before clicking "Create", we pass `disabled={isReadOnly}` to prevent running. But `isReadOnly` evaluates to `false` for normal workflows, and with `??`, that `false` was being treated as "explicitly enabled", which meant it completely bypassed the `canRun` check.

So a user could have a workflow that wasn't saved yet (`canRun: false`), but because `isReadOnly` was `false`, the button would still be enabled.

Switching to `||` means the button is disabled if *either* condition is true. The parent says so OR the workflow can't be run. This felt like the safer default for a "Run" button.

That said, if there's a use case where we need `disabled={false}` to force-enable the button regardless of `canRun`, we could make this more explicit, maybe separate props like `forceEnabled` or rename things to be clearer.

What do you think?

doc-han: Alright. makes sense. so if I'm getting it right. disabled should always depend on both conditions and never just one of them.

PR #4249, opened 2025-12-17
feat: make AI chat sessions collaborative

### LINKED ISSUE #4247: Better handling for multi-user chats
## Background

1. Chats are scoped by project, any user with <privs> can send messages in a session.
2. Right now, users can't see which messages _they sent_ and which were sent by other users -> bring back the indicator (small, subtle text, not big icon.)
3. Right now, users don't see messages sent by other users in real time, they only see responses -> ensure that messages sent are broadcast so they appear in all users interface
4. Right now, users don't see "message processing" icon and are able to send ANOTHER message while the first message is pending -> this is a no-no, you cannot send multiple messages on top of one another (regardless of which user sent the message)

## Acceptance criteria

- [ ] broadcast messages as they are sent so they appear for all users in that chat session
- [ ] add user indicator to sent messages (design coming)
- [ ] ensure that "loading" `...` appears for all users in that chat session when waiting on a reply
- [ ] ensure that sending a new message is blocked for all users in that chat session when waiting on a reply

elias-ba:
## Description

This PR makes AI chat sessions collaborative, allowing multiple users to participate in the same session in real-time.

When one user opens an AI chat session, other users with project access can now see and join the same conversation. Messages are broadcast instantly to everyone viewing the session, and each message shows who sent it via avatar tooltips. While the AI is processing a response, all users see the loading state and have their input blocked to prevent conflicting messages.

This PR also fixes several bugs: duplicate message content no longer returns the wrong message ID, switching between job and workflow chat modes no longer causes session type mismatches, and multiple users applying AI-generated workflows simultaneously no longer creates duplicate nodes on the canvas.

Closes #4247

## Validation steps

1. Make sure you have two user sessions. The easiest way is to log in as super@openfn.org in one browser window and as demo@openfn.org in another browser window. Consider using an incognito window.
2. Make sure the two users share the same project
3. Let's call your two users User A and User B
4. With User A, create a new workflow and save it
5. With User B, access the same workflow
6. Create a chat session with User A
7. Open the AI Assistant with User B and see the chat session created by User A
8. With User B, open the session
9. Note how User B sees all the messages
10. With User A, send a new message to the AI
11. See how User B sees the message sent, the user attribution under the message (Sent by User A), and the AI Assistant's response loader
12. Note how both chat sessions are disabled while waiting for the AI's response
13. Note how when the AI response arrives, it's broadcast to all users
14. Do more of this and try different combinations
15. Make the AI Assistant generate workflow code
16. Apply the workflow using one of the users
17. See the workflow changes applied correctly
18. Try this same experience in the Job chat too
19. Switch sessions, switch jobs, do different combinations
20. Also try this experience in the legacy AI Assistant

If you'd rather skip all this, see my quick demo here: https://www.loom.com/share/aca906a6e1824936a8c5f4cd3aeeb1e4

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -37,7 +37,7 @@ import {
 } from '../hooks/useWorkflow';
 import { useKeyboardShortcut } from '../keyboard';
 import { notifications } from '../lib/notifications';
-import type { JobCodeContext, SessionSummary } from '../types/ai-assistant';
+import type { JobCodeContext } from '../types/ai-assistant';
 import { Z_INDEX } from '../utils/constants';
 import {
   prepareWorkflowForSerialization,
@@ -371,40 +371,8 @@ export function AIAssistantPanelWrapper() {
     });
   }, [updateSearchParams, aiStore, aiMode]);
 
-  // Listen for new AI sessions created by other users (via workflow channel broadcast)
-  useEffect(() => {
-    const handleAISessionCreated = (event: CustomEvent<SessionSummary>) => {
-      const session = event.detail;
-
-      // Only add if we're viewing the session list (no active session)
-      // and the session matches the current context
-      const currentState = aiStore.getSnapshot();
-
-      if (!currentState.sessionId && session) {
-        // Check if session matches current mode/context
-        const matchesContext =
-          (currentState.sessionType === 'job_code' &&
-            session.session_type === 'job_code') ||
-          (currentState.sessionType === 'workflow_template' &&
-            session.session_type === 'workflow_template');
-
-        if (matchesContext) {
-          aiStore._prependSession(session);
-        }
-      }
-    };
-
-    document.addEventListener(
-      'ai_session_created',
-      handleAISessionCreated as EventListener
-    );
-    return () => {
-      document.removeEventListener(
-        'ai_session_created',
-        handleAISessionCreated as EventListener
-      );
-    };
-  }, [aiStore]);
+  // Note: AI session creation events are now handled by AIAssistantStore._connectChannel
+  // which receives events directly from the workflow channel
 
   const sendMessage = useCallback(
     (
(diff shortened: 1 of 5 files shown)


elias-ba:
Thanks for catching these @taylordowns2000 ! Both are now fixed:

1. When someone creates a new AI session, it now broadcasts to the workflow channel so other users see it pop into their session list. Actually this was implemented initially but I had to chase down a bug where we were broadcasting to the wrong channel topic (workflow: instead of workflow:collaborate:), but it should be working now.
2. When you click retry, we now broadcast the status change so the button disappears for everyone viewing the session, not just the person who clicked it.

taylordowns2000 (approved):
@elias-ba , this looks great (and i think we should merge) but i noticed that if i shut down  my server after sending a message, then restarted, the session (for both users) gets into a permanently-stuck state.

there's no way to recover. it shows the `...` waiting forever, but never lets me cancel or send another message. i have to create a new session. should we open this as a new bug?

elias-ba:
@taylordowns2000 that's a great catch. Not likely to happen in prod. But I would love to investigate it and (1) either fix (2) or open a follow up issue. Let me time box it !

elias-ba:
Hey @taylordowns2000, so I looked at this bug and it's super interesting!

Actually this is a bug as you mentioned, but the solution is not super trivial because we can't simply "auto-fail messages stuck in processing on join" - in collaborative sessions, another user might have legitimately sent a message that's actively being processed by the AI. We need a way to distinguish between:
- A legitimately processing message (AI is working on it right now)
- A stuck/orphaned message (server crashed, no one's working on it)

These are the options we have:
1. **Time-based recovery** - Only fail messages that have been in `:processing` for > 5 minutes (check timestamp)
2. **Verify Oban job exists** - On session load, check if the processing job is actually in the Oban queue
3. **Client-side timeout** - After X minutes of processing, show a warning with retry/cancel options

I've opened #4260 to track this and I think it's better we work on it separately. This is a very rare scenario anyway (requires server restart at the exact moment of AI processing) and wouldn't happen easily in prod under normal conditions, so it shouldn't delay the merge of this PR.

taylordowns2000:
All makes sense. Thanks @elias-ba . Please merge once you get a code review on this.

doc-han (approved):

Issue #4241, opened 2025-12-17
Maybe remove snapshot_id from workorders


taylordowns2000:
Following on #4186 , we might want to remove `snapshot_id` from workorders. The big question is this: **Why is it there?**

## The Docs
Work Orders (Key Concept): https://docs.openfn.org/documentation/get-started/terminology#work-order
Work Order Status: https://docs.openfn.org/documentation/monitor-history/status-codes#work-order-status

## The canonical OpenFn user experience is this

1. create "generate patient ID" workflow (wf1:v1)
2. receive work order (wo1) to "generate patient ID" for "Stu"
3. run (r1) for wo1 fails because of a syntax error in step 3
4. fix the syntax error (now we are on wf1:v2)
5. run (r2) for wo1 succeeds
6. receive work order (wo2) to "generate patient ID" for "Aleksa"
7. run (r3) for wo2 succeeds

###  Now you have:

- 1 workflow (with 2 versions),
- 2 work orders in a "success" state (you have generated patient IDs for 2 citizens),
- and you have 3 runs (2 succeeded, 1 failed).

**What's the rationale for having a snapshot_id on the work_order? If it's some sort of denormalization/performance thing, maybe we rename it to snapshot_for_initial_run to disambiguate?**

**_At one time,_** it might have been necessary to ensure that the _first run_ is executed with the appropriate snapshot. This makes good sense to me. If the work order is created at 7pm, when the workflow is on version 1 and it's number 423,973 in the queue, when the first run is finally created at 9pm, the workflow is now on version 2. We want to ensure that the initial run for the WO is executed with the version that was in place when the order to do the work (i.e., workorder) was created.

**_But @lmac-1 points out_** that the Work Order and its first Run are created synchronously in the same database transaction:

```ex
lib/lightning/work_orders.ex:97-120
def create_for(%Trigger{} = trigger, attrs, opts \\ []) do
  Multi.new()
  |> Multi.insert(:dataclip, fn _ -> ... end)
  |> Multi.insert(:workorder, fn %{dataclip: dataclip} ->
      build_for(trigger, attrs)  # ← Creates BOTH WO and Run together
    end)
  |> Repo.transaction()
end

lib/lightning/work_orders.ex:201-269
def build_for(%Trigger{} = trigger, attrs) do
  snapshot = Snapshot.get_current_for(attrs[:workflow])  # ← Fetched once
  
  build(attrs)
[...]

PR #4252, opened 2025-12-17
Fix: display correct workflow version for retried runs in history 

### LINKED ISSUE #4186: Correctly show version of run under work order with multiple runs with different versions

At the moment, if you have two runs underneath a work order with different versions, the collab editor doesn't correctly show the version number in the dropdown. 

**Tell us about your idea - describe the solution you'd like.**

When you click on a run underneath a work order, it should show the version that the run was executed on. This is currently broken after refreshing on work orders with multiple runs with different versions:

https://www.loom.com/share/09bbafb4587d4aa386f9d13d766c2a51

**Additional context**

There is some good context and discussion from @stuartc on the comments of this issue https://github.com/OpenFn/lightning/issues/4121
We might be pulling the version number from the work order rather than the run. 

Note this comment: https://github.com/OpenFn/lightning/pull/4235#issuecomment-3661343542 
This PR could also clean up the fix for this PR.

lmac-1:
## Problem

Work orders were incorrectly linked to snapshots in the history UI, showing `workorder.snapshot.lock_version` instead of `run.snapshot.lock_version`. This caused confusion when retrying runs:

**Scenario:**
1. Create a work order on workflow v1
2. Update workflow to v2
3. Retry the work order → executes with v2 (latest snapshot)
4. History panel shows v1 ❌ (wrong - original work order version)
5. Should show v2 ✅ (correct - retry execution version)

The root issue: version display was tied to the work order level, but **only runs should have versions** since each run can execute with a different snapshot (especially retries). This will be considered in #4241 .

## Solution

Move the `version` field from work order serialisation to run serialisation throughout the stack.

### Backend Changes (Elixir)
- Move version from work order to run in `workflow_channel.ex` serialisation
- Update preload: `[:snapshot, runs: :snapshot]` → includes run snapshots
- Add defensive nil check: `if(run.snapshot, do: run.snapshot.lock_version, else: 0)`
- Add clarifying comment on `workorder.snapshot_id` with future plan

### Frontend Changes (TypeScript/React)
- Move `version` from `WorkOrderSchema` to `RunSummarySchema` in Zod types
- Update `useVersionMismatch` hook to find and read `run.version` directly
- Add null/undefined guards in `CollaborativeWorkflowDiagram` for `run.version`
- Simplify test mocks with `@ts-expect-error` for partial fixtures

Closes #4186

## Validation steps

1. Create a work order
      - Create/trigger a new work order (ie trigger a run)
      - Note the version shown in version dropdown for the run in history panel (for example, v5)
2. Update workflow
      - Make changes to the workflow
      - Save -> verify that the version number in dropdown increments (for example, v6)
3. Retry the work order
      - Click retry on the original work order
      - You can retry a work order from this page in History: 
       image
      - New run executes with the latest workflow version
4. Check history panel
      - Original run shows its original version ✅ (e.g. v5)
      - Retry run shows the updated version ✅  (e.g. v6)
6. Verify version mismatch banner
      - If viewing latest workflow and selecting an older run -> banner appears
      - If viewing an old snapshot and selecting a run with matching version -> no banner
      - There should be no "flash" of the banner for situations that don't have a banner

Screenshots of different runs under the same work order with different versions for clarity:
image
image

Also, try going into existing workflows you have with multiple runs under a work order. Make sure everything is working as expected. Also feel free to test outside of these steps to make sure that things are working.

1. There is sometimes a little flash in number in the version dropdown when we switch versions. This is already existing in main and is something that we could improve in a separate PR/issue.

2. We are also looking into to remove `snapshot_id` from work orders as a 2nd phase fix for this problem as it seems like the field isn't necessary. There is more discussion and context in #4241.

## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning/work_orders.ex
```diff
@@ -532,7 +532,7 @@ defmodule Lightning.WorkOrders do
     from(wo in WorkOrder,
       where: wo.id in ^workorder_ids,
       order_by: [desc: wo.last_activity],
-      preload: [:snapshot, :runs]
+      preload: [:snapshot, runs: :snapshot]
```
taylordowns2000, line 535: I'm worried about this — why are we limiting the preload for runs here to only `:snapshot`? What else relies on this preload?

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1263,24 +1264,28 @@ defmodule LightningWeb.WorkflowChannel do
 
   defp format_work_order_for_history(wo) do
     # Preload if needed
-    wo = Repo.preload(wo, [:snapshot, :runs])
+    wo = Repo.preload(wo, runs: :snapshot)
```
taylordowns2000, line 1266: same here - we definitely don't need anything else from runs?

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1263,24 +1264,28 @@ defmodule LightningWeb.WorkflowChannel do
...
+
     %{
       id: run.id,
       state: run.state,
       error_type: run.error_type,
       started_at: run.started_at,
-      finished_at: run.finished_at
+      finished_at: run.finished_at,
+      version: if(run.snapshot, do: run.snapshot.lock_version, else: 0),
+      work_order_id: run.work_order_id
```
doc-han: I don't see us using `work_order_id` anywhere. do we have some future plans of making use of this id on runs? If not, I suggest we exclude it.

lib/lightning/work_orders.ex
doc-han: That's because approx. all the values we use from runs to be shown in the history panel aren't relationships and don't need preloading. `started_at`, `finished_at`, `error_type` - these values just sit in the table and don't need to be preloaded. currently, `snapshot` will be the only relationship we're making use of from a run.

lib/lightning_web/channels/workflow_channel.ex
doc-han: I guess so, we don't need to preload any other relationship but other values come through

doc-han (changes requested):
I think if we're not making use of `work_order_id` then let's not pass that data through until there's some need for it.

lib/lightning_web/channels/workflow_channel.ex
taylordowns2000: @doc-han , is it used for these bits? (or does that somehow come from another part of the app?)

`b2fc...` is the work order

image
image

doc-han: nope.
- Image 1 comes from a different data source
- Image 2 is from run history but the `workorder_id` already exists in the data being passed to the front.

Our history data structure is
```ts
WorkOrder = {
	id: ...
	runs: Array<runs>
}
```

we get the id form Image 2 from the workorder itself and not the runs in it.

lib/lightning/work_orders.ex
lmac-1: Yep @doc-han that's right. To address your concern @taylordowns2000, we aren't limiting anything in the preload from before, we're adding `:snapshot` to the runs association. 

**Before**: 
`preload: [:snapshot, :runs]`: preloads work order snapshots and runs (with only their direct table fields)

**After**: 
`preload: [:snapshot, runs: :snapshot]`: preloads work order snapshots, runs, and each run's associated snapshot

We need `run.snapshot.lock_version` to fix the bug where retried runs showed the wrong version. 

Side note: We're preloading work order snapshots (the first `:snapshot`) but we I don't think we  actually use them in the history serialisation. We could remove it, but since we're discussing deleting the work order snapshot field entirely in #4241 , I'll add a note about that cleanup there.

lib/lightning_web/channels/workflow_channel.ex
lmac-1: We've gone from `[:snapshot, :runs]` to `[runs: :snapshot]`, which changes what gets preloaded: 

**Before**: `Repo.preload(wo, [:snapshot, :runs])`:
- Work order snapshot
- Runs (with direct fields only - no associations)

**After:** `Repo.preload(wo, runs: :snapshot)`:
- Runs (with direct fields: id, state, error_type, started_at, finished_at, work_order_id, etc.)
- Each run's snapshot association (so we can access run.snapshot.lock_version)

This removes the work order snapshot preload (which we don't use) and adds the run snapshot preload (which we need for the version field). 

@doc-han @elias-ba can you confirm this is correct? I'm still learning Ecto preloading semantics and want to make sure I understand it properly.

doc-han: correct.

lib/lightning_web/channels/workflow_channel.ex
lmac-1: Thanks @doc-han for explaining. I've verified what you have said too, and it's not needed. Will remove!

lmac-1: Done! Requesting re-review, please @doc-han

assets/test/collaborative-editor/stores/createHistoryStore.test.ts
```diff
@@ -214,14 +214,15 @@ describe('createHistoryStore', () => {
...
-          version: 29,
           runs: [
             {
               id: 'f3218e57-df40-4a41-b22c-dcdfdf94650e',
               state: 'success',
               error_type: null,
               started_at: '2025-10-23T20:59:58.293382Z',
               finished_at: '2025-10-23T21:00:02.293382Z',
+              version: 29,
+              work_order_id: 'e2107d46-cf29-4930-b11b-cbcfcf83549d',
```
doc-han: residue

assets/test/collaborative-editor/stores/createHistoryStore.test.ts
```diff
@@ -259,14 +260,15 @@ describe('createHistoryStore', () => {
...
-          version: 29,
           runs: [
             {
               id: runId,
               state: 'success',
               error_type: null,
               started_at: '2025-10-23T20:59:58.293382Z',
               finished_at: '2025-10-23T21:00:02.293382Z',
+              version: 29,
+              work_order_id: 'e2107d46-cf29-4930-b11b-cbcfcf83549d',
```
doc-han: residue

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1246,15 +1246,16 @@ defmodule LightningWeb.WorkflowChannel do
...
           Enum.map(worder.runs, fn run ->
             %{
               id: run.id,
               state: run.state,
               error_type: run.error_type,
               started_at: run.started_at,
-              finished_at: run.finished_at
+              finished_at: run.finished_at,
+              version: run.snapshot.lock_version,
+              work_order_id: run.work_order_id
```
doc-han: residue

lmac-1: Done

assets/test/collaborative-editor/stores/createHistoryStore.test.ts
lmac-1: Done

assets/test/collaborative-editor/stores/createHistoryStore.test.ts
lmac-1: Done

PR #4289, opened 2026-01-06
Unsaved changes indicator

### LINKED ISSUE #3682: Implement Unsaved Changes Indicator (Red Dot) in React
### User story

As a workflow editor user, I want to see a visual indicator when I have unsaved changes so that I know my work isn't persisted yet.

### Details

Currently the unsaved changes indicator (red dot) is implemented in LiveView using Ecto changeset change tracking. This issue covers migrating the visual indicator to React while establishing proper change tracking.

**Current Implementation:**
- Indicator component: `with_changes_indicator/1` at `edit.ex:1192-1204`
- Tracking logic: Uses `@changeset.changes |> Enum.any?()`
- Two display locations: wraps workflow name/toggle in top bar (`edit.ex:166-228`) and wraps save button in inspector panel (`edit.ex:402-413`)
- Visual: Red dot (`bg-danger-500`, 12px diameter) positioned absolute top-right
- Change detection flow: form changes → `send_form_changed` → `handle_info({:form_changed, ...})` → changeset update

**Components Status:**
- Red dot indicator - **needs migration to React**
- Change tracking - **needs implementation in React state**

**Summary:**
This issue creates a reusable React component for displaying the unsaved changes indicator (red dot). The component will be a simple visual wrapper that receives a boolean prop from LiveView indicating whether there are unsaved changes. The server (LiveView) will remain authoritative for change tracking via Ecto changesets, and React will purely handle the visual presentation.

**Depends on:** #3681 (TypeScript types)

### Implementation notes

1. Create `UnsavedChangesIndicator.tsx` component
2. Add `hasUnsavedChanges: boolean` prop passed from LiveView
3. Server maintains authoritative dirty state via changeset
4. Component renders red dot when `hasUnsavedChanges === true`
5. Match visual design: 12px red circle, absolute positioned top-right with -4px offset
6. Use data attribute `data-is-dirty="true"` for testing

**Change Detection Options:**
- **Option A (Recommended):** Server-side tracking - LiveView sends `hasUnsavedChanges` boolean based on changeset
- **Option B:** Client-side tracking - React maintains dirty state based on form edits (more complex)

**Visual Requirements:**
- Red circle: `w-3 h-3` (12px)
- Background: `bg-danger-500`
- Position: `absolute -m-1 top-0 right-0 z-10`
- Rounded: `rounded-full`

### Release notes

- Unsaved changes indicator (red dot) migrated to React component

### User acceptance criteria

- [ ] Red dot appears when workflow has unsaved changes
- [ ] Red dot disappears after successful save
- [ ] Indicator shows in both locations (top bar and inspector panel)
- [ ] Visual design matches existing implementation
- [ ] Indicator appears/disappears without page reload
- [ ] Data attribute `data-is-dirty` reflects state correctly

doc-han:
## Description

This PR shows a red dot indicator on the save button when there are unsaved changes in a workflow.
Screenshot 2026-01-08 at 8 38 42 AM

Closes #3682

## Validation steps

1. Open a workflow
2. Make some changes to a job/edge
3. Do you see a red dot showing that some unsaved changes exist? we expect a yes

If you're seeing a red dot right upon entering a workflow, it means there already exist unsaved changes(between what's in the DB and what's on ydoc).

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning_web/channels/workflow_channel_broadcast_test.exs
+++ b/test/lightning_web/channels/workflow_channel_broadcast_test.exs
@@ -1,0 +1,483 @@
+defmodule LightningWeb.WorkflowChannelBroadcastTest do
+  @moduledoc """
+  Tests for workflow_saved broadcast behavior after save and save_and_sync operations.
+
+  These tests verify that when a user saves a workflow, all other connected users
+  receive a broadcast with the updated workflow state, allowing them to update their
+  base workflow state and show they have no unsaved changes.
+  """
+  use LightningWeb.ChannelCase
+
+  import Lightning.CollaborationHelpers
+  import Lightning.Factories
+  import Mox
+
+  setup :verify_on_exit!
+
+  setup do
+    Mox.stub(Lightning.MockConfig, :check_flag?, fn
+      :require_email_verification -> true
+      _flag -> nil
+    end)
+
+    # Set global mode for the mock to allow cross-process calls
+    Mox.set_mox_global(LightningMock)
+    # Stub the broadcast calls that save_workflow makes
+    Mox.stub(LightningMock, :broadcast, fn _topic, _message -> :ok end)
+
+    user = insert(:user)
+    project = insert(:project, project_users: [%{user: user, role: :owner}])
+    workflow = insert(:workflow, project: project)
+
+    {:ok, _, socket} =
+      LightningWeb.UserSocket
+      |> socket("user_#{user.id}", %{current_user: user})
+      |> subscribe_and_join(
+        LightningWeb.WorkflowChannel,
+        "workflow:collaborate:#{workflow.id}",
+        %{"project_id" => project.id, "action" => "edit"}
+      )
+
+    on_exit(fn ->
+      ensure_doc_supervisor_stopped(socket.assigns.workflow.id)
+    end)
+
+    %{socket: socket, user: user, project: project, workflow: workflow}
+  end
+
+  describe "save_workflow broadcasts workflow_saved to other users" do
+    setup %{socket: socket, user: user, project: project, workflow: workflow} do
+      # Create a second user and socket to simulate another collaborator
+      user2 = insert(:user)
+      insert(:project_user, project: project, user: user2, role: :editor)
+
+      {:ok, _, socket2} =
+        LightningWeb.UserSocket
+        |> socket("user_#{user2.id}", %{current_user: user2})
+        |> subscribe_and_join(
+          LightningWeb.WorkflowChannel,
+          "workflow:collaborate:#{workflow.id}",
+          %{"project_id" => project.id, "action" => "edit"}
+        )
+
+      %{socket: socket, socket2: socket2, user: user, user2: user2}
+    end
+
+    test "broadcasts workflow_saved with lock_version and workflow to other users",
+         %{socket: socket, socket2: _socket2, workflow: workflow} do
+      # User 1 modifies and saves the workflow
+      session_pid = socket.assigns.session_pid
+      doc = Lightning.Collaboration.Session.get_doc(session_pid)
+      workflow_map = Yex.Doc.get_map(doc, "workflow")
+
+      Yex.Doc.transaction(doc, "test_update", fn ->
+        Yex.Map.set(workflow_map, "name", "Updated Workflow Name")
+      end)
+
+      # User 1 saves
+      ref = push(socket, "save_workflow", %{})
+
+      # User 1 gets reply
+      assert_reply ref, :ok, %{
+        saved_at: saved_at,
+        lock_version: new_lock_version,
+        workflow: reply_workflow
+      }
+
+      assert %DateTime{} = saved_at
+      assert new_lock_version == workflow.lock_version + 1
+      assert reply_workflow.id == workflow.id
+      assert reply_workflow.name == "Updated Workflow Name"
+      assert reply_workflow.lock_version == new_lock_version
+
+      # User 2 (socket2) receives broadcast with the same workflow
+      assert_broadcast "workflow_saved", %{
+        latest_snapshot_lock_version: broadcast_lock_version,
+        workflow: broadcast_workflow
+      }
+
+      assert broadcast_lock_version == new_lock_version
+      assert broadcast_workflow.id == workflow.id
+      assert broadcast_workflow.name == "Updated Workflow Name"
+      assert broadcast_workflow.lock_version == new_lock_version
+      assert broadcast_workflow.jobs == reply_workflow.jobs
+      assert broadcast_workflow.edges == reply_workflow.edges
+      assert broadcast_workflow.triggers == reply_workflow.triggers
+    end
+
+    test "broadcast includes all workflow associations (jobs, edges, triggers)",
+         %{socket: socket, socket2: _socket2} do
+      # Add a job to the workflow via Y.Doc
+      session_pid = socket.assigns.session_pid
+      doc = Lightning.Collaboration.Session.get_doc(session_pid)
+      jobs_array = Yex.Doc.get_array(doc, "jobs")
+      job_id = Ecto.UUID.generate()
+
+      job_map =
+        Yex.MapPrelim.from(%{
+          "id" => job_id,
+          "name" => "Test Job",
+          "body" => Yex.TextPrelim.from("fn(state => state)"),
+          "adaptor" => "@openfn/language-common@1.0.0",
+          "project_credential_id" => nil,
+          "keychain_credential_id" => nil
+        })
+
+      Yex.Doc.transaction(doc, "test_add_job", fn ->
+        Yex.Array.push(jobs_array, job_map)
+      end)
+
+      # Save the workflow
+      ref = push(socket, "save_workflow", %{})
+      assert_reply ref, :ok, %{workflow: reply_workflow}
+
+      # Verify reply contains the job
+      assert length(reply_workflow.jobs) == 1
+      assert Enum.any?(reply_workflow.jobs, fn job -> job.id == job_id end)
+
+      # User 2 receives broadcast with the job included
+      assert_broadcast "workflow_saved", %{
+        workflow: broadcast_workflow
+      }
+
+      assert length(broadcast_workflow.jobs) == 1
+      assert Enum.any?(broadcast_workflow.jobs, fn job -> job.id == job_id end)
+      job = Enum.find(broadcast_workflow.jobs, &(&1.id == job_id))
+      assert job.name == "Test Job"
+      assert job.adaptor == "@openfn/language-common@1.0.0"
+    end
+
+    test "broadcast allows other users to reset their unsaved changes indicator",
+         %{socket: socket, socket2: socket2, workflow: workflow} do
+      # Scenario: User 2 has local unsaved changes, User 1 saves different changes
+      # User 2 should receive the broadcast and be able to update their base state
+
+      # User 2 makes local changes (not saved)
+      session_pid2 = socket2.assigns.session_pid
+      doc2 = Lightning.Collaboration.Session.get_doc(session_pid2)
+      workflow_map2 = Yex.Doc.get_map(doc2, "workflow")
+
+      Yex.Doc.transaction(doc2, "user2_local_change", fn ->
+        Yex.Map.set(workflow_map2, "name", "User 2 Local Change")
+      end)
+
+      # User 1 makes and saves changes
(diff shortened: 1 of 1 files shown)


assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!workflow || !storeWorkflow) return { hasChanges: false };
+  return {
+    hasChanges: isDiffWorkflow(
+      transformWorkflow(workflow),
+      transformWorkflow(storeWorkflow as Workflow)
+    ),
+  };
+}
+
+// transform workflow to normalized structure for comparison
```
taylordowns2000, line 58: this bit... feels brittle. is there no way around hard-coding the structure we want here?

doc-han: This part doesn't really change the structure of a workflow. it makes sure that
1. we're picking only the necessary properties (properties that come from user input)
2. sorting arrays to be arranged in the same way for comparison

Example: when a workflow reaches the store, every job/edge has an additional key called `errors` which holds the errors for that specific job/edge. This is an example of property we wouldn't want to consider in our comparison.

assets/js/collaborative-editor/types/sessionContext.ts
```diff
@@ -102,6 +106,7 @@ export const SessionContextResponseSchema = z.object({
   workflow_template: WorkflowTemplateSchema.nullable(),
   has_read_ai_disclaimer: z.boolean(),
   limits: LimitsSchema.optional(),
+  workflow: z.any(), // to be fixed
```
lmac-1: What is this comment for? What should it be fixed to, and when? It would be better if we don't use `any()` here

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
```diff
@@ -0,0 +1,110 @@
...
+    typeof target === 'object'
+  ) {
+    const baseObj = base as Record<string, unknown>;
+    const targetObj = target as Record<string, unknown>;
+    const keys = [
+      ...new Set(Object.keys(baseObj).concat(Object.keys(targetObj))),
+    ];
+    const resp = keys.some(k => isDiffWorkflow(baseObj[k], targetObj[k]));
+    if (resp) {
+      console.log('diff:', baseObj, targetObj);
```
lmac-1: Should this console.log be removed?

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
lmac-1: Just noticed that `workflow.concurrency` and `workflow.enable_job_logs` aren't included in `transformWorkflow()`, so changes to these settings won't trigger the unsaved changes indicator. Should they? (I assume yes, since these are user-editable workflow settings) 

image

I guess the issue here is if we add new user-editable fields, we need to make sure they are manually added to this transform function, otherwise unsaved changes detection will fail silently. Not sure how we could make this more maintainable though..

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
```diff
@@ -0,0 +1,110 @@
...
+    triggers: state.triggers,
+    edges: state.edges,
+    positions: state.positions || {},
+    name: state.workflow?.name,
+  }));
+
+  if (!workflow || !storeWorkflow) return { hasChanges: false };
+  return {
+    hasChanges: isDiffWorkflow(
+      transformWorkflow(workflow),
```
lmac-1: Should this also have `as Workflow` ?

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -40,6 +40,7 @@ import { NewRunButton } from './NewRunButton';
 import { ReadOnlyWarning } from './ReadOnlyWarning';
 import { ShortcutKeys } from './ShortcutKeys';
 import { Tooltip } from './Tooltip';
+import { useUnsavedChanges } from '../hooks/useUnsavedChanges';
```
lmac-1: Low priority but there's a linting error because of the order of importing hooks, should be before `useWorkflow`

lmac-1 (changes requested):
Most of my review comments are low priority / nice-to-haves and shouldn't block merging.

But I noticed that "Allow console.log() usage" (`workflow.enable_job_logs`) and "Max concurrency" (`workflow.concurrency`) don't trigger the unsaved changes dot. This should be addressed before merging. 

image

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
doc-han: Nice catch. will deal with `workflow.concurrency` and `workflow.enable_job_logs` and request review once again. thanks.

>I guess the issue here is if we add new user-editable fields, we need to make sure they are manually added to this transform function, otherwise unsaved changes detection will fail silently. Not sure how we could make this more maintainable though..

I've thought of that but not had any good ideas so far. suggestions are welcomed

assets/js/collaborative-editor/types/sessionContext.ts
doc-han: will remove the comment. the value held by this field is currently not used or supposed to be used by anything. it's stale data holding the current state of saved workflow. used as base for comparison. I think there's no need need to strictly validate it.

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
doc-han: No need I guess. it's well inferred by the ts compiler unlike the one below which needs the type to be casted.

assets/js/collaborative-editor/types/sessionContext.ts
lmac-1: Btw I jumped straight into review comments but also wanted to say, this PR is great work 🤩. Love it!

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export function useUnsavedChanges() {
+  const { workflow } = useSessionContext();
+  const storeWorkflow = useWorkflowState(state => ({
+    jobs: state.jobs,
+    triggers: state.triggers,
+    edges: state.edges,
+    positions: state.positions || {},
+    name: state.workflow?.name,
+  }));
```
stuartc: This returns a new object every call, it looks like we will get lots of rerenders/calls to transformWorkflow.

Please verify this.

We can avoid recomputing and rendering with something like this:

```typescript
export function useUnsavedChanges() {
  const { workflow } = useSessionContext();

  // Individual selectors - stable references
  const jobs = useWorkflowState(state => state.jobs);
  const triggers = useWorkflowState(state => state.triggers);
  const edges = useWorkflowState(state => state.edges);
  const positions = useWorkflowState(state => state.positions);
  const name = useWorkflowState(state => state.workflow?.name);

  // Memoize comparison
  const hasChanges = useMemo(() => {
    if (!workflow) return false;
    const storeWorkflow = { jobs, triggers, edges, positions: positions || {}, name };
    return isDiffWorkflow(transformWorkflow(workflow), transformWorkflow(storeWorkflow));
  }, [workflow, jobs, triggers, edges, positions, name]);

  return { hasChanges };
}
```

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
stuartc: yep, either make it use our logger.debug so we can opt into it or remove it please.

assets/js/collaborative-editor/stores/createSessionContextStore.ts
```diff
@@ -293,6 +295,12 @@ export const createSessionContextStore = (
     notify('setLatestSnapshotLockVersion');
   };
 
+  const setBaseWorkflow = (workflow: unknown) => {
+    state = produce(state, draft => {
+      draft.workflow = workflow as any;
+    });
+  };
```
stuartc: Why isn't there a `notify` call here? All the other ones call notify afterwards.

I'm guessing it works because it's called right after 'setLatestSnapshotLockVersion' and it just happens to have it's effects in the state? But this seems like it could be a quiet issue waiting?

assets/js/collaborative-editor/types/sessionContext.ts
stuartc: @doc-han you mean it's not used by anything 'outside'. But it is used for comparison, thats a good enough reason to have typing on it no?

Like what is the shape of the value? Everything else has a type.

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -324,8 +327,9 @@ defmodule LightningWeb.WorkflowChannel do
          {:ok, workflow} <- Session.save_workflow(session_pid, user) do
       # Broadcast the new lock_version to all users in the channel
       # so they can update their latestSnapshotLockVersion in SessionContextStore
-      broadcast_from!(socket, "workflow_saved", %{
-        latest_snapshot_lock_version: workflow.lock_version
+      broadcast!(socket, "workflow_saved", %{
+        latest_snapshot_lock_version: workflow.lock_version,
+        workflow: workflow
```
stuartc: I'm wondering why broadcast_from! was changed to broadcast!. The main reason we use broadcast specifically is to do force us handle all the server changes atomically in the callback we get on save. 

So the 'saver' gets both the success and the new baseline workflow at the same time. We have the workflow in memory and ready to be able to render it out in the reply, and by broadcasting to
ourselves we send a second message leaving the saver (depending on latency) with a period of time where the workflow was saved but the baseline isn't updated.

It also splits responsibility between the feedback of "Saved successfully" and then later a listener does the real work.

I noted that `save_and_sync` doesn't have this change, meaning GitHub saves wouldn't update the baseline. Are we missing a test here?

### workflow_channel.ex - save_workflow handler

```elixir
{:reply, {:ok, %{
  saved_at: workflow.updated_at,
  lock_version: workflow.lock_version,
  workflow: workflow # ← Add this
}}, socket}
```

And change back to broadcast_from!

```elixir
broadcast_from!(socket, "workflow_saved", %{
  latest_snapshot_lock_version: workflow.lock_version,
  workflow: workflow
})
```

Then on the client, the save callback should call
`setBaseWorkflow(response.workflow)` directly, rather than relying o n the
broadcast listener.

This also means `save_and_sync` needs the same treatment - include workflow in
response and use the callback to update baseline.

An observation about the tests for the WorkflowChannel (not from your changes), is that we have essentially zero tests for broadcast events. Which would have caught the inconsistency with `save_and_sync`.

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
doc-han: resolved

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
doc-han: removed

assets/js/collaborative-editor/components/Header.tsx
doc-han: resolved

lib/lightning_web/channels/workflow_channel.ex
doc-han: resolved.

assets/js/collaborative-editor/stores/createSessionContextStore.ts
doc-han: resolved.

assets/js/collaborative-editor/hooks/useUnsavedChanges.ts
doc-han: resolved.

lib/lightning_web/channels/workflow_channel.ex
doc-han: I took the approach of sending it back as the socket response and then falling back to `broadcast_from!` as described in this comment.

assets/js/collaborative-editor/types/sessionContext.ts
doc-han: resolved.

doc-han:
Thanks for the reviews @lmac-1 @stuartc. All change requests resolved. Please take a look once again.

PR #4304, opened 2026-01-12
Fix: Render user messages as plain text in AI chat

### LINKED ISSUE #4288: Text & code formatting bug in AI Chat
When sending code _to_ the assistant, the interface attempts to format it as a code block 😱 but it fails - it starts formatting half-way through and it doesn't let the user scroll right-to-left

Image

@lmac-1 thinks she may know the culprit 🕵️‍♀️ - there could be some code in here that guesses ("code to markdown") when it sees code and creates this big copy/pastable div. TD thinks that (a) he doesn't like that we're guessing about when code starts and ends and (b) if this really is the proper way to do it, that we should only make this kind of guess about messages sent from the AI chat. User sent text should always be text. (Unless, of course, there's some way for the user to harness the ` or ``` characters to send as markdown.

lmac-1:
## Description

This PR fixes text rendering issues in AI assistant user messages where code and long strings were being displayed incorrectly:
- **Render user messages as plain text** instead of markdown to prevent formatting issues
- Use `whitespace-pre-wrap` to preserve user's newlines and spacing
- Add `overflowWrap: 'break-word'` to handle extremely long unbreakable strings
- Add flex constraints (`min-w-0`, `max-w-full`) to properly contain content within message bubbles

Closes #4288

### Problem 
User messages were being rendered through the markdown parser, which caused issues:

1. **Incorrect code block detection**: The markdown renderer has a heuristic (`includes('\n')`) that treats any code with newlines as a full code block
2. **Inconsistent behaviour**: Users don't write markdown consistently - they paste code casually without proper formatting
3. **Over-engineering**: Supporting markdown for user messages required complex heuristics to distinguish code blocks from prose

Example: A user typing `` `some\ncode` `` would incorrectly get the full `CodeBlock` component treatment.

### Solution: Plain text rendering

I decided to render user messages as plain text only because:
1. **Simpler and more predictable**: What users type is what they see - no surprise formatting
2. **Appropriate for use case**: Users are asking questions and pasting code snippets, not authoring rich documents
3. **Assistant messages still use markdown**: The AI-generated responses retain full markdown support with proper code blocks, which is where polished formatting matters most

## Validation steps

You should play around with the AI assistant at both workflow and job level and make sure that the chat works as expected in general (no regression). You should also test the following 2 cases:

**1. Original issue:**
1. Open the AI assistant panel in collaborative editor
2. In the message input, type a message and paste the following code: 

```
<p>We need the processBatch function here
for (const conceptName of state.missing_concepts) {
try {
console.log(<code>Exact search: ${conceptName}</code>);</p>
<pre class="relative group"><code>  // Use 'name' parameter for exact match
  const response = await http.get(`/ws/rest/v1/concept`, {
    query: { name: conceptName, v: 'full' }
  })(state);

  if (response.data &amp;&amp; response.data.results &amp;&amp; response.data.results.length &gt; 0) {
    // name parameter should return exact matches
    const exactMatch = response.data.results[0];
    console.log(`Found via exact name search: ${exactMatch.display}`);

    state.existing_omrs_concepts.push(exactMatch);
    foundViaExactSearch++;
  } else {
    console.log(`Still not found`);
  }
} catch (error) {
  console.log(`Search error: ${error.message}`);
}</code><div class="code-actions absolute top-2 right-2 flex items-center gap-1 opacity-0 group-hover:opacity-100 transition-opacity"><button type="button" class="rounded-md px-2 py-1 text-xs font-medium transition-all duration-300 ease-in-out bg-slate-300 text-white hover:bg-primary-600 hover:scale-105">COPY</button></div></pre>
```
Expected behaviour: the whole code block is plain text and wraps with the chat bubble. 
Previous behaviour: there would be some weird mid-formatting happening like in the image in #4288 

**2. Code broken mid-word**
1. Open the AI assistant
2. Paste code with long variable names or classNames, example: 

```
// Another super long line to make sure horizontal scroll works: const veryLongVariableName = { property1: "value1", property2: "value2", property3: "value3", property4: "value4", property5: "value5", property6: "value6", property7: "value7", property8: "value8" };

const multiLineObject = { extremelyLongPropertyNameThatGoesOnAndOnAndOn: "This is a very long string value that extends way beyond the normal width", anotherVeryLongPropertyNameForTesting: "Another extremely long value to ensure horizontal scrolling works properly in all scenarios" };

```

Expected behaviour: the message is plain text and wraps correctly. Try resizing the panel and make sure it doesn't break
Previous behaviour: the message would go off screen with no scroll if the chat panel was too narrow

1. To keep things simple, user messages are always plain text and wrap appropriately. 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,21 @@ and this project adheres to
 - Render AI Assistant user messages as plain text instead of markdown to prevent
   incorrect code block formatting
   [#4288](https://github.com/OpenFn/lightning/issues/4288)
+- Standardized "GitHub" brand capitalization throughout the UI
+  [#4313](https://github.com/OpenFn/lightning/issues/4313)
+- Optimized PromEx average claim duration query to use SQL aggregation instead
+  of fetching rows into memory
+  [#4317](https://github.com/OpenFn/lightning/pull/4317)
 
 ### Fixed
+
+- Add clickable GitHub Actions link to collaborative editor's Save & Sync
+  success toast [#4281](https://github.com/OpenFn/lightning/issues/4281)
+- Fixed credential auto-selection when creating a new credential from the
+  workflow editor's ConfigureAdaptorModal
+  [#4287](https://github.com/OpenFn/lightning/issues/4287)
+- Add index on `runs.finished_at` to fix UsageTracking ReportWorker timeouts
+  [#4308](https://github.com/OpenFn/lightning/pull/4308)
 
 ## [2.15.10] - 2026-01-12
 
--- a/assets/js/collaborative-editor/components/ConfigureAdaptorModal.tsx
+++ b/assets/js/collaborative-editor/components/ConfigureAdaptorModal.tsx
@@ -237,8 +237,7 @@ export function ConfigureAdaptorModal({
     else if (hasSchemaMatches) {
       setShowOtherCredentials(false);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isOpen, currentAdaptor, currentCredentialId]);
+  }, [isOpen, currentAdaptor, currentCredentialId, credentialSections]);
 
   // Get version options for current adaptor
   const versionOptions = useMemo(() => {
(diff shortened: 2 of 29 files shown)


midigofrank (approved):
Nicely done

PR #4310, opened 2026-01-13
Scrubber eats too much memory

### LINKED ISSUE #4307: Scrubber Consumes 45+ MB Memory per Run with Large Credentials
Monitoring shows Lightning.Scrubber processes consuming 45-50 MB of memory per run when credentials contain many sensitive values (~500+).

Image

Claude says:

> Root Cause
> File: lib/lightning/scrubber.ex:136-140
> 
> The cartesian_pairs/1 function creates N² combinations of all sensitive values:
> 
> elixir
> defp cartesian_pairs(items) do
>   Enum.flat_map(items, fn item ->
>     items |> Enum.map(&[item, &1])
>   end)
> end
> Example:
> 
> Credential with 549 sensitive values
> Creates 549 × 549 = 301,401 pairs
> Each pair gets base64 encoded
> Result: ~47 MB memory per Scrubber process

midigofrank:
## Description

This PR does 2 things:
- Limits the number of sensitive the credential body can have. This is configurable using `MAX_CREDENTIAL_SENSITIVE_VALUES`. By default, it's `50`
- Removes duplicate sensitive values to be scrubbed. This way, we don't unnecessary iterations

Closes #4307

## Validation steps

1. Start your server with `MAX_CREDENTIAL_SENSITIVE_VALUES=2 mix phx.server`
2. Try creating a raw json credential with multiple sensitive keys (more than 2). Please look at https://github.com/OpenFn/lightning/blob/main/lib/lightning/credentials/sensitive_values.ex#L11 to see safe keys. Safe keys should not count up

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning/credentials/credential_test.exs
+++ b/test/lightning/credentials/credential_test.exs
@@ -344,10 +344,30 @@ defmodule Lightning.Credentials.CredentialTest do
     } do
       # Mock config to limit to 5 sensitive values for easier testing
       Lightning.MockConfig
-      |> expect(:max_credential_sensitive_values, fn -> 5 end)
+      |> expect(:max_credential_sensitive_values, 2, fn -> 5 end)
+
+      # Create a body with exactly 1 safe value 5 sensitive values (at the limit)
+      acceptable_body = %{
+        "username" => "user@example.com",
+        "api_key_1" => "secret_1",
+        "api_key_2" => "secret_2",
+        "api_key_3" => "secret_3",
+        "api_key_4" => "secret_4",
+        "api_key_5" => "secret_5"
+      }
+
+      changeset =
+        CredentialBody.changeset(%CredentialBody{}, %{
+          credential_id: credential.id,
+          name: "production",
+          body: acceptable_body
+        })
+
+      assert changeset.valid?
 
       # Create a body with 6 sensitive values (exceeds limit of 5)
       large_body = %{
+        "username" => "user@example.com",
         "api_key_1" => "secret_1",
         "api_key_2" => "secret_2",
         "api_key_3" => "secret_3",
@@ -370,32 +390,6 @@ defmodule Lightning.Credentials.CredentialTest do
 
       assert error_message =~ "contains too many sensitive keys (6)"
       assert error_message =~ "Max allowed is 5"
-    end
-
-    test "allows credential body with acceptable number of sensitive values", %{
-      credential: credential
-    } do
-      # Mock config to limit to 5 sensitive values
-      Lightning.MockConfig
-      |> expect(:max_credential_sensitive_values, fn -> 5 end)
-
-      # Create a body with exactly 5 sensitive values (at the limit)
-      acceptable_body = %{
-        "username" => "user@example.com",
-        "api_key_1" => "secret_1",
-        "api_key_2" => "secret_2",
-        "api_key_3" => "secret_3",
-        "api_key_4" => "secret_4"
-      }
-
-      changeset =
-        CredentialBody.changeset(%CredentialBody{}, %{
-          credential_id: credential.id,
-          name: "production",
-          body: acceptable_body
-        })
-
-      assert changeset.valid?
     end
 
     test "validates foreign key constraint for credential_id" do


lib/lightning/config.ex
```diff
@@ -389,6 +389,12 @@ defmodule Lightning.Config do
       |> Keyword.get(:port, 2222)
     end
 
+    @impl true
+    def max_credential_sensitive_values do
+      Application.get_env(:lightning, Lightning.Scrubber, [])
+      |> Keyword.get(:max_credential_sensitive_values, 200)
```
elias-ba: Don't we want this to be 50 instead of 200 ?

lib/lightning/config/bootstrap.ex
```diff
@@ -698,6 +698,10 @@ defmodule Lightning.Config.Bootstrap do
            :broadcast_work_available,
            env!("BROADCAST_WORK_AVAILABLE", &Utils.ensure_boolean/1, true)
 
+    config :lightning, Lightning.Scrubber,
+      max_credential_sensitive_values:
+        env!("MAX_CREDENTIAL_SENSITIVE_VALUES", :integer, 50)
```
elias-ba, line 703: Should we add this to `.env.example` ? And also maybe to `DEPLOYMENT.md` ?

lib/lightning_web/live/credential_live/raw_body_component.ex
```diff
@@ -71,11 +73,42 @@ defmodule LightningWeb.CredentialLive.RawBodyComponent do
       {%{}, types}
       |> Ecto.Changeset.cast(params, [:value])
       |> maybe_add_error(valid?, error, touched)
+      |> validate_sensitive_values_count(body, touched)
```
elias-ba, line 76: This parses the body string again via `parse_for_validation/1`, but it was already parsed in `parse_body/1` above. Could we consider passing the already-parsed `body_json` to avoid double parsing ?

lib/lightning/credentials/credential_body.ex
```diff
@@ -36,7 +36,50 @@ defmodule Lightning.Credentials.CredentialBody do
...
+  ## Examples
+
+      iex> validate_sensitive_values_count(%{"password" => "secret"})
+      :ok
+
+      iex> validate_sensitive_values_count(body_with_many_secrets)
+      {:error, "contains too many sensitive keys (250). Max allowed is 200"}
+  """
+  def validate_sensitive_values_count(body) do
+    max_values = Lightning.Config.max_credential_sensitive_values()
```
elias-ba, line 58: What happens to existing credentials that already exceed this limit? If a user tries to update an unrelated field on a credential with 100+ sensitive values, will the update fail? Could we consider whether a migration or audit is needed to identify any existing credentials over the limit ?

lib/lightning/credentials/sensitive_values.ex
```diff
@@ -56,5 +56,6 @@ defmodule Lightning.Credentials.SensitiveValues do
       String.downcase(k) in @safe_keys || is_nil(v)
     end)
     |> Enum.map(fn {_k, v} -> v end)
+    |> Enum.uniq()
```
elias-ba, line 59: Good optimization. Just noting this subtly changes the return value. Previously duplicate values were included multiple times. Should be fine but worth a sanity check that no code path relied on the duplicate count.

elias-ba (changes requested):
Hey @midigofrank this is really a good solution for a hard problem. I have few observations and questions. Could you check them and let me know what you think ?

lib/lightning/config.ex
midigofrank: Great catch, should be 50. That's a left over code

lib/lightning/config/bootstrap.ex
midigofrank: I didn't think it was necessary given the default was already a higher a value. But yeah, let me include it in the `DEPLOYMENT.md` docs

lib/lightning_web/live/credential_live/raw_body_component.ex
midigofrank: No it doesn't. We need the body as a map and `body_json` is a binary. It uses the `body` and it only parses the body if it is a binary

lib/lightning/credentials/credential_body.ex
midigofrank: Nothing happens to them if they don't touch the body. In the prod db, the highest has `19`. It's only the irregular one that has `500` and it will be dealt with at a user level.

lib/lightning/credentials/sensitive_values.ex
midigofrank: Yeah, we're good. The scrubber looks for the values as it scrubs. Duplicate values would just mean that the scrubber would make extra iterations

lib/lightning/config/bootstrap.ex
stuartc: Agreeing with Frank here (disclaimer I rarely look at or use .env.example), while we usually throw in an exhaustive list of options in that file (essentially listing the defaults) it's such a niche option probably fine to not include it.

lib/lightning_web/live/credential_live/raw_body_component.ex
stuartc: I'm with Elias here, it does get parse twice when it's a string. we enter via validate_body, calling parse_body with a binary and validate_sensitive_value_count which called parse_for_validation. Also the the `parse_body` taking a string and parsing it as a map and then in another match it takes a non-empty map and encodes it as a string - thats hurting my brain!

stuartc: Thinking some more, while I'm sure it does get parsed twice when it's a string; the previous (i.e. current) implementation in this component smells a bit now around the validation logic.

We don't use CredentialBody.changeset here, I'm not sure why though. That would require a bit of work to change.

PR #4311, opened 2026-01-13
fix: auto-select newly created credentials in workflow editor

### LINKED ISSUE #4287: Newly created credential not automatically added
**Describe the bug** When I create a new credential for a step in a workflow, the credential is not automatically added to the step. I'm taken back to the select credential interface, which isn't _usually_ a problem, but becomes a big problem when I'm creating a generic credential type.

Why? Because I've just hit save and the thing I've created is _NEITHER_ selected _NOR_ displayed on the list view that I'm taken back to. I need to perform additional navigation to find the thing I just created, and then select it.

This should behave like it did in the legacy system... on create the credential you create should be selected automatically.

**Version number**

`v2.15.6`

- [x] Yes
- [ ] No

1. Go to a workflow with an HTTP step
2. Click connect
3. Click create new
4. Click advanced
5. Click raw
6. Save it
7. See the bug

elias-ba:
## Description

This PR fixes the credential auto-selection bug where newly created credentials from the ConfigureAdaptorModal were not automatically selected for the job.

Closes #4287

## Validation steps

1. Open the collaborative workflow editor
2. Click on a job to open the inspector
3. Click "Connect" to open the ConfigureAdaptorModal
4. Click "Create new credential"
5. Fill in and save the credential
6. **Verify**: The new credential should be automatically selected
7. Test both type of credentials

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,19 @@ and this project adheres to
 
 ### Changed
 
+- Standardized "GitHub" brand capitalization throughout the UI
+  [#4313](https://github.com/OpenFn/lightning/issues/4313)
+- Optimized PromEx average claim duration query to use SQL aggregation instead
+  of fetching rows into memory
+  [#4317](https://github.com/OpenFn/lightning/pull/4317)
+
 ### Fixed
 
 - Fixed credential auto-selection when creating a new credential from the
   workflow editor's ConfigureAdaptorModal
   [#4287](https://github.com/OpenFn/lightning/issues/4287)
+- Add index on `runs.finished_at` to fix UsageTracking ReportWorker timeouts
+  [#4308](https://github.com/OpenFn/lightning/pull/4308)
 
 ## [2.15.10] - 2026-01-12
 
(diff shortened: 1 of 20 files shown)


assets/js/collaborative-editor/components/inspector/JobForm.tsx
```diff
@@ -150,8 +151,12 @@ export function JobForm({ job }: JobFormProps) {
         keychain_credential_id: is_project_credential ? null : credentialId,
       });
 
-      // Reload credentials in the background so the list is up to date
-      void requestCredentials();
+      // Wait for credentials to load before reopening modal
+      // This ensures the newly created credential appears in the list
+      await requestCredentials();
```
lmac-1: Should this be wrapped with a try / catch block? If this fails, I think the user could get stuck with no feedback because the modal won't reopen?

lmac-1 (approved):
Changes look great, nice work @elias-ba! I added one small comment to improve but it shouldn't block merging.

assets/js/collaborative-editor/components/inspector/JobForm.tsx
elias-ba: Good catch, thanks Lucy! Added the try/catch - the modal will now reopen even if the credential refresh fails.

Issue #4322, opened 2026-01-15
Channels


taylordowns2000:
### Context

The most common reason for adopting the OpenHIM is for **observability**. Being able to deploy the OpenHIM between two existing services, and get instant visibility into the requests and responses flowing back and forth between them is extremely helpful in complex systems.

Within this observability there is also a common need for:
- **The ability to retry transactions** (or batches of transactions). This doesn't always make sense because the response has already been sent by that point though.
- **Alerting for failed transactions** so that they can be investigated.
- **Scheduled reporting** (daily, weekly, monthly) to identify trends and spot anomalies.

The next-most important reason for adoption is for **access control**. Being able to separate the credentials used for authenticating a client, and those used for making requests to the target service provides a lot of flexibility in managing access, especially when combined with channels which can limit requests to a single endpoint. 

_This is also one of the biggest challenges for adoption though because of the diversity of authentication methods out there and the complexity which comes along with them. Mutual TLS with self-signed certificates are a fairly common approach._

Then a few other features that are commonly used:
- **Transformation of request paths** from one format to another. This is particularly helpful for (backwards) compatibility with target services.
- Sending a single incoming request to multiple target services for parallel processing or archival.

### Proposed Approach

In some cases, the complexity and overhead of a full OpenFn workflow javascript workflow isn't required. An inbound request (GET, POST, PUT, PATCH, DEL) to an OpenFn project _channel_ would not need to spawn a new worklfow run, but would instead use native Elixir/Phoenix to:

1. check if the requester is valid (authenticate using the `webhook_auth_method` model which should be abstracted to become a "registered client" model which is used by both webhook triggers and channels... if they are kept separate, call it a `channel_access_method`?)
2. optionally add authentication to the request
3. forward the request to a predefined destination (preserving the path, the body, the query params, etc.)

The goal would be to have a blazing fast reverse-proxy plus some simple authentication.

Requests to these channels _could_ be captured as `runs` or they could be stored in a different model such as `channel_requests`. Either way, we'd want them to appear on the history page, searchable and auditable just like `runs`. See image below, where you can see the traditional history page has been extended to include a "Client" column which tells you who started the run or channel-request.

<img width="1076" height="484" alt="Image" src="https://github.com/user-attachments/assets/5401c7c6-3745-40ef-8712-c0dbe9ba647c" />

[...]

PR #4323, opened 2026-01-15
Simple picklist to control sync/async mode

### LINKED ISSUE #4321: UI to let users turn on/off sync mode for workflows
The PR https://github.com/OpenFn/lightning/pull/3785 provided the back-end changes required to use workflows in "sync mode" - returning a response to the triggering webhook request only when the whole run is completed.

Currently, there is no picklist option on the trigger object to control this. The issue here is to add a picklist option to control it. For the moment, the only available values should be:
1. Async (default)
2. Sync

Image

Note that _later_ there will be a "custom" mode, where a the content, status, and timing of a response can be selected at any position in the workflow, but for now we are only implementing the sync/async options in this picklist to control the `trigger.webhook_reply` value.

taylordowns2000:
## Description

This PR closes #4321 , adding the picklist for async/sync mode triggers. (It's the second part of https://github.com/OpenFn/lightning/pull/3785, which has now been used for a month or two.)

## Validation steps

1. Create a webhook triggered workflow
2. Open the trigger
3. Change to sync mode
4. Curl the webhook trigger
5. See the curl response include the final state of the workflow

1. This is _not_ custom configurable responses #3102 (that comes later, if required)
2. This is _not_ channels #4322 (that comes later, if required)

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning_web/live/workflow_live/edit_test.exs
+++ b/test/lightning_web/live/workflow_live/edit_test.exs
@@ -633,7 +633,7 @@ defmodule LightningWeb.WorkflowLive.EditTest do
       workflow: workflow
     } do
       {:ok, _view, html} =
-        live(conn, ~p"/projects/#{project.id}/w/#{workflow.id}",
+        live(conn, ~p"/projects/#{project.id}/w/#{workflow.id}/legacy",
           on_error: :raise
         )
 


assets/js/collaborative-editor/types/session.ts
```diff
@@ -70,6 +70,7 @@ export namespace Session {
     enabled: boolean;
     cron_expression: string | null;
     has_auth_method: boolean;
+    webhook_reply: 'before_start' | 'after_completion' | null;
```
lmac-1, line 73: Is `| null` needed here? I think this field will always have a variable as we set a default and the migration set a default (priv/repo/migrations/20251018124310_add_delayed_response_to_triggers.exs).

assets/js/collaborative-editor/types/trigger.ts
```diff
@@ -19,6 +19,10 @@ const webhookTriggerSchema = baseTriggerSchema.extend({
   type: z.literal('webhook'),
   cron_expression: z.null(),
   kafka_configuration: z.null(),
+  webhook_reply: z
+    .enum(['before_start', 'after_completion'])
+    .nullable()
```
lmac-1, line 24: Is `nullable()` necessary here as  we are setting defaults both here and also in migration `priv/repo/migrations/20251018124310_add_delayed_response_to_triggers.exs`?

lib/lightning/collaboration/workflow_serializer.ex
```diff
@@ -202,7 +202,8 @@ defmodule Lightning.Collaboration.WorkflowSerializer do
           "cron_expression" => trigger.cron_expression,
           "enabled" => trigger.enabled,
           "id" => trigger.id,
-          "type" => trigger.type |> to_string()
+          "type" => trigger.type |> to_string(),
+          "webhook_reply" => trigger.webhook_reply |> to_string()
```
elias-ba: The `to_string()` call will convert `nil` to the string `"nil"` if `webhook_reply` is not set: `"webhook_reply" => trigger.webhook_reply |> to_string()`. Should we consider guarding against `nil` ? Something like: `"webhook_reply" => trigger.webhook_reply && to_string(trigger.webhook_reply)`. I think this is important because existing webhook triggers may have webhook_reply: nil, and the frontend type expects `'before_start' | 'after_completion' | null`, not the string `"nil"`.

assets/test/collaborative-editor/components/inspector/TriggerForm.test.tsx
```diff
@@ -0,0 +1,303 @@
...
+    adaptorStore,
+    awarenessStore,
+  };
+
+  const mockLiveViewActions = {
+    pushEvent: vi.fn(),
+    pushEventTo: vi.fn(),
+    handleEvent: vi.fn(() => vi.fn()),
+    navigate: vi.fn(),
+  };
```
elias-ba, line 90: Where is the `vi` imported from ?

elias-ba (changes requested):
Hey @taylordowns2000 nice work on this one. I have few observations. What do you think ?

taylordowns2000:
@elias-ba , @lmac-1 , thank you! can you make the necessary changes and merge this please?

assets/js/collaborative-editor/types/session.ts
doc-han: `null` because this field might not exist for all trigger types. 
eg. when `type = cron` triggers wouldn't have the `webhook_reply` field at all. This typing is for a Trigger (a union of all the types).
Hence, the fields here that don't have `null` are actually available for all trigger types

lib/lightning/collaboration/workflow_serializer.ex
doc-han: Makes sense and agreed. I think when it's `nil` we should actually send `nil` to the frontend. it's expecting it.

doc-han: resolved.

assets/js/collaborative-editor/types/trigger.ts
doc-han: This one here is a bit tricky because we're typing for `type = webhook`. I agree there shouldn't be nullable here because we expect every `type = webhook` to have `webhook_reply` defined.

assets/test/collaborative-editor/components/inspector/TriggerForm.test.tsx
doc-han: We have `globals: true` in `vitest.config.ts` and hence we have access to `vi` globally.

doc-han:
Added some update to make sure unsaved changes considers this new field

PR #4331, opened 2026-01-19
Add configurable work_mem for claim query

stuartc:
## Description

This PR adds a configurable `work_mem` setting for the claim query to eliminate temp file writes during window function sorting. Production testing confirmed that 32MB work_mem eliminates the 15.5MB disk sort.

Adds a new environment variable `CLAIM_WORK_MEM` (default: `32MB`) that is set via `SET LOCAL work_mem` during claim transactions alongside the existing `plan_cache_mode` setting.

## Validation steps

1. Run the config and queue tests: `mix test test/lightning/config_test.exs test/lightning/config/bootstrap_test.exs test/lightning/runs/queue_test.exs`
2. Verify invalid values are rejected (e.g., `CLAIM_WORK_MEM=invalid` raises an error on startup)
3. In production, verify with `EXPLAIN ANALYZE` that sort uses `quicksort Memory` instead of `external merge Disk`

1. The validation regex accepts PostgreSQL memory formats: `\d+(kB|MB|GB|TB)` (case-insensitive)
2. The Multi step was renamed from `:force_custom_plan` to `:configure_session` since it now sets both settings

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning/config.ex
```diff
@@ -334,6 +334,11 @@ defmodule Lightning.Config do
       Application.get_env(:lightning, :per_workflow_claim_limit, 50)
     end
 
+    @impl true
+    def claim_work_mem do
+      Application.get_env(:lightning, :claim_work_mem, "32MB")
```
rorymckinley: @stuartc Given that we already set the default in bootstrap.ex, do we need to set  default here as well?

test/lightning/config_test.exs
```diff
@@ -127,6 +127,14 @@ defmodule Lightning.Configtest do
 
       assert expected == actual
     end
+
+    test "returns the claim work_mem setting" do
+      expected = Application.get_env(:lightning, :claim_work_mem, "32MB")
```
rorymckinley: @stuartc If we are keeping the 32MB default setting in Config as well, then it may be a good idea to not have the expected value be the same as the default. If I was to misconfigure the value lookup, I would fall into the default and my test would still pass but the logic would be broken.

test/lightning/runs/queue_test.exs
```diff
@@ -46,4 +46,17 @@ defmodule Lightning.Runs.QueueTest do
...
     assert %{worker_name: nil} = Lightning.Repo.reload!(run_2)
   end
+
+  test "configures session with work_mem and plan_cache_mode", %{run_1: _run_1} do
+    ref =
+      :telemetry_test.attach_event_handlers(self(), [[:lightning, :repo, :query]])
+
+    Queue.claim(1, Query.eligible_for_claim())
+
+    assert_receive {[:lightning, :repo, :query], ^ref, _measurements,
```
rorymckinley: @stuartc This is a very cool technique - is it possible to assert the value being passed to `SET LOCAL` as well?

rorymckinley (approved):
@stuartc Looks good, I added some small comments but nothing that is a showstopper.

PR #4333, opened 2026-01-19
feat: show version_history at workflow level

### LINKED ISSUE #4329: Provisioner API does not support version history for workflows
- When GETing state JSON through the provisoner, a Project has a `version_history` array. But I'm not sure what this is?? It should be removed
- Each workflow should have a `version_history` key
- We must ensure that the legacy `cli pull`  does not blow up because of the extra version key.

For now, I am quite happy for the provisioner to ignore the version key on upload (see #3393 for more on this)

doc-han:
## Description

This PR makes `version_history` available at workflow level

Closes #4329 

## Validation steps

1. _(How can a reviewer validate your work?)_

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to
 - Add configurable `CLAIM_WORK_MEM` environment variable (default: 32MB) to set
   PostgreSQL work_mem for claim query, eliminating temp file writes during
   window function sorting [#4331](https://github.com/OpenFn/lightning/pull/4331)
+- Show version_history for workflows and not projects
+  [#4329](https://github.com/OpenFn/lightning/issues/4329)
 
 ### Fixed
 


lib/lightning_web/controllers/api/provisioning_json.ex
```diff
@@ -23,6 +24,7 @@ defmodule LightningWeb.API.ProvisioningJSON do
 
   def as_json(%Project{} = project) do
     Ecto.embedded_dump(project, :json)
+    |> Map.delete(:version_history)
```
elias-ba: I see you're deleting `version_history` from the project-level JSON. Could you add a brief comment explaining why? Something like: version_history lives at workflow level, not project level. I think this can help future readers understand the intent without needing to trace through the code. What do you think ?

lib/lightning_web/controllers/api/provisioning_json.ex
```diff
@@ -52,30 +54,45 @@ defmodule LightningWeb.API.ProvisioningJSON do
...
+      )
+
+    # snapshots don't have version_history
+    if module == Workflow do
+      version_history =
+        case Map.get(base_map, :version_history) do
+          nil -> WorkflowVersions.history_for(workflow_or_snapshot)
+          [] -> WorkflowVersions.history_for(workflow_or_snapshot)
+          history -> history
+        end
```
elias-ba: Looking at the definition of `WorkflowVersions.history_for/1`, it already handles the nil/empty case. So if `workflow_or_snapshot.version_history` is already `nil` or `[]`, calling `WorkflowVersions.history_for(workflow_or_snapshot)` directly would handle both cases. The current code works but essentially duplicates the nil/empty check. Maybe we could simplify to: `version_history = WorkflowVersions.history_for(workflow_or_snapshot)` Unless there's a reason to avoid the function call when history is already populated? In that case the current approach makes sense as an optimization to skip the function call overhead.

lib/lightning_web/controllers/api/provisioning_json.ex
josephjclark: The project should never have a version history - I think that's an error (or a late change since the feature was added)

Can we go further and nuke it from the project schema?

josephjclark:
I would like to check that the existing CLI is not broken by these changes, since we've changed the shape of provisioner state

EDIT: confirmed! I can pull and deploy using the old CLI commands with version history. All good there

josephjclark:
Tested the behaviour and it works great - just the comments on the PR to resolve

lib/lightning_web/controllers/api/provisioning_json.ex
doc-han: resolved. I've removed `version_history` entirely from projects

doc-han: hence, no need for this deletion anymore.

lib/lightning_web/controllers/api/provisioning_json.ex
doc-han: Thanks man. I've simplified to what you suggested

doc-han:
@josephjclark this is pending your approval. Thanks

josephjclark:
This works great @doc-han  - just need to update the changelog an you can merge.

But you're right - immediately next I'm going to need to generate a version stamp on provisioner push!

doc-han:
> This works great @doc-han - just need to update the changelog an you can merge.
> 
> But you're right - immediately next I'm going to need to generate a version stamp on provisioner push!

updated changelog here

josephjclark:
@doc-han @elias-ba I've been working around this and I think I've seen a problem

We have a `workflow_versions` table, which is cool and what I expect

But we also have a `version_history`  column on the workflow table.

Why do we need both?

I was expecting that the `workflow_versions` table would be queried to get the history. Because we're also caching the list in the workflow, it means if I clear the versions table then old versions still exist.

PR #4348, opened 2026-01-23
Fix converging edges swapping positions when selected

### LINKED ISSUE #4328: Shifty Path and Step Behavior on Auto-Layout Mode
**Describe the bug** On auto-layout mode, clicking across paths from the same step causes steps to shift on the canvas

https://www.loom.com/share/302055f6f3b748f1a4523ef7ccef67af

**Version number** v2.15.11

- [x] Yes
- [ ] No

1. Go to to a workflow where multiple paths come from the same node
2. Ensure you're on auto-layout mode
3. Click the different paths coming from the node

lmac-1:
## Description

This PR fixes a bug where converging edges would swap positions when selected on the diagram. 

The issue was caused by `sortOrderForSvg` reordering edges based on selection state, which affected Dagre's layout decisions. The fix preserves stable edge order regardless of selection.

Before: 
image

After: 
image

See the bug in action: https://www.loom.com/share/302055f6f3b748f1a4523ef7ccef67af

Closes #4328

## Validation steps

1. Open workflow with multiple branching points
2. Make sure your diagram is on auto layout mode
3. Click edges from different parent nodes
4. Click edges that lead to converging paths
5. **Expected:** All paths maintain their positions

1. If you have bigger workflows - perhaps click around on the paths and make sure that everything behaves as expected
2. I suspected that this sorting algorithm was to ensure that the edges stayed on top in terms of z-index. I've verified both with and without the fix that the behaviour remains the same when there are edges that are on top of each other.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/workflow-diagram/styles.ts
+++ b/assets/js/workflow-diagram/styles.ts
@@ -222,6 +222,5 @@ export const sortOrderForSvg = (a: object, b: object) => {
     return -1;
   }
 
-  // Fix for #4328: Preserve stable edge order to prevent layout shifts on selection
   return 0;
 };
--- a/assets/js/workflow-diagram/util/update-selection.ts
+++ b/assets/js/workflow-diagram/util/update-selection.ts
@@ -34,7 +34,7 @@ export default (model: Flow.Model, newSelection: string | null) => {
 
   const updatedModel = {
     nodes: model.nodes.map(updateItem) as Flow.Node[],
-    // Sort edges by enabled status. Selection state does not affect order (see #4328).
+    // Sort edges by enabled status. Selection state does not affect order.
     edges: model.edges.map(updateItem).sort(sortOrderForSvg) as Flow.Edge[],
   };
 
--- a/assets/test/workflow-diagram/styles.test.ts
+++ b/assets/test/workflow-diagram/styles.test.ts
@@ -30,7 +30,7 @@ describe.concurrent('sortOrderForSvg', () => {
     expect(sorted[2].id).toBe('edge3');
   });
 
-  test('preserves stable order when selection state changes (fix for #4328)', () => {
+  test('preserves stable order when selection state changes', () => {
     // This test ensures that selecting edges doesn't cause layout shifts
     const edges = [
       { id: 'edge1', data: { enabled: true }, selected: false },


assets/js/workflow-diagram/styles.ts
```diff
@@ -222,5 +222,6 @@ export const sortOrderForSvg = (a: object, b: object) => {
     return -1;
   }
 
-  return a.selected - b.selected;
```
doc-han, line 225: Nice fix. 
I'm trying to understand why this code existed in the first place. Seems to have zero value and I can't understand why it was like this.

assets/js/workflow-diagram/styles.ts
```diff
@@ -222,5 +222,6 @@ export const sortOrderForSvg = (a: object, b: object) => {
...
+  // Fix for #4328: Preserve stable edge order to prevent layout shifts on selection
```
doc-han: can we remove this comment? I think git blame handles this.

assets/js/workflow-diagram/util/update-selection.ts
```diff
@@ -34,7 +34,7 @@ export default (model: Flow.Model, newSelection: string | null) => {
 
   const updatedModel = {
     nodes: model.nodes.map(updateItem) as Flow.Node[],
-    // Must put selected edge LAST to ensure it stays on top.
+    // Sort edges by enabled status. Selection state does not affect order (see #4328).
```
doc-han: I'm yet to figure out why this sorting was done in that way. but looks good to me.

lmac-1: Removed comment referring to 4328

assets/js/workflow-diagram/styles.ts
lmac-1: Done

lmac-1:
@doc-han thanks for the review. I've fixed the comments referring to the PR. Can you check it again? 

Also re the sorting function, I asked my good friend Claude and it seems to think that it "ensures disabled edges render behind enabled edges in the SVG." So I think it's best if we keep it in!

doc-han:
> @doc-han thanks for the review. I've fixed the comments referring to the PR. Can you check it again?
> 
> Also re the sorting function, I asked my good friend Claude and it seems to think that it "ensures disabled edges render behind enabled edges in the SVG." So I think it's best if we keep it in!

Yeah makes sense. I was more concerned as to why the line had `return a.selected - b.selected;` because with the sort function, getting there meant the two values being compared are equal, so I'm wondering why we needed to sort based on another criteria which is `.selected`

lmac-1:
@doc-han are there any further changes needed here or can this be approved?

doc-han:
Hi @lmac-1 sorry for the delay. I was doing a deep dive into this.

### Why `sortOrderForSvg`
1. React flow assigns z-index to edges based on their order in the edges array. eg. `[edge1, edge2]` edge2 will have a higher z-index
2. Hence, the sort function `sortOrderForSvg` existed to make sure that the selected edge will always be at the bottom of the edges array. so that when two edges overlap, selecting either will always put it above the other. 

### Side Effect of `sortOrderForSvg` with Dagre
We then had a side-effect, Dagre(a library we use for calculating layouts) also depends on the order of edges to determine which edge/node shows left-most when a parent has more than one children. 

So the function `sortOrderForSvg` actually existing for solving `z-index` but had a side-effect of altering how `Dagre` also calculated its positioning.

### This solution
The solution in this PR tries to make Dagre satisfied by ensuring that edges always reach Dagre in the same order but then breaks the z-indexing reason why `sortOrderForSvg` existed. 

### Regression on this branch
When two edge labels overlap on this branch, the selected one doesn't get `z-index` precedence.

### Existing Regression on main
Looking at the logic in `sortOrderForSvg`, it tells me the problem also exists when you enable/disable an edge, the nodes will shift. Hence, that's something we also need to take care of. 

### Proposed Solution
We handle the z-index well and get rid of `sortOrderForSvg`. I think this is a problem react-flow is also realizing recently, so they just added [a new way of handling z-indexing](https://reactflow.dev/whats-new). This will require updating our version of react-flow and using the new z-index feature they've introduced. This will solve both problems. 

[zIndexMode in ReactFlow](https://reactflow.dev/api-reference/types/z-index-mode)

doc-han:
In short, this change swaps out the bug for another one by breaking the below feature.

CPT2601281336-539x301 (1)

It also doesn't handle the case for when the edge is being enabled/disabled which also gets us to the same issue.

CPT2601281340-826x776


@lmac-1 let me know whether you'll want to take on this, else I could pick it up.

lmac-1:
@doc-han great catch! I think it would be better if you could pick this up if you have time. I'm flying to the UK tonight and won't be online much more this week.

doc-han:
> @doc-han great catch! I think it would be better if you could pick this up if you have time. I'm flying to the UK tonight and won't be online much more this week.

Safe flight. I'll pick it up then

lmac-1:
Thanks a lot @doc-han - I appreciate the deep dive on this one.

doc-han:
Closing in favour of a new PR that fixed everything.

PR #4353, opened 2026-01-26
Fix version dropdown showing dot for workflows with cron/kafka triggers

### LINKED ISSUE #4352: Regression: Version dropdown does not show current version in header for workflows with cron/kafka trigger
**Describe the bug** Versions have regressed. The version dropdown does not prepopulate to the current version/latest and instead shows a dot. **This is happening for workflows with cron / kafka triggers.**

If you're running on a tagged version, include the number: `v2.15.12-pre1`

- [x] Yes
- [ ] No

This is also on staging

**To Reproduce** Steps to reproduce the behaviour:

1. Open any workflow
2. Check the header and look at the version dropdown. Make a note of what it says.
3. Click on the version dropdown and change the version.

**Expected behaviour** It should say "latest" on first load or a version number if you're viewing an older version. When you click the version dropdown and change the version, the header should show the version you have changed it to. 
*Actual behaviour** The dropdown in the header always shows a dot with a grey background

**Screenshots**

Image

**Additional context** I've noticed that it's happening on some workflows and not all. Not sure what it is that makes the bug happen. Also if there is no version in the version dropdown there are a number of related issues: can't manually run, can't save, ai chat not working as expected (loading infinitely)

lmac-1:
## Description

The version dropdown was breaking for workflows with cron or kafka triggers due to schema validation failures in the unsaved changes detection. The database stores `webhook_reply` with default `:before_start` for **all triggers**, but the frontend Zod schema expects `webhook_reply: null` for cron/kafka triggers.

This fix:
- Adds custom `Jason.Encoder` to conditionally serialise `webhook_reply`
- Extracts logic into `webhook_reply_for_json/1` helper function
- Updates `WorkflowSerializer` to use the same helper
- Updates tests to expect `null` for non-webhook triggers

The fix only affects JSON serialisation; backend logic that checks `webhook_reply` continues to work with raw database values unchanged.

Closes #4352

## Validation steps

1. Go to a workflow with a cron trigger
2. Check the header version dropdown
3. It should correctly load to "latest" 
4. Updating the version in the dropdown updates the version
5. When the version = latest, you're able to use AI chat, save the workflow and manually kick off a run
6. Repeat the same for a workflow with a kafka trigger

1. **Database design consideration**: - The `webhook_reply` field has `default: :before_start` for **all** trigger types in the schema, including cron and kafka triggers where it's not applicable. This fix conditionally serializes it to `null` for non-webhook triggers. A cleaner approach would be removing the default or migrating to separate embedded schemas per trigger type - but that would be best picked up by someone with a deeper understanding of the backend.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/collaborative-editor/components/inspector/TriggerForm.tsx
+++ b/assets/js/collaborative-editor/components/inspector/TriggerForm.tsx
@@ -176,13 +176,11 @@ export function TriggerForm({ trigger }: TriggerFormProps) {
                   id={field.name}
                   value={field.state.value as string}
                   onChange={e => {
-                    field.handleChange(
-                      e.target.value as 'webhook' | 'cron' | 'kafka'
-                    );
                     const newType = e.target.value as
                       | 'webhook'
                       | 'cron'
                       | 'kafka';
+                    field.handleChange(newType);
                     // Get default values for the new trigger type
                     const defaultValues = createDefaultTrigger(newType);
                     // Update the entire trigger with default values for the new type


lib/lightning/workflows/trigger.ex
```diff
@@ -26,16 +26,25 @@ defmodule Lightning.Workflows.Trigger do
...
+        cron_expression: trigger.cron_expression,
+        type: trigger.type,
+        enabled: trigger.enabled,
+        webhook_reply: Trigger.webhook_reply_for_json(trigger)
+      }
+
+      Jason.Encode.map(map, opts)
+    end
+  end
+
```
doc-han: I don't seem to be optimistic about this.

lib/lightning/workflows/trigger.ex
```diff
@@ -64,6 +73,18 @@ defmodule Lightning.Workflows.Trigger do
...
+  The database schema has webhook_reply with default :before_start for all triggers,
+  but the frontend Zod schema expects webhook_reply: null for cron/kafka triggers.
+  This function ensures the JSON sent to the frontend matches the expected schema.
+  """
+  def webhook_reply_for_json(%__MODULE__{type: :webhook, webhook_reply: reply}),
+    do: reply
+
+  def webhook_reply_for_json(%__MODULE__{}), do: nil
+
```
doc-han: I think this could be an `if-else`

lib/lightning/collaboration/workflow_serializer.ex
```diff
@@ -227,7 +227,9 @@ defmodule Lightning.Collaboration.WorkflowSerializer do
           "id" => trigger.id,
           "type" => trigger.type |> to_string(),
           "webhook_reply" =>
-            trigger.webhook_reply && to_string(trigger.webhook_reply)
+            trigger
+            |> Lightning.Workflows.Trigger.webhook_reply_for_json()
+            |> then(fn reply -> reply && to_string(reply) end)
```
doc-han: I think this could be made simpler than this

lmac-1:
thanks @midigofrank this looks much better - I wasn't sure how to do migrations.

doc-han:
Thanks @midigofrank.

midigofrank:
Aaah, I didn't see your comments :sweat_smile:

midigofrank:
Aaah, and there are failing tests. Let me fix that

doc-han:
> @lmac-1 @doc-han I have added a migration to ensure that only `webhook` type triggers have `webhook_reply` field filled in. Also, in the changeset validation, we set the value to `nil` if it's not a `webhook`. I think this solves the initial bug from the source. Let me know your thoughts

@midigofrank the changes look good. I'll still have to play with it and make sure the frontend schema validations aren't freaking out before approving.

midigofrank:
Alright. @doc-han all the tests are now passing

doc-han:
**update:** @midigofrank @lmac-1 
We need to be testing this PR with `KAFKA_TRIGGERS_ENABLED=no` because Kafka isn't fully supported now. Hence, we validate that this issue doesn't exist for cron triggers then we're good to go

doc-han:
The current issue on this branch(also exists on main) is that, 
1. when you have `trigger=cron` on the TriggerForm and then set your cron config then save.
2. switch to `trigger=webhook` then switch back to `trigger=cron`. You're sent back to the default state of the cron form instead of the previously saved state of the form for cron. 

Planning on getting this done separately as that was introduced by our kafka PR.

lmac-1:
Amazing, thank you so much @midigofrank for stepping in here. Lovely stuff.

PR #4364, opened 2026-02-01
Update workflow history

josephjclark:
Following #4333 and my final comment, I really don't think version history should be saved to the Workflow table. It's too hard to maintain there, and we can't run compaction properly.

This PR forces the provisioner and sandbox merge to look up the actual version history from the table.

Also, for good measure, since all version histories changed recently I've added a migration to remove all existing history entries. Hard reset on all of them. The CLI won't mind - it's incompatible with those versions anyway.

Tests are AI generated and I haven't even looked yet - I'll take a closer look before opening up for review.

Changelog deliberately omitted

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/test/integration/cli_deploy_test.exs
+++ b/test/integration/cli_deploy_test.exs
@@ -357,7 +357,12 @@ defmodule Lightning.CliDeployTest do
         {key, expected_edge_state(edge)}
       end)
 
-    Map.merge(state, %{jobs: jobs, triggers: triggers, edges: edges})
+    Map.merge(state, %{
+      jobs: jobs,
+      triggers: triggers,
+      edges: edges,
+      version_history: Lightning.WorkflowVersions.history_for(workflow)
+    })
   end
 
   defp expected_trigger_state(trigger) do


priv/repo/migrations/20260201163115_remove_version_history_from_workflows.exs
```diff
@@ -0,0 +1,33 @@
...
+    # Clean slate: remove all workflow_versions rows since they were out of sync
+    execute "DELETE FROM workflow_versions"
+  end
+
+  def down do
+    alter table(:workflows) do
+      add :version_history, {:array, :string}, default: [], null: false
+    end
+
+    execute """
```
josephjclark, line 23: idk what this is

josephjclark:
I'm a bit worried about some of the diffs in this PR. I'll have to review more closely with a clear head tomrrow

priv/repo/migrations/20260201163115_remove_version_history_from_workflows.exs
```diff
@@ -0,0 +1,33 @@
+defmodule Lightning.Repo.Migrations.RemoveVersionHistoryFromWorkflows do
```
josephjclark, line 1: @midigofrank if I wanted to basically reset all the versions in history, just to clear the slate, then I guess you wouldn't do it as a migration?

Maybe we should drop this. I was going to do it just as a precaution - but if it's too complex we should probably skip it.

I suppose the other way to look at it is a job that removes all versions timestamped before a certain date

elias-ba: I agree we don't need to reset the version history here. The code works without it. `history_for/1` and `latest_hash/1` now query the table directly and handle empty results fine. But I think we still need this migration for the column drop and constraint removal, since the `version_history` field is removed from the schema, the column should go too. So I'd suggest keeping this migration but removing these two lines:

```elixir                                                                                                                                                                                                       
  # Clean slate: remove all workflow_versions rows since they were out of sync                                                                                                                             
  execute "DELETE FROM workflow_versions"
```

If we want a clean slate later, a separate mix task with a date cutoff sounds like the right approach.

elias-ba:
Hey @josephjclark, I saw a few Elixir tests failing and wanted to help fix them. I already pushed a commit to fix the pull test in cli_deploy_test.exs (the expected state was missing version_history).

But there's one remaining failure I think is important for you to be aware of. The provisioning API still returns `version_history` via `WorkflowVersions.history_for` at `provisioning_json.ex:82-84`. The CLI round-trips it back during deploy, and the server rejects it with: `"extraneous parameters: version_history"`

Since the field was removed from the Workflow schema, the server no longer accepts it. Two options:
1. Stop returning `version_history` from the provisioning API (remove lines 82-87 in provisioning_json.ex)
2. Keep returning it but strip it from incoming deploy payloads

What do you think?

midigofrank:
Err @josephjclark ,  I think I'm lost on the intention for this. Is it to reset and recalculate the hashes?

josephjclark:
@elias-ba thank you for catching this!

For now, we want the provisioner to ignore the version history sent by the CLI. Lightning totally owns the history.

So can we strip the history from incoming deploy payloads? That's a perfect solution for me

(later, we might allow the CLI to explicitly overwrite the history to enable manual pruning)

@midigofrank The problem here is that a workflow's history is tracked in two places: it's in a table of its own, obviously, but it's also sort of weirdly cached as a JSON field in the Workflow. I want to remove this JSON cache so that the history is entirely derived from the DB

midigofrank:
@josephjclark @elias-ba I have copied these changes to this commit here https://github.com/OpenFn/lightning/pull/4396/changes/1672988175f179af94d05c55c82df7271eb07b58

I'm closing this PR

PR #4387, opened 2026-02-05
Restore intellisense to the Collaborative Editor

### LINKED ISSUE #4386: code assist support got dropped somewhere
Details to follow

- Code assist 100% used to work based on adaptor typescript (I've just confirmed against an old branch)
- Both legacy editor and collab editor seem to be broken
- At first glance I can't see any support for it in collab editor at all

josephjclark:
## Description

This PR fixes intellisense and code complete in the collaborative editor (it fixes the legacy editor too)

I've basically asked Claude to copy across the legacy solution and do a little refactoring.

Closes #4386 

Note that magic metadata appears to not work. I'll look at that separately.

## Validation steps

- Open a step with the common adaptor
- Over over any operation. You should get a pop-up with docs about that function. try `fn()`!
- Go to a new line. Hit ctrl+space
- You should be presented with a list of valid operations for that adaptor

Also note that the only things in the pick list should be adaptor  functions - no basic JS clutter

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,19 @@ and this project adheres to
 
 ### Changed
 
+### Fixed
+
+## [2.15.14] - 2026-02-13
+
+## [2.15.14-pre] - 2026-02-11
+
+### Added
+
+- Single AI chat for workflows and jobs
+  [#4109](https://github.com/OpenFn/lightning/issues/4109)
+
+### Changed
+
 - Refactor CircleCI to build-then-fan-out pattern, compiling once then running
   checks in parallel to eliminate cache race conditions and reduce flaky tests
   [#4378](https://github.com/OpenFn/lightning/pull/4378)
@@ -27,6 +40,8 @@ and this project adheres to
 
 - Fixed code-assist widget
   [4386](https://github.com/OpenFn/lightning/issues/4386)
+- AI Assistant button now disabled when Apollo not configured, preventing silent
+  failures [#4354](https://github.com/OpenFn/lightning/issues/4354)
 - Version chip missing tooltips
   [#4389](https://github.com/OpenFn/lightning/pull/4389)
 - Fixed StaleEntryError when saving workflows where a job is replaced and its
--- a/assets/js/collaborative-editor/CollaborativeEditor.tsx
+++ b/assets/js/collaborative-editor/CollaborativeEditor.tsx
@@ -42,6 +42,7 @@ export interface CollaborativeEditorDataProps {
   'data-root-project-id'?: string;
   'data-root-project-name'?: string;
   'data-is-new-workflow'?: string;
+  'data-ai-assistant-enabled'?: string;
   // Initial run data from server to avoid client-side race conditions
   'data-initial-run-data'?: string; // JSON-encoded RunStepsData
 }
@@ -65,6 +66,7 @@ interface BreadcrumbContentProps {
   projectNameFallback?: string;
   projectEnvFallback?: string;
   isNewWorkflow?: boolean;
+  aiAssistantEnabled: boolean;
 }
 
 function BreadcrumbContent({
@@ -74,6 +76,7 @@ function BreadcrumbContent({
   projectNameFallback,
   projectEnvFallback,
   isNewWorkflow = false,
+  aiAssistantEnabled,
 }: BreadcrumbContentProps) {
   const projectFromStore = useProject();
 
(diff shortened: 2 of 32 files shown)


assets/js/collaborative-editor/components/CollaborativeMonaco.tsx
```diff
@@ -82,6 +95,24 @@ export const CollaborativeMonaco = forwardRef<
         insertSpaces: true,
         automaticLayout: true,
         fixedOverflowWidgets: true,
+        dragAndDrop: false,
```
josephjclark: TODO double check these options

assets/js/collaborative-editor/components/CollaborativeMonaco.tsx
```diff
@@ -354,6 +455,10 @@ export const CollaborativeMonaco = forwardRef<
 
   return (
     <div className={cn('relative', className || 'h-full w-full')}>
+      {/* Loading indicator */}
```
josephjclark: TODO remove this stupid comment

assets/js/editor/Editor.tsx
```diff
@@ -96,6 +96,7 @@ type Lib = {
 };
 
 async function loadDTS(specifier: string): Promise<Lib[]> {
+  console.log('> loadDTS', specifier);
```
josephjclark: TODO revert difs here

doc-han (changes requested):
MIGAAA!

It works but only after I've switched the adaptor. 
eg. 
1. I select a salesforce adaptor from the canvas, then I visit the job code
2. The intellisense doesn't seem to work
3. Now I switch adaptor from salesforce to http. I see intellisense working.

could it be that it's downloading?

josephjclark:
@doc-han I think the problem is `@latest`. If the adaptor version is set to `@latest` then types will fail to load. If a specific version is set then it's OK.

I'll take a little look but I suspect that's an older bug - a smear in the old greatness

josephjclark:
@doc-han I've raised out https://github.com/OpenFn/kit/issues/1245 - this is a long standing bug in kit. I may raise a fix separately, I think it's cheap

josephjclark:
@doc-han the severity of this shouldn't be too high: for new steps, when you pick an adaptor, it'll select the actual latest version, not `@latest`. So unless a) you have an old step which is saved to `@latest`, or b you specically choose `@latest`, it'll work just fine.

elias-ba:
Fixed the failing JS tests with two changes:

1. The Monaco mock was missing the `languages.typescript` API that CollaborativeMonaco now uses for intellisense. Updated both the global mock and the local one in the diff test.
2. Also added a mock for `loadDTS` it was making network calls during tests that failed silently, causing vitest to exit with code 1 even though all tests passed.

josephjclark:
Thank you @elias-ba !!!

josephjclark:
@doc-han @theroinaochieng would love to get this into the demo release tomorrow

doc-han (approved):

assets/js/collaborative-editor/components/CollaborativeMonaco.tsx
```diff
@@ -99,6 +124,27 @@ export const CollaborativeMonaco = forwardRef<
...
+        document.body.appendChild(overflowNode);
+        overflowNodeRef.current = overflowNode;
+
+        // Update editor options with overflow container
+        // @ts-ignore - overflowWidgetsDomNode exists but isn't in updateOptions type
+        editor.updateOptions({
+          overflowWidgetsDomNode: overflowNode,
+          fixedOverflowWidgets: true,
+        });
+      }
```
doc-han: Doesn't monaco have a default `fixedOverflowWidgets` which is good enough? that's what monaco uses to keep all our tooltip high in z-index. I"m wondering why we had to create a new node to be appended to body.

doc-han:
@josephjclark change log not added.

assets/js/collaborative-editor/components/CollaborativeMonaco.tsx
josephjclark: Great catch!

This is old legacy editor code, introduced when we had problems with the z index. Feels fine without it in the collab editor. I'll cut it out

josephjclark:
@theroinaochieng I'd like to merge this to main and release it alongside single-session AI stuff. It's got some important  fixes for demos IMO. It does increase the risk and QA effort of the release though

PR #4392, opened 2026-02-07
fix: Disable AI Assistant when Apollo not configured

### LINKED ISSUE #4354: Regression: AI Assistant still available when Apollo isn't set up
**Describe the bug** On a local Lightning build, the AI assistant is still available for users to type in messages despite not having Apollo set up locally. This causes users to send messages and wait for responses that won't come because of the missing set up. 

- [x] Yes
- [ ] No

lmac-1:
### Problem
AI Assistant button was clickable when Apollo wasn't configured, causing silent failures where messages sent but never received responses.

### Solution
- Disable AI chat button when `APOLLO_ENDPOINT` or `AI_ASSISTANT_API_KEY` are missing
- Show tooltip: "Your instance does not have build-time AI enabled. Contact your administrator or support@openfn.org to configure it."
- Disable Cmd+K keyboard shortcut

Closes #4354

## Validation steps

1. Verify that it solves the scenario defined in #4354 
2. Also locally, comment out `APOLLO_ENDPOINT` or `AI_ASSISTANT_API_KEY` environment variables and confirm that AI assistant button is disabled. There should be a tooltip text advising the user. Make sure you can't open panel either with Cmd/Ctrl + K

1. I haven't been able to replicate the exact case described by Roina re local builds.. so I just want someone else to confirm that this definitely fixes that scenario. 

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to
 
 - AI Assistant button now disabled when Apollo not configured, preventing silent
   failures [#4354](https://github.com/OpenFn/lightning/issues/4354)
+- Version chip missing tooltips
+  [#4389](https://github.com/OpenFn/lightning/pull/4389)
 - Fixed StaleEntryError when saving workflows where a job is replaced and its
   edge retargeted to a new job (e.g. via AI assistant)
   [#4383](https://github.com/OpenFn/lightning/issues/4383)
--- a/lib/lightning_web/components/layout_components.ex
+++ b/lib/lightning_web/components/layout_components.ex
@@ -458,18 +458,14 @@ defmodule LightningWeb.LayoutComponents do
         <div class="pt-2 pb-1">
           <LightningWeb.Components.Common.openfn_logo class="h-6 primary-light mx-auto" />
         </div>
-        <div class="text-[8px] primary-light opacity-50">
-          v{Application.spec(:lightning, :vsn)}
-        </div>
+        <LightningWeb.Components.Common.version_chip />
       </div>
       <%!-- Collapsed branding: centered --%>
       <div class="sidebar-branding-collapsed hidden h-14 text-center">
         <div class="pt-2 pb-1">
           <LightningWeb.Components.Common.openfn_logo_collapsed class="h-6 primary-light mx-auto" />
         </div>
-        <div class="text-[8px] primary-light opacity-50">
-          v{Application.spec(:lightning, :vsn)}
-        </div>
+        <LightningWeb.Components.Common.version_chip />
       </div>
       <div class="border-t border-white/10 mt-2">
         <button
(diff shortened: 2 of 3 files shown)


assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
```diff
@@ -66,7 +66,11 @@ import { MessageList } from './MessageList';
  * - Persists width in localStorage
  * - Syncs open/closed state with URL query param (?chat=true)
  */
-export function AIAssistantPanelWrapper() {
+export function AIAssistantPanelWrapper({
+  aiAssistantEnabled = false,
+}: {
+  aiAssistantEnabled?: boolean;
+} = {}) {
```
elias-ba: The = {} default destructuring can be removed since the prop is always passed, right ?

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -240,6 +242,13 @@ export function Header({
   // When ?v= is present, user is viewing a specific version (even if latest)
   const isPinnedVersion = params['v'] !== undefined && params['v'] !== null;
 
+  // Determine AI button disabled message based on priority
+  const aiButtonDisabledMessage = !aiAssistantEnabled
+    ? 'Your instance does not have build-time AI enabled. Contact your administrator or support@openfn.org to configure it.'
```
elias-ba, line 247: Why do we use the expression "build-time AI" here ? Is that clear to the end user ? I know that expression because I have heard it being used internally multiple times. Is it the case for the end user in Lesotho using this app ? Wouldn't they stay minutes thinking about that term and asking themselve what does this mean ? Why not just use the term "AI Assistant" ? cc @taylordowns2000

lmac-1: @taylordowns2000 I used the message you mentioned on Slack but let me know if it should be changed. I agree that "AI assistant" could be a bit clearer for the user. 

This error message will trigger when the env variables aren't set up for AI assistant - can someone also confirm that's the scenario Roina was reporting (with local builds, etc.)?

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
lmac-1: Done! Nice catch.

assets/js/collaborative-editor/components/Header.tsx
lmac-1: @theroinaochieng we just need clarification / confirmation of this message. If it's okay we can approve and merge, if not I can make the update.

PR #4393, opened 2026-02-09
feat: single ai chat

### LINKED ISSUE #4109: Single chat for workflows and jobs
When chatting with our AI to create a new workflow, I'd like the chat to go like this:

- Me: "Please build a workflow that gets data from a REST API, transforms it, then loads it into a table in Postgres"
- AI: "Great, here's the workflow. [`HTTP -> Common -> Postgres`.] Once this looks good, let's dive into step 1 to write the first GET request together: [Provide _HYPERLINK_ to step 1 IDE.] "
- Me: [Clicks link] "What now?"
- AI: "Please drop a link to the API documentation for this first REST API, indicating which endpoint you want to connect to."
- Me: [Pastes link.]
- AI: "Thanks. Here's a GET request to that API, with authentication handled. Note that you don't have a credential set up for this step so you'll need to do that by clicking here: [Link to pick/create credential.]"

While that's the direction we want to go, the key thing we must do in _this issue_ (which requires no/minor Apollo changes) is allow the chat session to extend beyond a single job. This will require some DB changes and some routing/handling changes on Lightning.

Despite the front-end looking like the same chat and same agent at all time, in the back we'll use different agents for this v0 approach. The user will be able to see all 10 messages in the same chat session. If they click on a message that relates to another entity (the WF itself or step 3, while looking at step 2) nothing special will happen now, but in the future we may want to help them instantly navigate to another screen.

| session | message | workflow | job | agent    | response type | handler              |
| ------- | ------- | -------- | --- | -------- | ------------- | -------------------- |
| a       | 1       | 123      |     | wf-chat  | {text, yaml}  | only update workflow |
| a       | 2       | 123      |     | wf-chat  | {text, yaml}  | only update workflow |
| a       | 3       | 123      |     | wf-chat  | {text, yaml}  | only update workflow |
| a       | 4       | 123      | 1   | job-chat | {text, js}    | only print text\*    |
| a       | 5       | 123      | 1   | job-chat | {text, js}    | only print text\*    |
| a       | 6       | 123      |     | wf-chat  | {text, yaml}  | only update workflow |
| a       | 7       | 123      | 2   | job-chat | {text, js}    | only print text\*    |
| a       | 8       | 123      | 3   | job-chat | {text, js}    | only print text\*    |
| a       | 9       | 123      |     | wf-chat  | {text, yaml}  | only update workflow |
| a       | 10      | 123      |     | wf-chat  | {text, yaml}  | only update workflow |

*Eventually, we'll handle (and SUGGEST via green/red diff) the new `js` code.

**Note well:**

1. The workflow agent still doesn't change job code.
2. The job agent still doesn't change workflow yaml.
3. This is not a big change and requires nothing to be done (maybe a little prompt engineering?) on Apollo.
4. Given the current setup, this should _just work_ ™️ as is -> small DB changes and small changes in how requests are made from the front-end chat to Lightning's backend are all that are required.

doc-han:
## Description

This PR turns the AI chat into a single session for when a user is both on the workflow or job mode.

Closes #4109 

## Validation steps
**From create a new workflow**
1. Create a new workflow
2. Select the create with AI in the middle of the screen
3. Ask the ai to create a workflow for you eg. "create a 3 node workflow using the common adaptor"
4. save the workflow
5. open one of the steps, be sure you're still seeing all the messages in your session
6. ask the ai to generate a job code for you. eg. "write a simple fn job code that logs hello world to the console"

This demonstrates creating a new workflow and then editing the step code using AI in a single session

**From existing workflow**
1. In an existing  workflow
2. ask the ai to make some changes to the workflow nodes
3. then open one of the steps, be sure you're still seeing all the messages in your session
4. ask the ai to generate a job code for you. eg. "write a simple fn job code that logs hello world to the console"

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning_web/channels/ai_assistant_channel.ex
+++ b/lib/lightning_web/channels/ai_assistant_channel.ex
@@ -860,7 +860,8 @@ defmodule LightningWeb.AiAssistantChannel do
         {:ok, job}
 
       {:error, :not_found} ->
-        {:error, "Job not saved. Please save before AI can work on it."}
+        {:error,
+         "Job not saved or deleted. Please save if unsaved for AI to work."}
     end
   end
 


josephjclark:
Two important UX issues that fall out of this work:
- #4412 
- #4411 

These are not within the scope of the issue - it's just stuff we didn't think about when scoping the work.

assets/js/collaborative-editor/hooks/useAIMode.ts
```diff
@@ -66,22 +72,19 @@ export function useAIMode(): AIModeResult | null {
...
-        storageKey: `ai-job-${selectedJobId}`,
-      };
+      page = 'job_code';
     }
 
+    const storageKey = workflow?.id
+      ? `ai-workflow-${workflow.id}`
+      : `ai-project-${project.id}`;
+
+    console.log('han:ai-mode', context, page);
```
josephjclark: needs removing when finished

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -171,6 +170,15 @@ export function useAIWorkflowApplications({
    */
   const handleApplyWorkflow = useCallback(
     async (yaml: string, messageId: string) => {
+      if (!aiMode || aiMode.page !== 'workflow_template') {
```
josephjclark, line 173: I can't seem to trigger this - I can apply workflows even if the code editor is open. Also this error probably needs to feed back to the user.

We could spin this out into a new issue, I just wanted to raise the point first

assets/js/collaborative-editor/hooks/useAIMode.ts
doc-han: removed

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
doc-han: This is a UX problem that arises with this implementation. 
1. we disable workflow generated code buttons(apply,...) when in job mode and vice verse.
2. make them clickable but report that "you can't apply to workflow while viewing a job" and vice versa.

decision to be made i guess

lib/lightning/ai_assistant/ai_assistant.ex
```diff
@@ -753,11 +768,23 @@ defmodule Lightning.AiAssistant do
     |> handle_save_message_result()
   end
 
-  defp prepare_message_attrs(message_attrs, session_id, code) do
+  defp maybe_put_job_id_from_session(attrs, session) do
+    is_assistant = Map.get(attrs, "role") == "assistant"
+
+    if is_assistant && session.job_id do
+      job = Lightning.Repo.get(Lightning.Workflows.Job, session.job_id)
```
elias-ba, line 775: The `Repo.get` here runs a query every time an assistant message is saved. Since `ChatSession` already has `belongs_to :job, Job`, we could preload `:job` when the session is loaded (in `get_session/1` and `get_session!/1`, which already preload `:user` and `:messages`) and then just use `session.job` here instead of fetching. Something like:

```elixir
if is_assistant && session.job_id do
    Map.put(attrs, "job", session.job)
else
    attrs
end
```

Not a big deal at current scale, but avoids the extra query for free.

lib/lightning/ai_assistant/ai_assistant.ex
```diff
@@ -753,11 +768,23 @@ defmodule Lightning.AiAssistant do
     |> handle_save_message_result()
   end
 
-  defp prepare_message_attrs(message_attrs, session_id, code) do
+  defp maybe_put_job_id_from_session(attrs, session) do
+    is_assistant = Map.get(attrs, "role") == "assistant"
```
elias-ba: This won't match when the role value is an atom. `build_workflow_message/1` returns `%{role: :assistant}`, which after key stringification becomes `%{"role" => :assistant}`, so `== "assistant"` evaluates to `false`. It seems Harmless right now since old workflow_template sessions won't have `session.job_id`, but a quick fix would be: `is_assistant = to_string(Map.get(attrs, "role")) == "assistant"`

lib/lightning_web/channels/ai_assistant_channel.ex
```diff
@@ -801,8 +826,10 @@ defmodule LightningWeb.AiAssistantChannel do
          params,
          socket
        ) do
-    message_attrs = build_message_attrs(user, content, limit_result)
-    opts = extract_message_options(socket.assigns.session_type, params)
+    job = params["job_id"] && Jobs.get_job!(params["job_id"])
```
elias-ba: Since the workflow canvas is collaborative, another user could delete or replace this job (e.g., via "apply workflow") while someone is mid-conversation. `get_job!` would crash the channel in that case. Might be worth using `get_job` with a friendly error reply instead. What do you think ?

lib/lightning/ai_assistant/ai_assistant.ex
doc-han: With the current implementation. session action doesn't have a job relationship. I didn't remove that from the schema because the old/legacy AI still has that relationship.

With the current implement only chat_message has that relationship with a job.

lib/lightning/ai_assistant/ai_assistant.ex
doc-han: sweet. thanks man!

lib/lightning_web/channels/ai_assistant_channel.ex
doc-han: Yeah makes a lot of sense. updating this.

doc-han: Updated to send a good error message

doc-han:
@elias-ba can you take another look. concerns resolved

PR #4396, opened 2026-02-09
Show specific diverged workflows in sandbox merge dialog

### LINKED ISSUE #3958: sandboxes: no version history is available on old workflows
It seems like old projects always show a version conflict warning when trying to merge back into main:

Image

This happens even when there are no changes between the sandbox and main. Literally create the sandbox then merge it back in.

I am seeing this on `joe-sync-1`, a project which existed before sandboxes were invented.

If you create a new Project and a new Workflow, you don't see this.

Pretty sure it's just a migration issue - these workflows do not have a version history and so are (rightly) flagging the error.

### LINKED ISSUE #4001: sandboxes: when projects have diverged, tell me which workflows are in conflict
This is a quick win  - we have this information, and users can then make better decisions

midigofrank:
## Description
- Add `diverged_workflows/2` function to return list of workflow names that have diverged between sandbox and target projects
- Update merge modal to display specific workflow names in a bullet list when divergence is detected  

Screenshot 2026-02-09 at 11 36 49

Fixes #4001
Fixes #3958 
Closes #4364 

## Validation steps

1. Create a sandbox project
2. Pick any workflow in the sandbox project and add a job
3. Go back to the sandboxes page for the project and click to merge the sandbox. 
4. The merge modal should list your workflow as diverged, like in the attached screenshot

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/priv/repo/migrations/20260218060534_drop_versions_history_from_workflows.exs
+++ b/priv/repo/migrations/20260218060534_drop_versions_history_from_workflows.exs
@@ -8,7 +8,7 @@ defmodule Lightning.Repo.Migrations.DropVersionsHistoryFromWorkflows do
     """
 
     alter table(:workflows) do
-      remove :version_history
+      remove_if_exists :version_history
     end
   end
 


midigofrank:
@josephjclark let me know your thoughts on the wording

josephjclark:
Something is up here Frank - I see the warning even if the branches haven't diverged

Divergence means that the parent project changed since the sandbox was forked. So the head version in the parent is not in the sandbox's history.

lib/lightning/projects/merge_projects.ex
```diff
@@ -839,6 +839,57 @@ defmodule Lightning.Projects.MergeProjects do
     end)
   end
 
+  @doc """
+  Returns the list of workflow names that have diverged between source and target projects.
+
+  Compares version hashes for workflows with matching names. A workflow is considered
```
josephjclark: Yep:

> A workflow is considered
  diverged if it exists in both projects but has different version hashes.

this isn't quite right: a workflow is diverged if the parent head is not in the workflow's history.

midigofrank: Hmm, great catch!! Let me add tests

midigofrank: @josephjclark I have fixed this but I think there is another issue we need to fix. Currently, we're squashing the hashes if the `source(cli, app)` is the same. This means the workflows will always diverge after making a change because we will squash the "cloned" hash

midigofrank: Heh, it's bad. We don't clone the versions at all when creating sandbox

midigofrank: This is actually #3958 .

josephjclark:
@midigofrank I don't want to merge this until we've done the demo release - so I'm gonna pause review today. I'll review all your sandbox PRs together (maybe EOD tomorrow)

lib/lightning/projects/merge_projects.ex
midigofrank: >Heh, it's bad. We don't clone the versions at all when creating sandbox

I lied, we do clone the versions during forking: https://github.com/OpenFn/lightning/blob/main/lib/lightning/projects/sandboxes.ex#L558

midigofrank:
@josephjclark mind giving this another review?

josephjclark:
Sorry Frank this still isn't looking right :(

* I crate a new sandbox
* I edit a step
* I try to merge the sandbox - it warns me that it's diverged (it hasn't - I've edited the sandbox without changing the OG)

I've just gone onto main and I can see that it's giving exactly the same warning.

Ok, so this PR is correct in that it now lists the names of any affected workflows.

But we've got another critical issue in that something is wrong in the divergence logic still. I'm going to raise this as a new issue.

midigofrank:
Hey @josephjclark , I think part of the reason why you keep getting diverged workflows is because we always squash the versions. The fix is in https://github.com/OpenFn/lightning/pull/4414 . Let me get it ready

midigofrank:
Hey @josephjclark @elias-ba , I have merged in the 2 sandbox branches here. Would you mind giving this a fresh review?

lib/lightning_web/controllers/api/provisioning_controller.ex
```diff
@@ -144,6 +147,15 @@ defmodule LightningWeb.API.ProvisioningController do
...
   end
 
+  defp ensure_workflows_have_versions(project) do
+    workflows =
+      Workflows.list_project_workflows(project.id,
+        include: [:jobs, :edges, :triggers]
+      )
+
+    Enum.each(workflows, &WorkflowVersions.ensure_version_recorded/1)
+  end
```
elias-ba, line 157: 1. If `ensure_version_recorded/1` returns `{:error, reason}` for any workflow, the error is silently dropped. This will always succeed because Enum.each returns `:ok` regardless. Can we consider using `Enum.each` with explicit error checking, or switch to something like:

```elixir
defp ensure_workflows_have_versions(project) do
  workflows = Workflows.list_project_workflows(project.id, include: [:jobs, :edges, :triggers])

  Enum.reduce_while(workflows, :ok, fn workflow, :ok ->
    case WorkflowVersions.ensure_version_recorded(workflow) do
      {:ok, _} -> {:cont, :ok}
      {:error, reason} -> {:halt, {:error, reason}}
    end
  end)
end
```

2. This calls `ensure_version_recorded` for each workflow, which calls `latest_version `(1 query) and potentially `record_version` (1 transaction with multiple queries) per workflow. For projects with many workflows, this could add noticeable latency. Not a blocker but worth noting.

lib/lightning/projects/sandboxes.ex
```diff
@@ -386,6 +387,8 @@ defmodule Lightning.Projects.Sandboxes do
 
   defp create_sandbox_workflows(parent, sandbox) do
     Enum.reduce(parent.workflows, %{}, fn parent_workflow, mapping ->
+      {:ok, _} = WorkflowVersions.ensure_version_recorded(parent_workflow)
```
elias-ba, line 390: This will raise a `MatchError` if `ensure_version_recorded` returns `{:error, _}`. While unlikely in practice, a DB issue or constraint violation would crash the caller. Can we consider using with or propagating the error ? I think same pattern is in _lib/lightning/projects/merge_projects.ex:96_ too

test/lightning_web/controllers/api/provisioning_controller_test.exs
```diff
@@ -1052,14 +1110,13 @@ defmodule LightningWeb.API.ProvisioningControllerTest do
       workflow = Lightning.Repo.get!(Workflow, workflow_id)
       updated_history = Lightning.WorkflowVersions.history_for(workflow)
 
-      # Due to squashing behavior, consecutive versions from same source replace each other
-      # So we still have 1 version, but the hash should be different
-      assert length(updated_history) == 1
-      assert [updated_version] = updated_history
-      assert String.starts_with?(updated_version, "cli:")
+      # we nolonger squash the first version
```
elias-ba, line 1113: "no longer" instead of "nolonger" ?

lib/lightning/workflow_versions.ex
```diff
@@ -89,24 +88,50 @@ defmodule Lightning.WorkflowVersions do
...
+  def ensure_version_recorded(%Workflow{} = workflow) do
+    latest_version = latest_version(workflow.id)
+
+    if latest_version do
+      {:ok, latest_version}
+    else
+      hash = generate_hash(workflow)
+      record_version(workflow, hash)
+    end
+  end
```
elias-ba, line 126: This has a small TOCTOU window. Two concurrent calls for the same workflow could both see `latest_version` as `nil` and both proceed to `record_version`.

Inside `record_version`, the `Ecto.Multi` transaction queries `latest_version` again, but without a row lock (`FOR UPDATE`), so under PostgreSQL's default `READ COMMITTED` isolation both transactions can see `nil` simultaneously. Since the unique index on (`workflow_id, hash`) was removed recently, there's no DB-level constraint preventing both inserts from succeeding, which could result in a duplicate row.

In practice the risk is very low because:
- `ensure_version_recorded` is only called during sandbox creation, merge, and provisioning, infrequent, user-initiated operations
- `generate_hash` is deterministic, so both calls produce the same hash, a duplicate row is cosmetically odd but functionally harmless for divergence detection

If we ever wanted to fully close this window, adding `lock: "FOR UPDATE"` to the `latest_version` query inside the `record_version` Multi would do it. Not suggesting it's needed for this PR though, just flagging for awareness.

lib/lightning_web/controllers/api/provisioning_controller.ex
midigofrank: Great catch, but I really didn't want to check it's failure. This was more of a side effect than a main thread thing. Remember, at this point someone is fetching the `projectState.json`

lib/lightning/projects/sandboxes.ex
midigofrank: Great catch, but error propagation at this point was kinda difficult. The existing pattern even with workflow creation was all matching `{:ok, var} = insert(workflow | edge | ...)` so I just used the same pattern

lib/lightning/workflow_versions.ex
midigofrank: Aaah, great idea!! If it ever pops up, we'll include it

Issue #4412, opened 2026-02-09
Workflow Editor: In Draft mode, don't allow the user to edit code


josephjclark:
When the user creates as new workflow, they're put into a draft state of the editor. Nothing is saved until the Create button is pressed.

This draft mode is fairly clear when picking from the library (I have criticisms but it's OK)

<img width="1239" height="969" alt="Image" src="https://github.com/user-attachments/assets/6765f91d-461a-4a81-adad-05962d5f985d" />

But in Create With AI modem, it's much more subtle:

<img width="1245" height="958" alt="Image" src="https://github.com/user-attachments/assets/24141e58-9c31-4eec-b8b1-7b00f0ee8313" />

Because the templates are hidden, and there's no overlay text, you don't really know that the workflow hasn't been saved, or that you're in draft mode. It's just that create button that gives it away.

And yet I can open steps and start editing their code - and at this point things start getting weird because we're sending diffs back to Lightning about a step that doesn't exist in the database.

So until a Workflow has been Created, we must not allow the user to open the full screen IDE

PR #4416, opened 2026-02-11
feat: allow assistant to work on unsaved jobs

### LINKED ISSUE #4417: Allow single-chat AI assistant for unsaved jobs.
### User story
Allow users to use the AI assistant on unsaved jobs. 

### Details
When a use has an unsaved job, they're unable to use the single-chat AI assistant on it as it will fail and throw an error.

doc-han:
## Description

This PR allows the AI assistant to work on unsaved jobs.

Closes #4417  #4412 

## Validation steps

**New unsaved workflow**
1. Generate a new workflow using AI, don't click "create" yet
2. enter one of the nodes and use the AI chat to write some job code even tho it's not save. 
3. we expect it to work

**Already saved workflow with new nodes**
1. You have an already saved workflow but then you add a new node manually or via AI chat.
2. enter that unsaved node and then write some job code
3. we expect it to work

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/test/lightning_web/channels/ai_assistant_channel_test.exs
+++ b/test/lightning_web/channels/ai_assistant_channel_test.exs
@@ -590,6 +590,56 @@ defmodule LightningWeb.AiAssistantChannelTest do
       assert message.status == "error"
       assert error == limit_error_message
     end
+
+    @tag :capture_log
+    test "delegates to handle_unsaved_job_message when job_id is provided but not found in DB",
+         %{
+           socket: socket,
+           user: user,
+           workflow: workflow,
+           project: project
+         } do
+      unknown_job_id = Ecto.UUID.generate()
+
+      session =
+        insert(:chat_session,
+          user: user,
+          project: project,
+          workflow: workflow,
+          session_type: "workflow_template",
+          job_id: nil,
+          meta: %{}
+        )
+
+      {:ok, _, socket} =
+        subscribe_and_join(
+          socket,
+          AiAssistantChannel,
+          "ai_assistant:workflow_template:#{session.id}",
+          %{}
+        )
+
+      with_testing_mode(:manual, fn ->
+        ref =
+          push(socket, "new_message", %{
+            "content" => "what does this do?",
+            "job_id" => unknown_job_id,
+            "job_name" => "Draft Job",
+            "job_body" => "fn(state => state)",
+            "job_adaptor" => "@openfn/language-common@latest",
+            "workflow_id" => workflow.id
+          })
+
+        assert_reply ref, :ok, %{message: message}
+        assert message.role == "user"
+
+        reloaded = AiAssistant.get_session!(session.id)
+        user_msg = Enum.find(reloaded.messages, &(&1.role == :user))
+
+        assert is_nil(user_msg.job_id)
+        assert user_msg.meta["unsaved_job"]["id"] == unknown_job_id
+      end)
+    end
   end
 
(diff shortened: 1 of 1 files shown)


assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
```diff
@@ -414,7 +414,7 @@ export function AIAssistantPanelWrapper({
         }
       } else {
         // important: determines what ai to be used
-        options = { ...options, job_id: aiMode?.context.job_id };
+        options = { ...aiMode?.context, ...options };
```
elias-ba, line 417: Would it be worth explicitly listing just the fields needed here (e.g.,` job_id, job_name, job_body, job_adaptor, workflow_id`) instead of spreading the entire context? It works fine as-is since the spread order is correct and extra fields like `follow_run_id` are only read during session join, not message sending, just wondering if being explicit might make the intent clearer?

lib/lightning_web/channels/ai_assistant_channel.ex
```diff
@@ -848,9 +858,61 @@ defmodule LightningWeb.AiAssistantChannel do
              socket}
         end
 
-      {:error, msg} ->
-        {:reply, {:error, %{type: "validation_error", errors: %{base: [msg]}}},
-         socket}
+      {:error, :job_not_found} ->
```
elias-ba, line 861: Now that `may_get_job` returns `{:error, :job_not_found}` as a specific atom, should we keep a catch-all `{:error, _}` clause here too? The previous code had `{:error, msg}` which would handle any unexpected error, without it, if `may_get_job` ever returns a different error, this would crash with a CaseClauseError.

doc-han: got that. makes sense 👍🏽
resolved

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
doc-han: the new implementation doesn't have a separate session for jobs. Hence, session join mostly will happen at the workflow level. things like `follow_run_id` need to be put in context at the point the user is interacting with a job.

eg. I came from workflow chat then entered a job chat. note: it's the same workflow chat actually. we don't have a separate job chat now. so we need to get the job context to the backend.

josephjclark:
This is probably unrelated @doc-han but here's another issue in the same area:

- Edit an unsaved job
- Generate code with AI
- Close the AI Assistant
- The inspector freaks out and leaves you with this

image

doc-han:
> This is probably unrelated @doc-han but here's another issue in the same area:
> 
> * Edit an unsaved job
> * Generate code with AI
> * Close the AI Assistant
> * The inspector freaks out and leaves you with this
> 
> image

noted and resolved.

elias-ba (approved):
Nicely done Commander Han ! ⚔️

josephjclark:
What about those failing tests? Do they need fixing or can we merge?

elias-ba:
Hey @josephjclark no failing tests but we could improve the coverage of the ai_assistant_channel.ex. Let me push a commit with one test that could do us good here.

doc-han:
@elias-ba dealing with that

doc-han:
@elias-ba codecov looking good now. thanks

Issue #4410, opened 2026-02-09
Channels: K6 load test scenarios


stuartc:
### User story

We need parameterisable load test scripts that exercise the channel proxy under realistic conditions, so we can validate throughput, latency, and memory behaviour before shipping.

### Details

K6 scripts that cover the key proxy scenarios: happy-path throughput, concurrent connections, large payloads (both request and response), mixed methods, auth variations, and slow sinks. Reports should cover throughput, latency percentiles, error rate, and memory usage.

### Implementation notes

**Dependencies:** #4401 (proxy — needs something to test against), #4409 (mock sink server)

**Notes:** Run against local dev instance initially.

**Slow client testing:** K6 can't simulate slow clients at the TCP level (trickling request body data or reading responses slowly). This scenario is important for validating the proxy's streaming/memory behaviour and will need a different tool — e.g., `curl --limit-rate` or a custom script. Include it as a scenario but note the tooling difference.

### Release notes

N/A — internal test tooling.

### User acceptance criteria

- [ ] Happy-path throughput scenario (sustained requests/sec at configurable concurrency)
- [ ] Concurrent connections ramp-up scenario (10 → 50 → 100+ VUs)
- [ ] Large request payload scenario (1MB bodies, verify memory stays flat)
- [ ] Large response payload scenario (mock sink returns multi-MB responses, verify memory stays flat)
- [ ] Slow client scenario (slow body upload, slow response read — may need a tool other than K6)
- [ ] Mixed HTTP methods scenario (GET, POST, PUT, PATCH, DELETE)
- [ ] Auth variation scenario (valid keys, invalid keys, missing keys)
- [ ] Slow sink scenario (using mock sink with configurable delay)
- [ ] Reports: throughput, p50/p95/p99 latency, error rate, memory usage

Issue #4409, opened 2026-02-09
Channels: Mock sink server


stuartc:
### User story

We need a configurable test target that simulates real-world sink behaviours so we can test the proxy under realistic conditions during development and in CI.

### Details

A Plug app or standalone script that can be started locally and configured to simulate various sink responses: delays, status codes, auth rejection, timeouts, large response bodies, and mixed behaviour patterns. Reusable for ongoing development and CI.

### Implementation notes

**Dependencies:** None — can be built independently.

**Blocks:** #4410 (K6 load tests)

### Release notes

N/A — internal test tooling.

### User acceptance criteria

- [ ] Configurable response delay (0-30s)
- [ ] Configurable response status codes (200, 4xx, 5xx)
- [ ] Configurable response body size (up to several MB)
- [ ] Auth rejection mode (returns 401/403)
- [ ] Timeout mode (never responds)
- [ ] Mixed behaviour mode (e.g., 80% fast 200s, 20% slow 503s)
- [ ] Can be started locally with a single command

Issue #4408, opened 2026-02-09
Channels: History page integration


stuartc:
### User story

We need `ChannelRequest` records to appear in the existing History page alongside `WorkOrder` records so that users have a single view of all activity in a project.

### Details

> **Blocked: needs designs and discovery.** This touches one of the most complex pages in Lightning. The current implementation is complex and changes can be tricky — performance is critical here. A holistic review of the page's queries, filters, and LiveView structure is needed before implementation begins. Budget time for discovery.

ChannelRequests must appear alongside WorkOrders in the History page, filterable by channel. This is the most complex UI story in the MVP.

**Cut line:** If History page integration proves too expensive for the timeline, a standalone "Channel Requests" list page is the simpler alternative — standard LiveView CRUD with filtering, achievable in a couple of days. The MVP could ship with the simpler version and upgrade to full History integration in a fast follow.

### Implementation notes

**Dependencies:** #4399 (schema), #4405 (observer populates request/event data)

**Key technical considerations:**
- History page currently shows only WorkOrders with complex filtering (workflow, status, date range)
- Adding a second record type affects: queries, pagination, sorting, detail views, filters
- Recommended approach for MVP: polymorphic query (merge in application). Union view if performance requires it.
- Pagination with mixed record types is the hardest part — cursor-based pagination across two tables needs care
- Performance is critical — the History page is heavily used and must remain fast

### Release notes

Channel requests now appear in the project History page alongside workflow runs.

### User acceptance criteria

- [ ] History page shows both WorkOrders and ChannelRequests, sorted by time (most recent first)
- [ ] ChannelRequests visually distinguishable from WorkOrders (icon, label, or type badge)
- [ ] Filter by Channel added to History page filters
- [ ] Existing filters (workflow, status, date range) continue to work; channel filter is additive
- [ ] ChannelRequest detail view accessible from History row showing: request (method, path, headers, body preview, body hash), response (status, headers, body preview, body hash), timing (started_at, completed_at, TTFB, total latency), status badge, link to channel config via snapshot
- [ ] Pagination works correctly with mixed WorkOrder + ChannelRequest results
- [ ] No regressions in existing History page functionality

Issue #4407, opened 2026-02-09
Channels: CRUD LiveViews


stuartc:
### User story

We need LiveView pages for creating, editing, listing, and deleting channels within a project so that users can configure and manage their channels through the UI.

### Details

Standard CRUD LiveViews but with integration points into project navigation and credential selection. This is the most handoff-friendly piece if another engineer becomes available.

### Implementation notes

**Dependencies:** #4399 (schema + context), #4400 (audit trail for delete logging)

**Can start as soon as schema exists** — does not need the proxy pipeline.

**Key technical considerations:**
- Channel list accessible from project navigation (alongside Workflows)
- Credential dropdowns should pull from project credentials — and respect the supported-type allowlist from the auth design (#4403, #4404)
- Follows existing Lightning LiveView patterns and component library
- Bodyguard authorization policies for channel CRUD (project membership check)

### Release notes

Users can now create, edit, and manage Channels from within a project.

### User acceptance criteria

<img width="2562" height="1564" alt="Image" src="https://github.com/user-attachments/assets/b2096f80-29f2-4f0e-81ac-14d02538f70a" />

- [x] Channel list view accessible from project navigation (alongside Workflows)
- [x] List shows: channel name, sink URL, enabled/disabled status, request count and last activity
- [x] Create form uses same component as credential creation overlay on the canvas. Uses the fields: name, sink URL, sink credential (dropdown from project credentials), source credential (dropdown), enabled toggle
- [x] Edit form pre-populated with current values; saves via `config_changeset/2` (bumps `lock_version`)
- [ ] ~Channel detail/summary view showing current config and link to History (filtered)~
- [x] Delete with confirmation modal; deletion logged in audit trail
- [x] Form validation: name required and unique within project, sink URL required and valid URL format
- [x] Project navigation updated to include Channels section
- [x] Follows existing Lightning LiveView patterns and component library
- [x] Bodyguard authorization policies for channel CRUD (project membership check)

brandonjackson:
✅ designs ready
philosophical note: we will be changing how these pages work in the near future. so for now: please [satisfice](https://en.wikipedia.org/wiki/Satisficing) on this design. stick to conventions from workflows list. re-use components where possible.

midigofrank:
Closed by: https://github.com/OpenFn/lightning/pull/4445

PR #4419, opened 2026-02-11
Channels: HTTP proxy with observability for Lightning

### LINKED ISSUE #4322: Channels

stuartc:
## Summary

Channels adds lightweight reverse-proxy functionality with observability to Lightning. Deploy a Channel between two systems and get instant visibility into requests and responses, with full audit trail.

- **Source**: Authenticates inbound requests (API key / Basic Auth via project credentials)
- **Sink**: Forwards to target system with credential injection
- **Observability**: Every proxied request logged with headers, body preview, SHA256 hash, timing
- **Snapshots**: Channel config captured at request time for auditable history

Streaming proxy powered by [philter](https://github.com/OpenFn/philter) — constant memory regardless of payload size.

Closes #4322 | Go-live target: 26 Feb 2026

### Phase 1 — Foundation
- [x] #4399 — Schema and context setup
- [x] #4400 — Audit trail integration

### Phase 2 — Core Proxy
- [x] #4401 — Routing and proxy plug (no auth, no observation)
- [x] #4403 — Source authentication
- [x] #4404 — Sink authentication

### Phase 3 — Observability
- [x] #4405 — Observer implementation and persistence
- [ ] #4406 — Channel snapshots

### Phase 4 — UI
- [x] #4407 — Channel CRUD LiveViews
- [x] #4408 — History page integration _(cut line: standalone list page if too expensive)_

### Phase 5 — Performance Confidence
- [x] #4409 — Mock sink server
- [x] #4410 — K6 load test scenarios

## Dependency Graph

```
#4399 Schema ──┬──→ #4401 Proxy ──→ #4403 Source auth
               │       │           #4404 Sink auth (parallel)
               │       │
               ├───────┤──→ #4405 Observer ──→ #4406 Snapshots
               │       │
               ├──→ #4400 Audit trail
               ├──→ #4407 Channel UI
               └───────┴──→ #4408 History page

#4409 Mock sink ──┐
                  ├──→ #4410 K6 load tests
#4401 Proxy ──────┘
```


stuartc:
We've found a critical bug in the philter proxy library. It incorrectly folds in the `host` header into outgoing request. While this is 'fixable' in Lightning, the proxy library has no business blindly merging `host` and other headers in like this. 

The effect is that _any_ site that has more than one domain/site being served (i.e. most servers) will reject the request because the host/site requested gets set as whatever the host for Lightning is.

A fix is underway.

PR #4422, opened 2026-02-12
fix: enforce external_id uniqueness on credentials

### LINKED ISSUE #4170: Credentials can have the same external id

elias-ba:
## Description

This PR fixes duplicate `external_id` values on credentials, which caused ambiguous keychain resolution at runtime. Two layers of validation are added:

1. **DB unique index on `(external_id, user_id)`** - a user can never own two credentials with the same `external_id`
2. **Application-level check per project** - two different users cannot add credentials with the same `external_id` to the same project

**⚠️ Attention**: The migration handles pre-existing duplicates by logging affected credentials and nullifying the `external_id` on newer duplicates before creating the index. @taylordowns2000 @stuartc I would love your input for this strategy ? Can we nullify users conflicting external_ids on their existing credentials in prod ? That would probably lead to their jobs failing due to keychain credentials resolution problems. Should we find those users and send them an email to inform them about this fix before it goes live ?

Closes #4170

## Validation steps

1. Create a credential with an `external_id`, save it
2. Create another credential with the same `external_id` - should see a form error
3. Have a second user add a credential with the same `external_id` to the same project - should see a form error
4. Credentials with no `external_id` (nil) should always be allowed

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ and this project adheres to
 
 ### Fixed
 
+- Enforce `external_id` uniqueness on credentials per user and per project to
+  prevent ambiguous keychain resolution
+  [#4170](https://github.com/OpenFn/lightning/issues/4170)
+
 ## [2.15.14] - 2026-02-13
 
 ## [2.15.14-pre] - 2026-02-11


elias-ba:
@stuartc @taylordowns2000 , Note on cross-project external_id uniqueness enforcement. The per-project `external_id` uniqueness check (preventing two credentials with the same external_id in the same project, even across users) is application-level only, it uses `Repo.exists?` before insert/update.

This PR closes the main bypass path: the check now also validates against descendant sandbox projects, so a credential added to a parent won't propagate into a sandbox that already has a conflicting external_id.

The one remaining limitation is a TOCTOU race condition: two concurrent requests can both pass the `Repo.exists?` check before either commits, allowing a duplicate to slip through. In practice this requires two users adding credentials with the same `external_id` to the same project at the exact same moment. That's really very rare !

If we want DB-level enforcement later, the cleanest option would be a denormalized `external_id` column on `project_credentials`, maintained by Postgres triggers (BEFORE INSERT to copy from credentials, AFTER UPDATE to propagate changes), with a unique partial index `(external_id, project_id) WHERE external_id IS NOT NULL`. This follows the existing trigger pattern in the codebase (search vector triggers). The tradeoff is added schema complexity and trigger maintenance.

For now, the app-level check is sufficient given the low collision likelihood.

stuartc (approved):
Good work dude!

I made a few changes:

- Removed PII from migration - dropped credential name from the duplicate-clearing query and log output so user data doesn't leak into logs
- Added `descendants_query/1` and `descendant_ids/1` to Projects; composable recursive CTE that finds all descendant project IDs in a single query, replacing
  the N+1 `list_workspace_projects` loop
- Updated `Credentials` to use `descendant_ids/1` - both `validate_external_id` and `propagate_credential_to_descendants` now use the single-query approach
- Added regression test for removing a credential from a conflicting project (confirming the deletion path works correctly)
- Updated CHANGELOG

stuartc:
As for your comments for me and Taylor, I'm not worried about the TOCTOU, it would require some tricky denormalisation to catch it, given how low traffic this code path is - the risk is super low. We could try locking for reads while we change things, so enforcement wouldn't be susceptible. 

I've checked our installs and there aren't any duplicates so the migration won't change anything (other than add an index). But the logic seems sound to me.

elias-ba:
Hey @stuartc thank you very much for your changes and for verifying the duplicates. I reviewed your commits and they are all solid and important changes. I think this can now be safely merged. Thanks again

PR #4436, opened 2026-02-18
fix: auto-convert project name in settings & consolidate code

### LINKED ISSUE #4437: bug: project settings form does not auto-convert project name to url safe names
## Description

The project settings form rejects project names containing uppercase characters or spaces (e.g., "DEP-burundi-datafi") with an "invalid format" error, while the project creation form accepts them by auto-converting to the url-safe format (`dep-burundi-datafi`).

This inconsistency makes it appear as a bug to users — they can create a project with a mixed-case name but cannot rename an existing project the same way.

## Expected behavior

The settings form should auto-convert the name to lowercase-hyphens-digits format, matching the creation form behavior, and show a preview of the converted name.

## Steps to reproduce

1. Go to an existing project's settings page
2. Try to change the project name to something with uppercase letters (e.g., "DEP-burundi-datafi")
3. Observe the validation error "has invalid format"

elias-ba:
## Description

Fixes the project settings form to auto-convert project names to URL-safe format (lowercase, hyphens, digits), matching the behavior of the project creation form. Previously, entering a name like "DEP-burundi-datafi" in settings would fail validation since the form submitted the raw input directly.

This PR also consolidates all duplicate `coerce_raw_name_to_safe_name` functions into `Project.form_changeset/2` and `Project.form_with_users_changeset/2`, using a virtual `:raw_name` schema field. This follows the pattern already established by `Collection.form_changeset`.

Along the way, a pre-existing bug was fixed in the dashboard creation modal where the save error path assigned to `:project_changeset` instead of `:changeset`, meaning validation errors from failed project creation were silently swallowed. Additionally, the `@name` assign is now initialized in `update/2` (previously it only got set after the first validation event), and whitespace inside the name preview badge `<span>` tags was cleaned up to remove a trailing space.

Closes #4437

### Project settings
1. Edit the project name with uppercase/spaces (e.g. "My Cool Project")
2. Verify the preview badge shows `my-cool-project`
3. Save and confirm it persists the slug name
4. Clear the name and verify a single "can't be blank" error (not duplicated)
5. Change retention settings and verify the name does not break

### Create New Project
1. Type "Hello World!" and verify badge shows `hello-world`
2. Clear the name and verify submit disables and a single error shows
3. Submit and verify the project is created with the slug name

### Create / Edit Sandboxes
1. Create with "My Sandbox" and verify badge shows `my-sandbox`
2. Create another with the same name and verify "Sandbox name already exists" error
3. Edit a sandbox and rename it

### Create / Edit Collections
1. Create with spaces/uppercase and verify slug preview
2. Edit an existing collection and verify name displays correctly

### Create / Edit Projects (superuser interface)
1. Edit name to uppercase and verify preview updates
2. Save and verify slug persists

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/projects/project.ex
+++ b/lib/lightning/projects/project.ex
@@ -148,6 +148,7 @@ defmodule Lightning.Projects.Project do
     ])
     |> validate_required([:raw_name])
     |> derive_name_from_raw_name()
+    |> validate_required([:name])
     |> set_default_env_for_root_projects()
     |> validate()
   end
@@ -173,6 +174,7 @@ defmodule Lightning.Projects.Project do
     ])
     |> validate_required([:raw_name])
     |> derive_name_from_raw_name()
+    |> validate_required([:name])
     |> cast_assoc(:project_users, required: true, sort_param: :users_sort)
     |> validate()
     |> validate_project_owner()
--- a/lib/lightning_web/components/pills.ex
+++ b/lib/lightning_web/components/pills.ex
@@ -95,10 +95,10 @@ defmodule LightningWeb.Components.Pills do
 
   def name_badge(assigns) do
     ~H"""
-    <%= if to_string(@field.value) != "" do %>
+    <span :if={to_string(@field.value) != ""}>
       {render_slot(@inner_block)}
-      <span class="ml-1 rounded-md border border-slate-300 bg-yellow-100 p-1 font-mono text-xs"><%= @name %></span>.
-    <% end %>
+      <span class="ml-1 rounded-md border border-slate-300 bg-yellow-100 p-1 font-mono text-xs">{@name}</span>.
+    </span>
     """
   end
 
--- a/test/lightning/projects/project_test.exs
+++ b/test/lightning/projects/project_test.exs
@@ -207,6 +207,30 @@ defmodule Lightning.Projects.ProjectTest do
       cs = Project.form_changeset(%Project{}, %{raw_name: "root-proj"})
       assert Ecto.Changeset.get_field(cs, :env) == "main"
     end
+
+    test "requires both raw_name and name when raw_name is blank" do
+      for attrs <- [%{raw_name: ""}, %{"raw_name" => ""}] do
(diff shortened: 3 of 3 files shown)


test/lightning_web/live/project_live_test.exs
```diff
@@ -1661,10 +1657,6 @@ defmodule LightningWeb.ProjectLiveTest do
           |> to_string()
       }
 
-      assert view
-             |> form("#project-settings-form", project: invalid_project_name)
-             |> render_change() =~ "has invalid format"
-
```
stuartc, line 1667: Rather than removing this, we should test for values that exercise other parts of the validation. The form now keeps the `:name` as a hidden field, so any errors that could result wouldn't be shown. I think that uniqueness failures would get hidden as well?

stuartc (changes requested):
Some issues, mostly around the validation not being tested - and errors not shown probably (`raw_name` doesn't have a schema so I can't see it getting errors?).

Also there is a smell around the copy pasting of `coerce_raw_name_to_safe_name`, not just in this PR, in other places. There are 5 different files with an almost copy paste of this kind of function. I first thought we should put the transform into the schema, but then thought that might cause rendering weirdness with the schema changing the input - `Collection` has a `form_changeset` function that handles the `:raw_name` and the model has that as a virtual field. Keeping the logic a model concern and not a view concern.

test/lightning_web/live/project_live_test.exs
elias-ba: Right ! that test checked that `name: "some name"` triggered "has invalid format", but since the form now auto-converts raw input to a valid slug, spaces/uppercase are expected input, not errors.

It's replaced by tests that exercise the actual validation paths: submitting a name like "!!!" that resolves to blank triggers the save error and verifies it surfaces on the form via `copy_error`. Uniqueness errors go through the same `copy_error` mechanism so they show up on `raw_name` too.

stuartc (approved):

PR #4413, opened 2026-02-10
Add confirmation modal for adaptor changes with credentials 

lmac-1:
## Description

When changing adaptors in the collaborative editor, users now see a confirmation modal warning that credentials will be reset. The modal only appears when:
- Changing to a different adaptor package (not just version)
- A credential is currently selected

Also optimizes Y.Doc synchronization to eliminate double-sync by ensuring version changes happen in a single operation rather than syncing @latest then the actual version.

Includes comprehensive test coverage for all confirmation scenarios.

## Validation steps

**Setup:**
1. Open collaborative editor and select a job with an adaptor
2. Click Connect and create a credential to add to that step

**1. Confirmation appears when changing adaptors WITH credentials:**
1. Click on the adaptor on the step to open adaptor picker
4. Select a different adaptor (e.g., switch from Salesforce to HTTP)
5. Expected: "Change Adaptor?" confirmation modal appears with warning
6. Click "Continue"
7. Expected: Credential clears, adaptor changes, version auto-selects latest version number

**2.  No confirmation when changing adaptors WITHOUT credentials:**
1. Clear any selected credential
2. Click "Change" and select different adaptor
3. Expected: No confirmation, adaptor changes immediately

**3. No confirmation when changing versions:**
1. With credential set, change version in dropdown (don't click "Change")
2. Expected: No confirmation, version changes immediately

**4. Cancel preserves state:**
1. With credential set, click "Change" → select different adaptor
2. Click "Cancel" on confirmation
3. Expected: Original adaptor and credential unchanged, modal stays open

## Additional notes for the reviewer

1. NOTE: Pressing Escape when the confirmation modal is open closes everything (confirmation + modal + inspector).
This is consistent with existing behaviour for delete/reset confirmations throughout Inspector. If we want to 
change this so Escape only closes the confirmation dialog (leaving the parent modal open), that should be 
handled in a separate issue as it would require reworking how AlertDialog integrates with the keyboard 
shortcut priority system across the application. 

## AI Usage

Please disclose whether you've used AI anywhere in this PR (it's cool, we just
want to know!):

- [x] I have used Claude Code
- [ ] I have used another model
- [ ] I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Pre-submission checklist

- [x] I have performed an AI review of my code (we recommend using `/review`
      with Claude Code)
- [ ] I have implemented and tested all related **authorization policies**.
[...]

PR #4446, opened 2026-02-22
Add confirmation modal for adaptor changes with credentials 

### LINKED ISSUE #4395: Changing adaptor type doesn't reset credentials
### Description:
When a user changes an adaptor type on a step that already has credentials configured (e.g., from asana to common), the old credentials remain attached to the step. This can create invalid configurations when the new adaptor is incompatible with the existing credential type.

### Current Behavior:

1. Create a step with a specific adaptor (e.g., asana)
2. Configure credentials for that step
3. Change the adaptor type to any other adaptor (e.g., common, salesforce, etc.)
4. Observe that old credentials remain attached despite potential incompatibility

Image

### Expected Behavior:

- Any time an adaptor is changed, credentials should be reset
- **User should see a confirmation modal before the change: "Warning: Changing adaptors will reset the credential for this step. Are you sure you want to continue?"**

lmac-1:
_Originally raised in #4413 but I missed adding this functionality to FullScreenIDE, so started the approach again from scratch with this information._

## Description

Fixes credential handling when changing job adaptors. When users select a different adaptor, a confirmation modal now warns that credentials will be reset. 

### Key changes:

- Confirmation modal appears only when: (1) changing to a different adaptor package, (2) a credential is currently selected
- Both `project_credential_id` and `keychain_credential_id` are reset on confirmation
- Created `useAdaptorChangeConfirmation` hook to share logic between `FullScreenIDE` and `JobForm`

### Improvements over PR #4413:

- Hook-based architecture eliminates code duplication
- Fixed FullScreenIDE crash when currentJob is null
- Fixed Continue button navigation using ref pattern
- Fixed modal z-index stacking
- Skips confirmation when selecting same adaptor
- Added 21 unit tests

Closes #4395

## Validation steps

**Setup:**
1. Open collaborative editor and select a job with an adaptor
2. Click Connect and create a credential to add to that step

**1. Confirmation appears when changing adaptors WITH credentials:**
1. Click on the adaptor on the step to open adaptor picker
4. Select a different adaptor (e.g., switch from Salesforce to HTTP)
5. Expected: "Change Adaptor?" confirmation modal appears with warning
6. Click "Continue"
7. Expected: Credential clears, adaptor changes, version auto-selects latest version number

**2.  No confirmation when changing adaptors WITHOUT credentials:**
1. Clear any selected credential
2. Click "Change" and select different adaptor
3. Expected: No confirmation, adaptor changes immediately

**3. No confirmation for same adaptor**
Set job to HTTP adaptor with credential
Click "Configure" → "Change" → Select HTTP again
✓ No confirmation, configure modal reopens

**4. No confirmation when changing versions:**
1. With credential set, change version in dropdown (don't click "Change")
2. Expected: No confirmation, version changes immediately

**5. Cancel preserves state:**
1. With credential set, click "Change" → select different adaptor
2. Click "Cancel" on confirmation
3. Expected: Original adaptor and credential unchanged, modal stays open

**Repeat the same tests for FullScreenIDE - Click "Code" and then change the adaptor from there**

1. NOTE: Pressing Escape when the confirmation modal is open closes everything (confirmation + modal + inspector).
This is consistent with existing behaviour for delete/reset confirmations throughout Inspector. If we want to 
change this so Escape only closes the confirmation dialog (leaving the parent modal open), that should be 
handled in a separate issue as it would require reworking how AlertDialog integrates with the keyboard 
shortcut priority system across the application. 

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/collaborative-editor/components/ide/FullScreenIDE.tsx
+++ b/assets/js/collaborative-editor/components/ide/FullScreenIDE.tsx
@@ -26,7 +26,6 @@ import * as dataclipApi from '../../api/dataclips';
 import { RENDER_MODES } from '../../constants/panel';
 import { useCredentialModal } from '../../contexts/CredentialModalContext';
 import { useMonacoRef } from '../../contexts/MonacoRefContext';
-import { useAdaptorChangeConfirmation } from '../../hooks/useAdaptorChangeConfirmation';
 import { useProjectAdaptors } from '../../hooks/useAdaptors';
 import {
   useCredentials,
@@ -58,8 +57,7 @@ import {
 import { isFinalState } from '../../types/history';
 import { edgesToAdjList, getJobOrdinals } from '../../utils/workflowGraph';
 import { AdaptorDisplay } from '../AdaptorDisplay';
-import { AdaptorSelectionModal } from '../AdaptorSelectionModal';
-import { AlertDialog } from '../AlertDialog';
+import { AdaptorSelector } from '../AdaptorSelector';
 import { CollaborativeMonaco, type MonacoHandle } from '../CollaborativeMonaco';
 import { RunBadge } from '../common/RunBadge';
 import { ConfigureAdaptorModal } from '../ConfigureAdaptorModal';
@@ -596,29 +594,6 @@ export function FullScreenIDE({
       openCredentialModal(adaptorName, credentialId, 'ide');
     },
     [openCredentialModal]
-  );
-
-  // Use confirmation hook for adaptor changes
-  const {
-    isAdaptorChangeConfirmationOpen,
-    handleAdaptorSelect: handleAdaptorSelectWithConfirmation,
-    handleConfirmAdaptorChange,
-    handleCloseConfirmation,
-  } = useAdaptorChangeConfirmation({
-    job: currentJob,
-    updateJob,
-    setIsAdaptorPickerOpen,
-    setIsConfigureModalOpen,
-  });
-
-  // Wrapper to clear adaptor picker state after selection
-  const handleAdaptorSelect = useCallback(
-    (adaptorName: string) => {
-      if (!currentJob) return;
-      setAdaptorPickerFromConfigure(false);
-      handleAdaptorSelectWithConfirmation(adaptorName);
-    },
-    [currentJob, handleAdaptorSelectWithConfirmation]
   );
 
   // Handler for adaptor changes - immediately syncs to Y.Doc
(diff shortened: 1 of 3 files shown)


assets/js/collaborative-editor/components/inspector/JobForm.tsx
```diff
@@ -398,6 +420,18 @@ export function JobForm({ job }: JobFormProps) {
...
+      <AlertDialog
+        isOpen={isAdaptorChangeConfirmationOpen}
+        onClose={handleCloseConfirmation}
+        onConfirm={handleConfirmAdaptorChange}
+        title="Change Adaptor?"
+        description="Warning: Changing adaptors will reset the credential for this step. Are you sure you want to continue?"
+        confirmLabel="Continue"
+        cancelLabel="Cancel"
+        variant="primary"
+      />
```
doc-han: We now have this `<AlertDialog />` being placed in two different places. I'm wondering wouldn't it have been best to have it in the `<ConfigureAdaptor />` component so that wherever adaptor modal is used, we automatically have this `<AlertDialog />` ?

assets/js/collaborative-editor/components/ide/FullScreenIDE.tsx
```diff
@@ -1371,6 +1378,18 @@ export function FullScreenIDE({
...
+          <AlertDialog
+            isOpen={isAdaptorChangeConfirmationOpen}
+            onClose={handleCloseConfirmation}
+            onConfirm={handleConfirmAdaptorChange}
+            title="Change Adaptor?"
+            description="Warning: Changing adaptors will reset the credential for this step. Are you sure you want to continue?"
+            confirmLabel="Continue"
+            cancelLabel="Cancel"
+            variant="primary"
+          />
```
doc-han: Being called at two different location. could go into adaptor modal

doc-han (commented):
I initially expected issue #4395 to require less code than it did.

This PR addresses two things:
1. Fixes the reported bug
2. Introduces a confirmation modal when switching adaptors (new feature)

Since this changes the user experience, it may be worth a UX review. @theroinaochieng.

I'm wondering whether 2. could've been a feature PR instead of being added to a bug fix. The new confirmation modal increases the number of clicks users have to be doing whenever they're switching adaptors with credentials(the default for production work).

assets/js/collaborative-editor/components/inspector/JobForm.tsx
lmac-1: Thank you for your feedback 🙏🏻. 

I don't think adding the alert dialog to `ConfigureAdaptorModal` component will work because: 
- It opens after the adaptor has already changed - it's for selecting the version/credential for an adaptor
- The confirmation needs to happen before we commit to the new adaptor change, not after (ie from `AdaptorSelectionModal`)

I also don't think it makes sense to put it into `AdaptorSelectionModal` because:
- `WorkflowDiagram` uses it to create new jobs (which don't have credentials and don't need confirmation)
- It would force `WorkflowDiagram` to pass `job`/`updateJob`/`setIsConfigureModalOpen` etc props it doesn't need
- Nested modals feel a bit messy

Instead, I created a composite `AdaptorSelector` component that bundles `AdaptorSelectionModal` + `AlertDialog` + confirmation hook. I also fixed the ESC problem! What do you think?

lmac-1: Btw @doc-han this is back for a re-review now, if you have time. Or perhaps you'd prefer to wait until @theroinaochieng or someone does a UX review first?

theroinaochieng (approved):

lmac-1:
@doc-han Roina has approved from UX perspective. When you have time, if you can give this code one more review and then I think we are good to merge. Thanks :)

lmac-1:
Hey @doc-han do you think you'll have time to review this today, or should we reassign to someone else? Thanks!

doc-han (approved):
🎉

PR #4473, opened 2026-02-26
Allow on-the-fly updates to adaptors, schemas, and icons

### LINKED ISSUE #220: Credential types list should be pre-populated (caching?)
- When I click "new credential" the list of available credential types (defined in adaptors _and_ custom types file) is pre-populated with adaptors the instance already knows about.
- When I click on an adaptor that doesn't have a JSON schema packaged from NPM, it tells me "hey, no schema defined... wanna create a raw JSON credential?")

Consider:
- what's the caching strategy for adaptors in general?
- how do we handle changes to adaptor credential schemas? (versions?)
- how/where do we define schemas that aren't linked to specific adaptors

### LINKED ISSUE #325: Build service to pull credential schemas from npm/github
Considerations:
1. Make sure the adaptor registry isn't going wild.
2. Maybe add entry to package.json.
3. Consider `load_adaptors_after_boot` function which each copy of the app calls after it starts up?
4. Consider how we define credential body schemas that are _NOT_ linked to adaptors. (Taylor thinks that _these_ are the ones that should live in `Lightning` (where dhis2 is now) while the adaptor-specific credential bodies get moved into the adaptor repos.)

### LINKED ISSUE #1996: Adaptor registry is out of sync when the app is running in a cluster
**Describe the bug**
 

- [ ] Yes
- [ ] No

happen.

### LINKED ISSUE #2209: Better Adaptor List Refreshing
**Current Issue:**
The adaptor list is only loaded once at server startup, requiring restarts to see new adaptors or versions. This is increasingly problematic as we add pre-release versions and more adaptors.

**Proposed Solution:**
Implement a dual strategy approach:

1. Use Cachex for high-concurrency access to the adaptor list
2. Implement a Strategy pattern with:
   - A clear behavior contract for retrieving adaptors
   - Separate implementations for NPM Registry and Local Repository sources
3. Add a periodic refresh mechanism that runs in the background
4. Keep the data source configuration options we already have

**Benefits:**
- No server restarts needed to see new adaptor versions
- Better performance for concurrent adaptor list access
- Clear separation of concerns between data retrieval and caching
- Easily extensible if we need to add other adaptor sources
- Compatible with both NPM registry and local development setups

**Implementation Approach:**
We'll refactor the existing AdaptorRegistry into a more modular system while maintaining backward compatibility. The registry will determine the appropriate strategy when starting, then periodically refresh using that same strategy in the background.

### LINKED ISSUE #3114: Update adaptor icons list for new adaptors
When we release a new adaptor, we have to restart lightning for it to show in the icons picklist

But even then we're not done - we have to do something magical to get the icon to appear on the diagram. I think it even takes a release? Otherwise we just see this:

Image

The ultimate aim of this and #2209 is to make it so that whenever an adaptor (new or existing) is released to npm, the sales and services teams can start using it with as little effort as possible

taylordowns2000:
image

## Description

Enables on-the-fly updates to the adaptor registry, credential schemas, and adaptor icons without requiring an application restart. Superusers can trigger refreshes from a new Settings > Maintenance admin page, and a configurable Oban cron job (ADAPTOR_REFRESH_INTERVAL_HOURS) keeps them in sync automatically across clustered nodes.

Closes #3114
Closes #2209
Closes #325 (wow what a golden oldie!)
Closes #1996
Closes #220

##  Changes

  - New modules: AdaptorIcons, CredentialSchemas, and AdaptorRefreshWorker — extract runtime refresh logic out of mix tasks into callable modules
  - AdaptorRegistry — refactored GenServer state from a bare list to a map (%{adaptors, cache_path, local_mode}); added refresh/1, refresh_sync/1, and PubSub-based cross-node sync via adaptor:refresh topic
  - MaintenanceLive — new LiveView under Settings with action buttons for each refresh operation, gated to superusers
  - Mix tasks (install_adaptor_icons, install_schemas) — slimmed down to thin wrappers around the new modules
  - Oban cron — AdaptorRefreshWorker runs on a configurable interval; skips in local adaptors mode; broadcasts to peer nodes on success
  - Minor .env.example typo fixes

## Validation steps

1. Clear you adaptor registry and credential schemas
2. Log in as a superuser
3. Note that you can't find adaptors
4. Note that you can't find credential schemas
5. Go to the admin dashbaord
6. Go to "Maintenance"
7. Click the buttons and check that the things come back!

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


josephjclark:
When I try and run this locally I get:
```
[notice] Application lightning exited: exited in: Lightning.Application.start(:normal, [:timex])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) No Oban instance named `Oban` is running and config isn't available.

            (oban 2.20.1) lib/oban/registry.ex:37: Oban.Registry.config/1
            (oban 2.20.1) lib/oban.ex:609: Oban.insert/3
            (lightning 2.15.15-pre) lib/lightning/application.ex:155: Lightning.Application.start/2
            (kernel 10.2.6) application_master.erl:349: :application_master.start_supervisor/3
            (kernel 10.2.6) application_master.erl:331: :application_master.start_the_app/5
            (kernel 10.2.6) application_master.erl:317: :application_master.start_it_new/7
```

Main is fine

I haven't looked closely at this, but Claude says:

```
The error is happening in schedule_adaptor_refresh/0 at [application.ex:160-166](vscode-webview://0t3qgbcac14hmcf78knd4046spna3jtbcqlru439rntvjiue4v02/lib/lightning/application.ex#L160-L166).

The function calls Oban.insert/3 immediately after the supervisor starts, but Oban hasn't fully initialized yet by that point in the startup sequence. Oban needs a moment after its supervisor child starts before it can accept job insertions.

The root cause is a race condition in the startup sequence:

Supervisor.start_link(children, opts) starts all children (including Oban) and returns {:ok, pid}
But Oban.insert/3 is called synchronously right after, before Oban's internal registry/config is fully ready
The unless condition also won't protect against this — both local_adaptors_enabled?() and the :test env check pass, so it tries to insert the job.

How to fix it: You'd typically schedule this via a small delay (schedule_in: 5 or similar) or use a Task that waits, or restructure so the Oban insert happens after a confirmed ready signal. The simplest fix is wrapping the insert in a Task.start/1 so it runs asynchronously after the supervisor has returned:


defp schedule_adaptor_refresh do
  unless Lightning.AdaptorRegistry.local_adaptors_enabled?() or
           Lightning.Config.env() == :test do
    Task.start(fn ->
      Lightning.AdaptorRefreshWorker.new(%{}, schedule_in: 0)
      |> Oban.insert()
    end)
  end
end
Or just use schedule_in: 5 (seconds) to give Oban time to initialize before the job is inserted.
```

not sure if this is coincidence or a problem on the branch?

taylordowns2000:
Unfortunately that's just what happens when you start the app, @josephjclark !

I reported it first on September 10th, 2025. @stuartc reported it again more recently on January 21st.

josephjclark (approved):
Tested this locally and it seems to work! The same mix commands still work as on main (so we don't need to change anything in deployments), and the buttons on the admin page work great.

I get Oban errors on this branch (but not others). Claude made some suggestions which made it go away. I honestly have no idea - an adult should take a look at it.

I'm a little discomforted that the logic of the schema install (and presumably others) has changed so much. Error handling looks different. I'm sure it's fine!

stuartc (changes requested):
Hey @taylordowns2000, thanks for tackling this -- you know it's something I've
wanted to address for a long time. I have a few architectural concerns specific
to clustered/ephemeral deployments.

## Rolling restarts: late nodes miss the refresh

The Oban worker uses `unique: [period: 3600]`, which means only the first node's
startup job gets accepted during a rolling deploy. The job can run on any node
-- including one that's about to be terminated. The PubSub broadcast only
reaches nodes that are alive and subscribed at that moment. Nodes that start
later get their Oban job deduped and never receive a broadcast, so they're stuck
with build-time data until the next cron run.

Rough timeline for a 3-node rolling deploy:

```
T=0   Old A, Old B, Old C running
T=1   New A starts -> Oban job accepted
T=3   Job runs on Old B (about to die)
T=8   Job finishes -> broadcasts -> New A receives it
T=10  Old B terminates (refreshed data lost)
T=12  New B starts -> Oban job DEDUPED -> no broadcast -> stuck with build-time data
T=17  New C starts -> same -> stuck
```

New A ends up with fresh icons/schemas, but New B and New C are stuck until the
next scheduled cron run (which could be hours away).

## Icon path is different in releases

`Plug.Static` serves from `:lightning`'s release priv dir (e.g.
`/app/lib/lightning-x.y.z/priv/static/`), but `adaptor_icons_path` in prod is
`"priv/static/images/adaptors"` -- a CWD-relative path that resolves to
`/app/priv/static/images/adaptors/`. These are different directories. Icons
written at runtime go to the cwd relative path but Phoenix serves from the
release priv dir, so refreshed icons get written to disk but are never actually
served to users.

The build-time mix task works because it runs before `mix release`, so the files
end up inside the release artifact. At runtime, `adaptor_icons_path` would need
to resolve via `:code.priv_dir(:lightning)` instead.

Also bare in mind that if we put files inside priv/static without finger
printint and updating the manifest file we lose cache control (i.e. cache
busting filenames), and gz encoding (not that big a deal I guess with compressed
images). So Plug.Static DOES do etags, and 304s, but if you have fingerprinted
paths then you can tell the browser to cache forever (1 year) and you never get
another request from the same client. Not _that_ big of a deal with adaptor
icons though.

## Ephemeral storage and the "go fetch yourself" pattern

On ephemeral containers, all runtime files are lost on restart. Each node
independently fetches from NPM, GitHub, and jsDelivr after receiving a broadcast
-- the broadcast says "go refresh yourself" rather than "data is available."
With 2-3 nodes this is ok, but it means each deploy triggers N independent
fetches from external services, and the data is immediately lost when that
container cycles.

## A possible alternative direction

When I was working on this, my thinking was to invert this for
clustered/ephemeral setups something like:

- **PubSub signals "invalidate your cache"** rather than "go fetch your own
  copy."
- **DB as single source of truth** for registry data and schemas (they're just
  JSON). One Oban worker fetches and writes to Postgres, all nodes read from DB
  with an ETS cache. No stale window on startup -- new nodes read from DB
  immediately. (My old unfinished branch used the filesystem as well, but I
  think it may be time to use the db)
- **Lazy caching proxy for icons** -- serve through a dynamic route (not
  `Plug.Static`) that fetches from github or wherever on first request with a
  TTL. This sidesteps the path divergence and ephemeral storage issues entirely,
  and avoids downloading the full adaptors tarball for icons that may never be
  requested. Or a mix of both.

This kind of stuff needs to be tested in an environment much closer to
production, the priv directory thing is super easy to miss in dev mode, it's not
the first time I've wanted something like this, but being able to build a proper
Elixir release image (or no image) easily and repeatibly would be super valuable
here.

Happy to chat about any of this -- I know this is tricky territory and if we
don't get it right it's gonna be a real pain, and equally awesome if we do.
Really glad to see this moving forward!

taylordowns2000:
Awesome feedback @stuartc . I think I've understood and implemented it and my basic click-testing seems to work. Claude description of the changes below:

## Overview

Replace the current filesystem + GenServer-state storage for adaptor registry, credential schemas, and adaptor icons with Postgres as single source of truth and ETS as a read-through cache. This eliminates three production issues:

1. rolling-restart stale data,
2. icon path divergence in releases, and
3. ephemeral storage data loss.

## Files Created (9 new)

  ```
┌──────────────────────────────────────────────────────────────────────┬─────────────────────────────────────┐
  │                                 File                                 │               Purpose               │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ priv/repo/migrations/20260308204728_create_adaptor_cache_entries.exs │ Migration for adaptor_cache_entries │
  │                                                                      │  table                              │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ lib/lightning/adaptor_data/cache_entry.ex                            │ Ecto schema                         │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ lib/lightning/adaptor_data.ex                                        │ DB context (put, get, put_many,     │
  │                                                                      │ delete)                             │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ lib/lightning/adaptor_data/cache.ex                                  │ ETS read-through cache              │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ lib/lightning/adaptor_data/listener.ex                               │ PubSub invalidation listener        │
  │                                                                      │ GenServer                           │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ lib/lightning_web/controllers/adaptor_icon_controller.ex             │ Lazy icon proxy (serves PNGs +      │
  │                                                                      │ manifest)                           │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │                                                                      │ Endpoint plug to intercept          │
  │ lib/lightning_web/plugs/adaptor_icons.ex                             │ /images/adaptors/* before           │
  │                                                                      │ Plug.Static                         │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ test/lightning/adaptor_data_test.exs                                 │ Context CRUD tests                  │
  ├──────────────────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ test/lightning/adaptor_data/cache_test.exs + listener_test.exs       │ Cache + listener tests              │
  └──────────────────────────────────────────────────────────────────────┴─────────────────────────────────────┘
```
## Files Modified (key changes)
```
  ┌─────────────────────────────────────────────────────────────────────┬──────────────────────────────────────┐
  │                                File                                 │                Change                │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │                                                                     │ Reads from ETS/DB cache instead of   │
  │ lib/lightning/adaptor_registry.ex                                   │ GenServer state; removed PubSub      │
  │                                                                     │ subscription to old topic;           │
  │                                                                     │ simplified startup                   │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │                                                                     │ Writes to DB, broadcasts             │
  │ lib/lightning/adaptor_refresh_worker.ex                             │ {:invalidate_cache, kinds},          │
  │                                                                     │ uniqueness 3600→60s, no icon refresh │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning/credential_schemas.ex                                 │ Added fetch_and_store/0 for DB       │
  │                                                                     │ writes                               │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │                                                                     │ Rewritten: manifest from registry +  │
  │ lib/lightning/adaptor_icons.ex                                      │ lazy icon fetch from GitHub (no      │
  │                                                                     │ tarball)                             │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning/credentials.ex                                        │ get_schema/1 reads from cache with   │
  │                                                                     │ filesystem fallback                  │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning_web/live/credential_live/credential_form_component.ex │ get_type_options reads from cache    │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning_web/live/maintenance_live/index.ex                    │ Uses new DB-backed refresh functions │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning_web/endpoint.ex                                       │ Added AdaptorIcons plug before       │
  │                                                                     │ Plug.Static                          │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ lib/lightning/application.ex                                        │ ETS table init + Listener GenServer  │
  │                                                                     │ in supervision tree                  │
  ├─────────────────────────────────────────────────────────────────────┼──────────────────────────────────────┤
  │ test/support/{conn_case,channel_case,data_case}.ex                  │ Sandbox allow for AdaptorRegistry +  │
  │                                                                     │ registry seeding                     │
  └─────────────────────────────────────────────────────────────────────┴──────────────────────────────────────┘
```
 ## Architecture
```
  External sources (NPM, GitHub, jsDelivr)
          ↓ (Oban worker fetches once)
      PostgreSQL (adaptor_cache_entries)
          ↓ (read-through)
      ETS cache (Lightning.AdaptorData.Cache)
          ↓ (PubSub invalidation across cluster)
      All nodes read from ETS → DB fallback
```

**Comment by taylordowns2000** [comment]
@theroinaochieng , since Stu's changes have been addressed here, maybe a pre-review from @midigofrank or @elias-ba would warm him up for a nice quick merge later. The last few days have been a nightmare with all those restarts and new versions we've had to release. Would love to get this done to create some peace and quiet for engineering!

File: lib/lightning/maintenance.ex [hunk 1]
```diff
@@ -0,0 +1,236 @@
+defmodule Lightning.Maintenance do
```
elias-ba: I think this module is unused. MaintenanceLive calls AdaptorRegistry, AdaptorIcons, and CredentialSchemas directly, and the mix tasks now go to those too. Am I missing a caller? If not let's remove it.

lib/lightning/adaptor_refresh_worker.ex
```diff
@@ -0,0 +1,85 @@
...
+    if errors == [] do
+      Logger.info("Scheduled adaptor refresh completed successfully")
+    else
+      Logger.warning(
+        "Scheduled adaptor refresh partially failed: #{inspect(errors)}"
+      )
+    end
+
+    :ok
+  end
```
elias-ba, line 78: Should we also rebuild the icon manifest here? The cron worker refreshes registry and schemas, but the manifest stays empty until someone clicks Install Adaptor Icons. The workflow diagram fetches adaptor_icons.json once and keeps the deferred promise for the JS runtime lifetime, so on a fresh deploy AdaptorIcon falls through to its letter-tile placeholder until a superuser intervenes (and a full page reload happens after). I think calling `AdaptorIcons.refresh_manifest/0` from here would close that gap without adding any upstream traffic, since the manifest is built from the in-DB registry. The PNG prefetch could stay manual.

lib/lightning/adaptor_data/cache.ex
```diff
@@ -0,0 +1,83 @@
...
+    Cachex.clear!(@cache)
+    :ok
+  end
+
+  @doc "Broadcast cache invalidation to all nodes."
+  def broadcast_invalidation(kinds) when is_list(kinds) do
+    Lightning.API.broadcast(
+      "adaptor:data",
+      {:invalidate_cache, kinds, node()}
+    )
```
elias-ba: Could we go through `Lightning.broadcast/2` here instead of `Lightning.API` directly? Every other call site in `lib/` uses the wrapper, and I think relying on the test setup stubbing `LightningMock` with `Lightning.API` is a bit fragile if someone later wants to mock the boundary.

lib/lightning/adaptor_data/listener.ex
```diff
@@ -0,0 +1,34 @@
...
+  use GenServer
+  require Logger
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @impl GenServer
+  def init(_opts) do
+    Lightning.API.subscribe("adaptor:data")
```
elias-ba: Same thought as cache.ex. I'd lean toward `Lightning.subscribe("adaptor:data")` for consistency with the rest of `lib/`

lib/lightning/credentials.ex
```diff
@@ -575,15 +575,21 @@ defmodule Lightning.Credentials do
...
+      %{data: data} ->
+        Credentials.Schema.new(data, schema_name)
 
-    File.read("#{schemas_path}/#{schema_name}.json")
-    |> case do
-      {:ok, raw_json} ->
-        Credentials.Schema.new(raw_json, schema_name)
+      nil ->
+        # Fall back to filesystem for backwards compatibility
+        {:ok, schemas_path} = Application.fetch_env(:lightning, :schemas_path)
```
elias-ba, line 590: I'm a little unsure about this fallback. `AdaptorIconController` has no equivalent fallback to the build-time priv path on cache miss (it goes straight to GitHub or the local repo if configured), so the two callers behave differently when the cache is empty. I think it would be clearer to either drop the fallback (and treat DB as the single source) or add an equivalent disk-read path on the icon side. Curious which way you'd prefer.

lib/lightning/config/bootstrap.ex
```diff
@@ -285,7 +291,23 @@ defmodule Lightning.Config.Bootstrap do
...
         else: []
 
-    all_cron = base_cron ++ cleanup_cron
+    adaptor_refresh_cron =
+      case Application.get_env(
+             :lightning,
+             :adaptor_refresh_interval_hours,
+             0
+           ) do
+        hours when is_integer(hours) and hours >= 24 ->
```
elias-ba, line 309: I think the >= 24 branch may surprise operators a bit. Both 24 and any larger value collapse to a single 4am UTC daily cron, while 12 gives `"0 */12 * * *"` which fires at midnight and noon UTC. The 4am pick is also not surfaced in `.env.example` or in the worker's moduledoc as far as I could find. Would it be worth either documenting the cutoff or normalising so `1..23` uses `"0 */h * * *"` and 24 or above is daily?

lib/lightning/adaptor_registry.ex
```diff
@@ -286,6 +328,33 @@ defmodule Lightning.AdaptorRegistry do
...
 
+  defp get_adaptors_from_cache do
+    case Cache.get("registry", "all") do
+      %{data: data} ->
+        Jason.decode!(data, keys: :atoms!)
+
+      nil ->
+        []
+    end
+  end
```
elias-ba, line 372: Quick question on hot-path cost: `AdaptorRegistry.all/0`, `versions_for/2`, and `latest_for/2` all land here in non-local mode and `Jason.decode!` runs against the full registry payload each time. I think for high-frequency callers that adds up. Could we cache the decoded form (e.g. a separate Cachex key invalidated when the listener fires)?

lib/lightning/maintenance_verify.ex
```diff
@@ -0,0 +1,276 @@
...
+            {:error, "manifest JSON invalid: #{inspect(other)}"}
+        end
+    end
+  end
+
+  defp http_icon do
+    with {:ok, %{data: manifest_json}} <-
+           wrap_not_found(Cache.get("icon_manifest", "all")),
+         {:ok, manifest} <- Jason.decode(manifest_json),
+         [{adaptor, _} | _] <- Enum.to_list(manifest) do
```
elias-ba, line 225: I think this with can fall through with a bare `[]` when manifest is empty (`Enum.to_list(%{}) is []`). `safe_call` returns that `[]` verbatim, and the outer case in `run/0` has no `[]` arm, so it raises a `CaseClauseError` outside safe_call's rescue and crashes the whole verify run with a traceback. Would adding an else arm here that returns {:error, "manifest empty"} be acceptable?

lib/lightning_web/controllers/adaptor_icon_controller.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  defp serve_icon(conn, data) do
+    conn
+    |> put_resp_content_type("image/png")
+    |> put_resp_header(
+      "cache-control",
+      "public, max-age=#{@icon_max_age}"
+    )
+    |> send_resp(200, data)
+  end
```
elias-ba, line 83: Small thing: `put_resp_content_type("image/png")` adds `"; charset=utf-8"` to a binary PNG response. I think `put_resp_header("content-type", "image/png")` would be more accurate for binary content. Not a blocker

test/support/conn_case.ex
```diff
@@ -76,6 +76,51 @@ defmodule LightningWeb.ConnCase do
         shared: not tags[:async]
       )
 
+    # Seed the Cachex cache directly with minimal adaptor registry data so
+    # AdaptorRegistry.all() never falls through to a DB query from the
+    # GenServer process (which would fail sandbox ownership checks in async tests).
+    registry_json =
```
elias-ba: This seed block is duplicated byte-for-byte in data_case.ex (line 73) and channel_case.ex (line 62). Could we extract it into a shared helper? I think it'd save us from drift if the registry shape ever changes.

elias-ba (changes requested):

lib/lightning/credentials.ex
taylordowns2000: see comment below - i think removing the fallback would break the fully offline path

lib/lightning/adaptor_registry.ex
taylordowns2000: interesting, see comment

taylordowns2000:
@elias-ba , thank you for the review! i put in everything but these:

**5. (schema fallback).** I'm keeping it, but you're right that the comment is misleading. It's not legacy — it's the offline-mode read path. Lightning ships pre-baked Docker images that run `mix lightning.install_schemas` at build time and copy `priv/schemas/` into the final image. Combined with `LOCAL_ADAPTORS=true`, that gives us a fully offline deployment where the registry comes from the local repo (in-memory GenServer state), icons come from `OPENFN_ADAPTORS_REPO/packages/*/assets/*.png`, and schemas come from the `priv/schemas/` fallback you flagged. The cron is skipped in local-adaptors mode, so DB never fills in offline. Without that branch, the credential UI breaks on offline deployments — which was an explicit goal of this PR. Changed comments to make this clearer.

The icon controller doesn't need an analogous `priv/` fallback because icons already have an offline path via the local repo. If you'd like strict symmetry (baking icons into the image too), happy to file a follow-up, but I think it's redundant with what we already have.

**7. (decode on hot path).** Good call, doing it. Adding a parallel `{"registry", "decoded"}` Cachex key in `get_adaptors_from_cache/0` — the existing listener already clears every key matching `{"registry", _}` so invalidation comes for free. Note this only affects non-local-mode deployments since `LOCAL_ADAPTORS=true` bypasses this read path entirely (GenServer serves from in-memory state).

lib/lightning/credentials.ex
elias-ba: Got it, the offline rationale makes sense and the updated comment makes it clear. Thanks.

lib/lightning/adaptor_registry.ex
elias-ba: Nice, the `fetch_local/2` split is exactly right.

elias-ba (approved):

elias-ba:
@taylordowns2000 I just approved. Huge work on this one 👏🏽. It solves a really big problem, I hope it lands to prod very soon in the interest of everyone.

PS: I think you will need to look at the codecov report. The coverage seems really low on some files. I don't think it's blocking tho, but maybe you should consider bumping the coverage a little bit more

stuartc:
Closing in favour of #4801, which takes a different route on caching, seeding and failure/fallthrough scenarios.

PR #4474, opened 2026-02-26
Paginate superuser users/projects lists with server-side search

### LINKED ISSUE #2913: Paginate superuser projects and users list views
The superuser list views for managing projects and users are not paginated, resulting in slow loading times and unnecessary loads on the server when showing the list of users or projects.

**Expectation**

Both project and users list views need pagination and sorting (at least) in order to improve the experience. What we would want (perhaps in this issue, or another) is the ability to search for users or projects by one or more fields (single search box searching on name/email for example).

sakibsadmanshajib:
## Description

This PR adds server-side pagination, sorting, and search to the superuser users/projects settings list views, and moves those list operations out of in-memory filtering into SQL-backed queries.

Closes #2913

## Validation steps

1. `mix format`
2. `mix test test/lightning/accounts_test.exs test/lightning/projects_test.exs test/lightning_web/live/user_live_test.exs test/lightning_web/live/project_live_test.exs --seed 0`
3. `mix test test/lightning/accounts/admin_search_params_test.exs test/lightning/projects/admin_search_params_test.exs --seed 0`

1. Params use embedded schemas (`Accounts.AdminSearchParams` / `Projects.AdminSearchParams`) with WorkOrders-style field names (`search_term`, `sort_by`, `sort_direction`).
2. Context APIs are `list_all_users/1` and `list_all_projects/1`, taking the typed params struct.
3. Boolean columns (`role`, `enabled`, `support_user`) are not sortable; follow-up can add toggles.
4. No trigram indexes in this PR (can add later if needed).

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have implemented and tested all related authorization policies. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [x] I have updated the changelog.


lib/lightning_web/live/user_live/table_component.ex
```diff
@@ -28,18 +37,23 @@ defmodule LightningWeb.UserLive.TableComponent do
...
     {:ok,
-     assign(socket,
-       users: list_users("", "email", "asc"),
-       sort_key: "email",
-       sort_direction: "asc",
-       filter: ""
-     )}
+     socket
+     |> assign(:table_params, @default_table_params)
+     |> assign_table_state(@default_table_params, page)}
```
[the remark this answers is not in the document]
sakibsadmanshajib: Addressed in commit 67ca8dbb69199a73f60717563480ca8c36c1cce6. mount/1 no longer calls Accounts.list_users_for_admin/1; it initializes an empty Scrivener page and update/2 performs the single fetch using URL-driven table_params.

lib/lightning/accounts.ex
```diff
@@ -160,6 +185,109 @@ defmodule Lightning.Accounts do
...
+
+  defp filter_admin_users(query, filter) do
+    search = "%#{filter}%"
+
+    where(
+      query,
+      [u],
+      ilike(u.first_name, ^search) or
+        ilike(u.last_name, ^search) or
+        ilike(u.email, ^search) or
```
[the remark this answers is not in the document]
sakibsadmanshajib: Addressed in commit 67ca8dbb69199a73f60717563480ca8c36c1cce6. The email filter now matches the index expression by casting to text: ilike(fragment("?::text", u.email), ^search), so the trigram index can be used for this predicate.

priv/repo/migrations/20260226162000_add_admin_search_trgm_indexes.exs
midigofrank, line 1: Let's leave the indexes out for now, we can always add them later when needed.

lib/lightning_web/live/user_live/index.ex
midigofrank, line 1: You could leave the parsing and default values to an ecto schema. Have a look at: https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex and see how the runs index page utilizes ecto changesets to handle the filtering

lib/lightning/accounts.ex
```diff
@@ -114,6 +114,31 @@ defmodule Lightning.Accounts do
...
+  @admin_user_allowed_sorts ~w(first_name last_name email role enabled support_user scheduled_deletion)
+  @admin_user_default_page_size 10
+  @admin_user_max_page_size 100
+
+  @doc """
+  Returns a paginated list of users for the superuser admin table with
+  server-side filtering and sorting.
+  """
+  @spec list_users_for_admin(map()) :: Scrivener.Page.t()
+  def list_users_for_admin(params \\ %{}) do
```
midigofrank: You could have the normalization logic by having an embedded schema and doing validations using ecto changesets. Have a look at: https://github.com/OpenFn/lightning/blob/main/lib/lightning/invocation.ex#L482 for inspiration

midigofrank (changes requested):
Hey @sakibsadmanshajib , great job here. The `superuser` interface needs some love, thank you for picking this up.

I have left some comments:

1. I don't see a need for indexes on those tables. They're usually tiny
2. Consider using using an embedded schema to have a single place for validations, like: https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex
3. With the embedded schema in place, you can try to refactor the list_users function, like: https://github.com/OpenFn/lightning/blob/main/lib/lightning/invocation.ex#L482

lib/lightning_web/live/user_live/index.ex
sakibsadmanshajib: Addressed in 5ab22f0c35: moved the admin table param parsing into embedded schemas in `Lightning.Accounts.AdminSearchParams` and `Lightning.Projects.AdminSearchParams`. The user and project LiveViews now reuse those modules for validation, defaults, and URL normalization.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in 5ab22f0c35: refactored `list_users_for_admin/1` and `list_projects_for_admin/1` to consume the embedded search param structs and paginate through `AdminSearchParams.pagination_opts/1`, so the query layer works from normalized params instead of ad hoc parsing.

priv/repo/migrations/20260226162000_add_admin_search_trgm_indexes.exs
sakibsadmanshajib: Addressed in 5ab22f0c35: removed the `20260226162000_add_admin_search_trgm_indexes.exs` migration, so this PR no longer adds trigram indexes for the admin users or projects tables.

lib/lightning/accounts/admin_search_params.ex
midigofrank, line 1: You don't need to stringify and normalize while using ecto changesets. The changesets will do this for you, look at https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex. You can read more about Ecto changesets in https://hexdocs.pm/ecto/Ecto.Changeset.html

lib/lightning/projects/admin_search_params.ex
midigofrank, line 1: You don't need to normalize these values while using Ecto Changeset

lib/lightning/accounts.ex
```diff
@@ -114,6 +115,20 @@ defmodule Lightning.Accounts do
     Repo.all(User)
   end
 
+  @doc """
+  Returns a paginated list of users for the superuser admin table with
+  server-side filtering and sorting.
+  """
+  @spec list_users_for_admin(map()) :: Scrivener.Page.t()
+  def list_users_for_admin(params \\ %{}) do
```
midigofrank:
```suggestion
  def list_all_users(%AdminSearchParams{} = params) do
```

The context function should take the typed params instead of building it inside the body. This way we establish a "contract" with the callers.  

We should also rename the function, to `list_all_users` or `list_users_for_superuser`. `admin` conflicts with the `ProjectUser` role

lib/lightning/accounts/admin_search_params.ex
midigofrank: Also, could you use the same field names as the workorders search params? The `filter --> search_term`,  `sort --> sort_by`, `dir --> sort_direction`

lib/lightning/accounts.ex
```diff
@@ -160,6 +175,48 @@ defmodule Lightning.Accounts do
...
+    do: order_by(query, [u], desc: u.disabled)
+
+  defp order_admin_users(query, "enabled", "desc"),
+    do: order_by(query, [u], asc: u.disabled)
+
+  defp order_admin_users(query, "scheduled_deletion", "asc"),
+    do: order_by(query, [u], asc_nulls_last: u.scheduled_deletion)
+
+  defp order_admin_users(query, "scheduled_deletion", "desc"),
+    do: order_by(query, [u], desc_nulls_first: u.scheduled_deletion)
```
midigofrank:
```suggestion
    do: order_by(query, [u], desc_nulls_last: u.scheduled_deletion)
```

Given `null` are non deleted users, we should show them last

lib/lightning/accounts.ex
```diff
@@ -160,6 +175,48 @@ defmodule Lightning.Accounts do
...
+      query,
+      [u],
+      ilike(u.first_name, ^search) or
+        ilike(u.last_name, ^search) or
+        ilike(u.email, ^search) or
+        ilike(fragment("?::text", u.role), ^search)
+    )
+  end
+
+  defp order_admin_users(query, "enabled", "asc"),
```
midigofrank: I don't think we need to sort by a boolean. This should be toggle somewhere on the form. Let's skip it for now

lib/lightning/accounts.ex
```diff
@@ -114,6 +115,20 @@ defmodule Lightning.Accounts do
...
+  @doc """
+  Returns a paginated list of users for the superuser admin table with
+  server-side filtering and sorting.
+  """
+  @spec list_users_for_admin(map()) :: Scrivener.Page.t()
+  def list_users_for_admin(params \\ %{}) do
+    params = AdminSearchParams.new(params)
+
+    User
+    |> filter_admin_users(params.filter)
```
midigofrank:
```suggestion
    |> search_users_query(params.filter)
```

lib/lightning/accounts.ex
```diff
@@ -114,6 +115,20 @@ defmodule Lightning.Accounts do
...
+    |> order_admin_users(params.sort, params.dir)
```
midigofrank:
```suggestion
    |> sort_users_query(params.sort, params.dir)
```

lib/lightning/accounts.ex
```diff
@@ -160,6 +175,48 @@ defmodule Lightning.Accounts do
     if User.valid_password?(user, password), do: user
   end
 
+  defp filter_admin_users(query, ""), do: query
+
+  defp filter_admin_users(query, filter) do
```
midigofrank:
```suggestion
  defp search_users_query(query, search_term) do
```

lib/lightning/accounts.ex
```diff
@@ -160,6 +175,48 @@ defmodule Lightning.Accounts do
...
+  defp order_admin_users(query, "enabled", "desc"),
+    do: order_by(query, [u], asc: u.disabled)
+
+  defp order_admin_users(query, "scheduled_deletion", "asc"),
+    do: order_by(query, [u], asc_nulls_last: u.scheduled_deletion)
+
+  defp order_admin_users(query, "scheduled_deletion", "desc"),
+    do: order_by(query, [u], desc_nulls_first: u.scheduled_deletion)
+
+  defp order_admin_users(query, "role", dir) do
```
midigofrank: I also don't see the need to sort by `role`. There are only 2 types of users, `:user and :superuser`. This should be a toggle

lib/lightning/accounts/admin_search_params.ex
```diff
@@ -0,0 +1,118 @@
...
+  @moduledoc """
+  Normalized query params for the superuser users table.
+  """
+
+  use Lightning.Schema
+
+  @primary_key false
+
+  @default_sort "email"
+  @allowed_sorts ~w(first_name last_name email role enabled support_user scheduled_deletion)
```
midigofrank:
```suggestion
  @allowed_sorts ~w(first_name last_name email scheduled_deletion)
```

lib/lightning/accounts.ex
```diff
@@ -160,6 +175,48 @@ defmodule Lightning.Accounts do
...
+    )
+  end
+
+  defp order_admin_users(query, "enabled", "asc"),
+    do: order_by(query, [u], desc: u.disabled)
+
+  defp order_admin_users(query, "enabled", "desc"),
+    do: order_by(query, [u], asc: u.disabled)
+
+  defp order_admin_users(query, "scheduled_deletion", "asc"),
```
midigofrank:
```suggestion
  defp sort_users_query(query, "scheduled_deletion", "asc"),
```

midigofrank (changes requested):
Hey @sakibsadmanshajib , sorry I took so long to get back to you. You're on the right track. I have left inline comments, but here is the tldr:

1. Use standard field names, like those in https://github.com/OpenFn/lightning/blob/main/lib/lightning/workorders/search_params.ex . Use `search_term` instead of `filter`, `sort --> sort_by`, `dir --> sort_direction`
2. Boolean state fields like `role`, `enabled`, and `support_user` don't need to be in the sort list, these are better handled as toggles in the form. We can track that work in a separate follow-up issue.
3. Ecto changesets can handle all the normalizations you need, so the custom `normalize_value` function isn't necessary. 
4. Context functions should receive the struct directly rather than raw map params
5. I've suggested some renaming — `admin` conflicts with the `ProjectUser` role, so we'll want to resolve that

Please go through the comments and apply the same changes to the projects context. Note that I haven't reviewed the LiveViews yet, II'll do that in the next round once the contexts are updated. In the meantime, it's worth looking at how the workorders page uses the filter changeset (https://github.com/OpenFn/lightning/tree/main/lib/lightning_web/live/run_live) and applying the same pattern to these pages.

Thanks again for picking this up, don't hesitate to reach out if anything is unclear.

lib/lightning/accounts/admin_search_params.ex
sakibsadmanshajib: Addressed in a4c2853752: AdminSearchParams now casts via Ecto changesets (search_term/sort_by/sort_direction/page/page_size) and coerces invalid values with changeset helpers instead of pre-cast normalize/stringify.

lib/lightning/projects/admin_search_params.ex
sakibsadmanshajib: Addressed in a4c2853752: Projects.AdminSearchParams now uses the same changeset cast/coerce pattern as accounts (no pre-cast normalize helpers).

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: renamed to `list_all_users/1` and it now requires `%AdminSearchParams{}`. Same change on projects as `list_all_projects/1`.

lib/lightning/accounts/admin_search_params.ex
sakibsadmanshajib: Addressed in a4c2853752: fields renamed to `search_term`, `sort_by`, and `sort_direction` to match WorkOrders.SearchParams.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: scheduled_deletion desc now uses `desc_nulls_last` so non-deleted users stay last.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: removed sorting by enabled (and support_user). Those columns are no longer sortable in the table UI.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: renamed to `search_users_query/2`.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: renamed to `sort_users_query/3`.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: helper renamed to `search_users_query/2` with `search_term`.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: removed role from allowed sorts and dropped role sort clauses. Role column is no longer sortable in the UI.

lib/lightning/accounts/admin_search_params.ex
sakibsadmanshajib: Addressed in a4c2853752: `@allowed_sorts` is now `~w(first_name last_name email scheduled_deletion)`.

lib/lightning/accounts.ex
sakibsadmanshajib: Addressed in a4c2853752: helpers renamed to `sort_users_query/3` (and matching projects helpers).

lib/lightning/accounts/admin_search_params.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  def new(params \\ %{})
+  def new(%__MODULE__{} = params), do: params
+  def new(nil), do: new(%{})
+
+  def new(params) when is_map(params) do
+    %__MODULE__{}
+    |> cast(params, [:search_term, :sort_by, :sort_direction, :page, :page_size])
+    |> update_change(:search_term, &trim_search_term/1)
+    |> drop_invalid_cast(:page)
```
sakibsadmanshajib, line 40: Found while reviewing this: a hand-edited URL like /settings/users?page=abc raises Ecto.InvalidChangesetError through apply_action!/2, because cast/3 records an "is invalid" error for values the field type rejects and the safe-default fallbacks only handle out-of-range integers. Pushed a fix in a8a3daa8b7 that drops invalid casts so page/page_size fall back to their defaults, plus regression tests in test/lightning/accounts/admin_search_params_validation_test.exs (the existing tests only used "-10"/"1000", which cast fine, so this path was untested).

lib/lightning/accounts.ex
```diff
@@ -200,6 +213,41 @@ defmodule Lightning.Accounts do
     if User.valid_password?(user, password), do: user
   end
 
+  defp search_users_query(query, ""), do: query
+
+  defp search_users_query(query, search_term) do
+    search = "%#{escape_like_pattern(search_term)}%"
```
sakibsadmanshajib, line 244: Search terms containing LIKE metacharacters were passing straight through: % matches everything, _ matches any single character, and a trailing backslash could make the pattern match nothing. Since the old in-memory filter used String.contains?/2, this was also a behavior regression. Fixed in a8a3daa8b7 by escaping % _ \\ before wrapping in %...%, with regression coverage in accounts_test.exs (searching "u_a" now only matches a literal "u_a"). Same fix applies to projects.

lib/lightning/projects.ex
```diff
@@ -281,6 +297,68 @@ defmodule Lightning.Projects do
     end
   end
 
+  defp search_projects_query(query, ""), do: query
+
+  defp search_projects_query(query, search_term) do
+    search = "%#{escape_like_pattern(search_term)}%"
```
sakibsadmanshajib, line 307: Same wildcard issue as the users search: %, _, and backslash in search_term acted as SQL pattern metacharacters instead of literal text. Escaped in a8a3daa8b7, with a projects_test.exs regression test ("percent%" only matches the project literally named that).

lib/lightning/projects.ex
```diff
@@ -281,6 +297,68 @@ defmodule Lightning.Projects do
...
+  defp sort_projects_query(query, "owner", sort_direction) do
+    direction = sort_direction_to_atom(sort_direction)
+
+    order_by(
+      query,
+      [p, _pu, owner],
+      [
+        {^direction,
+         fragment("concat_ws(' ', ?, ?)", owner.first_name, owner.last_name)},
+        asc: p.id
```
sakibsadmanshajib, line 348: Single-column ORDER BY makes OFFSET paging nondeterministic when rows tie on the sort key (Postgres gives no order guarantee for ties), so pages can repeat or skip rows. Added asc p.id tiebreakers to every branch in a8a3daa8b7, including the scheduled_deletion nulls-last branches, plus a 12-project tied-description test that walks both pages and asserts no overlap. Also did the same for users (asc u.id) at lib/lightning/accounts.ex:245.

lib/lightning/accounts/admin_search_params.ex
sakibsadmanshajib: Fixed in a8a3daa8b7 — verified 329 context/param tests pass including the new non-numeric regression tests.

lib/lightning/accounts.ex
sakibsadmanshajib: Fixed in a8a3daa8b7.

lib/lightning/projects.ex
sakibsadmanshajib: Fixed in a8a3daa8b7.

lib/lightning/projects.ex
sakibsadmanshajib: Fixed in a8a3daa8b7.

sakibsadmanshajib (commented):
Ran an adversarial review pass over this PR against #2913 and pushed fixes for the two real bugs I found in a8a3daa8b7 (details in the inline threads). Everything else checked out: superuser gating is intact on both views, sort params are strictly allowlisted before reaching SQL, page_size caps at 100, counts stay consistent with filters, and the deferrals on boolean sorts and trigram indexes are respected throughout.

Three smaller things worth a look, none blocking:

1. `search_term` has no length cap (`lib/lightning/accounts/admin_search_params.ex:38`, same in projects). One line of `validate_length(:search_term, max: 200)` would bound the ILIKE payload per request.

2. The delete flows lose table state: the projects "Delete" link navigates without table params and the user deletion modal's return route drops them, so deleting from page 2 of a filtered view lands you back at unfiltered page 1 after confirmation. The cancel-deletion handlers already preserve params correctly.

3. Commit 0398ae3dc1 bundles a cowlib CVE ignore into this PR; fine functionally, just unrelated to pagination.

Test evidence: 329 context/param tests pass including new regressions for the malformed-param raise, wildcard literalness, and tied-key paging determinism. The 6 failures I saw when running both LiveView suites locally reproduce identically on the pre-fix head (enacl NIF missing in my container plus ordering artifacts), so they're environmental, not from these changes.

lib/lightning_web/live/user_live/table_component.ex
```diff
@@ -28,18 +30,24 @@ defmodule LightningWeb.UserLive.TableComponent do
...
+     |> assign(:table_params, table_params)
+     |> assign_table_state(table_params, empty_page())}
   end
 
   @impl true
   def update(assigns, socket) do
-    {:ok, assign(socket, assigns)}
+    table_params = Map.get(assigns, :table_params, socket.assigns.table_params)
+    search_params = AdminSearchParams.new(table_params)
+    page = Accounts.list_all_users(search_params)
```
sakibsadmanshajib, line 45: Follow-up from a second review pass: update/2 re-runs list_all_users/1 on every assigns change, not just when table_params actually moves. Opening the delete modal, a flash message, or any parent re-render triggers the count + select queries again, which works against the performance goal that motivated this PR. Guarding on table_params equality (fetch only when Map.get(assigns, :table_params) differs from socket.assigns.table_params) would keep the fetch URL-driven. Not blocking; happy to leave it for a follow-up if you prefer.

sakibsadmanshajib: Left as a follow-up optimization; current behavior is correct, just refetches more than strictly needed.

Issue #4467, opened 2026-02-26
Github installations sometimes don't show on sandboxes


josephjclark:
User reported this on a sandbox on app.openfn.org:

<img width="2962" height="1894" alt="Image" src="https://github.com/user-attachments/assets/3de02231-c3e9-4c68-bc46-cc455bfc0cf1" />

No installation available for the sandbox.

But when I go to the same project, I DO see the GH installations!

<img width="627" height="382" alt="Image" src="https://github.com/user-attachments/assets/3096ca5c-9c3d-4f9a-8205-302ed204959b" />

What's going on here?

Refresh, opening a new tab and clearing the cache doesn't seem to help.

Version `v2.15.14-pre`

lmac-1:
FYI @josephjclark  - this isn't just on sandboxes, it's happening on every project for this user. I think we're silently swallowing errors when the GitHub installations API call fails. I've raised a [draft PR](https://github.com/OpenFn/lightning/pull/4488) that adds error handling so users see a proper error message instead of just an empty dropdown.

Before having her disconnect/reconnect her GitHub account, we should merge this and see what error shows up. That'll confirm what's actually broken (my guess is expired OAuth token). Then she can reconnect and we'll know if it fixes things.

PR #4488, opened 2026-03-04
Fix: display error messages when GitHub installations fail to load

lmac-1:
A user reported that they have an empty dropdown when trying to add a GitHub installation:

image

My theory is that an API call has failed (e.g due to expired OAuth tokens), and we don't handle these errors. This PR adds proper error handling that displays specific error messages and guides users to reconnect their GitHub account in Settings when needed.

## What this PR does

When the GitHub installations API call fails (e.g., due to expired OAuth tokens), users would see an empty dropdown with no explanation. This PR adds:

- **Error handling**: Displays specific error messages for different failure types (invalid OAuth tokens, generic GitHub errors, network errors)
- **User guidance**: Provides actionable steps like reconnecting GitHub account in Settings
- **Accessibility**: ARIA attributes (`role="alert"`, `aria-live="polite"`) for screen reader support
- **Test coverage**: Comprehensive tests for all error state scenarios

## Validation steps

Not a very easy one to validate/test but UI will look something like this, with the GitHub fields disabled, code error messages only visible on dev environments: 

image
image
image
image

If you really want to test the UI, you can ask Claude to comment out some code so you can see the UI in all conditions.

1. The plan for this PR is that once it's merged, I'll get the user to retest and see what error they are getting. Then get them to disconnect and reconnect their GitHub account in order to see if it gets fixed by that. 
2. If we can confirm that the reason that error is happening is because of the expired oauth token, we could look how to make that experience a bit better.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning_web/live/project_live/github_sync_component.html.heex
+++ b/lib/lightning_web/live/project_live/github_sync_component.html.heex
@@ -10,49 +10,52 @@
       phx-target={@myself}
     >
       <%!-- Token is expired/invalid --%>
-      <%= if show_invalid_oauth_error?(@installations) do %>
-        <.error_banner title="Unable to load GitHub installations">
-          <p>
-            Your GitHub authentication has expired or is invalid.
-          </p>
-          <p class="mt-2">
-            Go to
-            <.link
-              href={~p"/profile"}
-              class="font-medium text-yellow-800 hover:text-yellow-900 underline"
-            >
-              Settings
-            </.link>
-            and reconnect your GitHub account.
-          </p>
-        </.error_banner>
-      <% end %>
+      <.error_banner
+        :if={show_invalid_oauth_error?(@installations)}
+        title="Unable to load GitHub installations"
+      >
+        <p>
+          Your GitHub authentication has expired or is invalid.
+        </p>
+        <p class="mt-2">
+          Go to
+          <.link
+            href={~p"/profile"}
+            class="font-medium text-yellow-800 hover:text-yellow-900 underline"
+          >
+            Settings
+          </.link>
+          and reconnect your GitHub account.
+        </p>
+      </.error_banner>
       <%!-- Token valid but no installations (app uninstalled/never installed) --%>
-      <%= if show_no_installations_warning?(@installations) do %>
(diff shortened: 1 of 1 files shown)


lmac-1:
@midigofrank Hey Frank, thank you for the review! A bit of extra context: I verified with Aisha and it's not just happening for sandboxes, it also happens at the parent project level, I left a comment on #4467. Sorry, I should have put it in the additional notes for reviewer here too. When in a huddle for Aisha I also got her to go to GitHub to check installations and it didn't fix things. Laura also raised the same issue on Friday at project-level and I got her to remove and re-add her GitHub in settings and it fixed the drop down issue.

So perhaps there are 3 scenarios: 
1. GitHub is down
2. Access token is expired (will be fixed by removing and readding GitHub in settings)
3. User uninstalled the app from GitHub

Do you want me to do the 'get out of jail free' link for scenario 3? Also do you think that telling the user to go to their settings to remove and readd their GitHub account it a good thing to do to fix 2? Or is there a better way to do this?

midigofrank:
Aaah, that's good to know @lmac-1 . Yes, I think we should have the get out jail link at the top instead of leading them to the profile page. It is worth testing though

lmac-1:
@midigofrank this is back for review: 
- For the error with the expired tokens, I haven't used the get out of jail link and instead link to /profile page. This is because I tested the get out of jail link with Aisha and it didn't resolve the situation.
- Fixed a typo in the error message
- Cleaned up the logic of the error message to use a shared component

lmac-1:
@midigofrank do you have some time to review this today? Thanks!

midigofrank:
Nicely done @lmac-1 . I have approved this, but just noticed the test file is new. Is it not possible to use the existing test setup?

lmac-1:
@midigofrank Done! Tests are now in the existing file, following the style of the rest of the project. Thanks for flagging. I removed the other file.

lmac-1:
Thanks @midigofrank for your patience on this one :)

PR #4517, opened 2026-03-10
feat: streaming ai assistant messages

### LINKED ISSUE #3585: Add support for streaming text and status updates from Apollo AI service

### LINKED ISSUE #4575: AI Chat: auto-scroll when new content comes in
The new streaming AI mode will beautifully render text character-by-character as the AI assistant sends it down, resulting in a delightful user experience.

But if the assistant is streaming long messages, or if the user's window is very small, the content will not auto-scroll to the bottom. So content appears out of sight.

This isn't terribly hard to fix BUT sometimes the user might not want to auto-scroll. Maybe they're reading slowly. So I think the correct behaviour is:
* call scroll to bottom after every character (or after every 10m whatever)
* Disable auto-scroll as soon as the user does a manual scroll

I think we need auto-scroll enabled for every new message UNTIL the user scroll's themselves. So the behavior is a little subtle

### LINKED ISSUE #4576: AI Assistant: newline characters can briefly get rendered in streaming mode
Sometime when streaming answers down to the user, content can get rendered wierdly. Like you'll see `\n` characters come up.

This will be something to do with the incremental markdown parsing, which generally handles content very well but sometimes goes wrong. It might be very hard to fix.

## Repro Steps

When I run this query, the assistant gives me back a  long indented list. That list will include `\n` characters for a while, until it suddenly pops into shape
```
What is the difference between each and Array.each? Give me a lot of detail please
```

elias-ba:
## Description

This PR adds real-time streaming support for AI assistant responses. Instead of waiting for the full response to complete, users now see text stream in word-by-word. Structured changes (workflow YAML, job code diffs) are applied to the canvas/editor before the text explanation finishes streaming.

Previously, the AI would generate text first and code/YAML second, so users would read "I've updated your workflow..." and only see the changes seconds later. With a companion Apollo PR (https://github.com/OpenFn/apollo/pull/407) that reverses the generation order, Lightning now receives a `changes` SSE event early in the stream containing the resolved code or YAML, applies it immediately, and then streams the text explanation.

The backend handles SSE streaming events from Apollo (text chunks, status updates, structured changes), broadcasts them via PubSub, and forwards them through Phoenix Channels to WebSocket clients. It also fixes a Mint transport error crash during SSE streaming and a message_processor crash on non-map error tuples.

The frontend adds streaming state to the AI assistant store, registers event handlers in AIChannelRegistry, renders streaming content in the message list, and auto-applies workflow YAML / auto-previews job code diffs as soon as streaming changes arrive, with deduplication to prevent double-application when the final message lands.

Closes #3585
Closes #4576
Closes #4575 

## Validation steps

1. Send a message to the AI assistant. The response text should stream in word-by-word instead of appearing all at once after completion
2. While streaming, verify that a status indicator (e.g. "Thinking...") appears before text starts arriving
3. After the stream completes, verify the final message renders correctly with proper markdown formatting, code block, and Apply button
4. In workflow mode, ask the AI to create or modify a workflow. The canvas should update before the text explanation finishes streaming
5. In job code mode, ask the AI to modify code. The diff preview should appear in Monaco before the text finishes
6. Send a message without code context (e.g. "how does this adaptor work?"). Text should stream normally without any code preview
7. With a collaborator viewing the same session, verify that only the message author gets auto-apply/preview
8. Disconnect the network mid-stream or trigger an error. Verify graceful handling without crashes

This PR depends on a companion Apollo PR (https://github.com/OpenFn/apollo/pull/407) which reverses the generation order and introduces the new `changes` SSE event. To validate this work, you'll need Apollo running locally on the `tweaks-for-streaming` branch (or a deployed version with that branch merged). Without the Apollo changes, streaming will still work but structured changes won't arrive early. They'll only appear when the final message completes, as before.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning_web/controllers/api/ai_assistant_controller.ex
+++ b/lib/lightning_web/controllers/api/ai_assistant_controller.ex
@@ -118,9 +118,7 @@ defmodule LightningWeb.API.AiAssistantController do
   end
 
   defp authorize_access("workflow_template", project, user) do
-    project_user = Projects.get_project_user(project, user)
-
-    case Permissions.can(:workflows, :access_read, user, project_user) do
+    case Permissions.can(:workflows, :access_read, user, project) do
       :ok -> :ok
       {:error, _reason} -> {:error, :forbidden}
     end
@@ -156,12 +154,7 @@ defmodule LightningWeb.API.AiAssistantController do
   defp check_unsaved_job_access(_, _user), do: :ok
 
   defp check_workflow_access(workflow, user) do
-    project_user =
-      Enum.find(workflow.project.project_users, fn pu ->
-        pu.user_id == user.id
-      end)
-
-    case Permissions.can(:workflows, :access_read, user, project_user) do
+    case Permissions.can(:workflows, :access_read, user, workflow.project) do
       :ok -> :ok
       {:error, _reason} -> {:error, :forbidden}
     end


josephjclark:
@elias-ba this is looking pretty good (and I agree that it works better with your apollo changes, even if I have some concerns about the changes themselves).

The big thing I have here is that the streaming text comes in really really fast, and in really big chunks. It's stressful and kinda hard to read.

Do we need to buffer the response from the model, tokenise it, and then render token by token on a throttle? The worry with that of course is that the markdown will render all weird. It's difficult that we have to both render text smoothly, but also incrementally render markdown. Ideally we'd background-generate a markdown block, then incrementally render the html output.

Very happy to cut cloth - maybe the current rendering is the most appropriate place to release. Just want to thoroughly discuss options and costs before moving forward.

elias-ba:
@josephjclark Thanks for the feedback on the streaming speed. I've added a word-by-word rendering buffer that smooths out the bursty TCP chunks into a readable typing effect. The text now drains one word at a time at a configurable pace.

The speed is controlled by a single constant in `assets/js/collaborative-editor/lib/AIChannelRegistry.ts`:

```typescript
private static readonly WORD_INTERVAL_MS = 50;
```

This is the delay in milliseconds between each word. Lower = faster, higher = slower. Some reference points:
- `30` too fast, ~33 words/sec
- `50` feels a bit better, it's the current value, ~20 words/sec
- `75` a bit slower, ~13 words/sec
- `100` maybe too slow, 10 words/sec (even though I like it a lot)

When the model finishes generating, the buffer continues draining at normal pace and then seamlessly transitions to the final message - no jarring dump of text.

I also removed the `▍` cursor character that was being injected into the markdown content.

Would appreciate if you could test and let me know if 50ms feels right or if you'd prefer a different pace.

assets/js/collaborative-editor/components/MessageList.tsx
```diff
@@ -562,45 +623,47 @@ export function MessageList({
...
-                          : 'text-gray-400 hover:text-gray-600'
-                      )}
-                      title={
-                        copiedMessageId === message.id
-                          ? 'Copied!'
-                          : 'Copy message'
-                      }
-                    >
-                      <span
+                  {!isStreaming(message) && (
```
josephjclark: Ah I was wondering a bit about this - so we hide the buttons at the bottom of the message until the streaming has finished?

It's kind weird when that stuff pops in at the bottom but I think this is correct.

Actually, would it work better to render this stuff invisibly so that we're saving screenspace for it? Remove the pop of it landing?

assets/js/collaborative-editor/lib/AIChannelRegistry.ts
```diff
@@ -129,11 +133,82 @@ export class AIChannelRegistry {
   // Short enough to not waste resources, long enough to handle accidental clicks
   private cleanupDelayMs = 2_000; // 2 seconds
 
+  // Word-by-word streaming buffer — smooths out bursty TCP chunks into
```
josephjclark, line 136: Ah so this is letter by letter now

assets/js/collaborative-editor/lib/AIChannelRegistry.ts
```diff
@@ -129,11 +133,82 @@ export class AIChannelRegistry {
...
+  // a readable typing effect. Drains one word per tick at a fixed interval.
+  private streamingBuffer = '';
+  private streamingDrainPos = 0;
+  private streamingDrainTimer: ReturnType<typeof setInterval> | null = null;
+  // ms between letters. 15ms ≈ 65 chars/sec.
+  private static readonly WORD_INTERVAL_MS = 15;
```
josephjclark: And this is maybe LETTER_INTERVAL_MS (that's my bad!)

elias-ba:
Thanks @josephjclark ! All addressed. The hooks bug was a `useMemo` sitting after an early return in MessageList - when the session cleared and messages became empty, React saw fewer hooks on re-render. Moved it before the return and that fixes it. Also renamed `WORD_INTERVAL_MS` to `LETTER_INTERVAL_MS` as you suggested, and changed the message footer to use `invisible` instead of conditional rendering during streaming so the space is reserved and there's no pop when it lands.

assets/js/collaborative-editor/components/MessageList.tsx
elias-ba: Nice catch and suggestion. Done

elias-ba:
> Nicely done @elias-ba

Thanks a lot @midigofrank. I think you forgot to hit that approve button lol

josephjclark:
One small problem I see now @elias-ba is that if you get a long message streaming down to you (or you have a  very small window), the page doesn't scroll down as it renders.

This is a fiddly behaviour because you probably need to NOT auto-scroll if the user scrolls. So maybe we'll spin this out into an issue to resolve separately?

josephjclark:
I've also just seen an issue where newline characters get rendered for a little while. It happened when rendering a bulleted list. It sorted itself out in the end but was a bit untidy there

Maybe try this query if you want to look at it:
```
hey what is the difference between each and Array.each? Give me a lot of detail please
```

josephjclark (approved):
I've found two minor issues, but I think this PR is now at the point where the UI is better WITH streaming than without it

I think we should ship this but raise issues for improvement later

elias-ba:
Thank you very much @josephjclark. I will look at those minor UI/UX stuff you flagged. Thanks for the eagle eyes. If easy to fix I will push a commit here before merging this PR. If not easy to fix I will open a follow up issue and merge this one. Thanks man

josephjclark:
@elias-ba let's just merge this - I'll raise out two issues now

josephjclark:
https://github.com/OpenFn/lightning/issues/4576
https://github.com/OpenFn/lightning/issues/4575

josephjclark:
great job @elias-ba !

PR #4531, opened 2026-03-15
Save final state for runs; flexible input for "next" cron run

### LINKED ISSUE #4485: Flexible "next input state" for cron-triggered runs
Sometimes users need to be able to set a "cursor" for the next cron run in a downstream step...  i.e., not in the very first step of the workflow. Currently, this is impossible since the "next input state" for a cron run is hard coded to be the output of the first step in the workflows' last run.

Giving the user more flexibility would allow them to deliver more use cases via OpenFn

Image

## Tech notes

See: https://github.com/OpenFn/lightning/blob/main/lib/lightning/invocation.ex#L219-L242

This function is important. When a workflow is run via cron trigger, each time it runs via cron trigger it takes the output state from the LAST successful run of the cron-triggered job (i.e., the first job/first step in the workflow) as its input. sometimes, users want to use the final state of the whole run (in a multi-step, multi-job workflow) as the input of the next cron run.

Please propose several ways that we could change the system so that a cron triggered workflow can take its next initial state from a final state of any specific STEP in the workflow (with the first step as the default and current behaviour for backwards compatibility) or from the final state of the whole RUN (useful if the workflow has multiple leaf nodes).

# Plan: Add `cron_cursor_job_id` to Cron Triggers

## Context

When a cron-triggered workflow runs, it needs to determine the input state for
the next run. Previously, the scheduler always used the output of the **first
job** (the cron edge's target) from the last successful step. Now that runs track
a `final_dataclip_id`, we can offer a better default: use the final state from
the last run.

This change adds a `cron_cursor_job_id` field to triggers so users can select
which job's output to use as the cron cursor:

- **`nil` (new default)** — use `final_dataclip_id` from the last run (the final
  output state of the entire workflow execution)
- **Specific job ID** — use the output of the last successful step for that job

A backfill migration sets `cron_cursor_job_id` on all existing cron triggers to
the target job of their cron edge, preserving their current behavior.

### 1. Migration — add column + backfill existing triggers

Create `priv/repo/migrations/TIMESTAMP_add_cron_cursor_job_id_to_triggers.exs`:

```elixir
alter table(:triggers) do
  add :cron_cursor_job_id,
    references(:jobs, type: :binary_id, on_delete: :nilify_all),
    null: true
end

# Backfill: set cron_cursor_job_id on all existing cron triggers to the
# target_job_id of their cron edge. This preserves current behavior for
# existing workflows (use first job's output) while letting NEW triggers
# default to nil (use final run state).
flush()

execute """
  UPDATE triggers
  SET cron_cursor_job_id = edges.target_job_id
  FROM edges
  WHERE edges.source_trigger_id = triggers.id
    AND triggers.type = 'cron'
"""
```

**Why the backfill matters:** Before this change, the scheduler always called
`last_state_for_job(target_job.id)` — using the first job's output. After this
change, `nil` means "use `final_dataclip_id` from the last run." Without the
backfill, every existing cron workflow would silently switch from using the first
job's output to using the final run state, which could break workflows that
depend on the first job's output as their cursor.

### 2. Trigger schema — `lib/lightning/workflows/trigger.ex`

- Add `belongs_to :cron_cursor_job, Lightning.Workflows.Job` to schema
- Add `:cron_cursor_job_id` to `cast_changeset/2` cast list
- In `validate_by_type/1` for `:webhook` and `:kafka`, clear it:
  `put_change(:cron_cursor_job_id, nil)`
- For `:cron`, leave it as-is (nullable, optional)

### 3. Snapshot embed — `lib/lightning/workflows/snapshot.ex`

- Add `field :cron_cursor_job_id, :binary_id` to the inline
  `embeds_many :triggers` block (~line 53)
- `@trigger_fields` auto-includes it from `Trigger.__schema__(:fields)` — no
  other change needed for cast

### 4. Scheduler — `lib/lightning/workflows/scheduler.ex`

Modify `invoke_cronjob/1` to branch on `cron_cursor_job_id`:

```elixir
defp invoke_cronjob(%Edge{target_job: job, source_trigger: trigger}) do
  with %{project_id: project_id} <- job.workflow,
       :ok <- WorkOrders.limit_run_creation(project_id) do
    case trigger.cron_cursor_job_id do
      nil ->
        # New default: use final_dataclip_id from the last run
        last_run_final_state(trigger)

      job_id ->
        # Explicit job selected: use that job's last successful step output
        last_state_for_job(job_id)
    end
    |> case do
      nil ->
        # No prior state found — use empty global dataclip
        WorkOrders.create_for(trigger,
          dataclip: %{type: :global, body: %{}, project_id: project_id},
          workflow: job.workflow
        )

      dataclip ->
        WorkOrders.create_for(trigger,
          dataclip: dataclip,
          workflow: job.workflow
        )
    end
  end
end
```

Add a new private function `last_run_final_state/1` that finds the most recent
successful run for the trigger's workflow and returns its `final_dataclip_id`:

```elixir
defp last_run_final_state(trigger) do
  # Find the last successful run for work orders created by this trigger,
  # and return its final_dataclip.
  from(r in Run,
    join: wo in assoc(r, :work_order),
    where: wo.trigger_id == ^trigger.id,
    where: r.state == :success,
    where: not is_nil(r.final_dataclip_id),
    order_by: [desc: r.finished_at],
    limit: 1,
    preload: [:final_dataclip]
  )
  |> Repo.one()
  |> case do
    nil -> nil
    run -> run.final_dataclip
  end
end
```

### 5. TypeScript types — `assets/js/collaborative-editor/types/trigger.ts`

- Add `cron_cursor_job_id: z.string().uuid().nullable().default(null)` to
  `cronTriggerSchema`
- Add `cron_cursor_job_id: z.null().default(null)` to `webhookTriggerSchema` and
  `kafkaTriggerSchema`
- Add `cron_cursor_job_id: null` to default values in `createDefaultTrigger` for
  all types

### 6. TriggerForm UI — `assets/js/collaborative-editor/components/inspector/TriggerForm.tsx`

Inside the `currentType === 'cron'` block, after the cron expression field, add
a new "Cron Input Source" select:

- Label: "Cron Input Source" with tooltip: "Select which step's output to use as
  the input for each cron-triggered run"
- Options:
  - `null` → "Final run state (default)" — uses `final_dataclip_id` from the
    last successful run
  - One option per job in the workflow, shown by name
- The field binds to `form.Field name="cron_cursor_job_id"`
- Populate job options from `useWorkflowState(state => state.jobs)`

### 7. Invocation UI helper — `lib/lightning/invocation.ex`

Update `get_next_cron_run_dataclip/2` and `list_dataclips_with_cron_state/3` to
accept/use the cursor job ID so the dataclip panel shows the correct "next cron
run" indicator. When `cron_cursor_job_id` is nil, the indicator should reflect
the final run state rather than a specific job's output.

### 8. Tests

- **Scheduler test**: Test that:
  - `cron_cursor_job_id = nil` uses `final_dataclip_id` from last successful run
  - `cron_cursor_job_id = some_job_id` uses that job's last successful step
    output (old behavior)
  - Both cases fall back to empty global dataclip when no prior run exists
- **Trigger changeset test**: Test new field is cast, cleared for non-cron types
- **Migration**: Verify via `mix ecto.migrate` / `mix ecto.rollback`
- **Backfill**: Verify existing cron triggers get `cron_cursor_job_id` set to
  their edge's target_job_id

## Files to modify

| File | Change |
|------|--------|
| `priv/repo/migrations/new` | New migration with column + backfill |
| `lib/lightning/workflows/trigger.ex` | Add field + cast |
| `lib/lightning/workflows/snapshot.ex` | Add field to embed |
| `lib/lightning/workflows/scheduler.ex` | Branch on `cron_cursor_job_id`, add `last_run_final_state/1` |
| `lib/lightning/invocation.ex` | Update UI dataclip helper |
| `assets/js/collaborative-editor/types/trigger.ts` | Add to Zod schemas |
| `assets/js/collaborative-editor/components/inspector/TriggerForm.tsx` | Add dropdown |
| `test/lightning/workflows/scheduler_test.exs` | Add test cases |
| `test/lightning/workflows/trigger_test.exs` | Add changeset tests |

## Verification

1. `mix ecto.migrate` — migration runs cleanly, existing cron triggers backfilled
2. `mix test test/lightning/workflows/scheduler_test.exs` — scheduler tests pass
3. `mix test test/lightning/workflows/trigger_test.exs` — trigger tests pass
4. `mix format && mix credo --strict --all` — no warnings
5. Start dev server, open workflow editor, select a cron trigger, verify dropdown
   appears with "Final run state (default)" and all jobs listed
6. Save workflow, verify `cron_cursor_job_id` persists and snapshot captures it
7. Verify an existing cron trigger shows the first job selected (from backfill),
   not "Final run state"

## For Consideration: Multi-leaf / Multi-path Workflow Warnings

When `cron_cursor_job_id` is `nil` (using final run state), the system relies on
`Run.final_dataclip_id` — a single value per run. However, some workflow
topologies can produce ambiguous or surprising final states:

1. **Multiple leaf nodes** — A workflow that branches into two terminal jobs
   (e.g., "Send to DHIS2" and "Send to Sheets") will have two final steps. Which
   one's output becomes `final_dataclip_id`?

2. **Multiple paths to one leaf node** — If a job can be reached via two
   different paths (e.g., success path and fallback path), the final dataclip
   depends on which path was taken.

### Possible approaches

- **UI warning**: When a user selects "Final run state" for a cron trigger and
  the workflow has multiple leaf nodes or converging paths, show a warning:
  _"This workflow has multiple possible final states. Consider selecting a
  specific step to avoid unpredictable cron input."_

- **Force explicit selection**: In workflows with ambiguous final states, disable
  the "Final run state" option entirely and require the user to pick a specific
  job.

- **Do nothing for now**: Document the behavior and let advanced users figure it
  out. The backfill ensures all existing workflows are pinned to a specific job
  anyway, so this only affects newly created triggers.

### How to detect

The workflow DAG is available in the editor. Leaf nodes are jobs with no outgoing
edges. The check would be: count leaf nodes in the workflow; if > 1, flag it.
For converging paths, check if any job has multiple incoming edges.

**Decision needed:** Which approach to take? The warning approach is the least
disruptive and still guides users toward correct configuration.

taylordowns2000:
Please look at this [super quick slide](https://docs.google.com/presentation/d/1iYYnQ-IKCmosB6p8gj04CcXl7BxQnzP3p_CnmSLhJTs/edit?slide=id.p#slide=id.p) show for quick context. When testing this PR, note that it goes _best_ with https://github.com/OpenFn/kit/pull/1306 so you can see the performance enhancements, but you're welcome to test on the current worker too.

## Overview

This PR saves the run's final state and lets users choose which step's output the cron scheduler uses as input for the next execution. Closes #4485.

Final state preservation: When the worker sends run:complete, we now persist the run's final state by setting `final_dataclip_id` on the runs table. Previously final_state was only used for webhook response broadcasting, then discarded. The handler accepts two mutually exclusive fields from the worker:
  - final_dataclip_id — reuse an existing step output dataclip (single-leaf workflows, no duplication)
  - final_state — a new map to persist as a separate dataclip (multi-leaf workflows with merged output)

  Neither field is required — backward compatible with the current worker.

  Cron input source: Each cron trigger now has an optional cron_cursor_job_id. When set, the scheduler uses that specific job's last successful step output. When null (the new default for fresh triggers), it uses the run's final_dataclip_id — the merged final state from above. A backfill migration points all existing cron triggers at their first downstream job, preserving current behavior.

##  Changes

  - Migration: Adds nullable final_dataclip_id FK on runs → dataclips
  - Migration: Adds nullable `cron_cursor_job_id` FK on triggers → jobs, with backfill for existing cron triggers
  - Run schema: New `belongs_to :final_dataclip`; `complete/2` casts the new field
  - CompleteRun handler: `resolve_final_dataclip/2` either reuses an existing ID or inserts a new dataclip. Respects `save_dataclips: false`. Wrapped in Repo.transact
  - Run channel: Injects project_id into the run:complete payload so new dataclips can be associated with the correct project
  - Trigger schema: New belongs_to :cron_cursor_job; cleared automatically when type changes to webhook/kafka
  - Scheduler: Replaced `last_state_for_job/1` with `Invocation.get_next_cron_run_dataclip/1`, which branches on `cron_cursor_job_id` — either `last_run_final_dataclip/1` (final run state) or `last_successful_step_dataclip/1` (specific job output)
  - Invocation: Refactored cron dataclip resolution to be trigger-aware instead of job-aware
  - TriggerForm UI: New "Cron Input Source" dropdown listing workflow jobs, with tooltip
  - ManualRunPanel: Fixed stale-closure bug in cron auto-selection; disableAutoSelection now respects `manuallyUnselectedDataclip` instead of always being true in fullscreen mode
  - Workflow serializer/snapshot: Round-trips `cron_cursor_job_id` through Y.Doc and snapshots

##  Performance considerations

  - Single-leaf workflows (most common): Worker sends `final_dataclip_id` pointing to the last step's output dataclip. Zero additional storage — just sets the FK.
  - Multi-leaf workflows: One extra dataclip INSERT per run. The merged body doesn't exist elsewhere, so this is net-new data.
  - Transaction scope: CompleteRun now wraps in Repo.transact. For the ID-reuse path this adds no extra queries. For the final_state path it's one INSERT + one UPDATE in a short transaction.
  - I was super worried about a race condition where the `step:complete` with the dataclip would be processed _after_ the `run:complete` with a reference to the dataclip ID, but Claude tells me this isn't gonna happen... in a channel all events are processed in order.

##  How to test

  1. Run any workflow — `final_dataclip_id` will be null until the worker is updated (backward compatible)
  2. After worker change: run a single-leaf workflow, verify `final_dataclip_id` equals the last step's `output_dataclip_id` (no duplicate)
  3. After worker change: run a branching workflow, verify `final_dataclip_id` points to a new dataclip with the merged output
  4. Open a cron-triggered workflow — the trigger inspector should show a "Cron Input Source" dropdown
  5. Default ("Final run state") uses the run's `final_dataclip_id`; selecting a specific job uses that job's last successful step output
  6. Verify the manual run panel auto-selects the correct next cron dataclip

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/test/lightning/projects_test.exs
+++ b/test/lightning/projects_test.exs
@@ -897,6 +897,82 @@ defmodule Lightning.ProjectsTest do
       assert Repo.get(Lightning.Workflows.Workflow, workflow.id)
     end
 
+    test "deletes runs used as cron input sources" do
+      project = insert(:project, history_retention_period: 7)
+
+      %{triggers: [trigger], jobs: [job | _rest]} =
+        workflow = insert(:simple_workflow, project: project)
+
+      now = Lightning.current_time()
+      snapshot = insert(:snapshot, workflow: workflow)
+
+      # Mode A: trigger with cron_cursor_job_id: nil; run has final_dataclip_id set
+      assert is_nil(trigger.cron_cursor_job_id)
+
+      final_dataclip = insert(:dataclip, project: project)
+
+      workorder_a =
+        insert(:workorder,
+          workflow: workflow,
+          last_activity: Timex.shift(now, days: -8),
+          trigger: trigger,
+          dataclip: build(:dataclip),
+          snapshot: snapshot,
+          runs: [
+            build(:run,
+              state: :success,
+              starting_trigger: trigger,
+              dataclip: build(:dataclip),
+              final_dataclip: final_dataclip,
+              finished_at: Timex.shift(now, days: -8),
+              steps: [build(:step, job: job)]
+            )
+          ]
+        )
+
+      # Mode B: trigger with cron_cursor_job_id pointing to a job;
+      # run has a step with output_dataclip_id set
+      {:ok, trigger} =
+        trigger
+        |> Ecto.Changeset.change(cron_cursor_job_id: job.id)
+        |> Repo.update()
+
+      output_dataclip = insert(:dataclip, project: project)
+
+      workorder_b =
+        insert(:workorder,
+          workflow: workflow,
+          last_activity: Timex.shift(now, days: -8),
+          trigger: trigger,
+          dataclip: build(:dataclip),
+          snapshot: snapshot,
+          runs: [
+            build(:run,
+              state: :success,
+              starting_trigger: trigger,
+              dataclip: build(:dataclip),
+              finished_at: Timex.shift(now, days: -8),
+              steps: [
+                build(:step,
+                  job: job,
+                  exit_reason: "success",
+                  output_dataclip: output_dataclip
+                )
+              ]
+            )
+          ]
+        )
+
+      assert :ok =
+               Projects.perform(%Oban.Job{args: %{"type" => "data_retention"}})
+
+      refute Repo.get(WorkOrder, workorder_a.id)
+      refute Repo.get(Run, hd(workorder_a.runs).id)
+
+      refute Repo.get(WorkOrder, workorder_b.id)
+      refute Repo.get(Run, hd(workorder_b.runs).id)
+    end
+
     test "does not incorrectly delete runs that reference older snapshots" do
       project = insert(:project, history_retention_period: 7)
 


midigofrank:
@taylordowns2000 is this ready for review?

taylordowns2000:
> @taylordowns2000 is this ready for review?

It is, @midigofrank , but Stu's doing it with me this morning. Thanks for looking out :-)

PR #4542, opened 2026-03-19
Refactor runs-related cascades (because run is king from an audit perspective)

### LINKED ISSUE #4538: Change cascade delete relationships around runs/dataclips
# Cascade Relationships Audit: work_orders, runs, steps, run_steps, dataclips

### work_orders

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `workflow_id` | workflows | `CASCADE` | NOT NULL | **Fine.** Delete a workflow → delete its work orders. Correct ownership. |
| `trigger_id` | triggers | `SET NULL` | NULL | **Fine.** Trigger deleted → work order survives, FK nulled. |
| `dataclip_id` | dataclips | `SET NULL` | NULL | **Fine.** Already fixed in 2023. Dataclip deleted → work order survives. |
| `snapshot_id` | workflow_snapshots | `RESTRICT` | NULL | **Fine.** Already fixed in 2025. Can't delete a snapshot that's in use. |

### runs

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `work_order_id` | work_orders | `CASCADE` | NOT NULL | **Fine.** Delete a work order → delete its runs. Correct ownership. |
| `dataclip_id` | dataclips | `CASCADE` | NOT NULL | **Problem.** Delete a dataclip → deletes the run. Should be `SET NULL`. |
| `starting_trigger_id` | triggers | `CASCADE` | NULL | **Questionable.** Delete a trigger → deletes all runs that started from it. Destroys audit trail. |
| `starting_job_id` | jobs | `CASCADE` | NULL | **Questionable.** Delete a job → deletes all runs that started from it. Destroys audit trail. |
| `created_by_id` | users | `CASCADE` | NULL | **Questionable.** Delete a user → deletes all runs they created. Destroys audit trail. |
| `snapshot_id` | workflow_snapshots | `RESTRICT` | NULL | **Fine.** Can't delete a snapshot in use. |

### steps

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `job_id` | jobs | `NO ACTION` | NOT NULL | **Fine.** Already fixed in 2024. Job deletion blocked if steps reference it. |
| `credential_id` | credentials | `NO ACTION` | NULL | **Fine.** Credential deletion blocked if steps reference it. |
| `input_dataclip_id` | dataclips | `CASCADE` | NOT NULL | **Problem.** Delete a dataclip → deletes the step. Should be `RESTRICT`. |
| `output_dataclip_id` | dataclips | `CASCADE` | NULL | **Problem.** Delete a dataclip → deletes the step. Should be `SET NULL`. |
| `snapshot_id` | workflow_snapshots | `RESTRICT` | NULL | **Fine.** Already fixed in 2025. |

### run_steps (join table)

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `run_id` | runs | `CASCADE` | NOT NULL | **Fine.** Delete a run → delete the join record. |
| `step_id` | steps | `CASCADE` | NOT NULL | **Fine.** Delete a step → delete the join record. |

### log_lines

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `run_id` | runs | `CASCADE` | NULL | **Fine.** Delete a run → delete its logs. |
| `step_id` | steps | `CASCADE` | NULL | **Fine.** Delete a step → delete its logs. |

### dataclips

| Column | → References | on_delete | Nullable | Verdict |
|--------|-------------|-----------|----------|---------|
| `project_id` | projects | `CASCADE` | NOT NULL | **Fine.** Delete a project → delete its dataclips. |

---

### Definitely wrong (3 constraints)

| FK | Current | Should Be | Business Reason |
|----|---------|-----------|-----------------|
| `runs.dataclip_id` | `CASCADE` | `SET NULL` + nullable | Deleting an input dataclip shouldn't destroy the run's audit record. The 7-day dataclip wipe / 30-day history retention scenario makes this concrete — if anyone ever `DELETE`s instead of `UPDATE`s during wipe, you lose 23 days of run history. |
| `steps.input_dataclip_id` | `CASCADE` | `RESTRICT` | A step's existence proves work happened. Deleting the input data it consumed shouldn't erase that proof. `RESTRICT` forces correct deletion order (steps first, then dataclips) which the cleanup code already follows. |
| `steps.output_dataclip_id` | `CASCADE` | `SET NULL` | Same reasoning — the step is the primary audit record. If its output dataclip gets cleaned up, the step should survive with a null output reference. |

### Worth discussing (3 constraints on runs)

| FK | Current | Could Be | Tradeoff |
|----|---------|----------|----------|
| `runs.starting_trigger_id` | `CASCADE` | `SET NULL` | Currently, deleting a trigger wipes all runs that started from it. If a trigger is reconfigured or removed during workflow editing, that's audit trail destruction. But triggers are soft-deleted via snapshots, so this may never fire in practice. |
| `runs.starting_job_id` | `CASCADE` | `SET NULL` | Same — deleting a job wipes all manual-run history. Again, snapshots should prevent this in practice, but the constraint is a landmine. |
| `runs.created_by_id` | `CASCADE` | `SET NULL` | Deleting a user destroys all runs they manually created. This is almost certainly wrong — when someone leaves an organization, you don't want their run history to vanish. |

### Already correct (everything else)

The ownership chains are right:
- `workflow → work_orders → runs → run_steps` all cascade in the correct parent-owns-child direction
- `run/step → log_lines` cascade correctly (parent deletion cleans up child logs)
- `project → dataclips` cascade correctly
- Snapshots use `RESTRICT` (already fixed in a 2025 migration)
- `steps.job_id` and `steps.credential_id` use `NO ACTION` (already fixed in 2024)

---

### Tier 1 — Fix now (the 3 dataclip cascades)

These invert the ownership model and create a real risk during data retention. Dataclips are the _content_ consumed/produced by runs and steps — they are not the primary audit record. The cascade direction should respect that hierarchy.

### Tier 2 — Fix soon (`runs.created_by_id`)

Clear business impact: removing a user from the system shouldn't destroy run history. Changing to `SET NULL` is low-risk.

### Tier 3 — Low urgency (`runs.starting_trigger_id`, `runs.starting_job_id`)

The snapshot system means triggers and jobs are effectively never hard-deleted while runs reference them. These cascades are technically wrong but unlikely to fire in practice. Worth fixing for correctness but not urgent.

The hook keeps firing because `.context/` is untracked. Since CLAUDE.md says never commit it, I'll add it to `.gitignore` to silence the hook.

taylordowns2000:
## Description

This PR enforces the "run is king" principle at the database level and handles a UI edge case when referenced resources have been deleted.

###  Database constraints

 - 4 existing CASCADE constraints changed to RESTRICT: `runs.dataclip_id`, `runs.created_by_id`, `steps.input_dataclip_id`, `steps.output_dataclip_id` — deleting a dataclip or user that is still referenced by runs/steps is now rejected by the DB
 - `starting_job_id` and `starting_trigger_id` are intentionally left as bare UUID columns with no FK constraint — the snapshot system preserves full job/trigger data for every run, so live rows can be freely deleted from workflows without affecting audit history
 - 3 indexes added on `runs.created_by_id`, `runs.starting_trigger_id`, `runs.starting_job_id` for query performance

###  Application changes

 - `Accounts.delete_user/1` now returns `{:error, changeset}` with a descriptive message instead of cascading deletes when the user has associated runs
 - Added a comment on Run schema explaining why `starting_job` and `starting_trigger` have no FK constraint

###  UI: deleted node handling

Edit: to be addressed in a later PR, maybe

 - ~The collaborative workflow editor now detects when a URL references a job or trigger that no longer exists in the workflow (e.g., navigating from an old run to a since-deleted node) and shows a "Resource not found" warning notification instead of silently selecting nothing~

###  Tests

 - Replaced the "removes any associated Run and RunStep records" and "removes any associated LogLine records" tests with a single "cannot delete a user who has associated runs" test that asserts the new RESTRICT behavior

Closes #4538

##  Validation steps

1. Run `mix ecto.migrate` and confirm migration succeeds
2. Run `mix ecto.rollback` and `mix ecto.migrate` to confirm clean round-trip
3. Run the purge/deletion tests: `mix test test/lightning/setup_utils_test.exs:604 test/lightning/setup_utils_test.exs:631 test/lightning/projects_test.exs:237 test/lightning/projects_test.exs:802 test/lightning/projects_test.exs:900 test/lightning/accounts_test.exs:599 test/lightning/runs_test.exs:376 test/lightning/runs_test.exs:993`
4. Run the full test suite to check for unexpected RESTRICT violations

  1. The migration uses @disable_ddl_transaction with NOT VALID + VALIDATE CONSTRAINT to avoid blocking writes on large tables. The ADD ... NOT VALID is instant (metadata only), and VALIDATE takes a SHARE UPDATE EXCLUSIVE lock that allows concurrent reads and writes.
  2. All existing deletion code already deletes in the correct order (steps before dataclips, work_orders before triggers/jobs), so no application code changes were needed beyond the delete_user changeset. The RESTRICT constraints now enforce that ordering — any future code that gets it wrong will get a constraint violation instead of silently cascading away audit history.
  3. The `starting_trigger_id` and `starting_job_id` FK constraints were silently dropped by the `20240405131241_add_snapshot_to_workorders` migration (it used modify `:starting_trigger_id`, `:binary_id` which strips the FK). This PR intentionally does not restore them — instead, these columns are left as bare UUIDs because the snapshot system makes the FK unnecessary, and omitting it allows jobs/triggers to be freely deleted from workflows.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to
 
 ### Fixed
 
+- Proper warn & error for exceeding max dataclip size
+  [#4524](https://github.com/OpenFn/lightning/issues/4524)
 - Copying api tokens doesn't work on unsecure non-localhost contexts
   [PR#4551](https://github.com/OpenFn/lightning/pull/4551)
 - Fix AI assistant authorization for support users on projects with support
--- a/assets/js/collaborative-editor/api/dataclips.ts
+++ b/assets/js/collaborative-editor/api/dataclips.ts
@@ -157,8 +157,15 @@ export async function submitManualRun(
   );
 
   if (!response.ok) {
-    const error = (await response.json()) as { error?: string };
-    throw new Error(error.error || 'Failed to submit manual run');
+    if (response.status === 413) {
+      throw new Error('Dataclip is too large. Reduce the size and try again.');
+    }
+    let errorMessage = 'Failed to submit manual run';
+    try {
+      const error = (await response.json()) as { error?: string };
+      errorMessage = error.error || errorMessage;
+    } catch {}
+    throw new Error(errorMessage);
   }
 
   return response.json() as Promise<ManualRunResponse>;
--- a/assets/js/collaborative-editor/components/ManualRunPanel.tsx
+++ b/assets/js/collaborative-editor/components/ManualRunPanel.tsx
@@ -10,7 +10,9 @@ import { useURLState } from '#/react/lib/use-url-state';
 import _logger from '#/utils/logger';
 
 import { FilterTypes } from '../../manual-run-panel/types';
-import CustomView from '../../manual-run-panel/views/CustomView';
+import CustomView, {
+  DEFAULT_MAX_DATACLIP_SIZE_BYTES,
+} from '../../manual-run-panel/views/CustomView';
 import EmptyView from '../../manual-run-panel/views/EmptyView';
 import ExistingView from '../../manual-run-panel/views/ExistingView';
 import type { Dataclip } from '../api/dataclips';
@@ -19,6 +21,7 @@ import { RENDER_MODES, type RenderMode } from '../constants/panel';
 import { useActiveRun, useFollowRun } from '../hooks/useHistory';
 import { useRunRetry } from '../hooks/useRunRetry';
 import { useRunRetryShortcuts } from '../hooks/useRunRetryShortcuts';
+import { useAppConfig } from '../hooks/useSessionContext';
 import { useCanRun } from '../hooks/useWorkflow';
 import { useKeyboardShortcut } from '../keyboard';
 import type { Workflow } from '../types/workflow';
@@ -80,6 +83,10 @@ export function ManualRunPanel({
   customBody: customBodyProp,
   disableAutoSelection = false,
 }: ManualRunPanelProps) {
+  const appConfig = useAppConfig();
+  const maxDataclipSizeBytes =
+    appConfig?.max_dataclip_size_bytes ?? DEFAULT_MAX_DATACLIP_SIZE_BYTES;
+
   const [selectedTabInternal, setSelectedTabInternal] =
     useState<TabValue>('empty');
   const [selectedDataclipInternal, setSelectedDataclipInternal] =
@@ -209,6 +216,7 @@ export function ManualRunPanel({
     onRunSubmitted: onRunSubmitted,
     edgeId: edgeId || null,
     workflowEdges: workflow.edges,
+    maxDataclipSizeBytes,
   });
 
   const followedRunStep = useMemo(() => {
@@ -473,6 +481,7 @@ export function ManualRunPanel({
             }
           }}
           renderMode={renderMode}
+          maxDataclipSizeBytes={maxDataclipSizeBytes}
         />
       )}
       {selectedTab === 'existing' && (
--- a/assets/js/collaborative-editor/hooks/useRunRetry.ts
+++ b/assets/js/collaborative-editor/hooks/useRunRetry.ts
@@ -75,6 +75,7 @@ export interface UseRunRetryOptions {
   onRunSubmitted: ((runId: string, dataclip?: Dataclip) => void) | undefined;
   edgeId: string | null;
   workflowEdges?: Workflow.Edge[];
+  maxDataclipSizeBytes?: number;
 }
 
 export interface UseRunRetryReturn {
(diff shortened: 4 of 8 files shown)


taylordowns2000:
Will tidy up the failures now. Thank you!

Then... I think the little nice-to-have I tried to add here (warn users when they have an impossible trigger/job IDs in the url) is more trouble than it's worth, and definitely out of scope for this set of DB changes. I'll remove all of that.

PR #4544, opened 2026-03-19
Hotfix: new snapshots for updated cron workflows 

### LINKED ISSUE #4545: Backwards compatibility cron-triggered workflow migration doesn't snapshot
We can't just update the workflow. We've got to trigger a new snapshot also.

Here's the insufficient migration:
```ex
defmodule Lightning.Repo.Migrations.AddCronCursorJobIdToTriggers do
  use Ecto.Migration

  def up do
    alter table(:triggers) do
      add :cron_cursor_job_id,
          references(:jobs, type: :binary_id, on_delete: :nilify_all),
          null: true
    end

    flush()

    # Backfill existing cron triggers to preserve current behavior:
    # point at the first downstream job so the old per-step lookup is used.
    execute("""
    UPDATE triggers
    SET cron_cursor_job_id = workflow_edges.target_job_id
    FROM workflow_edges
    WHERE workflow_edges.source_trigger_id = triggers.id
      AND triggers.type = 'cron'
    """)
  end

  def down do
    alter table(:triggers) do
      remove :cron_cursor_job_id
    end
  end
end
```

taylordowns2000:
Creates new snapshots for workflows with cron triggers that have a `cron_cursor_job_id` set.The previous migration added cron_cursor_job_id to triggers and backfilled it, but existing snapshots were created before that column existed. Since runs execute against snapshots, the cron_cursor_job_id was effectively invisible at runtime until a new snapshot is captured.

Closes #4545

## Validation steps

1. Run the migration
2. Verify that new snapshots are created for workflows with cron triggers

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/.claude/agents/phoenix-elixir-expert.md
+++ b/.claude/agents/phoenix-elixir-expert.md
@@ -29,6 +29,7 @@ You are a **battle-tested Elixir/Phoenix architect** with deep expertise in the 
 - Implement complex transactional operations using Ecto.Multi
 - Optimize performance with indexes, query planning, and schemaless queries
 - Handle migrations safely in production environments
+- Don't use application code in migrations; use pure SQL
 
 **Lightning Platform Specialization:**
 - Work within Lightning's DAG-based workflow architecture


taylordowns2000:
@midigofrank rejected: 
```
Can we ask claude to generate an SQL instead?
[5:26 PM]      |> Workflow.touch()
      |> repo().update!()
      |> Snapshot.create()

These might break in future incase we change the functionality of .touch and  Snapshot.create
```

midigofrank:
Wait don't merge yet @taylordowns2000 . I need to confirm if the positions should stay `null` or be empty

priv/repo/migrations/20260319155941_re_snapshot_cron_cursor_job_id.exs
```diff
@@ -0,0 +1,97 @@
...
+
+    # Create new snapshots for those workflows
+    execute("""
+    INSERT INTO workflow_snapshots (id, workflow_id, name, lock_version, positions, jobs, triggers, edges, inserted_at)
+    SELECT
+      gen_random_uuid(),
+      w.id,
+      w.name,
+      w.lock_version,
+      COALESCE(w.positions, '{}'::jsonb),
```
midigofrank:
```suggestion
      w.positions,
```

The positions should be copied as is. `null` means `no positions`

PR #4562, opened 2026-03-24
Fix version-stuck: reconcile SharedDocs via Collaborate.start after provisioner import

### LINKED ISSUE #4535: Merging sandbox into workflow with unsaved changes sometimes causes stuck version
**Describe the bug** 

When merging a sandbox into a workflow that has unsaved changes, we are **sometimes** seeing a bug where the version number gets stuck. There is a workaround. Users can save the unsaved changes that are still hanging and then re-apply their changes. 

Update: this same issue has been reported when a user deploys via OpenFn CLI. 

If you're running on a tagged version, include the number: `v2.15.16`

- [x] Yes
- [ ] No

It's not easy to reproduce. Doesn't happen in every scenario but it is somewhere around this: 
1. User creates a new workflow
2. User creates a sandbox of the workflow
3. User adds a new job in the original workflow and doesn't save. Keep the tab open. Call this tab A.
4. In a new tab (tab B), user goes to sandbox and adds a new job in the sandbox. Saves.
5. Merge the sandbox into main in tab B.
6. In tab B, open the workflow.
7. Observe that it loads with v1 rather than latest but the URL has no v= parameter
8. Tab A still shows "latest"
9. In Tab A, if you click on v1, it will show you actually what v1 is (the workflow before we added any job)
10. In Tab A, if they click latest, it updates to v1 and shows the workflow of v1 with unsaved changes on top. 
11. You can't get back to latest unless you click "latest" and click Save. 

**Expected behavior** Version shouldn't get stuck. 

lmac-1:
## Description

This PR fixes a class of bugs where the collaborative editor ends up stuck on a stale version of a workflow after an external write (sandbox merge or CLI deploy).

### The bug

When the provisioner (or sandbox merge) writes a new workflow version to the database, live editor sessions are never notified. Two failure modes:
1. **Someone is online:** the SharedDoc stays alive but is never updated. Users with the workflow open see the pre-merge state, and their next save silently overwrites the provisioner's work. The unsaved-changes indicator (red dot) stays on permanently because `baseWorkflow` is never refreshed.
2. **Nobody is online:** the stale `DocumentState` (the persisted Y.Doc snapshot) is loaded the next time a user opens the workflow, giving them the pre-merge state. The correct DB content is ignored.

### The fix

Both cases now go through the same code path.

After the provisioner's DB transaction commits, `WorkflowReconciler.reconcile_workflow_from_db/2` is called for each updated workflow. The reconciler calls `Collaborate.start/1` (joining an existing live SharedDoc if one is running, starting a new one if not), applies a state-based diff via `apply_db_state_to_doc`, then calls `Session.stop`. The shutdown chain flushes the corrected state to `DocumentState` before any real user opens the workflow.

After `Session.stop`, the reconciler broadcasts `:workflow_updated_externally` on PubSub. Each connected `WorkflowChannel` process receives this and pushes a `workflow_saved` event to its client — the existing frontend handler updates `baseWorkflow` and clears the unsaved-changes indicator. No frontend changes were needed.

**Supporting changes:**

- `Collaborate.start/1` now retries once (after a 1ms sleep) if `Session.init` returns `:shared_doc_not_found`. This handles a 0ms auto_exit race: when the reconciler's `Session.stop` triggers the last observer to leave, the SharedDoc dies via a 0ms timer. A subsequent `Collaborate.start` can find a dying process still registered in `:pg` before the timer fires. The retry absorbs this and all callers benefit.

- `Session` now accepts a `ProjectRepoConnection` as the `user` option. Presence tracking is guarded by `is_struct(user, User)` — system actors skip it entirely.

- The previous approach (`invalidate_document_states` inside the provisioner transaction, which deleted stale `DocumentState` rows) is removed. It only handled the nobody-online case and required the provisioner to know about the online/offline distinction.

Closes #4535

## Validation steps

1. **Someone online during a sandbox merge:**
   - Open a workflow in the editor (Tab A)
   - Make a sandbox for the project in Tab B
   - In Tab A, make some changes to the workflow but don't save.
   - In Tab B, make a different change to the workflow, save and then merge the sandbox.
   - Confirm Tab A's Y.Doc is updated in real time (new jobs/edges appear) **unsaved changes will be deleted, this is expected**
   - Confirm the unsaved-changes indicator (red dot) clears after the merge
   - Confirm saving from Tab A produces the merged state, not the pre-merge state

2. **Nobody online during a sandbox merge / CLI deploy:**
   - Make sure the workflow has some unsaved changes and then close it
   - Ensure no sessions are open for the workflow (close all tabs)
   - Run a sandbox merge or `Provisioner.import_document` with changes
   - Open the workflow in a new tab
   - Confirm the new tab shows the merged state, not the pre-merge state

3. **GitHub sync (ProjectRepoConnection actor):**
   - Trigger a GitHub sync that updates a workflow
   - Confirm the same behaviour as steps 1 and 2 depending on whether anyone
     was online

1. @stuartc, if the reconciler part is not correct, I might need you to jump in to help a bit as it's a bit over my head and I fear we will be going in circles.
2. At the moment, if a user is active with unsaved changes when an external change happens, their changes will be lost. This PR does not address this but can be discussed in a follow-up issue to improve the user experience. 
3. **Commits 1–5 on this branch are the old approach** (deleting `DocumentState` rows in the provisioner transaction). They are superseded by commits 6–7. The final state of the code is what matters; the history reflects the iteration. Happy to squash if you'd prefer a cleaner history.
4. **Concurrent sandbox merges (not fixed here):** For CLI deployments (`allow_stale: false`), DB optimistic locking ensures at most one provisioner succeeds per workflow — so at most one reconciler ever runs. For sandbox merge (`allow_stale: true`), two concurrent merges to the same workflow could both succeed and both run a reconciler. `Collaborate.start/1` has a pre-existing TOCTOU race (check-then-act on the SharedDoc lookup) that could cause one reconciler to fail silently; the other would still complete correctly. This is not introduced by this PR and is a realistic edge case only under concurrent sandbox merges. A follow-on improvement would be for the reconciler to re-load the workflow from DB by ID rather than using the struct passed in at transaction time, so the last writer always applies current state.
5. **No frontend changes.** The existing `workflowSavedHandler` in`createSessionContextStore.ts` already handles `workflow_saved` by calling`setBaseWorkflow()` and `setLatestSnapshotLockVersion()`.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lmac-1:
@stuartc would love to hear your thoughts on everything! I will be tying up test cases and doing some more testing tomorrow.

lmac-1:
@stuartc I am doing my own review on this now and adding tests for provisioner. Feel free to shout if there's something alarming in the meantime.

lmac-1:
I've completed testing of the CLI ✅

lib/lightning/collaboration/workflow_reconciler.ex
```diff
@@ -50,6 +53,365 @@ defmodule Lightning.Collaboration.WorkflowReconciler do
...
+      [
+        {"name", db_job.name || ""},
+        {"adaptor", db_job.adaptor},
+        {"project_credential_id", db_job.project_credential_id},
+        {"keychain_credential_id", db_job.keychain_credential_id}
+      ]
+      |> Enum.flat_map(fn {key, db_val} ->
+        case Yex.Map.fetch(ydoc_job, key) do
+          {:ok, ^db_val} -> []
+          _ -> [{:set_field, ydoc_job, key, db_val}]
```
lmac-1, line 243: What came up in my review: 

The `_ ->` branch can't distinguish "provisioner changed this" from "user has unsaved edits here". DB is truth.

This is acknowledged in the PR (note 2). The correct long-term fix would require the provisioner to write *through* the Y.Doc, producing real CRDT operations so the CRDT can merge provisioner intent with user unsaved changes. That's a much bigger change and is tracked as a follow-up.

lmac-1:
@elias-ba @stuartc I have not manually tested the validation steps for GitHub sync that updates a workflow still. Will work out how to test this properly later this morning.

lmac-1:
@stuartc I synced with @elias-ba on this and he flagged about the issue of losing unsaved changes for a user actively editing (which we are already aware of). He's going to dive into the code more and share his thoughts.

I think something to bear in mind when reviewing this is to make sure that it won't be too difficult to add something on top of it to make unsaved changes work more nicely (like the rebase we discussed as a follow-on issue). At the very least we might want to notify a user via toast that their workflow has been updated due to external change.

He also suggested that we should maybe sync with Joe next week to talk through next steps / priority / if there are next steps as he might have some opinions or thoughts about this already.

Issue #4574, opened 2026-03-27
Abort or interrupt a running run


taylordowns2000:
I'd like to be able to cancel a run that's already claimed, started, or in progress. Sometimes something goes wrong and you know you want to stop it right away, but a long run might take 45 minutes to complete or, worse yet, a crashed worker might not lead to your run being marked as "lost" for 60 minutes.

## Proposal
1. Click into a running run, and you should find a "cancel" button.
2. Select a bunch of work orders from history in state "running" and you should be able to cancel any associated runs for those work orders (in the same way you can cancel `pending` runs in #1622 )
3. Doing so should send a new event to the worker, telling it to abort, cease & desist, etc., etc.
4. The worker should then send back some tidying up data - last logs, final state with a cancellation reason, etc.
5. Basically, we don't want to shut the door from the lightning side... we want to tell the worker to stop and clean up immediately.

taylordowns2000:
@jfeldman-openfn , @AishaHassen , here's the other issue btw.

PR #4578, opened 2026-03-28
Cancel available runs

### LINKED ISSUE #1622: Cancel runs individually or in bulk
I'd like to be able to cancel runs if they are enqueued, claimed, started, or in progress. Sometimes something goes wrong and you _know_ you want to stop it right away, but a long run might take 45 minutes to complete or, worse yet, a crashed worker might not lead to your run being marked as "lost" for 60 minutes.

1. Click into a running run, and you should find a "cancel" button.
2. Select a bunch of work orders from history in state "pending" or "running" and you should be able to cancel any associated runs for those work orders that are in a pending state.

taylordowns2000:
## Description

Closes #1622 by allowing users to cancel available runs, essentially removing them from the queue (`:available -> :cancelled`) and changing the state of their associate work orders from `:pending -> :cancelled`.

**Action availability by work order state:**

| WO State                                                                                                          | Available Actions   |
| ----------------------------------------------------------------------------------------------------------------- | ------------------- |
| `:pending` (runs `:available`)                                                                                    | Cancel              |
| `:running`                                                                                                        | None (all disabled) |
| All final states (`:success`, `:failed`, `:crashed`, `:killed`, `:exception`, `:lost`, `:cancelled`, `:rejected`) | Retry, Retry from   |

## Validation steps

(Probably easiest to start your server with `RTM=false iex -S mix phx.server` so runs don't run.)

1. Enqueue a lot of runs
2. Cancel all runs related to a work order one by clicking on the work order row button
3. Cancel a single run by clicking on the run item button (both in the list and on the run detail page)
4. Select multiple work orders and cancel all runs for those selected by clicking on the "Cancel" multi-item action button.
5. Select work orders with mixed states and note that you can't retry or cancel them.
6. Select work orders with only running states and note that you can't retry or cancel them.
7. Select work orders with non-running, non-pending states and note that you can retry them.
8. Select work orders with pending states and note that you can cancel them.
9. Select a work order with pending state, open the cancel modal, let it start running, and then confirm the cancellation; note it doesn't cancel it.
10. Navigate to the run detail page for an enqueued run. Cancel it via the button.

1. ‼️ Yes, I've also removed `FifoQueue.dequeue` and `Run.dequeue` since these functions _deleted_ runs! They were in the code base, but (thankfully) never used. You can't delete runs.
2. ⚠️ Note that Stu and I discussed "requeueing" runs, but in the end I realized that this breaks the "one-way state machine" that's fundamental to our audit trail. Left it out for now to obey the current rules; we can add it as a feature later with more thought to audit compliance.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning/work_orders.ex
+++ b/lib/lightning/work_orders.ex
@@ -509,7 +509,8 @@ defmodule Lightning.WorkOrders do
   For small batches, cancels synchronously via atomic UPDATE. Returns
   `{:ok, count}` where count is the number of runs cancelled.
   """
-  @spec cancel_many([WorkOrder.t()], keyword()) :: {:ok, non_neg_integer()}
+  @spec cancel_many([WorkOrder.t()], keyword()) ::
+          {:ok, non_neg_integer()} | {:error, any()}
   def cancel_many([], _opts), do: {:ok, 0}
 
   def cancel_many([%WorkOrder{} | _rest] = work_orders, opts) do


taylordowns2000:
> Thanks. I approve, but would like to request that before you wrap up, you create a separate ticket describing the work that is needed to _fully_ support the cancel operation for jobs that are both in the queue _and_ running. To me this only gets us halfway.

So far this has been tracked here: https://github.com/OpenFn/lightning/issues/4574

I feel like that's what we need. (And we could prio this fairly high, if you're keen to have the "whole" solution sooner rather than later.)

lib/lightning_web/live/run_live/index.ex
```diff
@@ -540,6 +556,111 @@ defmodule LightningWeb.RunLive.Index do
...
+  def handle_event("cancel-run", %{"run_id" => run_id}, socket) do
+    if socket.assigns.can_run_workflow do
+      case Runs.get(run_id) do
+        nil ->
+          {:noreply, put_flash(socket, :error, "Run not found.")}
+
+        run ->
+          case Runs.cancel_run(run) do
+            {:ok, _run} ->
+              {:noreply, put_flash(socket, :info, "Run cancelled.")}
```
elias-ba: Could this allow cancelling a run from another project? `Runs.get/1` is unscoped - it fetches any run by ID regardless of project. The `cancel` (work order) path is safe because `cancel_available_for_work_orders` joins on `w.project_id == ^project_id`, but this path doesn't have an equivalent check. Would something like verifying `Runs.get_project_id_for_run(run) == socket.assigns.project.id` before calling `cancel_run` close this gap?

lib/lightning_web/live/run_live/show.ex
```diff
@@ -341,6 +364,46 @@ defmodule LightningWeb.RunLive.Show do
...
+  def handle_event("cancel-run", %{"run_id" => run_id}, socket) do
+    if socket.assigns.can_run_workflow do
+      case Runs.get(run_id) do
+        nil ->
+          {:noreply, put_flash(socket, :error, "Run not found.")}
+
+        run ->
+          case Runs.cancel_run(run) do
+            {:ok, _run} ->
+              {:noreply, put_flash(socket, :info, "Run cancelled.")}
```
elias-ba: Same question as before, should we verify the run belongs to `socket.assigns.project` before cancelling? A crafted event with a run_id from another project would pass the `can_run_workflow` check (which only validates the user's role in *their* project).

lib/lightning_web/live/run_live/index.ex
```diff
@@ -631,6 +752,62 @@ defmodule LightningWeb.RunLive.Index do
...
+
+    case work_orders do
+      [] ->
+        {:ok, 0}
+
+      _ ->
+        WorkOrders.cancel_many_async(
+          work_orders,
+          project_id: socket.assigns.project.id
+        )
```
elias-ba: Should we pattern match on the return value here? If `Oban.insert` fails for some reason, the caller still returns `{:async, count}` and the user sees "Cancelling runs... in the background" even though nothing was enqueued.

lib/lightning_web/live/run_live/index.ex
```diff
@@ -540,6 +556,111 @@ defmodule LightningWeb.RunLive.Index do
      |> put_flash(:error, error_message)}
   end
 
+  def handle_event("bulk-cancel", attrs, socket) do
+    redirect_to =
+      ~p"/projects/#{socket.assigns.project.id}/history?#{%{filters: socket.assigns.filters}}"
+
+    with true <- socket.assigns.can_run_workflow,
+         result <- handle_bulk_cancel(socket, attrs) do
```
elias-ba: Is the bare `result <-` intentional here? Unlike `{:ok, ...} <-`, this will always match - so if `handle_bulk_cancel` returns an `{:error, _}`, it would fall through to the `case result do` block and hit no clause (CaseClauseError). The retry handler uses `with :ok <- ...` which guards properly. Should this be `{:ok, _} = result <- ...` or similar?

lib/lightning_web/live/run_live/show.ex
```diff
@@ -341,6 +364,46 @@ defmodule LightningWeb.RunLive.Show do
...
+      end
+    else
+      {:noreply,
+       put_flash(
+         socket,
+         :error,
+         "You are not authorized to perform this action."
+       )}
+    end
+  end
```
elias-ba, line 402: This handler is nearly identical to the one in `index.ex`. Would it make sense to extract the shared logic into a helper (maybe in a shared module or in `Runs` itself) to avoid the duplication? Not blocking, just wondering if it'd be easier to maintain.

lib/lightning_web/live/run_live/components.ex
```diff
@@ -588,6 +588,130 @@ defmodule LightningWeb.RunLive.Components do
...
+            phx-key="escape"
+            phx-click-away={hide_modal(@id)}
+            class="hidden relative transform overflow-hidden rounded-lg bg-white px-4 pb-4 pt-5 text-left shadow-xl transition-all sm:my-8 sm:w-full sm:max-w-lg sm:p-6"
+          >
+            <div id={"#{@id}-content"} class="mt-3 text-center sm:mt-5">
+              <h3
+                class="text-base font-semibold leading-6 text-gray-900"
+                id={"#{@id}-title"}
+              >
+                Cancel Work Orders
```
elias-ba: Since the action cancels the *runs* (and the work order state changes as a side effect), would "Cancel Pending Runs" be more precise here? The body text already says "pending runs" so it's a bit inconsistent with the title.

taylordowns2000:
This is great, @elias-ba . I think your concerns have been addressed here:

  | # | Issue | Fix | Files |                                                                                                                                                                                                                                                               
  |---|-------|-----|-------|                                                                                                                                                                                                                                                               
  | 1 | **Auth gap** — `Runs.get/1` unscoped | Added `Runs.get_for_project/2` that joins through `work_order → workflow` to verify project ownership | `runs.ex` |                                                                                                                          
  | 2 | **`with result <-` always matches** | Replaced `with` with `if`/`case` so `{:error, _}` is properly handled | `index.ex` |                                                                                                                                                          
  | 3 | **Oban insert failure ignored** | Pattern match on `cancel_many_async` return; propagate errors | `index.ex` |                                                                                                                                                                      
  | 4 | **Duplicated cancel-run logic** | Extracted `CancelHelper.cancel_run/2` used by both LiveViews | `cancel_helper.ex`, `index.ex`, `show.ex` |                                                                                                                                        
  | 5 | **Modal title inconsistency** | Changed "Cancel Work Orders" → "Cancel Pending Runs" | `components.ex` |

taylordowns2000:
> One question (cc @stuartc): the "selected" cancel path runs synchronously in the LiveView process, while rerun always goes through Oban. I assume that's intentional since cancel is just a single UPDATE query bounded by page size - but wanted to confirm: was there a deliberate decision to keep it sync for small batches, or should we align it with the rerun pattern for consistency?

It was a specific decision to keep it sync for small batches. To my knowledge, Stu was not part of this decision. Happy for you to go wherever you want with it depending on your/Brandon's/Stu's cost:benefit analysis.

elias-ba:
Thanks for the quick response, Taylor. That makes sense to me - a bounded sync UPDATE is the right call here. Merging! 🚀

PR #4590, opened 2026-04-01
Fix copy token tooltip clipping and icon flicker

### LINKED ISSUE #2463: UI inconsistency when copying user API token
When copying user tokens, we see a few issues: 
- The "Copied" text that shows when the icon is clicked is behind the container and feels like it's cut out (See image below)
- When the user clicks the copy icon, it disappears. 

Image

Image

m4cd4r4:
## Description

Fixes #2463 - UI inconsistency when copying the user API token.

Two bugs existed on the Personal Access Tokens page (`/profile/tokens`):

1. **No visible 'Copied!' feedback** - the copy button had no `phx-hook="Tooltip"` ancestor, so `showCopiedTooltip()` in the Copy hook silently no-opped (`this.el.closest('[phx-hook="Tooltip"]')` returned null). The only feedback was a flash message at the bottom of the screen.

2. **Icon flicker on click** - `phx-then={JS.push("copy")}` sent a server event on every click, which triggered a LiveView diff/repaint that caused the clipboard icon to flicker.

**Fix:** Wrap the copy button in a `<span phx-hook="Tooltip">` with `aria-label="Copy to clipboard"` and `data-hide-on-click="false"`. The existing `showCopiedTooltip()` mechanism then works correctly - tippy.js shows 'Copied!' for 1.5s and resets. This matches the established pattern in `channel_live/form_component.ex`.

The server-side `handle_event("copy", ...)` handler and `phx-then` attribute are removed - feedback is now handled client-side, consistent with other copy UIs in the codebase.

Changed `overflow-auto` to `overflow-hidden` on the token container - this prevents unexpected scrollbars while keeping rounded corner clipping.

## Validation steps

1. Go to `/profile/tokens` and generate a new token
2. Hover the clipboard icon - tooltip reads "Copy to clipboard"
3. Click the clipboard icon - tooltip immediately shows "Copied!" for 1.5s, then resets
4. Paste the token somewhere to confirm the clipboard contents are correct
5. Confirm the clipboard icon does not flicker or disappear during steps 2-4

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


midigofrank (approved):
Nicely done @m4cd4r4

PR #4591, opened 2026-04-01
Standardize UUID validation across all REST API controllers

### LINKED ISSUE #4588: Standardize UUID validation across all REST API controllers
### User story

As an API consumer, when I accidentally pass a malformed identifier (e.g. a typo, a slug, or a non-UUID string) to any API endpoint, I expect a clear `400 Bad Request` response telling me the parameter is invalid. Instead, most endpoints return a `500 Internal Server Error` with no useful message, which makes me think the server is broken rather than that my request was wrong. This is inconsistent with the behavior on `/api/workflows` endpoints, which do return proper errors.

### Details

UUID validation before database queries is only implemented in 2 of the 9 API controllers:

- **`WorkflowsController`** validates UUIDs across all actions via a private `validate_uuid/1` function
- **`CredentialController`** validates UUIDs only in its `delete` action via its own duplicate private `validate_uuid/1` function

The remaining 7 controllers pass user-supplied ID parameters directly into Ecto/Repo queries with no format validation:

| Controller | Unvalidated UUID parameters |
|---|---|
| `JobController` | `id` (path), `project_id` (query) |
| `ProjectController` | `id` (path) |
| `RunController` | `id` (path), `project_id` (path), `workflow_id` (query), `work_order_id` (query) |
| `WorkOrdersController` | `id` (path in show, query filter in index), `project_id` (path), `workflow_id` (query) |
| `LogLinesController` | `project_id`, `workflow_id`, `job_id`, `work_order_id`, `run_id` (all query) |
| `ProvisioningController` | `id` (path and body) |
| `AiAssistantController` | `job_id`, `project_id`, `workflow_id` (all query) |
| `CredentialController` (gaps) | `project_id` in `index` and `create` actions |

When a malformed UUID (e.g. `"notauuid"`) reaches Ecto, it raises either `Ecto.Query.CastError` or `Postgrex.Error`. The `FallbackController` has no clause to handle either exception, so Phoenix returns a generic 500 response.

**Test coverage is nearly absent.** Only a single test across all API controller test files (`credential_controller_test.exs:842`) exercises a malformed UUID input. Even `WorkflowsController`, which has correct production validation, has no corresponding test. All other "not found" tests use `Ecto.UUID.generate()` (valid UUID format, just nonexistent) — these verify 404 behavior but do not test the malformed-input path.

### Implementation notes

- **Shared helper:** Extract a single `validate_uuid/1` function into `lib/lightning_web/controllers/api/helpers.ex`, which already exists as a home for shared API utilities. The existing implementation pattern using `Ecto.UUID.dump/1` is sound and can be reused. An alternative approach would be a Plug that validates UUID-shaped path/query params before the action is called, keeping validation logic out of individual action functions.
- **Replace duplicates:** Remove the two private `validate_uuid/1` functions in `WorkflowsController` and `CredentialController` and replace their call sites with the shared version.
- **Apply consistently:** Audit every controller action that accepts a UUID parameter (path or query) and add validation at the entry point of each action. The table above can serve as a checklist.
- **Error response:** `FallbackController` already handles `{:error, :bad_request}` → 400, so the shared helper just needs to return that tuple on validation failure.
- **Test helper:** Consider a shared ExUnit helper (e.g. `assert_returns_400_for_invalid_uuid(conn, method, path)`) to reduce boilerplate when adding malformed-UUID tests across all controller test files. Each controller should have at least one test per action confirming that malformed UUIDs return 400 with a meaningful error body.

### Release notes

Fixed a bug where passing a malformed (non-UUID) identifier to most API endpoints returned a 500 Internal Server Error instead of a 400 Bad Request. UUID parameter validation is now applied consistently across all API controllers.

### User acceptance criteria

- [ ] Passing a malformed UUID as a path parameter (e.g. `GET /api/jobs/notauuid`) returns `400 Bad Request` with a JSON error body — not 500
- [ ] Passing a malformed UUID as a query parameter (e.g. `GET /api/runs?project_id=notauuid`) returns `400 Bad Request` with a JSON error body — not 500
- [ ] All 9 API controllers validate all UUID parameters consistently
- [ ] The two existing private `validate_uuid/1` implementations are replaced by a single shared helper
- [ ] Every controller action that accepts a UUID parameter has at least one test asserting that a malformed UUID returns 400
- [ ] Existing behavior preserved: valid-format UUIDs referencing nonexistent resources still return 404
- [ ] No unhandled `Ecto.Query.CastError` or `Postgrex.Error` exceptions from UUID parameter handling in any API controller

m4cd4r4:
## Summary

Closes #4588.

Malformed UUID path/query params currently reach `Repo.get` and raise `Ecto.Query.CastError`, which Phoenix turns into a 500 response. This extracts a shared `validate_uuid/1` helper and applies it consistently across all REST API controllers.

**Changes:**

- Add `validate_uuid/1` to `LightningWeb.API.Helpers` - returns `:ok` or `{:error, :bad_request}`
- `WorkflowsController` - private `validate_uuid/1` now delegates to the shared helper (preserves existing 422 message behaviour via `maybe_handle_error`)
- `CredentialController` - replaces private `validate_uuid/1` with shared helper; malformed credential `id` now returns 400 instead of 404
- `JobController`, `ProjectController`, `RunController`, `WorkOrdersController`, `ProvisioningController`, `AiAssistantController` - add `validate_uuid/1` calls before any DB lookup that would otherwise raise `Ecto.Query.CastError`

**Tests:** UUID validation test cases added to all affected controller test files.

## AI disclosure

- [x] I used Claude Code to assist with this contribution

## Test plan

- [ ] New tests in each controller test file assert 400 for `not-a-uuid` inputs
- [ ] Existing tests continue to pass
- [ ] `mix test test/lightning_web/controllers/api/` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -281,6 +281,9 @@ and this project adheres to
   [PR#4551](https://github.com/OpenFn/lightning/pull/4551)
 - Fix AI assistant authorization for support users on projects with support
   access enabled [#4571](https://github.com/OpenFn/lightning/issues/4571)
+- REST API controllers now return 400 instead of 500 for malformed UUID
+  parameters. Extracts a shared `validate_uuid/1` helper used across all API
+  controllers. [#4588](https://github.com/OpenFn/lightning/issues/4588)
 
 ## [2.16.0] - 2026-03-24
 
--- a/lib/lightning_web/controllers/api/ai_assistant_controller.ex
+++ b/lib/lightning_web/controllers/api/ai_assistant_controller.ex
@@ -9,6 +9,7 @@ defmodule LightningWeb.API.AiAssistantController do
   alias Lightning.Policies.Permissions
   alias Lightning.Projects
   alias Lightning.Workflows
+  alias LightningWeb.API.Helpers
   alias LightningWeb.Channels.AiAssistantJSON
 
   action_fallback LightningWeb.FallbackController
@@ -75,7 +76,9 @@ defmodule LightningWeb.API.AiAssistantController do
   end
 
   defp get_resource("job_code", %{"job_id" => job_id}) do
-    {:ok, job_id}
+    with :ok <- Helpers.validate_uuid(job_id) do
+      {:ok, job_id}
+    end
   end
 
   defp get_resource("job_code", _params) do
@@ -83,9 +86,11 @@ defmodule LightningWeb.API.AiAssistantController do
   end
 
   defp get_resource("workflow_template", %{"project_id" => project_id}) do
-    case Projects.get_project(project_id) do
-      nil -> {:error, :not_found}
-      project -> {:ok, project}
+    with :ok <- Helpers.validate_uuid(project_id) do
+      case Projects.get_project(project_id) do
+        nil -> {:error, :not_found}
+        project -> {:ok, project}
+      end
     end
   end
 
--- a/lib/lightning_web/controllers/api/credential_controller.ex
+++ b/lib/lightning_web/controllers/api/credential_controller.ex
@@ -25,6 +25,7 @@ defmodule LightningWeb.API.CredentialController do
   alias Lightning.Policies.Permissions
   alias Lightning.Policies.ProjectUsers
   alias Lightning.Projects
+  alias LightningWeb.API.Helpers
 
   action_fallback LightningWeb.FallbackController
 
@@ -61,7 +62,8 @@ defmodule LightningWeb.API.CredentialController do
   def index(conn, %{"project_id" => project_id}) do
     current_user = conn.assigns.current_resource
 
-    with project when not is_nil(project) <- Projects.get_project(project_id),
+    with :ok <- Helpers.validate_uuid(project_id),
+         project when not is_nil(project) <- Projects.get_project(project_id),
          :ok <-
            ProjectUsers
            |> Permissions.can(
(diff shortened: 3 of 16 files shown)


m4cd4r4:
Rebased onto current `main` (was 61 commits behind, CONFLICTING) and addressed the failing `test_elixir` job.

**Conflicts resolved:**
- `lib/.../api/run_controller.ex` (`show`): merged with the new `%Lightning.Run{}` struct match so we get UUID validation **plus** the stricter return-shape check.
- `lib/.../api/work_orders_controller.ex` (`show`): same pattern for `%Lightning.WorkOrder{}`.
- `lib/.../api/project_controller.ex` (`show`): kept the `%Lightning.Projects.Project{}` struct match introduced upstream.
- `lib/.../api/job_controller.ex` (`show`): switched to the new `Jobs.get_job(id, include: [...])` API (no bang, single preload) that landed upstream.
- `test/.../api/run_controller_test.exs`: kept both the new upstream `describe "show"` block and this PR's `describe "malformed UUID params"` block.

**Test failure root cause:**
The `index` action in `credential_controller.ex` had an `else` clause that handled `nil` and `{:error, :unauthorized}` but not `{:error, :bad_request}`, so `Helpers.validate_uuid/1` returning `:bad_request` for a malformed `project_id` blew up the `with` (`WithClauseError`). Added the missing `{:error, :bad_request} -> {:error, :bad_request}` arm so it falls through to `FallbackController`, which already returns 400. Matches the pattern this PR already uses in `delete`.

The second failure in the prior CI (`test list_dataclips_for_job/3 doesn't return a dataclip if the wrong text is entered` in `Lightning.InvocationTest`) is unrelated to this PR (touches no invocation code) - flagging in case it's a known flake.

CI is re-running now.

PR #4593, opened 2026-04-01
Fix collaborative_editor_url incorrectly mapping trigger/edge s param to job

### LINKED ISSUE #4375: Map collab editor params to legacy editor
#4372 changed the links in the history page to use the collab editor params. This now means that if you have chosen to use the legacy editor, if you click to inspect a run step, the code editor won't open.

This issue makes it possible to convert the collab params to the legacy params

_Originally posted by @doc-han in https://github.com/OpenFn/lightning/pull/4372#pullrequestreview-3746016660_

m4cd4r4:
## Summary

Closes #4375

When a user switches from the legacy editor to the collaborative editor,
the `s` query param (selected node) is converted to the appropriate collab
param (`job=`, `trigger=`, or `edge=`). The conversion depends on
`determine_selection_type/2`, which pattern-matches on atom-keyed socket
assigns (`selected_trigger`, `selected_job`, `selected_edge`).

The bug: `collaborative_editor_url/2` was passing the URL params map
(string-keyed) as the 4th argument to `convert_param`, not socket assigns.
Since the URL params never contain `:selected_trigger` etc., the cond always
fell through to the default `"job"` case - so triggers and edges were always
mapped to `job=` instead of `trigger=` or `edge=`.

### Changes

- `collaborative_editor_url/3` now takes an optional `assigns` arg (defaults
  to `%{}` for backward compat) and passes it to `convert_param` instead of
  the URL params map
- `edit.ex` now passes `socket.assigns` when calling `collaborative_editor_url`
  from the `switch_to_collab_editor` event handler
- Three new tests cover the fixed trigger/edge/job resolution with assigns

## Checklist

- [x] I have added tests to cover my changes
- [x] I have used an AI assistant (Claude Code) to help produce this code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -272,6 +272,9 @@ and this project adheres to
 
 ### Fixed
 
+- Fix `collaborative_editor_url/2` incorrectly mapping trigger/edge selections
+  to `job` param when switching from legacy to collaborative editor
+  [#4375](https://github.com/OpenFn/lightning/issues/4375)
 - Consider manual runs for "next cron run input" via the
   `last_run_final_dataclip` function
   [#4584](https://github.com/OpenFn/lightning/issues/4584)
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -1644,7 +1644,7 @@ defmodule LightningWeb.WorkflowLive.Edit do
       })
 
     collaborative_url =
-      Helpers.collaborative_editor_url(params, socket.assigns.live_action)
+      Helpers.collaborative_editor_url(params, socket.assigns.live_action, socket.assigns)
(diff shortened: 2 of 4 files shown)


m4cd4r4:
The `test_elixir` failure is a pre-existing flaky test (`sorting workflows by last_workorder_updated_at desc`) unrelated to this change - it touches no workflow sorting code. The lint failure has been fixed in the latest push (line too long in `edit.ex`).

m4cd4r4:
Rebased onto current `main` (was 61 commits behind) - both commits cherry-pick cleanly and CI is re-running. Happy to also split the line-length fix into the previous commit if a single-commit PR is preferred. Mentioning in case this slipped off the review queue. 🙏

midigofrank:
Closing this because legacy editor is being deprecated

PR #4594, opened 2026-04-01
Allow special characters and non-Latin scripts in job names

### LINKED ISSUE #4577: [P3] Support special characters in workflow step names
Problem:
As raised by @lmac-1  [in #product-feedback](https://openfn.slack.com/archives/C0394D0GBUN/p1774622488122969), non-english users struggle to define names without special characters. 

Image

Solution Proposal: 
Remove this special character validation function since we treat step names as labels in the yaml now.

m4cd4r4:
## Summary

Closes #4577

Workflow step names are stored as labels in YAML exports, not as
identifiers. The existing regex (`/^[a-zA-Z0-9_\- ]*$/`) was unnecessarily
blocking non-English users from entering names in their native scripts
(Arabic, French, Chinese, etc.) or using characters like `@` or `'`.

Removes the character-set restriction from three locations:

- `lib/lightning/workflows/job.ex` - Elixir changeset validation
- `assets/js/workflow-diagram/nodes/PlaceholderJob.tsx` - PlaceholderJob inline validation
- `assets/js/collaborative-editor/types/job.ts` - Zod schema for collaborative editor

The 100-character length limit and blank-name check are preserved.

## Validation steps

1. `mix format`
2. `mix test test/lightning/workflows/job_test.exs --seed 0`

## Checklist

- [x] I have added tests to cover my changes
- [x] I have used Claude Code to help produce this code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to
 
 ### Fixed
 
+- Allow special characters and non-Latin scripts in workflow step names
+  [#4577](https://github.com/OpenFn/lightning/issues/4577)
 - Consider manual runs for "next cron run input" via the
   `last_run_final_dataclip` function
   [#4584](https://github.com/OpenFn/lightning/issues/4584)
--- a/assets/js/collaborative-editor/types/job.ts
+++ b/assets/js/collaborative-editor/types/job.ts
@@ -20,7 +20,6 @@ export const JobSchema = z
       .string()
       .min(1, "Job name can't be blank")
       .max(100, "Job name shouldn't be longer than 100 characters.")
-      .regex(/^[a-zA-Z0-9_\- ]*$/, "Job name can't include special characters.")
       .transform(val => val.trim()), // Auto-trim whitespace like backend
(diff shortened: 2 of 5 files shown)


m4cd4r4:
The `test_elixir` failure is a pre-existing flaky test (`sorting workflows by last_workorder_updated_at desc`) unrelated to this change - it touches no workflow sorting code. Lint passed on this PR.

josephjclark:
This needs thinking through. Why did we have character restrictions in the first place? Are there any characters which we truly can't support? Are there database limitations? I know we generate an id slug from the name the user enters, is that robust to non-latin characters? Is it adequately tested?

We need real assurance that this is stable before we can merge it.

stuartc:
Agreed, @josephjclark the original purpose of that validation was to ensure they are url-safe. While we only address them by uuid currently, the project yaml files reference by keys derived (or directly) from the name. So the surface area and tradeoffs aren't so quite so cut and dry to drop the name. Although we do allow spaces currently which means they technically aren't url-safe (would need `%20`).

lmac-1:
Closing due to inactivity for 7 days

PR #4597, opened 2026-04-01
feat: export webhook_reply and cron_cursor_job in provisioner

### LINKED ISSUE #4587: Support `webhook_reply` for triggers in the provisioner
There is currently no way to update the `webhook_reply` via the provioner

midigofrank:
## Description

Include `webhook_reply` and `cron_cursor_job_id` in the provisioning JSON API response and YAML export. 

Closes #4587

## Validation steps

1. In the project settings, when you export the project.yaml, it should contain the `webhook_reply` for `webhook` type workflows and `cron_cursor_job` for `cron` type workflows 
2. If you go to view your workflow as yaml, you should see them as well.
3. Try creating a workflow via the YAML import, it should pick up the set values for `webhook_reply` and `cron_cursor_job`

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning/export_utils.ex
+++ b/lib/lightning/export_utils.ex
@@ -160,7 +160,7 @@ defmodule Lightning.ExportUtils do
   defp merge_edge_common_fields(json, edge, target_job) do
     json
     |> Map.merge(%{
-      target_job: hyphenate(target_job.name),
+      target_job: job_key(target_job),
       condition_type: edge.condition_type |> Atom.to_string(),
       enabled: edge.enabled,
       node_type: :edge
--- a/test/integration/cli_deploy_test.exs
+++ b/test/integration/cli_deploy_test.exs
@@ -240,8 +240,15 @@ defmodule Lightning.CliDeployTest do
         |> with_job(reply_job)
         |> with_edge({webhook_trigger, reply_job}, condition_type: :always)
 
-      cron_trigger = build(:trigger, type: :cron, cron_expression: "0 6 * * *")
+      # Build the job first so we have its id
       cursor_job = build(:job, name: "cursor job", body: "fn(state => state)")
+
+      cron_trigger =
+        build(:trigger,
+          type: :cron,
+          cron_expression: "0 6 * * *",
+          cron_cursor_job_id: cursor_job.id
+        )
 
       cron_workflow =
         build(:workflow, name: "cron cursor workflow", project: nil)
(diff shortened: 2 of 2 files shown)


lib/lightning/workflows/workflow.ex
```diff
@@ -44,9 +44,12 @@ defmodule Lightning.Workflows.Workflow do
     field :enable_job_logs, :boolean, default: true
     field :positions, :map
 
+    # the ordering of edges, triggers and jobs are intentional
+    # ecto reverses the relations when inserting. so jobs->triggers->edges
+    # triggers depend on jobs for cron_cursor_job_id
```
stuartc, line 49: This feels really brittle dude, there must be a better way to deal with this. This feels like the kind of thing that could silently break from a change in Ecto no?

I'm glad the tests reflect the behaviour of this change (bit wary it might only be the provisioner import that fails, shouldn't the cli_deploy tests also fail without this change)

lib/lightning/export_utils.ex
```diff
@@ -348,6 +370,10 @@ defmodule Lightning.ExportUtils do
     }
   end
 
+  defp job_key(job) do
+    hyphenate(job.name)
+  end
+
```
stuartc, line 376: Not sure this level of abstraction is necessary, but not gonna fight it. If this is going to stay, I think there is a `hyphenate` call that should use `job_key` as well in `merge_edge_common_fields`.

test/integration/cli_deploy_test.exs
stuartc, line 1: This pattern exists elsewhere but it can be simplified:

`build(:job)` generates a UUID, you can set `cron_cursor_job_id` at build time and let the has_many ordering handle the rest:

```elixir
# Build the job first so we have its id
cursor_job = build(:job, name: "cursor job", body: "fn(state => state)")

cron_trigger =
  build(:trigger,
    type: :cron,
    cron_expression: "0 6 * * *",
    cron_cursor_job_id: cursor_job.id
  )

cron_workflow =
  build(:workflow, name: "cron cursor workflow", project: nil)
  |> with_trigger(cron_trigger)
  |> with_job(cursor_job)
  |> with_edge({cron_trigger, cursor_job}, condition_type: :always)

project =
  insert(:project,
    name: "webhook-reply-and-cron-cursor-project",
    project_users: [%{user: user, role: :owner}],
    workflows: [webhook_workflow, cron_workflow]
  )

# No reload + update needed — the has_many ordering ensures jobs
# are inserted before triggers.
```

This removes the Repo.reload / Repo.update! dance and doubles as a test that the has_many ordering in workflow.ex actually works. I verified it passes.

lib/lightning/workflows/workflow.ex
midigofrank: Yeah, sadly the only way around it is to use `Ecto.Multi` and insert them in order. 

No need for alarm though, we have been depending on this ordering to handle inserting the `edges`.  I can't find the source where I read about it. We should plan on refactoring `save_workflow` to use `Ecto.Multi` instead

PR #4613, opened 2026-04-10
Support sandboxes in collections

### LINKED ISSUE #3548: sandboxes: support collections
[Spec drafted by @brandonjackson 30 March 2026]

# Problem Statement
**Sandboxes** do not properly support **Collections**. When running a Workflow in a Sandbox, users can only access production collections data.

### Current Behavior
When a sandbox is provisioned from a parent project, collections are not created or linked in the sandbox. Workflows that reference collections by name (e.g., `collections.get('my-collection')`) will fail with "not found" errors in the sandbox because no collection with that name exists there.

Additionally, collection names are currently globally unique across the entire Lightning instance (enforced by `unique_index(:collections, [:name])`), which means a sandbox cannot even create a collection with the same name as one in its parent project.

### Impact

Services teams are unwilling to use sandboxes on projects that use collections. This is a blocker for sandboxes adoption and must be resolved before public release.

# Requirements

## Auto-create empty collections in sandbox on provision

* When a sandbox is forked, create empty Collection records matching each parent collection's name
* This ensures job code referencing `collections.get('my-collection')` resolves correctly in the sandbox
* Collections should be empty (no items copied) — sandbox is for testing, not data duplication

## Scope collection name uniqueness to project (not global)

* Enable both the main branch and multiple sandboxes to all have collections with the same name in the same project (rather than enforcing global uniqueness)

## Sandbox merge merges configuration not data

* When merging a sandbox back into its parent, collection data is never merged / never overwrites parent data
* New collections created in sandbox are created in parent sandbox as empty collections
* Existing collections deleted in sandbox are deleted in parent project 

# Implementation Details

* Update collection uniqueness constraints so that the name is only unique within a project.
* Presumably this means adding a project key to the database and migrate all collections to include a matching key
* Accept a version flag on the collections API. Probably via header. 
* Update the v1 version to search for a collection by name, and return `409` if more than one match is found
* Add a new v2 collections API which uses the path `collections/:project-id/:name`
* Pass the owning project id to the options of the Run sent to the worker (as `options.project_id`)
* On sandbox fork, initialise new empty collections and add to the new project
* On sandbox merge, add/remove collections to the target project (but do not sync data)

# Out of Scope

There are several ideas which we could/should do later which are out of scope:
* Support sharing collections across environments, eg namespace syntax like main::my-collection to explicitly reference a collection from a different environment
* Collection data seeding, Option to copy a subset of collection items when provisioning a sandbox (for test data)

# Archive: Original Ticket

Reminder to ensure  that collections will work in sandboxes

I haven't thought about this enough: but we need to ensure that collections are namespaced and isolated for each sandbox.

So in job code I do: `collections.get('my-collection')`, but  the main project and staging sandbox access different variations of that.

So collection names are not unique within a workspace.

When looking up a credential by name we probably need to know what sandbox you're in, so that we can find correct collection to use

josephjclark:
## Description

Sandboxes can't support projects that use collections because collection names are globally unique, and the collections API has no way to disambiguate between a parent and its sandbox. This PR relaxes the uniqueness constraint to per-project and lets the API scope lookups by project via an optional `?project_id=<uuid>` query param.

On the lifecycle side, empty collection records are now cloned into a sandbox when it's provisioned so that `collections.get('my-collection')` resolves correctly without copying production data. When a sandbox is merged back into its parent, collection names are synchronised (new collections in the sandbox are created empty in the parent, collections missing from the sandbox are removed from the parent). Collection data is never synced.

Existing unscoped calls keep working whenever a name is unambiguous. When a bare-name lookup finds the same name across multiple projects, the API returns 409 Conflict pointing clients at the `?project_id=` query param. The query-param approach was chosen during a Friday sync with @stuartc, @josephjclark and @midigofrank : it keeps the existing routes in `router.ex`, does not require a version bump, and leaves room for future scopes like `?env=dev` without another API version.

Closes #3548

### Setup: create a project with a collection

1. Create a new project called **Cat Shelter Registry**
2. Navigate to **Project Settings > Collections** and create a collection called `cats`
3. Create a workflow called "Fetch Cat Facts" with:
   - A **cron trigger** (any schedule is fine, we will run it manually)
   - A single **job** using `@openfn/language-http@latest` with this code:
     ```js
     get('https://catfact.ninja/facts?limit=10');

     each('$.data.data[*]', fn(state) => {
       const fact = state.data;
       const id = fact.fact.substring(0, 20).replace(/\s/g, '-').toLowerCase();
       return collections.set('cats', `fact-${id}`, {
         fact: fact.fact,
         length: fact.length
       })(state);
     });
     ```
4. Run the workflow manually by clicking **Run** on the workflow page
5. Navigate to **Project Settings > Collections**. You should see: `cats` listed with a storage size greater than 0
6. Click the **download** button next to the `cats` collection. Open the downloaded JSON file. You should see 10 items, each with a `fact` and `length` field.

This is your **parent project** with production data. Keep it open.

### 1. Sandbox provision clones empty collections

1. From the Cat Shelter Registry project, navigate to **Sandboxes**
2. Click **Create Sandbox** and name it "sandbox-test"
3. Once created, click into the sandbox to open it
4. Inside the sandbox, navigate to **Project Settings > Collections**

**Verify**: The `cats` collection is listed. Click the **download** button. Open the JSON file -- it should be an empty array `[]`. The sandbox has the collection, but none of the parent's 10 cat facts were copied.

### 2. Collection data is never merged back

1. Inside the sandbox from step 1, open the "Fetch Cat Facts" workflow
2. Edit the job to fetch breed data instead:
   ```js
   get('https://catfact.ninja/breeds?limit=5');

   each('$.data.data[*]', fn(state) => {
     const breed = state.data;
     return collections.set('cats', `breed-${breed.breed}`, {
       breed: breed.breed,
       country: breed.country,
       coat: breed.coat
     })(state);
   });
   ```
3. Run the workflow manually
4. Navigate to **Project Settings > Collections** inside the sandbox. Click **download** on `cats`. Open the JSON -- you should see 5 breed items.
5. Now go back to **Sandboxes** and **merge** the sandbox into the parent
6. After the merge, open the **parent project** (Cat Shelter Registry)
7. Navigate to **Project Settings > Collections**. Click **download** on `cats`.

**Verify**: Open the JSON. You should still see the original 10 cat fact items (with `fact` and `length` fields). None of the sandbox's 5 breed items (with `breed`, `country`, `coat` fields) should be present. Collection data is never merged.

### 3. New collections in sandbox are created in parent on merge

1. Create a new sandbox from Cat Shelter Registry called "sandbox-breeds"
2. Open the sandbox and navigate to **Project Settings > Collections**
3. Create a new collection called `cat-breeds`
4. Open the workflow and edit the job:
   ```js
   get('https://catfact.ninja/breeds?limit=10');

   each('$.data.data[*]', fn(state) => {
     const breed = state.data;
     return collections.set('cat-breeds', breed.breed, {
       country: breed.country,
       origin: breed.origin,
       coat: breed.coat
     })(state);
   });
   ```
5. Run the workflow. Go to **Project Settings > Collections** and download `cat-breeds` -- you should see 10 breed items.
6. Go back to **Sandboxes** and **merge** this sandbox into the parent
7. Open the **parent project** and navigate to **Project Settings > Collections**

**Verify**: You should now see two collections listed: `cats` and `cat-breeds`. Click **download** on `cat-breeds` and open the JSON. It should be an empty array `[]`. The collection was created in the parent, but the sandbox's breed data was not copied over.

### 4. Collections deleted in sandbox are deleted from parent on merge

1. In the parent project, go to **Project Settings > Collections** and create a collection called `temp-data`
2. Create a new sandbox called "sandbox-cleanup"
3. Open the sandbox and navigate to **Project Settings > Collections**. You should see both `cats` and `temp-data` listed.
4. Delete the `temp-data` collection from the sandbox
5. Confirm only `cats` remains in the sandbox's collections list
6. Go back to **Sandboxes** and **merge** this sandbox into the parent
7. Open the **parent project** and navigate to **Project Settings > Collections**

**Verify**: `temp-data` is no longer listed. Only `cats` (and `cat-breeds` if you did step 3) remain. The collection and all its items were removed from the parent because it was deleted in the sandbox.

### API test setup

Steps 5-7 use curl to test the API directly. First, get a token and project ID:

1. In Lightning, click your profile icon (top right) and go to **Tokens**
2. Click **Generate New Token**, give it a name, and copy the token value
3. In your terminal, save it for the curl commands:
   ```bash
   export TOKEN="paste-your-token-here"
   ```
4. To get the project ID, open the Cat Shelter Registry project in Lightning. The URL will look like `http://localhost:4000/projects/abc123-def456-.../w`. The UUID after `/projects/` is your project ID:
   ```bash
   export PROJECT_ID="abc123-def456-..."
   ```

### 5. Ambiguous name returns 409 (API)

This requires an unmerged sandbox so that `cats` exists in two projects. If you merged all sandboxes in previous steps, create a new one first.

```bash
# Unscoped lookup -- the name "cats" exists in multiple projects
curl -s -H "Authorization: Bearer $TOKEN" \
     http://localhost:4000/collections/cats | jq .
```

**Verify**: The response is **409 Conflict**:
```json
{
  "error": "Multiple collections found with this name. Add ?project_id=<uuid> to disambiguate."
}
```

Now add `?project_id=` to resolve it:
```bash
curl -s -H "Authorization: Bearer $TOKEN" \
     "http://localhost:4000/collections/cats?project_id=$PROJECT_ID" | jq .
```

**Verify**: The response is **200** with the parent's cat fact items.

### 6. Invalid project_id rejected (API)

```bash
curl -s -o /dev/null -w "%{http_code}\n" \
     -H "Authorization: Bearer $TOKEN" \
     "http://localhost:4000/collections/cats?project_id=not-a-uuid"
```

**Verify**: The response status is **400**.

### 7. Backwards compatibility with unscoped calls

Delete all sandboxes first (so `cats` only exists in one project), or use a project with a uniquely named collection.

```bash
curl -s -H "Authorization: Bearer $TOKEN" \
     http://localhost:4000/collections/cats | jq .
```

**Verify**: The response is **200** with the collection items, exactly as it worked before this change. No query param is needed when the collection name is unambiguous.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning_web/controllers/collections_controller.ex
```diff
@@ -25,6 +25,75 @@ defmodule LightningWeb.CollectionsController do
...
 
+  defp api_version(conn) do
+    case get_req_header(conn, "x-api-version") do
+      ["2"] -> :v2
+      _ -> :v1
+    end
+  end
+
+  def dispatch(conn, %{"path" => path}) do
+    case api_version(conn) do
```
josephjclark: We'll need some error handling here if an unsupported version is requested

lib/lightning_web/controllers/collections_controller.ex
```diff
@@ -25,6 +25,75 @@ defmodule LightningWeb.CollectionsController do
 
   @valid_params ["key", "cursor", "limit" | @timestamp_params]
 
+  defp api_version(conn) do
+    case get_req_header(conn, "x-api-version") do
+      ["2"] -> :v2
+      _ -> :v1
```
elias-ba: Let's reduce the scope of this to only fallback to v1 for when no version is specified. If a specific version is given and not supported throw 400 BAD REQUEST

lib/lightning_web/controllers/collections_controller.ex
```diff
@@ -92,19 +168,109 @@ defmodule LightningWeb.CollectionsController do
...
+      case stream_chunked(conn, items_stream, response_limit) do
+        {:error, conn} -> conn
+        {:ok, conn} -> conn
+      end
+    end
+  end
 
+  # V2 actions
+
+  def put(conn, project_id, name, key, %{"value" => value}) do
```
elias-ba: Maybe let' try to see if we can make these functions a little bit dryer.

josephjclark:
Testing and QA notes:

- The existing worker & collections adaptor should continue to work against the new lightning server, no changes needed
- If you create two collections with the same id, and request that id through the legacy API, you should get an error
- After updating the adaptor and worker, this branch should return the correct collection (even for duplicate collection names)

lib/lightning_web/live/sandbox_live/index.ex
```diff
@@ -814,6 +815,44 @@ defmodule LightningWeb.SandboxLive.Index do
     end
   end
 
+  defp sync_collections(sandbox, parent) do
```
elias-ba: Move this to the projects/sandboxes.ex

josephjclark:
@elias-ba one thing I didn't think to check is access control. A project should not be able to access the collections of another project.

How can we enforce this?

Using the CLI you can send any access token you want and so long as the project is within that access token's scope, you're good.

In the worker, we send a JWT which I believe is scoped only to access for that workflow. 

That must be sufficient right? We just look at the scope of the access token?

elias-ba:
Good question @josephjclark , and yes, the access token scope is sufficient. The controller already enforces it.

Every action calls `authorize/2` which hits `Lightning.Policies.Collections.authorize/3`. That policy has two clauses:

- For a user subject, it delegates to `ProjectUsers.authorize(:access_project, user, ...)` which checks membership against the collection's project
- For a run subject, it directly compares `Runs.get_project_id_for_run(run) == collection.project_id`

So a CLI user with a PAT can only touch collections in projects they're a member of, and a worker's run token can only touch collections in the same project as the run it was issued for. There's a test that specifically covers the worker case (`"with a run token, cannot access a different project's collection"` returns 401) and another for the user case (`"returns 401 when user does not have access to the project"`).

One subtle thing worth noting: v2 makes this tighter by design. With v1, a request with a stale or leaked token could at least look up a collection by name globally; v2 requires you to know (and be authorized for) the project_id upfront.

elias-ba (approved):
Hey @josephjclark this is now looking good to me and it has all things we discussed this morning. Please do test / QA this extensively and let me know if you find anything weird

lib/lightning_web/controllers/collections_controller.ex
elias-ba: Done

midigofrank (changes requested):
@elias-ba this looks great, I'm a little worried with the approach to remove the routes from the `router`.  One thing that worries me is that the `collections_router` plug hijacks  all actions and manually calls the controller, in my mind I'm expecting it handle the `x-api-version` and `backward compatibility` only. I don't like that we're losing visibility of the routes from the router.ex 

I don't know the exact discussion you had with @stuartc , so I have added him as a reviewer as he has more context. Everything else looks perfect.

josephjclark:
We're gonna resolve the API issue with this approach:
```
api/collections/name?project_id=abc
```
- non breaking change
- doesn't require version changes
- We can add api/collections/name?env=dev later
- If project_id is excluded we search by name and return 409 (or whatever) if there are multiple results

josephjclark:
I've run careful QA on this, coupled with supporting implementations in kit and adaptors. And it does the expected things in all cases. It's fully compatible with main, and when everything is updated then collections are properly scoped.

Here's my [QA worksheet](https://docs.google.com/spreadsheets/d/1xaqtX-bVQYbGCjdZwSGWaR1oPYYxsMkbFEoJ37K_f4A/edit?usp=sharing)


I'm just going to do a bit of CLI stuff but I think we can probably merge this in? Do we need an extra review from someone @elias-ba ?

elias-ba:
Hey @josephjclark thanks a lot for the QA, that's really great. Well we need another approval to be able to merge (unless we bypass that rule and someone with higher privelegges merges this cc @stuartc @midigofrank). But I did already the code review part and I am happy with this guy.

midigofrank:
Please don't merge yet @josephjclark @elias-ba . We have a pending release

elias-ba:
@midigofrank you're already a reviewer on this one. Feel free to approve and merge anytime is best

midigofrank (approved):
Nicely done @elias-ba

PR #4626, opened 2026-04-15
Fix: non-map state creates lost runs every time

### LINKED ISSUE #4627: Non-map final state leads to lost
This reliably produces a lost run:
```js
fn(state => {
  console.log("click run to take down the worker");
  return 42;
})
```

taylordowns2000:
## Description

This PR makes it so we don't automatically "lose" all runs when the user choses to return something other than a JSON object. To repro the bug, run this job code:

```js
fn(state => {
  return 42;
})
```

Closes https://github.com/OpenFn/lightning/issues/4627

## Validation steps

1. Check out (and run!) the new `red_team_test.exs` file

@josephjclark , @stuartc , why do we let the worker successfully complete runs where the final state is an integer or a string? I'm OK wrapping it in "value" but a cleaner approach seems to be making sure it fails/crashes on the worker side since `42` isn't a valid final_state.

(Queue up philosophical music... is `42` a valid final state? No, it's not for me... but if we thought of the worker as independent from "OpenFn" then maybe `42` could be a totally fine input state, or output state.)

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


josephjclark:
We should discuss the philosophy (perhaps not here?)

In the `runtime`, 42 is a perfectly valid state. I won't take arguments on that - the runtime is just JSON in, JSON out.

We  could be strict in the worker and ensure that an object is returned. I'd be happy implementing that instead (it would be a better fix) - so long as you can convince me of _why_?

Lightning has never really cared about output state before (that's always been weird to me but fine). It's starting to care now though, what with sync workflows and configurable cron. So who cares if a workflow returns 42 or true or an array? I expect 99.9% of use cases to return `{ count: 42 }` or `{ success: true }`, but even so I don't know why we need to have an opinion on that.

lib/lightning/runs/handlers.ex
```diff
@@ -473,10 +473,13 @@ defmodule Lightning.Runs.Handlers do
       Dataclip.new(%{
         id: dataclip_id,
         project_id: project_id,
-        body: output_dataclip |> Jason.decode!(),
+        body: output_dataclip |> Jason.decode!() |> ensure_map(),
```
josephjclark, line 476: `42` is valid JSON AFAIK. Why do we need to ensure a map? Like where does Lightning actually care? Isn't the value arbitrary?

test/integration/red_team_test.exs
```diff
@@ -0,0 +1,178 @@
+defmodule Lightning.RedTeamTest do
```
josephjclark: I've never heard of this red team stuff until now, and I can't be the only one

If we're going to keep this file, can we just name this file for what it is, rather than giving it a code name? Sorry to be boring but sometime boring is how I feel

taylordowns2000: Sure thing. Change incoming.

lib/lightning/runs/handlers.ex
taylordowns2000: This might be the philosophical question :-)

Since the very beginning, AFAIK, we've said that state is a JSON object (not _just_ JSON) so I'm not actually sure what might break if we allowed `null` or `true` or `42`. Let's see what @stuartc has to say here!

josephjclark: I think we're gonna merge this just to cover the lost runs case (which I don't think is an active problem?)

But I do want to talk about this properly -  I really think we should support any JSON value.

If we are going to insist on an object, maybe we should throw an error and crash the run or something. And document it clearly.

And if we insist on an object value, should we also update runtime to match?

taylordowns2000: Awesome. Yeah I bet this was more a simple miscommunication back in the day than some serious, deep rooted stance. Let's decide if we want to support values or objects, enforce it across the stack, and move onto the next one! Saw the calendar invite. Thank you.

josephjclark (approved):

lib/lightning/runs/handlers.ex
elias-ba: This line still says final_state is "a new map". After the PR, non-map values are accepted too, right ?

elias-ba (approved):
Hey @taylordowns2000 this looks good to me. Maybe just check that doc line I mentioned in my comment. But that's not blocking, so approving.

elias-ba:
You'll need to handle the conflict also

lib/lightning/runs/handlers.ex
elias-ba: Done

PR #4630, opened 2026-04-16
Adjust initial AI streaming status

hanna-paasivirta:
## Description

This PR changes the initial progress status shown to the user when starting a conversation with the AI assistants from "Generating response..." to "Thinking..." . This change is required because we are adding a more detailed stream of status updates from Apollo, where "Generating response" would feel out of place. https://github.com/OpenFn/apollo/pull/452

## Validation steps

1. Start a conversation with any AI assistant. Check that the first status that gets shown is "Thinking..."
 
## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/collaborative-editor/components/MessageList.tsx
+++ b/assets/js/collaborative-editor/components/MessageList.tsx
@@ -773,7 +773,7 @@ export function MessageList({
                 <span className="inline-block w-1.5 h-1.5 bg-gray-400 rounded-full animate-bounce [animation-delay:0.3s]" />
               </div>
               <span className="text-xs text-gray-400 italic">
-                {streamingStatus || 'Generating response...'}
+                {streamingStatus || 'Thinking...'}
               </span>
             </div>
           </div>
--- a/assets/js/collaborative-editor/components/MessageList.tsx
+++ b/assets/js/collaborative-editor/components/MessageList.tsx
@@ -11,6 +11,15 @@ import { Tooltip } from '../../components/Tooltip';
 
 const STREAMING_MESSAGE_ID = '__streaming__' as const;
 
+const LOADING_STATUSES = [
+  'Thinking about the question...',
+  'Working on it...',
+  'Processing your request...',
+  'Examining your question...',
+  'Taking a look...',
+  'Looking into it...',
+] as const;
+
 const PROSE_CLASSES =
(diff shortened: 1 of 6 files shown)


josephjclark:
Should we cycle it, rather than using a static value?

How long does it take to get the first response back from apollo?

Should we say `Calling assistant` or something to imply "we're uploading your question"? Because it's not _really_ thinking yet, it's just taking the input and preparing to think.

Maybe that doesn't matter much but "Thinking" is a) boring b) not very insightful. So I'm just wondering if there's something cheap we can use instead

hanna-paasivirta:
@josephjclark I think it's more frustrating to wait for a system to connect than to wait for it to do something properly, so I'm not sure about "calling assistant".

What if we took the opportunity to reinforce the voice of the assistant here with a more playful pool like:

* Warming up
* Cracking knuckles
* Rolling up sleeves
* Gathering thoughts
* Getting oriented
* Reading the room
* Taking a breath
* Sharpening the pencil
* Opening the toolbox
* Checking the toolbox
* Putting the kettle on
* Laying out the tools
* Unrolling the blueprints
* Doing the boring-but-important part
* Looking before leaping
* Doing the plumbing
* Getting on with it
* Doing the quiet work
* Checking the pipes
* Laying the groundwork
* Pouring the foundation
* Wiring it up
* Greasing the gears
* Oiling the hinges

josephjclark:
@hanna-paasivirta those are great fun, and in keeping with the OpenFn corporate voice. But I also don't think I want the AI assistant to have that same voice? It's too fun, and I know I'm a grumpy old man but I don't want it to be fun.

This also raises a wider point about what we want the assistant's voice to be like, for which we'll have to canvas more opinions.

For now (as we just discussed) let's have a simple pool like `Reading the question... thinking about the question...`

hanna-paasivirta:
It appears for a bit too long for "reading" to make sense (it makes me think why does it read slower than me). I've gone with "Thinking about the question...", "Working on it...", "Processing your request...", "Examining your question...", "Taking a look...", and "Looking into it..." They don't make 100% sense in all situations but I think they're adequate.

assets/js/collaborative-editor/components/MessageList.tsx
```diff
@@ -416,6 +425,14 @@ export function MessageList({
   const [expandedYaml, setExpandedYaml] = useState<Set<string>>(new Set());
   const [copiedMessageId, setCopiedMessageId] = useState<string | null>(null);
 
+  // Pick a new random status each time loading starts so it stays
+  // stable across re-renders within one loading session.
+  const loadingStatus = useMemo(
+    () =>
+      LOADING_STATUSES[Math.floor(Math.random() * LOADING_STATUSES.length)],
+    [isLoading]
+  );
```
elias-ba, line 434: Hey @hanna-paasivirta this is a very interesting one. It caught my eyes and I went to the [React docs of useMemo](https://react.dev/reference/react/useMemo) and saw that the factory "should be a pure function" and that React reserves the right to throw away cached values (useMemo caveats). `Math.random()` inside the factory breaks that contract, and in Strict Mode dev the factory runs twice per render too, which muddies the intent a bit. It works fine today, but it's leaning on behavior React documents as non-guaranteed. I think a cleaner shape uses the "store information from previous renders" pattern from the [useState docs](https://react.dev/reference/react/useState):

```javascript
const pickStatus = () => LOADING_STATUSES[Math.floor(Math.random() * LOADING_STATUSES.length)];

const [loadingStatus, setLoadingStatus] = useState(pickStatus);
const wasLoading = useRef(isLoading);

if (isLoading && !wasLoading.current) {
  setLoadingStatus(pickStatus());
}
wasLoading.current = isLoading;
```

That should provide the same behavior, new random phrase each time loading starts and stable within a session, but the randomness lives in useState where impurity is fine, and the trigger is an explicit false -> true transition.

I don't think this is a blocker tho. I just learnt something and wanted to share and maybe it could help make the code better

josephjclark: Yeah I second this - it's a weird usage of useMemo.

@elias-ba what if we did this on `sendMessage` - when the message is sent, set the streaming status to one of these random messages. No need for a transient local state now, and we guarantee a new one will be picked each time.

elias-ba: Yeah, that works cleanly. `sendMessage` already funnels both branches through `aiStore.setMessageSending()`, so the natural seam is to seed `streamingStatus` there (or a sibling action called right next to it). Then MessageList just renders `streamingStatus` with no fallback, and the server's `streaming_status` push still overwrites freely.

One wrinkle: `retryMessage` doesn't go through `setMessageSending`, it flips `isLoading` on its own path. If we want retries to get a fresh phrase too we'd seed there as well, otherwise a retry would show an empty loading indicator until the server pushes status.

Happy with this direction if @hanna-paasivirta is.

josephjclark (approved):

assets/js/collaborative-editor/components/MessageList.tsx
josephjclark: @elias-ba I've made the changes and approved. Do you concur? Can we get this merged?

elias-ba (approved):

PR #4632, opened 2026-04-17
Sandbox-aware Project Settings page

### LINKED ISSUE #3398: sandboxes: update Settings page
In the Project Settings page, there is a small number of settings for sandboxes. 
Settings include:

- environment
- destroy sandboxes on merge
- long lived sandboxes
- Access control defaults

This issue is likely to be deleted, unless more options manifest

elias-ba:
## Description

This implements the per-page approach we agreed on in #3398. The Project Settings page now behaves differently when you're inside a sandbox: each tab tells you whether your changes will sync on merge, and the UI for some tabs changes shape to match what makes sense for a sandbox (no Delete button, MFA read-only, webhook security managed at the parent level, etc.).

The categorization, agreed in the issue thread:

| Tab | Category | What it means |
|---|---|---|
| `project` | Local | Editable but does not sync on merge. Sandbox Identity panel with a link to the parent. Delete button hidden (sandboxes are deleted from the sandboxes UI). |
| `credentials` | Editable | Editable, will sync on merge. |
| `collections` | Editable | Editable, will sync on merge (handled in #4613). |
| `webhook_security` | Inherited | Auth methods are shared with the sandbox and enforced on its webhook triggers, but can only be created, edited, or deleted from the parent project. |
| `collaboration` | Local | Editable but does not sync on merge. Adds a parent admin floor rule (see below). |
| `security` (MFA) | Inherited | Read-only, value cloned from parent at provision time, MFA toggle disabled. |
| `vcs` | Local | Editable but does not sync on merge. |
| `data-storage` | Local | Editable but does not sync on merge. |
| `history-exports` | Local | Action button stays, banner explains scope. |

A few things worth knowing about how this is implemented, because they affected the scope:

**The merge layer is already protective.** I started worrying that "Local" and "Inherited" were just UI promises and that a user changing data retention in a sandbox would actually push that value back to the parent on merge. After tracing through `MergeProjects.merge_project/3` and `Provisioner.import_document/4`, that turns out not to be the case. The merge document only takes `[:name, :description, :env, :color]` from the **target** (not the source), and the provisioner only casts `[:id, :name, :description]` on the project itself. Project-level fields like `requires_mfa`, `concurrency`, `retention_policy`, `history_retention_period`, `dataclip_retention_period` aren't in the cast list, so they never propagate. Webhook auth methods, project_users, and the GitHub repo connection aren't in the merge document at all. I added regression tests in `sandboxes_test.exs` to lock this in so a future change to MergeProjects or the Provisioner doesn't accidentally start syncing them.

**The parent admin floor rule is enforced server-side, not just in the UI.** A user who is admin or owner on any ancestor project (walking the parent chain, so admin on grandparent counts) cannot be removed from a sandbox. The UI disables the Remove button with a tooltip explaining why, but `Projects.delete_project_user!/1` also raises an `ArgumentError` if you try to do it directly, so a curl bypass or a future code path can't get around it. There's a unit test covering both paths.

**Webhook auth methods stay enforced but can't be managed from the sandbox.** Auth methods flow down from the parent and are still applied to the sandbox's webhook triggers, but the sandbox UI points users back to the parent for create/edit/delete. The tab is still present so users can find it; the panel renders a notice and a link to the parent project's webhook security tab instead of the full management table.

Closes #3398

### Setup

1. Create a project called **Cat Shelter Registry**.
2. In **Project Settings**, set MFA to required on the Security tab and change History Retention to 7 days under Data Storage. We'll use these later to confirm they don't get overwritten.
3. On the **Collaboration** tab, add a teammate as `admin` on this project.
4. Go to **Sandboxes** and create a sandbox called **sandbox-test**. Open it.

You're now inside the sandbox. Every test below happens from here unless I say otherwise.

### 1. Sandbox Setup tab (Local)

Navigate to **Project Settings** in the sandbox. The first tab should be **Setup**.

What you should see:

- A blue banner at the top: *"Changes you make here only apply to this sandbox and do not sync to the parent project on merge."*
- The section title says **Sandbox setup**
- The first card is **Sandbox Identity**
- Below the description, a line: *"Identifies this sandbox within its parent: [Cat Shelter Registry]"* with the parent name as a link
- The Export project button is still there
- The danger zone at the bottom reads **"The danger zone"** with a **Delete sandbox** button. Clicking it opens a confirmation modal that requires typing the sandbox name; submitting calls `Sandboxes.delete_sandbox/2` and redirects to the root project's workflows page with the flash *"Sandbox <name> and all its associated descendants deleted"*

### 2. Credentials and Collections (Editable)

These are the two tabs where changes flow back. We'll create one of each in the sandbox now and confirm in step 7 that they show up in the parent after merge.

**Credentials**

1. Click the **Credentials** tab. You should see a green banner: *"Changes you make here will sync to the parent project on merge."*
2. Click **New credential**, pick any type (raw JSON is quickest), and create a credential called **sandbox-only-credential**.
3. In another browser tab, open the parent project (**Cat Shelter Registry**) → Credentials. Confirm **sandbox-only-credential** is **not there yet** — it lives only in the sandbox until merge.

**Collections**

1. Click the **Collections** tab. Same green banner.
2. Click **New collection**, name it **sandbox-only-collection**, and save.
3. In the parent project's Collections tab, confirm **sandbox-only-collection** is **not there yet**.

### 3. Webhook Security (managed in parent)

Click **Webhook Security**. Instead of the usual auth methods table, you should see a notice that reads something like:

> *Webhook authentication is managed in the parent project. Methods are shared with this sandbox and enforced on its webhook triggers, but can only be created, edited, or deleted from the parent.*

Followed by a link: *"Manage them in the parent project ([Cat Shelter Registry])"* pointing back to the parent's webhook security tab.

### 4. Collaboration and the parent admin floor rule

Click **Collaboration**. You should see a Local banner.

The collaborators table shows the current sandbox members. The teammate you added as admin in step 3 of Setup should also be listed (they were inherited when the sandbox was provisioned).

Now hover over the **Remove Collaborator** button next to that admin. The button should be **disabled**, and the tooltip should read:

> *Cannot remove a user who is admin or owner on the parent project*

For comparison, add another teammate to the **sandbox** as `editor` (not admin on the parent). Their Remove button should be enabled normally.

### 5. Security tab (Inherited)

Click **Security**. You should see:

- A yellow banner: *"These settings are inherited from the parent project ([Cat Shelter Registry]) and cannot be changed here."*
- The MFA toggle is visible but **disabled** (cursor shows `not-allowed` on hover, the switch reflects the parent's value)
- Clicking the toggle does nothing

The value you see should match what you set on the parent in step 2 of Setup.

### 6. The other Local tabs

Click through **Sync to GitHub**, **Data Storage**, **History Exports**. Each should show the blue Local banner at the top. The forms remain editable. This is intentional: you can change these in the sandbox to support sandbox-specific workflows, but the values do not sync back.

### 7. Merge: what flows back, what doesn't

This is the part that worried me when I started, so worth confirming hands-on.

Still in the sandbox:

1. Go to **Data Storage** and change History Retention to 90 days (or anything different from the parent's 7).
2. Go to **Setup** and change Project Concurrency to 1.
3. Open the parent project's settings briefly and note the starting values — History Retention 7 days, MFA required, no concurrency override.
4. From the Sandboxes page, **merge** the sandbox into the parent.
5. Open the parent project (**Cat Shelter Registry**) and walk through each tab.

**Credentials (Editable — should sync):**
- **sandbox-only-credential** should now appear in the parent's Credentials tab.

**Collections (Editable — should sync):**
- **sandbox-only-collection** should now appear in the parent's Collections tab (per #4613).

**Setup (Local — should not sync):**
- Project name and description unchanged.
- Concurrency unchanged (didn't become 1).

**Data Storage (Local — should not sync):**
- History Retention is still 7 days (didn't change to 90).
- I/O retention / policy unchanged.

**Security (Inherited — should not sync):**
- MFA is still required (didn't get unset).

Editable categories flow back, Local and Inherited categories don't. That asymmetry is what `sandboxes_test.exs` locks in at the merge-document/provisioner level.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/components/sandbox_settings_banner.ex
+++ b/lib/lightning_web/components/sandbox_settings_banner.ex
@@ -1,0 +1,71 @@
+defmodule LightningWeb.Components.SandboxSettingsBanner do
+  @moduledoc """
+  Banner shown at the top of a Project Settings tab when the project is a
+  sandbox, communicating how changes on that tab will (or will not) flow
+  to the parent project on merge.
+
+  Three variants:
+
+    * `:local` — changes apply only to this sandbox and do not sync on merge
+    * `:editable` — changes will sync to the parent on merge
+    * `:inherited` — settings are read-only, managed in the parent project
+
+  ## Examples
+
+      <.sandbox_settings_banner variant={:local} />
+
+      <.sandbox_settings_banner
+        variant={:inherited}
+        parent_project={@parent_project}
+      />
+  """
+  use LightningWeb, :component
+
+  alias LightningWeb.Components.Common
+
+  attr :variant, :atom, required: true, values: [:local, :editable, :inherited]
+  attr :id, :string, required: true
+  attr :parent_project, :map, default: nil
+
+  def sandbox_settings_banner(%{variant: :local} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="info" class="border border-blue-200">
+      <:message>
+        Changes you make here only apply to this sandbox and do not sync to the parent project on merge.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  def sandbox_settings_banner(%{variant: :editable} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="success" class="border border-green-200">
+      <:message>
+        Changes you make here will sync to the parent project on merge.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  def sandbox_settings_banner(%{variant: :inherited} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="warning" class="border border-yellow-200">
+      <:message>
+        These settings are inherited from the parent project
+        <.parent_link :if={@parent_project} project={@parent_project} />and cannot be changed here.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  attr :project, :map, required: true
+
+  defp parent_link(assigns) do
+    ~H"""
+    (<.link
+      navigate={~p"/projects/#{@project.id}/settings"}
+      class="font-medium underline"
+    >{@project.name}</.link>)
+    """
+  end
+end
--- a/lib/lightning_web/components/sandbox_settings_banner.ex
+++ b/lib/lightning_web/components/sandbox_settings_banner.ex
@@ -1,0 +1,71 @@
+defmodule LightningWeb.Components.SandboxSettingsBanner do
+  @moduledoc """
+  Banner shown at the top of a Project Settings tab when the project is a
+  sandbox, communicating how changes on that tab will (or will not) flow
+  to the parent project on merge.
+
+  Three variants:
+
+    * `:local` — changes apply only to this sandbox and do not sync on merge
+    * `:editable` — changes will sync to the parent on merge
+    * `:inherited` — settings are read-only, managed in the parent project
+
+  ## Examples
+
+      <.sandbox_settings_banner variant={:local} />
+
+      <.sandbox_settings_banner
+        variant={:inherited}
+        parent_project={@parent_project}
+      />
+  """
+  use LightningWeb, :component
+
+  alias LightningWeb.Components.Common
+
+  attr :variant, :atom, required: true, values: [:local, :editable, :inherited]
+  attr :id, :string, required: true
+  attr :parent_project, :map, default: nil
+
+  def sandbox_settings_banner(%{variant: :local} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="info">
+      <:message>
+        Changes you make here only apply to this sandbox and won't sync to the parent project on merge.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  def sandbox_settings_banner(%{variant: :editable} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="success">
+      <:message>
+        Changes you make here will sync to the parent project on merge.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  def sandbox_settings_banner(%{variant: :inherited} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="warning">
+      <:message>
+        These settings are inherited from the parent project
+        <.parent_link :if={@parent_project} project={@parent_project} />and can't be changed here.
+      </:message>
+    </Common.alert>
+    """
+  end
+
+  attr :project, :map, required: true
+
+  defp parent_link(assigns) do
+    ~H"""
+    (<.link
+      navigate={~p"/projects/#{@project.id}/settings"}
+      class="font-medium underline"
+    >{@project.name}</.link>)
+    """
+  end
+end
--- a/lib/lightning_web/components/sandbox_settings_banner.ex
+++ b/lib/lightning_web/components/sandbox_settings_banner.ex
@@ -29,7 +29,7 @@ defmodule LightningWeb.Components.SandboxSettingsBanner do
 
   def sandbox_settings_banner(%{variant: :local} = assigns) do
     ~H"""
-    <Common.alert id={@id} type="info" class="border border-blue-200">
+    <Common.alert id={@id} type="info" class="border border-blue-300">
       <:message>
         Changes you make here only apply to this sandbox and do not sync to the parent project on merge.
       </:message>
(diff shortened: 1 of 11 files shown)


lib/lightning_web/components/sandbox_settings_banner.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  attr :variant, :atom, required: true, values: [:local, :editable, :inherited]
+  attr :id, :string, required: true
+  attr :parent_project, :map, default: nil
+
+  def sandbox_settings_banner(%{variant: :local} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="info">
+      <:message>
+        Changes you make here only apply to this sandbox and won't sync to the parent project on merge.
```
josephjclark: This banner could use a border or something - it feels a bit too etherial

Image

I also don't love the alignment and layout of that whole top corner. But I don't want to nitpick on that yet - maybe I'll circle back

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -62,20 +62,43 @@
...
           <div class="bg-white p-4 rounded-md space-y-4">
             <div>
               <h6 class="font-medium text-black">
-                Project Identity
+                {if @sandbox?, do: "Sandbox Identity", else: "Project Identity"}
               </h6>
               <small class="block my-1 text-xs text-gray-600">
-                This metadata helps you identify the types of workflows managed in this project and the people that have access.
+                <%= if @sandbox? do %>
+                  Identifies this sandbox within the parent project.
```
josephjclark: Maybe like `Identifies this sandbox within its parent:  <a>PARENT</a>`

lib/lightning_web/live/project_live/settings.html.heex
```diff
@@ -325,171 +358,198 @@
...
+        <%= if @sandbox? do %>
+          <.section_header
+            title="Webhook security"
+            subtitle="Webhook authentication methods that are used with the starting trigger in workflows."
+            permissions_message="webhook auth methods."
+            can_perform_action={false}
+          />
+          <div class="bg-white p-4 rounded-md">
+            <p class="text-sm text-gray-700">
+              Webhook security is not available in sandboxes for now. Manage these methods in the parent project <.link
```
josephjclark: I don't like the "for now" bit - let's keep roadmapping out of the app

Maybe `Webhook security is disabled for sandboxes` is a clearer message. Because it's not just the options we're disabling - we're disabling all security on the sandbox (that might need an issue and a separate PR?)

lib/lightning_web/components/sandbox_settings_banner.ex
josephjclark: For language I prefer `do not sync` to `won't sync`

lib/lightning_web/components/sandbox_settings_banner.ex
```diff
@@ -0,0 +1,71 @@
...
+    </Common.alert>
+    """
+  end
+
+  def sandbox_settings_banner(%{variant: :inherited} = assigns) do
+    ~H"""
+    <Common.alert id={@id} type="warning">
+      <:message>
+        These settings are inherited from the parent project
+        <.parent_link :if={@parent_project} project={@parent_project} />and can't be changed here.
```
josephjclark: tweak to `cannot be changed here`

josephjclark (commented):
Fantastic!

It feels a lot better to think about this stuff here than in the issue. The top banner is working great - setting per-group context really helps

Some minor tweaks in comments and maybe we'll want to tune the UI right at the end

elias-ba:
@josephjclark ready for another look. Addressed the five inline comments. One wording call to flag: you'd suggested *"Webhook security is disabled for sandboxes"*, but auth methods from the parent are still enforced on the sandbox's webhook triggers, so I went with *"Webhook authentication is managed in the parent project"* instead - happy to swap if you prefer yours.

elias-ba:
@josephjclark just a bit of nudge on this one too whenever you have 5 mins to QA. Also feel free to merge it in the Sandbox Collection branch

josephjclark:
Yep thanks @elias-ba - I'm at the end of my day now but I'll likely get this merged in the morning

josephjclark:
@elias-ba good catch on the auth thing - I've opened up #4643 to deal with this. For now let's keep your wording.

josephjclark (approved):

josephjclark:
Ack sorry I had a commit with some tiny style changes which failed to push earlier. I've just pushed it up now

midigofrank (approved):
Nicely done

PR #4634, opened 2026-04-17
feat: support v2 sync protocol

### LINKED ISSUE #4523: Update GH sync to support v2 protocol
The GitHub Sync UI and action are configured around CLI sync v1 - the `openfn pull` and `openfn deploy` commands

v2 CLI sync is designed to make the github experience much better by writing out workflows to the file system and using neater "state files".

The UI makes assumptions and explains concepts that only apply to v1, such as the path to config.json.

We should update this UI to enable use of v2 sync, with more relevant text and instruction.

‼️Super important! v2 only allows one openfn.yaml and one `workflows` folder. You can set up multiple syncs to a v2 repo, but only one pull (ie, only one connection that uses Save & Sync to trigger a pull into the repo  and commit). You can deploy a single project to multiple places though. Because of this we might want to add a one-way flag to prevent save and sync.

We don't actually need to change the action - it'll work just fine with the v2 sync if there's an `openfn.yaml` file in the repo. We might want to reengineer around the new APIs but I don't think there's any value.

Saying that, the initial sync action for v2 needs to initialise an empty project.yaml file  - so maybe we do need a v2 action.

doc-han:
## Description

This PR makes sync v2 protocol an option for syncing projects with Github. By default sync will use the old `config.json` style but users can switch to the new `openfn.yaml` style if they prefer.

**Currently defaults to v1 (legacy) but can be switched to v2 (new) style**

<table>
<tbody>
<tr>
    <th colspan='2'>Github Sync Setup Screen</th>
  </tr>
<tr>
    <td colspan='2'>
    Screenshot 2026-04-20 at 12 06 56 PM
    </td>
  </tr>
<tr>
    <th>Legacy</th>
    <th>New</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-04-21 at 6 48 44 AM
</td>
    <td>
Screenshot 2026-04-21 at 6 47 58 AM
</td>
  </tr>
</tbody>
</table>

Closes #4523 

## Validation steps

1. Try to sync a project with github. connect and select the repo you want to sync with
2. Click on Advanced: Use new `openfn.yaml` format
3. After connecting. sync the project.
4. Check whether your `openfn.yaml` is in github with nicely generated content like the example below

#### Example `openfn.yaml` content
```yaml
project:
  uuid: 43c34b65-50cf-4ec3-b480-32b3b17ac9c0
  endpoint: <your-instance-url>
  alias: 
  inserted_at: 2026-04-13T01:00:00Z
  updated_at: 2026-04-13T04:37:54Z
  id: sync
  name: sync
workspace:
  credentials: credentials.yaml
  dirs:
    projects: .projects
    workflows: workflows
  formats:
    openfn: yaml
    project: yaml
    workflow: yaml
```
## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning/version_control/project_repo_connection.ex
+++ b/lib/lightning/version_control/project_repo_connection.ex
@@ -22,6 +22,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
     field :branch, :string
     field :access_token, :binary
     field :config_path, :string
+    field :use_yaml_config, :boolean, default: false
     field :accept, :boolean, virtual: true
 
     field :sync_direction, Ecto.Enum,
@@ -56,7 +57,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
   end
 
   @required_fields ~w(github_installation_id repo branch project_id)a
-  @other_fields ~w(config_path)a
+  @other_fields ~w(config_path use_yaml_config)a
 
   def changeset(project_repo_connection, attrs) do
     project_repo_connection
@@ -125,6 +126,14 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
 
   def config_path(repo_connection) do
     repo_connection.config_path ||
-      "./openfn-#{repo_connection.project_id}-config.json"
+      if repo_connection.use_yaml_config do
+        openfn_yaml()
+      else
+        "./openfn-#{repo_connection.project_id}-config.json"
+      end
+  end
+
+  def openfn_yaml do
+    "openfn.yaml"
   end
 end
--- a/lib/lightning/version_control/project_repo_connection.ex
+++ b/lib/lightning/version_control/project_repo_connection.ex
@@ -127,13 +127,13 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
   def config_path(repo_connection) do
     repo_connection.config_path ||
       if repo_connection.sync_version do
-        openfn_yaml()
+        path_to_openfn_yaml()
       else
         "./openfn-#{repo_connection.project_id}-config.json"
       end
   end
 
-  def openfn_yaml do
+  def path_to_openfn_yaml do
     "openfn.yaml"
   end
 end
--- a/lib/lightning/version_control/project_repo_connection.ex
+++ b/lib/lightning/version_control/project_repo_connection.ex
@@ -22,7 +22,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
     field :branch, :string
     field :access_token, :binary
     field :config_path, :string
-    field :use_yaml_config, :boolean, default: false
+    field :sync_version, :boolean, default: false
     field :accept, :boolean, virtual: true
 
     field :sync_direction, Ecto.Enum,
@@ -57,7 +57,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
   end
 
   @required_fields ~w(github_installation_id repo branch project_id)a
-  @other_fields ~w(config_path use_yaml_config)a
+  @other_fields ~w(config_path sync_version)a
 
   def changeset(project_repo_connection, attrs) do
     project_repo_connection
@@ -126,7 +126,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
 
   def config_path(repo_connection) do
     repo_connection.config_path ||
-      if repo_connection.use_yaml_config do
+      if repo_connection.sync_version do
         openfn_yaml()
       else
         "./openfn-#{repo_connection.project_id}-config.json"
--- a/lib/lightning/version_control/version_control.ex
+++ b/lib/lightning/version_control/version_control.ex
@@ -11,7 +11,6 @@ defmodule Lightning.VersionControl do
   alias Ecto.Multi
   alias Lightning.Accounts.User
   alias Lightning.Extensions.UsageLimiting
-  alias Lightning.Projects.Project
   alias Lightning.Repo
   alias Lightning.VersionControl.Audit
   alias Lightning.VersionControl.Events
--- a/lib/lightning/version_control/version_control.ex
+++ b/lib/lightning/version_control/version_control.ex
@@ -18,8 +18,8 @@ defmodule Lightning.VersionControl do
   alias Lightning.VersionControl.GithubError
   alias Lightning.VersionControl.ProjectRepoConnection
   alias Lightning.VersionControl.VersionControlUsageLimiter
-  alias Lightning.Workflows.Workflow
   alias Lightning.Projects.Project
+  alias Lightning.Workflows.Workflow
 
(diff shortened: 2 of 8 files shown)


doc-han:
### Problems
1. Why don't we allow users to generate a new github project from our platform when doing sync? (because if I need it on a new repo, I'll have to manually create that on github)
2. When a repo has no content yet. We're unable to find the branches on it. i.e we can't find main on the repo

doc-han:
@josephjclark suggests we keep v1 as default and then allow users to switch to v2 style

lib/lightning_web/live/project_live/github_sync_component.ex
```diff
@@ -683,6 +683,45 @@ defmodule LightningWeb.ProjectLive.GithubSyncComponent do
...
+
+  defp config_format_toggle(assigns) do
+    ~H"""
+    <div class="mt-4">
+      <details class="group">
+        <summary class="cursor-pointer text-sm text-gray-500 hover:text-gray-700 select-none list-none flex items-center gap-1">
+          <.icon
+            name="hero-chevron-right-mini"
+            class="h-4 w-4 transition-transform group-open:rotate-90"
+          /> Advanced: use legacy JSON config format
```
josephjclark, line 697: This is a really nice idea, and  we need to bank it because when we make v2 the default, this is probably how we'll want to handle it.

But for now we actually want to leave v1 the default until v2 has had a bit more testing.

lib/lightning/version_control/project_repo_connection.ex
```diff
@@ -125,6 +126,14 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
 
   def config_path(repo_connection) do
     repo_connection.config_path ||
-      "./openfn-#{repo_connection.project_id}-config.json"
+      if repo_connection.use_legacy_config do
```
josephjclark, line 129: Calling v1 "legacy" internally is still the correct thing to do though :)

josephjclark:
Sample v2 filesystem after sync has been established: https://github.com/doc-han/openfn-new

lib/lightning_web/live/project_live/github_sync_component.ex
doc-han: resolved

lib/lightning/version_control/version_control.ex
```diff
@@ -625,6 +633,33 @@ defmodule Lightning.VersionControl do
...
+    project = Repo.get!(Project, repo_connection.project_id)
+
+    inserted_at = DateTime.to_iso8601(project.inserted_at)
+    updated_at = DateTime.to_iso8601(project.updated_at)
+
+    """
+    project:
+      uuid: #{project.id}
+      endpoint: #{LightningWeb.Endpoint.url()}
+      alias: #{project.env}
```
josephjclark: Ooh - interesting one. The alias should be the project name if this is a sandbox, or otherwise `main`

I don't expect a lot of syncs to be set up with sandboxes. Maybe we should just not set an alias for now? Let users set it up later if they want

josephjclark: Actually, the first pull will cause a lot of this metadata to be updated. Maybe its' best to just set the uuid and endpoint and let the CLI own the rest?

lib/lightning/version_control/project_repo_connection.ex
```diff
@@ -22,6 +22,7 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
     field :branch, :string
     field :access_token, :binary
     field :config_path, :string
+    field :use_yaml_config, :boolean, default: false
```
josephjclark: Can we rename this to `sync_version`, where `false` is `1` and `true` is `2`

lib/lightning/version_control/version_control.ex
doc-han: resolved. lightning now only creates `openfn.yaml` with 
```
project:
  uuid:
  endpoint:
```

lib/lightning/version_control/project_repo_connection.ex
doc-han: renamed

lib/lightning/version_control/version_control.ex
```diff
@@ -625,6 +633,16 @@ defmodule Lightning.VersionControl do
     )
   end
 
+  defp openfn_yaml(repo_connection) do
+    project = Repo.get!(Project, repo_connection.project_id)
```
midigofrank: You don't need to fetch the `project` if all you need is the `project.id`. Why not use `repo_connection.project_id`

lib/lightning/version_control/project_repo_connection.ex
```diff
@@ -125,6 +126,14 @@ defmodule Lightning.VersionControl.ProjectRepoConnection do
...
     repo_connection.config_path ||
-      "./openfn-#{repo_connection.project_id}-config.json"
+      if repo_connection.sync_version do
+        openfn_yaml()
+      else
+        "./openfn-#{repo_connection.project_id}-config.json"
+      end
+  end
+
+  def openfn_yaml do
```
midigofrank:
```suggestion
  def path_to_openfn_yaml do
```

Can we rename this? It's not the yaml itself, just the path.

midigofrank (changes requested):
Nicely done @doc-han , left a few change requests. Also, is there a link we can add that describes what this new yaml format is all about? Maybe  community post or just the docs

lib/lightning/version_control/version_control.ex
doc-han: nice. was assuming it wasn't the same id. making change

doc-han: resolved

lib/lightning/version_control/project_repo_connection.ex
doc-han: resolved

josephjclark (approved):
Great @doc-han, I'm happy with this

I think we should run careful testing against staging. I've not done any kind of QA beyond watching you screen. It's not actually that complicated though so we should be alright

josephjclark:
QA notes:

- Setting up a GH sync works exactly the same as it is in production, with (almost no) change in the UI
- After sync is set up, the settings will tell you you're using sync v1
- And the sync works for push and pull
- Now you can set up a GH sync but tick the box to use V2
- In v2 sync, your github should contain:
  - an openfn.yaml file
  - a workflows folder with all job code beautifully expanded into files
  - a .projects folder with a yam file called something like `main@openfn.org.yaml`
  - No config.json, state.json or project.yaml file from v1
- After sync is established, the sync should tell you it's using v2 style
- Sync should work for push and pull

midigofrank (approved):
Nicely done

PR #4659, opened 2026-04-23
Sandbox merge UI update

### LINKED ISSUE #4651: sandboxes: it's not clear that sandboxes will be deleted on merge
It is not super clear right now that when merging a sandbox, it'll get deleted. I think we lost the emphasis on that after a bunch of UI  changes.

Here is the current merge UI

Image

This got services into trouble today because they didn't realise the sandbox would be removed. They lost work, and there is no recovery.

I think this UI needs a clean up tbh. Too many words. The checkbox makes it clear what will happen on merge so we don't need all the words on top.

And we should remove the red warning - no-one reads it after the first time and it's needlessly scary.

I also think we should do a soft delete instead of a hard one, see https://github.com/OpenFn/lightning/issues/4649  This would make deletion undoable and far less consequential.

josephjclark:
## Description

* Made the wordy description of what's happening smaller - users can still find it but it's de-emphasized. The UI tells you what is happening
* Simplify the warning message at the bottom and change it from an error UI to a warning UI
* Changed workflows show a green "Changed" badge in the workflow list
* Strengthen the warnining on diverged workflows
* Workflows changed in the sandbox are now pre-selected by default when opening the merge modal
* Merge button is disabled with a tooltip when no workflows are selected

New UI:

image

Closes #4651

## Validation steps

1. Test the following states for a workflow in a sandbox:
  a. Change the workflow in the sandbox (a  green change badge)
  b. Change the workflow in the main project (an amber diverge badge)
  c. Remove a workflow in the sandbox (a  red Removed badge)
  d. Add workflow from the project
2. Validate that the changes actually happened as expected on merge
3. Should be able to toggle changes in the UI
4. Unmodified workflows are deselected in the UI by default

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning_web/live/sandbox_live/components.ex
+++ b/lib/lightning_web/live/sandbox_live/components.ex
@@ -268,41 +268,6 @@ defmodule LightningWeb.SandboxLive.Components do
             will be lost.
           </p>
 
-          <%= if @descendant_count > 1 do %>
-            <Common.alert
-              id="merge-descendants-alert"
-              type="warning"
-              header="Child sandboxes will be closed"
-            >
-              <:message>
-                <p class="mb-2">
-                  The following {@descendant_count} sandboxes will be permanently closed:
-                </p>
-                <ul class="list-disc list-inside space-y-1 ml-2 mb-3">
-                  <li :for={descendant <- @descendants}>
-                    {descendant.name}
-                  </li>
-                </ul>
-                <p>
-                  Consider merging child sandboxes into
-                  <strong>{@sandbox.name}</strong>
-                  first to preserve their work before merging up.
-                </p>
-              </:message>
-            </Common.alert>
-          <% end %>
-
-          <%= if @descendant_count == 1 do %>
-            <Common.alert id="merge-single-descendant-alert" type="warning">
-              <:message>
-                <strong>{List.first(@descendants).name}</strong>
-                will also be closed. Consider merging it into
-                <strong>{@sandbox.name}</strong>
-                first to preserve its work.
-              </:message>
-            </Common.alert>
-          <% end %>
-
           <div class="border border-gray-200 rounded-md overflow-hidden">
             <div class="flex items-center justify-between px-3 py-2 bg-gray-50 border-b border-gray-200">
               <span class="text-sm font-medium text-gray-700">
@@ -372,6 +337,13 @@ defmodule LightningWeb.SandboxLive.Components do
           >
             <:message>
               This action cannot be undone.
+              <div :if={@descendant_count == 1} class="mt-2">
+                Child sandbox <strong>{List.first(@descendants).name}</strong>
+                will also be permanently closed.
+              </div>
+              <div :if={@descendant_count > 1} class="mt-2">
+                Its {@descendant_count} child sandboxes will also be permanently closed.
+              </div>
             </:message>
           </Common.alert>
 
--- a/lib/lightning_web/live/sandbox_live/components.ex
+++ b/lib/lightning_web/live/sandbox_live/components.ex
@@ -221,7 +221,7 @@ defmodule LightningWeb.SandboxLive.Components do
     >
       <:title>
         <div class="flex items-start justify-between">
-          <span class="font-bold">Merge</span>
+          <span class="font-bold">Merge Sandbox</span>
           <button
             type="button"
             class="rounded-md bg-white text-gray-400 hover:text-gray-500 focus:outline-none"
@@ -258,7 +258,7 @@ defmodule LightningWeb.SandboxLive.Components do
             </div>
           </div>
 
-          <p class="text-gray-700">
+          <p class="text-xs text-gray-700">
             This will overwrite the selected workflows in
             <strong>
               {get_selected_target_label(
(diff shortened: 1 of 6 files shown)


josephjclark:
Looking into those failing tests

doc-han (approved):

elias-ba:
@josephjclark instead of the warning emoji could we use a hero-icon class to stay consistent with icons we use in the app ? Are you happy if I push a commit that does that ?

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -258,30 +258,32 @@ defmodule LightningWeb.SandboxLive.Components do
...
             </strong>
-            with the versions from <strong>{@sandbox.name}</strong>,
+            with the versions from sandbox <strong>{@sandbox.name}</strong>,
             then delete <strong>{@sandbox.name}</strong>
+            <%= if @descendant_count == 0 do %>
+              .
+            <% end %>
             <%= if @descendant_count == 1 do %>
-              and its child sandbox <strong>{List.first(@descendants).name}</strong>
+              and its child sandbox <strong>{List.first(@descendants).name}.</strong>
```
elias-ba: Do we want the dot (`.`) in the `strong` tag or is this a typo ?

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -376,17 +385,23 @@ defmodule LightningWeb.SandboxLive.Components do
...
           <Common.alert
             id="merge-beta-warning"
-            type="danger"
-            header="This action cannot be undone"
+            type="warning"
+            header="This sandbox will be deleted after merging"
           >
             <:message>
-              Sandbox merging is in beta. For production projects, use the CLI to merge locally and preview changes first.
-              Collection names will be synced: new collections are added (empty) to the target, and collections missing from the sandbox are removed from the target. Collection data is never merged.
```
elias-ba, line 384: @josephjclark isn't this text too important to delete ? I mean the part about collections being merged empty ?

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -258,30 +258,32 @@ defmodule LightningWeb.SandboxLive.Components do
...
             <%= if @descendant_count == 1 do %>
-              and its child sandbox <strong>{List.first(@descendants).name}</strong>
+              and its child sandbox <strong>{List.first(@descendants).name}.</strong>
             <% end %>
             <%= if @descendant_count > 1 do %>
-              and its {@descendant_count} child sandboxes
-            <% end %>.
-            Any changes in
+              and its {@descendant_count} child sandboxes.
+            <% end %>
```
elias-ba: Small readability thought on this block: the period ends up in three different places depending on `@descendant_count` (standalone in the `== 0` branch, inside `<strong>` in the `== 1` branch, at the end of the text in the `> 1` branch). Would a `case` with the period placed once at the end read cleaner? Something like:

```
then delete <strong>{@sandbox.name}</strong><%= case @descendant_count do %>
    <% 0 -> %>
    <% 1 -> %> and its child sandbox <strong>{List.first(@descendants).name}</strong>
    <% n -> %> and its {n} child sandboxes
<% end %>.
```

Not blocking tho, happy either way.

elias-ba:
The 'Select all' being to the right of the table and the checkboxes being to the left looks a bit weird to me. I am wondering if it could be on the same vertical axis (left side) of the checkboxes.

lib/lightning_web/live/sandbox_live/components.ex
josephjclark: I changed this deliberately because there's an ugly space in prod if there are no dependents (the most common case)

image

Can you see it? In `then delete staging .`

Annoys the heck out of me.

I'm open to other solutions (claude said it was literally impossible to fix :joy: ), but I'm putting UX over devx here. Maybe worth a comment to this effect though

josephjclark:
@elias-ba Re the emoji yes of course - feel free to stick the icon in

Re Select All... maybe? But it's also weird on the left because there's nowhere to put it. It's out of scope for this work. If you want to try something very quickly, and if I super love it, we can do it. But I recommend leaving it alone for now.

lib/lightning_web/live/sandbox_live/components.ex
josephjclark: I don't think so?

We don't say or anywhere that collection data will be merged. It isn't copied when forking. Docs explain that collection data is not merged (not that anyone reads those!). I think this is fairly intuitive.

If we really do have to say something, it shouldn't be in a block paragraph of warning text like this. No-one reads that stuff. Any important messages have to be really really terse and clear.

elias-ba:
Yeah @josephjclark I think we should leave it alone for now.

elias-ba:
Hey @josephjclark, while testing I noticed the single-descendant alert mostly repeats the paragraph now that your copy explicitly names the child. Fits your de-cluttering direction - OK if I drop it ?

Screenshot 2026-04-24 at 23 01 16

<br />

Screenshot 2026-04-24 at 23 07 58

lib/lightning_web/live/sandbox_live/components.ex
elias-ba: 😂 Claude is liar. Got it better using function component and a `phx-no-format`. I also added a regression test to lock it. Also added a short inline comment explaining the constraint.

lib/lightning_web/live/sandbox_live/components.ex
elias-ba: Fair enough, happy to leave it. If it comes up from users we can add a terse one-liner later.

josephjclark:
@elias-ba great catch thank you - I'll drop it and ask you to review :)

elias-ba (approved):
Lovely 🌻

Issue #4618, opened 2026-04-13
Clearer Step Panel Design


brandonjackson:
### Clearer Step Panel Buttons

- Run from Here right aligned
- Code button replaced with Expand secondary button, left aligned
- Delete button replaced with Trash Can secondary button, left aligned

<img width="349" height="287" alt="Image" src="https://github.com/user-attachments/assets/f2455ee3-9b81-45e2-8fad-0925ac483d31" />

### Make Connect Primary CTA 
- If connecting a credential is required for an adaptor and one has not been connected:
- Switch Run from Here button to be secondary style (no blue colour) to focus attention on the Connect CTA

<img width="377" height="326" alt="Image" src="https://github.com/user-attachments/assets/59096e02-2963-4242-a3ae-a9a76cafcd08" />

doc-han:
The tricky part around this has to do with giving focus to "connect" when the adaptor requires a credential. currently, we're unable to really tell whether an adaptor requires a credential to function or not.

Issue #4617, opened 2026-04-13
Run From Here Button Text


brandonjackson:
- Change run button on step panel to say "Run From Here"
- <img width="312" height="265" alt="Image" src="https://github.com/user-attachments/assets/e066da34-bdcd-416d-9c1b-2f6747bdf3b4" />
- Clicking "Run From Here" always opens the Run with Custom Input panel, which also should still say "Run From Here"
- <img width="324" height="453" alt="Image" src="https://github.com/user-attachments/assets/84af8dff-9c47-48f3-8476-ed3cdc57bf91" />

Issue #4616, opened 2026-04-13
Run with Custom Input


brandonjackson:
- Experience opens when you click the new "Run from Custom Input" button in the run button dropdown (instead of every time you click run)
- Change title from “Run from Trigger” to “Pick a Custom Input”
- Change second tab's text to New

<img width="593" height="690" alt="Image" src="https://github.com/user-attachments/assets/b0a5a52c-18b2-4ea1-b28c-11e3f91f4986" />

elias-ba:
Hey @brandonjackson should we remove the empty tab ? I am wondering whether it makes sense to leave it there. It seems like that's what the primary Run button option is doing, right ? cc @doc-han

Issue #4615, opened 2026-04-13
One Click (Split) Run Button


brandonjackson:
Dropdown Contents:
* [Outline Play Button Icon] Run with Custom Input
* [FUTURE] [Outline Beaker Icon] Test Run in Sandbox

ACs
* When user clicks the primary Run CTA, start the run immediately
* Reuse the Split Button component for the retry run button in the IDE
* Update component so that dropdown uses non-bolded text

<img width="622" height="318" alt="Image" src="https://github.com/user-attachments/assets/4c438dd2-5b57-4add-91ac-444343f0d92e" />

doc-han:
Notes: 
1. When user click "Run" we want to run immediately. Meaning we run with an empty dataclip
2. When they click "Run with custom input" we then show them the dataclip panel.

doc-han:
The sandbox stuff isn't very straightforward. hence, I'll create a new for it.
#4657

PR #4665, opened 2026-04-27
feat: Better Run Experience

### LINKED ISSUE #4614: Better Run Experience
# Spec

Status: One-Pager

## Hypothesis
If we make running a workflow a clearer + more rewarding experience then users will want to do it more often

## Problem Observations

- It's confusing to see a button called "Run" which doesn't actually run.
- 80% of users want to run with the default input 80% of the time. Running with a custom input or existing input is a power user feature. But OpenFn requires specifying the input 100% of the time. Creating an extra click. Even if it’s a cron job (w/o inputs).
- When you click run, [it takes you to a new screen](https://github.com/OpenFn/lightning/issues/4198) (leaving canvas).
- Two run buttons are shown when IDE is open.

doc-han:
## Description

This is a PR for the EPIC on better run experience. All sub-issues will be merged into this till it's ready to be merged to main

Closes #4614 

## Validation steps

This Epic PR contains multiple issues and validation will be for all the sub-issues involved.

#### One click split button #4615 
The run button is now a split button with a dropdown on the right
- Clicking on the run button itself should instantly run a workflow with a default/empty dataclip
- Clicking the dropdown on the run button and the selecting "Run with custom input" should allow you to select a dataclip before running the workflow

#### Run with custom input #4616 
- Whenever you click the "Run with custom input" button. the run panel should show up with the title "Pick a custom input" instead of eg. "Run from trigger"

#### Run from Here Button text #4617 
- The text on the run panels should now be "Run From Here" instead of just run. to show that a run begins for the active step

#### Clearer Step panel design #4618 
This changes the look of the step panel a bit
- "Run From Here" button moves to the right.
- "Delete" and "Code" buttons become icons to the left
- When an adaptor is selected that requires a credential to be useful, the "Run From Here" button become gray coloured while the connect button for the adaptor becomes primary colored to give it more focus. Once a credential is attached, the connect button becomes gray instead while the "Run From Here" button steals the focus by becoming primary coloured.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/collaborative-editor/constants/adaptors.ts
+++ b/assets/js/collaborative-editor/constants/adaptors.ts
@@ -1,0 +1,12 @@
+// Adaptors that don't require a credential to run
+export const ADAPTORS_WITHOUT_CREDENTIALS = [
+  'common',
+  'http',
+  'testing',
+  'ping',
+  'fhir',
+  'fhir-4',
+  'fhir-eswatini',
+  'fhir-fr',
+  'fhir-ndr-et',
+];
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,9 @@ and this project adheres to
   [#4615](https://github.com/OpenFn/lightning/issues/4615)
 - Clearer step panel button design with icon-only secondary buttons for Code and
   Delete. [#4618](https://github.com/OpenFn/lightning/issues/4618)
+- "Pick a custom input" panel and renamed "New" tab when opening the manual run
+  panel from the canvas Run dropdown.
+  [#4616](https://github.com/OpenFn/lightning/issues/4616)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,7 +88,6 @@ and this project adheres to
   [#4510](https://github.com/OpenFn/lightning/issues/4510)
 - Worker plan payload now includes `project_id` so workers can scope callbacks
   (e.g. the collections API) to the project that owns the run.
-- Bumped local worker to 1.24.0
 - Updated the Merge Sandbox UI to be cleaner, clearer, and only include changed
   workflows by default [#4651](https://github.com/OpenFn/lightning/issues/4651)
 - Sandbox deletion (manual or after merge) is now soft. The sandbox and its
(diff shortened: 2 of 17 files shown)


assets/js/collaborative-editor/constants/adaptors.ts
```diff
@@ -0,0 +1,12 @@
...
+  'common',
+  'http',
+  'testing',
+  'ping',
+  'fhir',
+  'fhir-4',
+  'fhir-eswatini',
+  'fhir-fr',
+  'fhir-ndr-et',
+];
```
elias-ba, line 12: @doc-han I experienced a lot of long discussions about this with the team. I am a lot in favor of having this list as first pass and then later on find a better way to represent this information (should come from the adaptor package and not hardcoded) but I think this is so important that @brandonjackson and @taylordowns2000 and @theroinaochieng and maybe @josephjclark should be aware of it. Currently the team uses credentials for almost anything (the ethiopian implementation team even used credentials recently as environment variables store). We found out that this kind of things may block people more than it gives them value. Let's talk to the product team please before we ship ! what do you think ?

elias-ba (changes requested):

assets/js/collaborative-editor/constants/adaptors.ts
josephjclark: Yeah I don't love this, but I actually recommended the pattern to Farhan

Having seen the UX design on this PR I'm not keen on _why_ we're doing this. And as it 's for a UX niceity, rather than a core UX feature, I'm not very keen at all.

I liked @midigofrank's comments this morning. If we have a really clear need to recognise adaptors that MUST have a schema to run, let's declare that in the adaptor somehow. Lightning always uses the latest adaptor schema version (which is problematic actually!) so once we set the flag on the monorepo, we  can update the UI here to match.

But before we do that I want to be super clear that this is a desired change, because I'm really not sold on the UX proposition here.

taylordowns2000:
Sounds like a quick trip down memory lane would help! Please @doc-han , Tyrell, familiarize with https://github.com/OpenFn/lightning/issues/2844 and our Q2 2024 planned feature 🤣  — **"Credential Versions":** What happens if someone changes the required SCHEMA for an adaptor. Do all the credentials for that adaptor on Lightning then become invalid? (I think so! Yes. They all break instantly.) We might need to lock credentials on lightning to specific versions of adaptors. (Add a "raw" view escape hatch for credentials that are not on current version, so we don't lock users into a can't get out of jail situation.)

assets/js/collaborative-editor/constants/adaptors.ts
josephjclark: Happy with the list design, just want hear the UX design being validated

doc-han:
The credential bug as discussed on this issue should be blocking and still be resolved separately from this.

elias-ba (approved):
Hey @doc-han great work man 👏🏽. I am approving this guy now. I think the credentials whitelist / blacklist thing is bigger than the scope of this work and shouldn't block it. Maybe can you open a follow up issue for it so that we don't forget to come back to that problem and find a solution ?

josephjclark (approved):

PR #4666, opened 2026-04-27
Configurable webhook responses

### LINKED ISSUE #4479: Configure a custom status code and response body for both successful and failed/crashed runs
### User story
As a workflow builder using Sync mode, I want to configure a custom status code and response body for both successful and failed/crashed runs, so that the system calling my webhook receives exactly the data format and signal it needs.

### Acceptance criteria:

1. A "Configure Response" button appears on the trigger node when Sync mode is selected.
2. The button opens a modal with four fields: success status code, success body, failure/crash status code, and failure/crash response body.
3. Fields accept either a hardcoded value or a state.* dot-path expression.
4. When a run completes, the configured values are evaluated against the final run state and returned as the webhook response.
5. If a state.* path fails to resolve, the default sync response is returned instead.
6. Success config is used when the run state is :success; failure config is used for all other terminal states.

### LINKED ISSUE #4480: Extend the Trigger schema to store custom response configuration
### Details
The current trigger record has a webhook_reply enum but no fields to store what the custom response should actually say. We need to add four new nullable fields to the triggers database table and Ecto schema: success_status_code, success_body, failure_status_code, and failure_body. When any of these are nil, the existing default sync response behaviour applies.

### Acceptance criteria

1. Migration runs cleanly forward and backward.
2. Trigger struct includes the four new fields.
3. A trigger can be saved and retrieved with these values populated.

### LINKED ISSUE #4481: Apply custom response config to the webhook response at runtime
### User story
When a sync-mode webhook run completes, the response currently returns a fixed 201 with the full final state. We need to check whether the trigger has custom response config and, if so, use those values to shape the HTTP response instead.

### User acceptance criteria

1. Hardcoded status codes and bodies are returned correctly when set.
2. state.* dot-path expressions are resolved against the final run state.
3. If a path fails to resolve, the default sync response is returned.
4. The existing default sync response behaviour is fully preserved when no custom config is set.

### LINKED ISSUE #4482: Add "Configure Response" button and modal to the trigger form
### User story
When Sync mode is selected in the trigger form, a "Configure Response" button should appear. Clicking it opens the "Configure Webhook Response" modal. 

Image

For the modal: The title should be "Configure Webhook Response" and the subtitle "Set the status code and body returned by this webhook." The four fields are styled text inputs (monospace/code font, dark background) — not full code editor instances. 

Image

### User acceptance criteria

1. "Configure Response" button is visible only when Sync mode is selected.
2. Clicking the button opens the modal.
3. Modal title and subtitle match the spec exactly.
4. All four fields render with grey placeholder text showing their defaults.
5. Entering a value makes the text white and shows the clear button.
6. Clear button resets the field to empty (restoring the placeholder).
7. Save writes all four values to the trigger doc.
8. Cancel discards changes.
9. Button and modal do not appear when Async mode is selected.

### LINKED ISSUE #4483: Warn user before switching from Sync back to Async
### User story
As a workflow builder, I want to be warned before I accidentally lose my custom response configuration when switching from Sync back to Async, so I don't have to re-enter settings I've carefully defined

### User acceptance criteria

1. Switching from Sync to Async when any custom config field is populated triggers a warning: "Switching to async will clear your custom response configuration. This can't be undone."
2. The warning offers Continue and Cancel options.
3. Choosing Continue clears all custom config fields and completes the mode switch.
4. Choosing Cancel leaves the mode as Sync and all config values unchanged.
5. No warning is shown if all custom config fields are empty.

### LINKED ISSUE #4484: Support project export, import, and GitHub sync
### User story
As a deployment operator or developer using the CLI, I want my custom response configuration to survive project export, import, and GitHub sync, so that the behaviour I've configured in one environment is faithfully reproduced in another.

### User acceptance criteria

1. webhook_reply, success_status_code, success_body, failure_status_code, and failure_body all survive a full project export → import round-trip.
2. The same fields survive a GitHub sync operation.
3. The potential bug where webhook_reply is silently dropped on export is confirmed and fixed if still present.

### LINKED ISSUE #4598: Support custom webhook response in the provisioner
https://github.com/OpenFn/lightning/pull/4597 adds support for `:before_start and :after_completion`.

This should touch 3 places:

1. Workflow import via YAML, Copy workflow as YAML, Workflow templates
2. Project YAML download via the provisioner
3. Project state download via the provisioner

midigofrank:
## Description

Adds support for configurable webhook responses in sync (after completion) mode.

Previously, when a webhook trigger ran in sync mode, the response was always the final run state with a fixed `201` status code. This PR allows operators to:

- **Configure custom HTTP status codes** per trigger — separate codes for success and error runs — via the trigger inspector UI (Options → Response Status)
- **Override the response status and/or body at runtime** by setting `webhookResponse` in a job's final state, with partial overrides now supported (only `status`, only `body`, or both)
- `webhookResponse` is captured from each `step:complete` event as the run progresses. If multiple steps emit a value, **the last one wins** (until the response is sent on completion). A step that does not set `webhookResponse` does not clear a previously captured one.

**Defaults**
- Success runs default to `201`
- Errored runs default to `500`

**Response metadata headers** (`x-meta-work-order-id`, `x-meta-run-id`) are now documented and included on all webhook responses (both async and sync).

Closes #4479
Closes #4480
Closes #4481
Closes #4482
Closes #4483
Closes #4484
Closes #4598

This PR has a docs sibling: https://github.com/OpenFn/docs/pull/765

## Validation steps

1. Create a webhook-triggered workflow and set **Response Mode → Sync (After Completion)**
2. Under **Options**, click **Configure Response Status** and add a **Success Status Code** (e.g. `200`) — verify it is returned on a successful run
3. Add an **Error Status Code** (e.g. `422`) — trigger a failing run and verify the status code is returned
4. Remove the configured status codes via the trash icon — verify success falls back to `201` and error falls back to `500`
5. In a job, set `webhookResponse: { status: 202 }` (no body) — verify the custom status is used and the body is the final run state
6. Set `webhookResponse: { body: { ack: true } }` (no status) — verify the custom body is used and the status comes from the config (or `201` if none set)
7. Set both `status` and `body` in `webhookResponse` — verify both are used
8. Set `webhookResponse` in two different steps with different values — verify the **last** step's value is what gets returned
9. Set `webhookResponse` to a non-map value (e.g. a string) — verify it is silently ignored and defaults apply
10. Set `webhookResponse` with an invalid `status` type (e.g. a string) — verify the malformed response message is returned
11. Switch back to **Async (Before Start)** — verify the response configuration is cleared and the trigger responds immediately
12. Check that `x-meta-work-order-id` and `x-meta-run-id` headers are present on both async and sync responses

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


doc-han (approved):
Everything is this PR works just fine. we just need to be aware that when there's multiple leaf nodes. the value for `_webhookResponse` will exist on the leaf node and not the top level of the response data. 

### When There's a single leaf node.
<table><thead>
  <tr>
    <th>Workflow</th>
	<th>Code</th>
    <th>Output</th>
  </tr></thead>
<tbody>
  <tr>
    <td>

Image

</td>
    <td>

Image

</td>
	<td>

<pre><code>{
  "text": "Some data"
}</code></pre>
</td>
  </tr>
</tbody>
</table>

### When There's a multiple leaf nodes.
<table><thead>
  <tr>
    <th>Workflow</th>
	<th>Code</th>
    <th>Output</th>
  </tr></thead>
<tbody>
  <tr>
    <td>

Image

</td>
    <td>

Image

Image

</td>
	<td>

<pre><code>{
  "43d3a775-7cee-45ce-8feb-c951079d07cb": {
    "_webhookResponse": {
      "body": {
        "text": "Other data"
      }
    },
    "data": {
      "name": "Farhan-node-1"
    },
    "request": {
      "headers": {
        "accept": "*/*",
        "host": "localhost:4000",
        "user-agent": "curl/8.7.1"
      },
      "method": "POST",
      "path": [
        "i",
        "5bfa3aa9-e4e3-4249-8b05-95840365f761"
      ],
      "query_params": {}
    }
  },
  "dff9d8bd-4cd0-45ca-8c14-aaa404f1b394": {
    "_webhookResponse": {
      "body": {
        "text": "Some data"
      }
    },
    "data": {
      "name": "Farhan-node-1"
    },
    "request": {
      "headers": {
        "accept": "*/*",
        "host": "localhost:4000",
        "user-agent": "curl/8.7.1"
      },
      "method": "POST",
      "path": [
        "i",
        "5bfa3aa9-e4e3-4249-8b05-95840365f761"
      ],
      "query_params": {}
    }
  }
}</code></pre>
</td>
  </tr>
</tbody>
</table>

doc-han:
> Everything is this PR works just fine. we just need to be aware that when there's multiple leaf nodes. the value for `_webhookResponse` will exist on the leaf node and not the top level of the response data.
> 
> ### When There's a single leaf node.
> Workflow	Code	Output
> Image	Image	
> ```
> {
>   "text": "Some data"
> }
> ```
> 
> ### When There's a multiple leaf nodes.
> Workflow	Code	Output
> Image	Image Image	
> ```
> {
>   "43d3a775-7cee-45ce-8feb-c951079d07cb": {
>     "_webhookResponse": {
>       "body": {
>         "text": "Other data"
>       }
>     },
>     "data": {
>       "name": "Farhan-node-1"
>     },
>     "request": {
>       "headers": {
>         "accept": "*/*",
>         "host": "localhost:4000",
>         "user-agent": "curl/8.7.1"
>       },
>       "method": "POST",
>       "path": [
>         "i",
>         "5bfa3aa9-e4e3-4249-8b05-95840365f761"
>       ],
>       "query_params": {}
>     }
>   },
>   "dff9d8bd-4cd0-45ca-8c14-aaa404f1b394": {
>     "_webhookResponse": {
>       "body": {
>         "text": "Some data"
>       }
>     },
>     "data": {
>       "name": "Farhan-node-1"
>     },
>     "request": {
>       "headers": {
>         "accept": "*/*",
>         "host": "localhost:4000",
>         "user-agent": "curl/8.7.1"
>       },
>       "method": "POST",
>       "path": [
>         "i",
>         "5bfa3aa9-e4e3-4249-8b05-95840365f761"
>       ],
>       "query_params": {}
>     }
>   }
> }
> ```

I think the first example is able to identify `_webhookResponse` easily on the output json and narrow down to it's body but for the multiple leaf nodes. you'll that field living nested in the response for the specific nodes.

midigofrank:
Oof, thanks @doc-han . That is a really important catch. In this implementation, I assumed all nodes could write to the root of the state. @josephjclark is there no way to always write to the root irregardless of the node structure? Is this where we now introduce a helper function?

josephjclark:
@midigofrank all nodes do write to the root of state. The problem is that multiple nodes have returned a final root state. What the runtime does - all it can do really - is aggregate them. 

If node a and b fire, which is the final state? :shrug:

No helper function can help with that.

The options I see are:
- Use the first or last response that was written to state. You have to listen to `step:complete` to work this out
- Re-write the runtime in such a way that it knows what final state is
- Don't use state to carry the response and use a special event to send the response instantly

josephjclark:
And I really don't like the underscore key :(

midigofrank:
Thanks @josephjclark @doc-han , one thing I love about this discovery is that it would still be a dilemma regardless of whether we handle it in lightning or the worker. 

> Use the first or last response that was written to state. You have to listen to step:complete to work this out

I love this option. The only problem is that dataclips are in json string form when sent via `step:complete`. This would mean we have to decode the dataclip in all `step:complete` events which is not ideal. Also what if the worker drops the output dataclip because the user has the `zero-persistence` policy on dataclips?

> Re-write the runtime in such a way that it knows what final state is

If we're going to touch the worker, I would rather have it be that we're introducing the apis to send the response via an event.

option 4 here would be for lightning to go one level deep trying to look for `_webhookResponse`. This would be the least effort but I don't like it either.

@josephjclark how much effort is needed to introduce the new event api to the worker?

josephjclark:
Zero persistence is a great point too. I'll remind myself how that works but it might be the final nail in the coffin for using state.

I'll think about a design for an event, I think that's the only practical option left.

midigofrank:
@josephjclark while at it, please remember that there are possibly 2 apis that we will need to support:

1. when the run finishes
2. during step execution

We can have a single event api for the worker and let lightning figure out when to respond depending on the config in the `trigger`, i.e. `:after_completion` or `:custom/during_execution`.

josephjclark:
@midigofrank here's what I propose

Let the worker process events coming out of the runtime, find the `_webhookResponse` key on `step:complete`, and when the run is over, send the response to Lightning via a new key on the `run:complete` payload.

So basically the _worker_ creams the magic key off state, and sends it to Lightning via a pre-agreed protocol at the _end_ of a run.

It needs to work  like this:
* The worker reads the `event.state` object of `step:complete` in `src/events/step-complete`
* If it finds a key called `_webhookResponse`, it extracts that object and writes it to `context.state.webhookResponse` (`context.state` is internal worker state for the run, it is not user state returned by steps. Confusing I know!)
* You can choose whether the first or last response is the one that is used (either always set the key, or only set it the first time).
* In the `run:complete` handler in `src/events/run-complete`, look in `context.state` to see if `webhookResponse` was set
* If it IS set, update the `step:complete` payload we send to lightning to include `webhook_response`
* Type definitions for `RunState` and `RunCompletePayload` need updating
* When the run has completed, Lightning will then read the webhook response returned on that payload, and respond via http to the caller.

When we later add support for early responses, we'll add `webhook_response` to the `StepCompletePayload` and ensure that we _only_ send this value to Lightning once.

This works fine with zero persistence. The worker receives the full state object from the runtime regardless of whether zero persistence is set. So it can read from user state and update the internal run state, no problem. Zero persistence just prevents the state being sent to Lighting. Not a problem if we're re-imagining this as a Worker feature.

The alternative would be do add a function somewhere which instantly sends the event out to lightning. We'd add like a `custom:response` event and a generic transport for the runtime to issue application-events from a run. This would have to be done via a function call in the job code. It's hard for us to reason about this in the workflow diagram, hard for users to do, and hard for us to engineer (where does the function come from? We can't add it to an adaptor so the worker must inject it somehow). I don't want to go down that road.

I'm serious about the underscored `_webhookResponse` key though. Users will be setting that in job code. Users have to type that ugly key out when they build their state objects. I really really think making it an underscore key is problematic. Can we think of another example of an underscore key that's used today?

midigofrank:
@josephjclark I love this, I have 3 questions:

1. I don't fully understand point 3. How will users choose whether it's the first or last call that wins? Is it that the last call will always win?
2. If we later support early responses, how will the worker know that it should send `webhook_response` on `step:complete` and not `run:complete`? Or do we leave this for lightning to decide when to respond
3. I think we should know which step sent the `webhook_response`, even if it is sent on `run:complete`. I'm seeing a possibility in the future where we show the webhook responses payload in the UI, so it might be benefitial to include that metadata early in the design



Regarding the leading underscore, it's just a general convention for private/internal stuff. In elixirland, you'll find libraries using leading underscores to track their internal state. I'm guessing because we're allowing users to set it, then it's no longer "private".  @taylordowns2000 do you know if it's already being used somewhere in openfn?

josephjclark:
> 1. I don't fully understand point 3. How will users choose whether it's the first or last call that wins? Is it that the last call will always win?

Users don't choose - you as the architect will choose :)

> 2. If we later support early responses, how will the worker know that it should send webhook_response on step:complete and not run:complete? Or do we leave this for lightning to decide when to respond

I think we decide that later, but I think the worker would _always_ send webhook_repsonse on `step:complete` (rather than `run:complete`), and Lightning decides when to actually trigger the request (maybe option driven). But I think it's quite intuitive for users that the response is sent at the end of the step when it is set. So if I set the response on step 2, the response will be sent before step 3 starts. And if a user wants to send the response at the very end, they just set the state key on the final step. 

Farhan actually flagged to me that we should send the response on step:complete _anyway_, rather than wait for the run to complete.

> 3. I think we should know which step sent the webhook_response, even if it is sent on run:complete.

Cool, let's add that to the return payload

> Regarding the leading underscore, it's just a general convention for private/internal stuff.

Sure but if that key is part of the contract between the worker and lightning, it's not private.

I think in openfn the underscore is more used for system keys (things used by the platform) and less for private ones. But I suspect that's actually quite an old idea - I don't think there's any modern underscore keys used by Lightning, the worker, the runtime, or the adaptors.

midigofrank:
Thanks @josephjclark , I agree with @doc-han that we should just send it on `step:complete` . This is future proof and lets lightning decide when to respond depending on the configuration set in the trigger:

- `before_start`: lightning doesn't wait for the worker, responds immediately
- `after_completion` (what we're doing currently):  lightning waits for `run:complete` in order to respond. In this mode, the `webhook_response` that comes in last `step:complete` should be used (last one wins)
- `after_job_completion` (future): lightning responds back with the first `webhook_response` in `step:complete` (first one wins)

josephjclark:
Worker `@openfn/ws-worker@1.25.0` has the support for this. At the time of writing its released to npm but not deployed to prod or staging.

assets/js/yaml/schema/workflow-spec.json
```diff
@@ -62,6 +62,14 @@
               "type": ["string", "null"],
               "enum": ["before_start", "after_completion", "custom", null]
             },
+            "webhook_response": {
```
midigofrank:
```suggestion
            "webhook_response_config": {
```

assets/js/yaml/types.ts
```diff
@@ -87,11 +91,17 @@ export type SpecCronTrigger = {
...
+  success_code?: number | null;
+  error_code?: number | null;
+};
+
 export type SpecWebhookTrigger = {
   id?: string;
   type: 'webhook';
   enabled: boolean;
   webhook_reply: string | null;
+  webhook_response?: WebhookResponseConfig | null;
```
midigofrank:
```suggestion
  webhook_response_config?: WebhookResponseConfig | null;
```

PR #4670, opened 2026-04-28
Soft-delete sandboxes with grace-period purge

### LINKED ISSUE #4649: sandboxes: maybe save deleted sandboxes for N days, and allow undo
Right now, when a sandbox is deleted, it is dropped instantly from the database.

I've had whiffs of pushback about this from services.

One thing that happen is you can merge a sandbox, only merging Workflows A and B. But after the merge you realise you also wanted to include workflow C!

Or maybe you merged the sandbox to main, but you want to double check the changes (perhaps a bug crept in on main and you want to track what's going on).

Or maybe it was just deleted in error and someone is freaking out.

Two features that might help this are:
- Flag  sandboxes as deleted and leave them for N days before fully dropping them (just like we do with credentials)
- Allow a delete to be cancelled or undone

elias-ba:
## Description

Sandbox deletion, both manual and automatic-after-merge, now happens softly. The sandbox and its descendant subtree are scheduled for deletion and stay around for a grace period (configurable via `PURGE_DELETED_AFTER_DAYS`, default 7 days) before the existing project purge worker permanently removes them. This mirrors how credentials and root projects already work, so accidental deletions can be recovered by a workspace administrator from `/settings/projects`.

The merge modal also has a new "Delete sandbox after merging" toggle, default on. Turning it off keeps the sandbox alive after the merge for ongoing work, picking up the opt-out behaviour proposed in the issue's comments.

End-user cancel UI is intentionally not in this PR. We did not want a sandbox-only recovery affordance while project owners still can't self-cancel either; #4673 tracks adding self-serve cancel for both projects and sandboxes together so they stay consistent.

Closes #4649

## Validation steps

1. Create a sandbox, delete it from the sandboxes tab. Confirm the sandbox disappears from the list, the database row has its scheduled-deletion timestamp set, and triggers in the subtree are disabled.
2. From the superuser `/settings/projects` page, click "Cancel deletion" on the sandbox row. Confirm the sandbox reappears under the parent's sandboxes tab and the timestamp is cleared on the whole subtree.
3. Merge a sandbox into its parent with the toggle on (default). Confirm the flash mentions the merge succeeded plus the deletion was scheduled, and the sandbox is hidden from the list afterwards.
4. Merge a sandbox with the toggle off. Confirm the merge happens, the sandbox stays in the list, and nothing was scheduled for deletion.
5. Create a parent sandbox with descendants, schedule it for deletion, advance time past the grace window, and trigger the purge worker. Confirm the entire subtree is purged.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/sandbox_live/components.ex
+++ b/lib/lightning_web/live/sandbox_live/components.ex
@@ -290,7 +290,7 @@ defmodule LightningWeb.SandboxLive.Components do
           <div class="space-y-2">
             <div class="flex items-baseline justify-between">
               <span class="text-xs font-medium uppercase tracking-wide text-gray-500">
-                Workflows
+                Workflows to merge
               </span>
               <span class="text-xs text-gray-500">
                 {MapSet.size(@selected_workflow_ids)} of {length(@source_workflows)} selected
--- a/lib/lightning_web/live/sandbox_live/components.ex
+++ b/lib/lightning_web/live/sandbox_live/components.ex
@@ -144,7 +144,9 @@ defmodule LightningWeb.SandboxLive.Components do
 
       <section class="space-y-4">
         <p class="text-gray-700">
-          Deleting a sandbox permanently removes its workflows, triggers, versions, keychain clones, and dataclips.
+          Deleting a sandbox removes it (along with its descendants) from
+          OpenFn. Workflows, triggers, versions, keychain clones, and
+          dataclips will be permanently removed once the grace period ends.
           <%= if @sandbox.is_current do %>
             You are currently viewing this project.
             After deletion, you'll be redirected to <strong>{@root_project.name}</strong>.
@@ -152,8 +154,10 @@ defmodule LightningWeb.SandboxLive.Components do
           To confirm, type the sandbox name below.
         </p>
 
-        <div class="bg-red-50 border border-red-200 rounded-md p-3">
-          <p class="text-sm text-red-800">This action cannot be undone.</p>
+        <div class="bg-amber-50 border border-amber-200 rounded-md p-3">
+          <p class="text-sm text-amber-800">
+            The sandbox will be scheduled for deletion and retained for {grace_period_label()} before being permanently removed. Contact a workspace administrator if you need it restored within that window.
+          </p>
         </div>
 
         <.form
@@ -178,7 +182,7 @@ defmodule LightningWeb.SandboxLive.Components do
               disabled={!@changeset.valid?}
               {if !@changeset.valid?, do: [tooltip: "Type the sandbox name to enable"], else: []}
             >
-              Delete Sandbox
+              Delete sandbox
             </.button>
             <.button
               theme="secondary"
@@ -192,6 +196,14 @@ defmodule LightningWeb.SandboxLive.Components do
       </section>
     </.modal>
     """
+  end
+
+  defp grace_period_label do
+    case Lightning.Config.purge_deleted_after_days() do
+      nil -> "the configured grace period"
+      1 -> "1 day"
+      days when is_integer(days) -> "#{days} days"
+    end
   end
 
   attr :open?, :boolean, required: true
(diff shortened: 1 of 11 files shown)


midigofrank (approved):
Nicely done

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -344,22 +361,31 @@ defmodule LightningWeb.SandboxLive.Components do
...
-                Child sandbox <strong>{List.first(@descendants).name}</strong>
-                will also be permanently closed.
-              </div>
-              <div :if={@descendant_count > 1} class="mt-2">
-                Its {@descendant_count} child sandboxes will also be permanently closed.
-              </div>
-            </:message>
-          </Common.alert>
+          <div class="space-y-2">
+            <.input
```
josephjclark: Oh no! I didn't mean for this to be implemented now. Sorry I wish I'd seen earlier.

This basically enables long-lived sandboxes, which is something I'm not prepared to enable just yet

My point in the issue was that _if_ we add long running sandboxes, the soft delete is still a useful thing to include. And therefor it's not blocking or anything.

This needs deleting or splitting out into a separate branch

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -144,16 +144,20 @@ defmodule LightningWeb.SandboxLive.Components do
 
       <section class="space-y-4">
         <p class="text-gray-700">
-          Deleting a sandbox permanently removes its workflows, triggers, versions, keychain clones, and dataclips.
+          Deleting a sandbox removes it (along with its descendants) from
```
josephjclark: I think this is too wordy. And not nicely formatted.

Can we try this:

```
Deleting a sandbox removes it (along with its descendants) from OpenFn.

Workflows, triggers, versions, keychain clones, and dataclips will be permanently removed.

To confirm, type the sandbox name below.

[This sandbox will be retained for 7 days before being permanently removed. Contact a workspace administrator if you need it restored within that window.]
```

The stuff in square brackets is the boxout

lib/lightning_web/live/sandbox_live/components.ex
```diff
@@ -144,16 +144,20 @@ defmodule LightningWeb.SandboxLive.Components do
...
             After deletion, you'll be redirected to <strong>{@root_project.name}</strong>.
           <% end %>
           To confirm, type the sandbox name below.
         </p>
 
-        <div class="bg-red-50 border border-red-200 rounded-md p-3">
-          <p class="text-sm text-red-800">This action cannot be undone.</p>
+        <div class="bg-amber-50 border border-amber-200 rounded-md p-3">
+          <p class="text-sm text-amber-800">
+            The sandbox will be scheduled for deletion and retained for {grace_period_label()} before being permanently removed. Contact a workspace administrator if you need it restored within that window.
```
josephjclark: After deleting a sandbox, I can't see it anywhere. I'm owner of this project and it doesn't show up.

What is a "workspace administrator"? Is that language we use elsewhere?

A project scheduled for deletion seems to still show up in my Projects list. Shouldn't it be the same for Sandboxes?

Also worth nothing that if I delete a sandbox, and then try and create a new sandbox with the same name, I get an error like `dev already exists`. But I can't see it in my sandboxes list

josephjclark (changes requested):
Nice work @elias-ba , but a few things jump out at me. See comments but as a summary:

* I'm worried that only superusers can revert deleted sandboxes. Removing a sandbox isn't like removing a project - I think we need to be a bit more relaxed about this in the UI?
* I didn't intend the "delete sandbox" checkbox to be included here. That basically enables long running sandboxes and I'm not really to expose that functionality yet - not before a bit more thinking and testing. Please revert this.

elias-ba:
Hey @josephjclark, I think I've addressed all your comments. Toggle reverted, modal copy is your version, and scheduled sandboxes now show in the list with a self-serve Cancel-deletion button (gated by delete permission). The three telemetry events from Slack are wired too. Could you take another look?

josephjclark (approved):
@elias-ba I think this is great. It works as I expect (although I haven't waited 7 days to see if the deletion actually goes through)

per slack I' m recommending a little tweak to make the UI more like this:

Image

I'm also going to raise an issue to add a "delete now" option and do a bigger UX/UI review of some of these screens.

In any case I'm quite happy to merge this. Thanks man!

PR #4672, opened 2026-04-28
Add pagination to credentials and project credentials page

### LINKED ISSUE #4301: Add pagination to credentials and project credentials page
The credentials and project credentials list pages do not have paginations. When a user or a project has many credentials the tables become super long. We should:

1. Limit the table to 10 credentials, oauth clients, and keychain credentials per page
2. Maybe reorganize the pages to show credentials first, keychain credentials second, and oauth clients last.
3. We could also put oauth clients behind a collapsible section to avoid always seeing them in that page (not all users wanna see them)

Image

sharleenawinja:
## Description

This PR adds pagination to the credentials, keychain credentials, and OAuth clients tables on the credentials list pages (10 items per page, with Previous/Next control). OAuth clients are now in a collapsible section, collapsed by default. Tables are ordered: credentials first, keychain credentials second, OAuth clients last.

Closes #4301

## Validation steps

1. Go to `/credentials` with more than 10 credentials — confirm the table paginates

2. Confirm the OAuth Clients section is collapsed by default and the toggle shows/hides it.

3. If there are more than 10 OAuth clients, expand the section and confirm pagination works.

4. Verify the same behaviour on **Project Settings → Credentials**.

## AI Usage

want to know!):

- [X] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [X] I have performed an AI review of my code (we recommend using `/review`
      with Claude Code)
- [X] I have updated the **changelog**.
- [X] I have ticked a box in "AI usage" in this PR

### DIFF
--- a/lib/lightning/scrivener/query_paginater.ex
+++ b/lib/lightning/scrivener/query_paginater.ex
@@ -90,7 +90,7 @@ defimpl Scrivener.Paginater, for: Ecto.Query do
   defp aggregate(
          %{
            group_bys: [
-             %Ecto.Query.QueryExpr{
+             %{
                expr: [
                  {{:., [], [{:&, [], [source_index]}, field]}, [], []} | _
                ]
--- a/lib/lightning_web/live/credential_live/credential_index_component.ex
+++ b/lib/lightning_web/live/credential_live/credential_index_component.ex
@@ -7,6 +7,8 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
   alias Lightning.OauthClients
   alias Lightning.Policies
 
+  @page_size 10
+
   @impl true
   def mount(socket) do
     {:ok,
@@ -16,10 +18,14 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
        can_create_project_credential: false,
        credential: nil,
        credentials: [],
+       credentials_page: 1,
        current_user: nil,
        keychain_credentials: nil,
+       keychain_credentials_page: 1,
        oauth_client: nil,
        oauth_clients: [],
+       oauth_clients_expanded: false,
+       oauth_clients_page: 1,
        project: nil,
        project_user: nil,
        projects: [],
@@ -86,6 +92,24 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
      socket
      |> assign(active_modal: nil, credential: nil, oauth_client: nil)
      |> load_credentials()}
+  end
+
+  def handle_event(
+        "change_page",
+        %{"table" => table, "page" => page, "container_id" => container_id},
+        socket
+      ) do
+    page = if is_integer(page), do: page, else: String.to_integer(page)
+    key = :"#{table}_page"
+
+    {:noreply,
+     socket
+     |> assign(key, page)
+     |> push_event("scroll-to-top", %{id: container_id})}
+  end
+
+  def handle_event("toggle_oauth_clients", _params, socket) do
+    {:noreply, update(socket, :oauth_clients_expanded, &(!&1))}
   end
 
   def handle_event("show_modal", %{"target" => "new_credential"}, socket) do
@@ -401,6 +425,22 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
      socket
      |> put_flash(:error, "You are not authorized to perform this action")
      |> push_patch(to: socket.assigns.return_to)}
+  end
+
+  defp paginate(list, page) when is_list(list) do
+    total = length(list)
+    total_pages = max(1, div(total + @page_size - 1, @page_size))
+    page_num = max(1, min(page, total_pages))
+    slice = Enum.slice(list, (page_num - 1) * @page_size, @page_size)
+
+    page_info = %{
+      page_number: page_num,
+      page_size: @page_size,
+      total_entries: total,
+      total_pages: total_pages
+    }
+
+    {slice, page_info}
   end
 
   defp list_credentials(user_or_project) do
--- a/lib/lightning_web/live/credential_live/credential_index_component.ex
+++ b/lib/lightning_web/live/credential_live/credential_index_component.ex
@@ -7,7 +7,13 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
   alias Lightning.OauthClients
   alias Lightning.Policies
 
-  @page_size 10
+  @empty_page %{
+    entries: [],
+    page_size: 0,
+    total_entries: 0,
+    page_number: 1,
+    total_pages: 0
+  }
 
   @impl true
   def mount(socket) do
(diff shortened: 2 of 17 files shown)


sharleenawinja:
@midigofrank could you please review this PR when you have a moment? Thanks!

midigofrank:
Hey @sharleenawinja , thank you. I will review this today

josephjclark:
Hi @sharleenawinja, thank you for this! There are two tests failing in CI, I think just because some of the HTML structures have changed. Would you mind taking a look before Frank reviews this properly? You can see the errors here: https://app.circleci.com/pipelines/github/OpenFn/lightning/17326/workflows/2b9ed83e-0bac-4f14-9180-2958451d29ac/jobs/93782/tests

midigofrank (changes requested):
Hey @sharleenawinja , thank you for picking this up, you've picked a tough one, pagination in a page with multiple lists is an unchartered territory in lightning.

The credentials page is gnarly, it has multiple lists and the lists are shared in both the credentials page and the project settings page. 

I see that you've chosen to do the pagination in the UI, I think that's understandable given the complexity of the page. I'm just not sure if that's the direction we want to take given that every other place does pagination using db queries. @elias-ba what do you think about this approach?

I also noticed that you created a new pagination component, could you try and reuse the one in https://github.com/OpenFn/lightning/blob/main/lib/lightning_web/pagination.ex ?

Also, test this in both pages (project settings and credentials page)

sharleenawinja:
@midigofrank Thanks for the context. I'll rework the pagination to use DB queries to stay consistent with the rest of the app, and I'll reuse the existing pagination component instead of the one I added. I'll update the PR once that's done.

sharleenawinja:
@josephjclark @midigofrank @elias-ba I have switched to DB-level pagination and reused the existing pagination_bar component. Each table now paginates via URL params. For the pipeline, the elixir tests job is failing due to pre-existing flaky tests.

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -797,6 +865,34 @@ defmodule LightningWeb.ProjectLive.Settings do
     end
   end
 
+  defp map_credentials(%Scrivener.Page{} = page) do
```
elias-ba, line 868: This block (`map_credentials` / `add_credential_display_fields` / `map_oauth_clients` / `add_oauth_client_display_fields`) is now identical to the one in `credential_live/index.ex`. It lived in one place inside `CredentialIndexComponent` before the refactor pulled loading up into the two parents. Could we lift these into a shared home so they don't drift, maybe `Credentials.with_display_fields/1` in the context, or a small helper both LiveViews call? Non-blocking, but it's the thing I'd most like sorted before merge.

lib/lightning_web/live/project_live/settings.ex
```diff
@@ -128,9 +129,33 @@ defmodule LightningWeb.ProjectLive.Settings do
        active_menu_item: :settings,
        can_receive_failure_alerts: can_receive_failure_alerts,
        collaborators_to_invite: [],
+       credentials_page: %{
```
elias-ba, line 132: Small one: these placeholder pages are hand-written maps (and there's another as `@empty_page` in `CredentialIndexComponent`), while the real values from `apply_action` are `%Scrivener.Page{}` structs. Safe today since `:index` always overwrites before render, but it's easy for the shapes to drift apart. Maybe one `empty_page()` helper, or just `%Scrivener.Page{page_number: 1}`, so there's a single source of truth?

lib/lightning/scrivener/query_paginater.ex
```diff
@@ -90,7 +90,7 @@ defimpl Scrivener.Paginater, for: Ecto.Query do
   defp aggregate(
          %{
            group_bys: [
-             %Ecto.Query.QueryExpr{
+             %{
```
elias-ba, line 93: This one I'd like to think through with you. Loosening `%Ecto.Query.QueryExpr{}` to a bare `%{expr: ...}` works, but it now matches any map with an `expr` key rather than the actual struct. I think the precise fix is `%Ecto.Query.ByExpr{}`, that's what `group_by` builds in Ecto 3.13, which keeps the type guard. It's also a shared file every grouped-paginated query flows through, with no test hitting it directly. Could we name the struct and add a small test for the grouped-count path? More a robustness worry than a blocker.

lib/lightning_web/live/credential_live/credential_index_component.ex
```diff
@@ -45,47 +57,26 @@ defmodule LightningWeb.CredentialLive.CredentialIndexComponent do
...
+    {:ok, socket |> assign(assigns)}
   end
 
   @impl true
   def handle_event("close_active_modal", _params, socket) do
     {:noreply,
      socket
-     |> assign(active_modal: nil, credential: nil, oauth_client: nil)
-     |> load_credentials()}
+     |> assign(active_modal: nil, credential: nil, oauth_client: nil)}
```
elias-ba, line 75: More of a question than a problem. `close_active_modal` used to reload the lists; now it just clears the modal, so a create/edit only appears if the save handler remembers to `push_patch` to `return_to` (the delete paths now do). It works in the tests, I just want to be sure the create-then-close path always routes back through `handle_params` so the new row shows up, rather than leaning on each handler to remember. Did the new-credential flow refresh cleanly when you tested it?

lib/lightning_web/live/credential_live/index.ex
```diff
@@ -34,4 +67,32 @@ defmodule LightningWeb.CredentialLive.Index do
...
+  end
+
+  defp add_credential_display_fields(credential) do
+    project_names = Map.get(credential, :projects, []) |> Enum.map(& &1.name)
+
+    environment_names =
+      credential |> Map.get(:credential_bodies, []) |> Enum.map(& &1.name)
+
+    credential
+    |> Map.put(:project_names, project_names)
```
elias-ba, line 82: Non-blocking, and carried over from the old component rather than something you introduced: we're adding `:project_names`/`:environment_names` onto the `%Credential{}` struct with `Map.put`, which quietly turns it into a struct with keys that aren't in the schema. Fine for display, but since we're already reshaping this, deriving them in the template (or a small view struct) would keep the schema honest. Happy to leave it for a follow-up.

elias-ba (commented):
Hey @sharleenawinja, really nice work on this 🙌 I pulled it down, ran the credentials and oauth suites and read through it carefully. It reads well, and the scoping and tests look right to me.

Nothing blocking from me. A few things I'd love us to look at, mostly questions:

- **A bit of duplication I'd like to consolidate** (inline). The display-mapping helpers are now copy-pasted into both `index.ex` and `settings.ex`; they used to live once inside the shared component. This is the one I'd most like tidied.
- **The paginater change** (inline). Works, but I think we can name the actual Ecto struct instead of matching a bare map, and it's worth a test since it's a shared file.
- **A couple of smaller things** (inline): the repeated empty-page maps, the `Map.put` display fields, and a question about whether create/edit reliably refreshes the list now that `close_active_modal` no longer reloads.

Honestly it's close. Lift the duplicated helpers into one home and I'm happy. Thanks for taking this one on 🙏

elias-ba:
Oh, and the red `test_elixir`, I don't think that one's on you. The two failures are both outside what this PR touches. A rebase on `main` and a re-push should give it a fresh run. If either still fails after that, let us know.

PR #4685, opened 2026-04-30
Autoscale collection storage column and make it sortable

### LINKED ISSUE #4684: Collection storage column shows 0 for collections under 1 MB
## What's broken

When you open a project and go to Settings, then Collections, each collection has a "Used Storage (MB)" column. Right now that column shows `0` for any collection that holds less than 1 MB of data. The number is correct in the database, the page just rounds it down to zero whole megabytes.

The same column exists on the admin "all collections" page and has the same behaviour.

## How it surfaced

Caught during QA of #4613. The sandboxes-in-collections work has a validation step that says you should see storage size greater than 0 after running a small test workflow. The test data was a few kilobytes, so the column showed `0` and it looked like writes weren't being recorded. They were, the storage just wasn't visible at that scale.

## What to fix

Two things:

1. Render each value in the most natural unit for its size. A small collection should read "1.4 KB", a medium one "320 MB", a large one "2.1 GB". Same convention file systems and most cloud dashboards use.
2. Drop "(MB)" from the column header so it just says "Used storage". Otherwise the header lies whenever a value is shown in KB or GB.

We did consider keeping a fixed unit:

- KB only works for small collections, but turns a 600 MB collection into "600,000 KB", which reads worse.
- MB with decimals (e.g. "0.001 MB") is still ugly for sub-MB values, and the exact confusion that surfaced this issue (zero for tiny payloads) would still happen at one-decimal precision.

Autoscaling reads cleanly at any size and prevents this from biting again.

Loom video here: https://www.loom.com/share/edeb8b65ffe7486fa539c5b8fab27853

elias-ba:
## Description

The Collections panel under Project Settings (and the admin collections index) used to render the storage column as integer megabytes, so any collection holding less than one megabyte displayed as `0`. The underlying counter was always correct, the rendering just rounded down. This PR replaces the integer-MB rendering with autoscaled units (B, KB, MB, GB, TB) via a shared helper, drops the hardcoded "(MB)" from the column header, and makes the column sortable alongside Name. Default sort is by name, ascending.

Closes #4684

## Validation steps

Loom video here for faster QA: https://www.loom.com/share/2caf6a6cb2434158a14534689aa19e0c

1. Open a project and navigate to **Settings**, then **Collections**.
2. Confirm the column header reads "Used storage" (no unit in the title).
3. Confirm a collection with a small payload (a few kilobytes) now shows its size as e.g. "1.5 KB" rather than "0".
4. Click the **Name** column header. Order toggles between ascending and descending.
5. Click the **Used storage** column header. Order toggles between ascending and descending and the chevron indicator follows the active column.
6. Repeat on the admin collections index at `/settings/collections` (superuser).
7. To exercise the full unit range, vary the storage size of a few collections from a few hundred bytes up to several terabytes and confirm each renders in the natural unit (B, KB, MB, GB, TB).

The default sort for the per-project collections list changes from "newest first" to "by name (ascending)". The two internal callers of that list function (sandbox provisioning and merge sync) only consume names and byte-size totals, so they are not order-sensitive.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,18 @@ and this project adheres to
 
 ### Changed
 
+- Project Settings, Collections panel: rename the "Used Storage (MB)" column to
+  "Used storage" and render values with autoscaled units (B, KB, MB, GB, TB)
+  instead of integer megabytes. The column is now sortable, alongside the
+  existing Name column. Same change applied on the admin collections index.
+  [#4684](https://github.com/OpenFn/lightning/issues/4684)
+
 ### Fixed
+
+- Collection storage on Project Settings, Collections no longer shows `0` for
+  collections holding less than one megabyte of data. The underlying counter was
+  always correct, the rendering now reflects values at any scale.
+  [#4684](https://github.com/OpenFn/lightning/issues/4684)
 
 ## [2.16.3-pre] - 2026-04-30
 
--- a/lib/lightning/collections.ex
+++ b/lib/lightning/collections.ex
@@ -42,12 +42,15 @@ defmodule Lightning.Collections do
     Repo.all(from(c in Collection, order_by: ^order_by, preload: ^preload))
   end
 
-  @spec list_project_collections(Project.t()) :: [Collection.t(), ...] | []
-  def list_project_collections(%Project{id: project_id}) do
+  @spec list_project_collections(Project.t(), keyword()) ::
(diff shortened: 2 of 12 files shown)


taylordowns2000 (approved):
Fantastic. Thank you!

theroinaochieng (approved):

PR #4687, opened 2026-05-01
Fix wrong sort order at UTC month boundaries and pre-release versions

elias-ba:
## Description

Several places in the codebase used `Enum.sort_by(.., :asc | :desc)` (or the equivalent `Enum.sort/2` form, or `<=`/`>=` as a comparator) on a key whose value was a `DateTime` or a `Version` struct. That form does Erlang structural term comparison, which on these structs walks fields alphabetically and disagrees with chronological / semver order. Same root cause across four surfaces.

- **Workflow list, "Latest Work Order" column** (project dashboard). Sorted incorrectly when the most recent work orders for two workflows fell on either side of a UTC month boundary.
- **Projects list, "Last Updated" column, support-user view**. Same shape, same trigger. Non-support users were unaffected; they go through a SQL `order_by`.
- **Admin tables that sort on date columns** ("Created at" and "Scheduled deletion" on the admin projects table, "Scheduled deletion" on the admin users table). All go through a shared `TableHelpers.sort_items/4` helper, which used `<=`/`>=` (structural compare) and produced the same inversion at month boundaries.
- **Adaptor version dropdown**. Listed pre-release versions above their corresponding stable release because the sort relied on structural compare of parsed version structs.

Each fix maps the sort key through a domain-aware comparator: `DateTime` becomes integer microseconds (or ISO 8601 strings inside the shared table helper), version strings go through `Version.compare/2`. The dashboard, projects-overview, and admin-table fixes have real user impact at month boundaries. The adaptor version fix is a real-but-currently-zero-impact correction (OpenFn adaptors essentially always ship plain `X.Y.Z` semver, so today no user is hitting it; included here because it is the same root cause).

## Validation steps

1. Run the affected test files together:

   ```
   mix test \
     test/lightning/dashboard_stats_test.exs \
     test/lightning/projects_test.exs \
     test/lightning_web/live/helpers/table_helpers_test.exs \
     test/lightning_web/live/job_live/adaptor_picker_test.exs
   ```

   All tests pass on this branch, including four new regression tests that pin chronological / semver ordering across boundaries.

2. Quick spot-check in `iex -S mix`:

   ```elixir
   # DateTime sort, dashboard / projects family
   stats = [
     %Lightning.DashboardStats.WorkflowStats{workflow: %{name: "older"}, last_workorder: %{updated_at: ~U[2026-04-30 23:10:00Z]}},
     %Lightning.DashboardStats.WorkflowStats{workflow: %{name: "newer"}, last_workorder: %{updated_at: ~U[2026-05-01 01:10:00Z]}}
   ]
   Lightning.DashboardStats.sort_workflow_stats(stats, :last_workorder_updated_at, :desc) |> Enum.map(& &1.workflow.name)
   # On this branch: ["newer", "older"]. On main: ["older", "newer"].

   # Admin tables sort helper
   items = [
     %{name: "April", inserted_at: ~U[2026-04-30 23:10:00Z]},
     %{name: "May", inserted_at: ~U[2026-05-01 01:10:00Z]}
   ]
   LightningWeb.Live.Helpers.TableHelpers.sort_items(items, "inserted_at", "desc", %{"inserted_at" => :inserted_at}) |> Enum.map(& &1.name)
   # On this branch: ["May", "April"]. On main: ["April", "May"].

   # Version sort, adaptor family
   LightningWeb.JobLive.AdaptorPicker.sort_versions_desc(["1.0.0", "1.0.0-beta", "1.0.0-alpha"])
   # On this branch: ["1.0.0", "1.0.0-beta", "1.0.0-alpha"]. On main: ["1.0.0-beta", "1.0.0-alpha", "1.0.0"].
   ```

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ and this project adheres to
   in-memory sort over project rows previously relied on structural comparison of
   timestamps and could invert when projects' last activity straddled a month
   boundary in UTC.
+- Sort admin tables chronologically when the active column is a date column
+  (admin projects "Created at" and "Scheduled deletion", admin users "Scheduled
+  deletion"). The shared in-memory sort helper previously used structural
+  comparison on date values and could invert across a month boundary in UTC.
 
 ## [2.16.3-pre] - 2026-04-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,18 +24,20 @@ and this project adheres to
 - Sort the project workflow list chronologically by Latest Work Order. The
   column previously sorted incorrectly when the most recent work orders for two
   workflows fell on either side of a month boundary in UTC.
+  [#4687](https://github.com/OpenFn/lightning/pull/4687)
 - Sort the adaptor version dropdown per semver. Pre-release versions previously
   appeared above the corresponding stable release because the sort relied on
   structural comparison of parsed version structs rather than
-  `Version.compare/2`.
+  `Version.compare/2`. [#4687](https://github.com/OpenFn/lightning/pull/4687)
 - Sort the support-user projects overview chronologically by Last Updated. The
   in-memory sort over project rows previously relied on structural comparison of
   timestamps and could invert when projects' last activity straddled a month
-  boundary in UTC.
+  boundary in UTC. [#4687](https://github.com/OpenFn/lightning/pull/4687)
 - Sort admin tables chronologically when the active column is a date column
   (admin projects "Created at" and "Scheduled deletion", admin users "Scheduled
   deletion"). The shared in-memory sort helper previously used structural
   comparison on date values and could invert across a month boundary in UTC.
+  [#4687](https://github.com/OpenFn/lightning/pull/4687)
 - Prevent crash when an unsupported data type from `credential-schema.json` is
   loaded for building a credential schema from an adaptor. Fall-back to
   `:string` type, log warning and alert Sentry.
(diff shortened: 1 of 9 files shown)


taylordowns2000 (approved):
Approved, but please update the changelog (link to the issue or the PR itself) and consider these before merging:

1. What was using the DateTime that you're now converting to Unix? Do all those downstream sites (maybe UI displays of the dates?) now need updating too?
2. Does the `nil -> 0` branch map missing timestamps to the Unix epoch? Mixing "absent" with "1970-01-01" would not be great.

midigofrank (changes requested):
Nicely done @elias-ba , I'm just not sure why we need the `0` and `1`, 
why can't we do `DateTime.to_unix("1970-01-01")` when it's `nil`? The tuple makes it feel complicated

elias-ba:
@midigofrank thanks for the review man! I totally hear you on the tuple feeling complicated, and I went back and forth on it myself. Let me explain where the `0` / `1` shape came from.

There's a business requirement we need to honor: a nil row and a hypothetical real `~U[1970-01-01]` row must not tie at the comparator level. The reason your suggestion of mapping nil to `DateTime.to_unix(~U[1970-01-01 00:00:00Z], :microsecond)` doesn't quite get us there is that the result of that conversion is exactly `0`, which is the same value a real `~U[1970-01-01]` row would produce. So the two cases would collapse to the same sort key.

The tuple is the smallest change I could find that keeps both worlds:
- Second slot is unix microseconds, so the chronological compare works exactly the same as your suggestion.
- First slot tags nil vs present, so the two cases stay structurally distinct regardless of what value lands in the second slot.

We also shipped this same shape in the billing app for the same root-cause class, so I'd like to keep it consistent here too.

midigofrank:
Aaah, okay. Makes sense

midigofrank (approved):

PR #4688, opened 2026-05-01
Sort adaptor version dropdown per semver

elias-ba:
## Description

The adaptor version dropdown could list pre-release versions above their corresponding stable release because the sort relied on structural comparison of parsed version structs rather than semver. Switch to `Version.compare/2` so the order matches semver in every case.

This is a real-but-currently-zero-impact bug. OpenFn adaptors essentially always ship plain `X.Y.Z` semver, so no user is hitting this today. Opening as a draft so the fix is documented and reviewable if pre-release adaptors ever land in the registry.

## Validation steps

1. Run `mix test test/lightning_web/live/job_live/adaptor_picker_test.exs`. All 3 tests pass on this branch, including a new regression test that pins the pre-release ordering.

2. To eyeball the bug directly, run `iex -S mix` and paste:

   ```elixir
   versions = ["1.0.0", "1.0.0-beta", "1.0.0-alpha"]
   LightningWeb.JobLive.AdaptorPicker.sort_versions_desc(versions)
   ```

   On this branch you get `["1.0.0", "1.0.0-beta", "1.0.0-alpha"]` (release first, semver-correct). On `main`, the equivalent line `Enum.sort_by(versions, &Version.parse/1, :desc)` returns `["1.0.0-beta", "1.0.0-alpha", "1.0.0"]` (pre-releases ahead of release, the bug).

Found while investigating a related bug in the dashboard sort (#4687). Same family: `Enum.sort_by(.., :asc | :desc)` does structural compare on the key, which disagrees with domain-specific compare for types like `DateTime` and `Version`. Different surface, separate PR.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to
 
 ### Fixed
 
+- Sort the adaptor version dropdown per semver. Pre-release versions previously
+  appeared above the corresponding stable release because the sort relied on
+  structural comparison of parsed version structs rather than
+  `Version.compare/2`.
+
 ## [2.16.3-pre] - 2026-04-30
 
 ### Added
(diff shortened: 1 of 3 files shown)


elias-ba:
Consolidated into #4687. Same root cause as the dashboard and projects-overview sorts; one PR is cleaner than three.

PR #4711, opened 2026-05-06
Fix runs UI crash on unrecognised kill error types

### LINKED ISSUE #4709: Runs UI crashes when a step is killed with `StateTooLargeError`
## What's broken

When a step finishes with `exit_reason = "kill"` and `error_type = "StateTooLargeError"`, the runs UI crashes with a `CaseClauseError` while rendering the step icon. The project history page and the run detail page both fail on render whenever they include such a step. Workflow execution itself is unaffected. Only the rendering breaks.

The crash is in `LightningWeb.RunLive.Components.step_icon/1`, which has explicit clauses for `SecurityError`, `ImportError`, `TimeoutError`, and `OOMError`, but no clause for `StateTooLargeError` and no catch-all.

The same mapping is mirrored in the React `StepIcon` used by the collaborative editor (`assets/js/collaborative-editor/components/run-viewer/StepIcon.tsx`). It has a gray ellipsis fallback so it doesn't crash, but it renders the killed step as "pending", which is misleading.

## How it surfaced

Sentry on `v2.16.2`, both alerts from a single project that hit the new `StateTooLargeError` exit type from the worker:

- [LIGHTNING-1MM](https://openfn.sentry.io/issues/7459916390/) (run detail page)
- [LIGHTNING-1MN](https://openfn.sentry.io/issues/?query=LIGHTNING-1MN) (history page)

## What to fix

Two things, in the same change:

1. Add an explicit clause for `{"kill", "StateTooLargeError"}` in `step_icon/1`, mapping to the same icon as `OOMError` (`hero-exclamation-circle`, yellow), since both are resource-budget kills. The textual label next to the icon already shows the precise `error_type`, so the user can still tell the two apart.
2. Add a `{"kill", _}` catch-all that falls back to the same icon, so any future worker exit type degrades gracefully instead of crashing the LiveView. The `error_type` field is a free-form string set by the worker and Lightning has no enum constraining it, so this class of crash will keep recurring otherwise.

Mirror both changes in the React `StepIcon` so the collaborative editor renders consistently with the runs UI.

elias-ba:
## Description

The runs history page and the run detail page used to crash with a `CaseClauseError` when they tried to render a step killed with `StateTooLargeError` (or any kill error type the renderer didn't know about). The step icon mapping in both the LiveView (`LightningWeb.RunLive.Components.step_icon/1`) and the React `StepIcon` had explicit clauses for `SecurityError`, `ImportError`, `TimeoutError`, and `OOMError` but no fallback. This PR adds an explicit clause for `StateTooLargeError` (mapped to the same resource-budget icon as `OOMError`) and a kill catch-all so any future worker error type falls through to the same icon instead of crashing the page.

Closes #4709

## Validation steps

1. With this branch, view a run history or run detail page that contains a step with `exit_reason = "kill"` and `error_type = "StateTooLargeError"`. The page renders without crashing and the step shows a yellow exclamation-circle icon. The textual `kill:StateTooLargeError` label remains alongside.
2. The same applies to any unrecognised kill error type the worker may produce in the future (covered by the regression test).
3. Run the relevant test files:

   ```
   mix test test/lightning_web/live/run_live/components_test.exs
   ```

   and

   ```
   cd assets && npx vitest run test/collaborative-editor/components/run-viewer/StepIcon.test.tsx
   ```

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


assets/js/collaborative-editor/components/run-viewer/StepIcon.tsx
```diff
@@ -3,8 +3,7 @@ interface StepIconProps {
   errorType: string | null;
 }
 
-// Ported from lib/lightning_web/live/run_live/components.ex:504-555
-// Matches the LiveView step_icon implementation exactly
+// Mirrors LightningWeb.RunLive.Components.step_icon/1.
```
taylordowns2000, line 6: seeing the word "mirrors" makes me nervous. does this mean the same list or same logic is repeated in both places?

taylordowns2000 (approved):
looks good, but have a question about the mirroring - is that going to be unsustainable? (i.e., not DRY?)

assets/js/collaborative-editor/components/run-viewer/StepIcon.tsx
elias-ba: I think you're right. It smells a bit. This is probably a pattern we do have in some places of the code base. Especially areas where we do have similar components in the react world. I don't think we should address this now but I agree with you.

midigofrank (approved):

PR #4717, opened 2026-05-07
Fix cron scheduler tick discards on single trigger failure

### LINKED ISSUE #4716: Cron scheduler discards entire tick when a single trigger query times out
### User story

As someone running cron-triggered workflows, I expect all my scheduled workflows to fire each minute, regardless of what's happening with other tenants' workflows on the same instance.

### Details

`Lightning.Workflows.Scheduler` Oban jobs in the `:scheduler` queue were being discarded with PostgreSQL error 57014 (`canceling statement due to user request`).

The scheduler runs once per minute and loops over every enabled cron trigger. For each one it queries the database for a dataclip to seed the next run. When a query hit Postgrex's 15s `DATABASE_TIMEOUT`, the exception propagated out of the bare `Enum.each` and every trigger after it in that batch was silently skipped. Users with workflows later in the loop got no runs.

We wrote a script to replay a scheduler tick with per-trigger timings to diagnose the problem. A few things showed up:

- Per-trigger queries were scanning millions of rows. For workflows with no successful run history, `EXPLAIN ANALYZE` showed a full backward scan through 2.3M rows, taking 8–25s, to confirm there was no match.
- The slow scan was caused by Postgres switching from a custom plan (fast, bounded to the workflow's work orders) to a generic plan after 5 executions of the same prepared statement. The generic plan chose the wrong index for the access pattern.
- Soft-deleted workflows were included in the cron trigger enumeration — their triggers were still marked enabled even though the workflow had been deleted via the provisioner.

### Release notes

Cron-triggered workflows now run reliably on each tick. A failure on one trigger is logged and reported but no longer causes other scheduled workflows to be skipped.

### User acceptance criteria

- All enabled cron triggers are attempted on each scheduler tick, even if earlier triggers in the same tick encounter errors
- Scheduler failures appear in logs and Sentry with enough context (edge, trigger, job, workflow) to diagnose

stuartc:
## Description

Fixes a bug where a single slow trigger query caused the scheduler to silently skip every subsequent trigger in that minute's tick.

`Lightning.Workflows.Scheduler` runs once a minute and loops over all enabled cron triggers. When a query exceeded Postgrex's 15s database timeout, the exception propagated out of the bare `Enum.each` and all remaining triggers in the batch were dropped without any error recorded.

Three things were wrong:

**No error isolation.** One bad trigger killed the whole tick. The fix wraps each trigger in a `rescue` that logs the failure, reports to Sentry with full context (edge, trigger, job, workflow IDs), and lets the loop continue.

**Slow queries on the cron dataclip path.** The two queries that look up the seed dataclip for a trigger had no supporting composite indexes, causing scans of millions of rows per trigger. Two migrations add the right composites — both are idempotent for indexes of the same name.

**Prepared statement plan-cache flip.** Even with the indexes in place, Postgres switches from a custom plan (bounded, fast) to a generic plan after 5 executions of the same prepared statement on a connection. For workflows with no successful runs, the generic plan scans the entire `runs` table backward looking for nothing — 8–25 seconds per trigger. `prepare: :unnamed` on the two affected `Repo.one` calls pins each query to a custom plan.

As a side fix: soft-deleted workflows processed via the provisioner were leaving their triggers enabled, so the scheduler kept enumerating them. The query now filters on `deleted_at`, and the provisioner disables triggers and notifies the kafka pipeline supervisor on soft-delete, matching what the UI path already did.

Closes #4716

## Validation steps

1. To see the plan-cache flip that caused the slow queries, run the following against your database. Pick any workflow ID that has few or no successful runs:

```sql
PREPARE bench(uuid, varchar) AS
  SELECT d.*
  FROM runs r
  INNER JOIN work_orders wo ON wo.id = r.work_order_id
  INNER JOIN dataclips d ON d.id = r.final_dataclip_id
  WHERE wo.workflow_id = $1 AND r.state = $2 AND d.wiped_at IS NULL
  ORDER BY r.finished_at DESC LIMIT 1;

-- Generic plan — what named prepared statements eventually flip to:
SET plan_cache_mode = force_generic_plan;
EXPLAIN (ANALYZE, BUFFERS)
  EXECUTE bench('<workflow_id>', 'success');
-- Expect: Index Scan Backward on runs_finished_at_index, many rows, slow

-- Custom plan — what prepare: :unnamed always uses:
SET plan_cache_mode = force_custom_plan;
EXPLAIN (ANALYZE, BUFFERS)
  EXECUTE bench('<workflow_id>', 'success');
-- Expect: Nested Loop via runs_work_order_id_state_finished_at_index, <1ms

DEALLOCATE bench;
RESET plan_cache_mode;
```

2. Confirm scheduler failures appear in Sentry with `tags: {type: "scheduler"}` and the four context IDs (edge, trigger, job, workflow) rather than as unattributed Oban job discards.

3. Soft-delete a workflow via the provisioner and confirm its cron trigger no longer appears in the scheduler's next tick.

- The `rescue` in `safe_invoke_cronjob/1` is intentionally `rescue` and not `catch`. `Repo.rollback` uses `throw` internally but that throw is consumed by the transaction wrapper in `Runs.enqueue/1` — it never reaches this frame. `:exit` signals from genuine infrastructure death should propagate to Oban's failure surface rather than get swallowed here.
- `prepare: :unnamed` has a small per-query planning cost. For ~30–50 cron trigger queries per minute this is negligible.

## AI Usage

- [x] I have used Claude Code

## Pre-submission checklist

- [ ] I have implemented and tested all related **authorization policies** *(no auth changes)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to
   renderer doesn't recognise (such as `StateTooLargeError`); unknown kill
   reasons fall back to the resource-budget icon instead.
   [#4709](https://github.com/OpenFn/lightning/issues/4709)
+- Cron scheduler now attempts all triggers each tick, even if one fails. A
+  single slow or erroring trigger previously aborted the entire batch.
+  [#4716](https://github.com/OpenFn/lightning/issues/4716)
 
 ## [2.16.3-pre2] - 2026-05-07
 
--- a/lib/lightning/invocation.ex
+++ b/lib/lightning/invocation.ex
@@ -235,6 +235,12 @@ defmodule Lightning.Invocation do
   Scopes by workflow (not trigger) so that manual runs are also considered.
   """
   def last_run_final_dataclip(%Trigger{workflow_id: workflow_id}) do
+    # `prepare: :unnamed` forces a custom plan per workflow_id. With named
+    # prepared statements, Postgres flips to a generic plan after 5
+    # executions which scans `runs_finished_at_index` backward and
+    # post-filters on state — catastrophic for workflows with sparse
+    # successes (full backward scan looking for nothing). The custom plan
+    # bounds via the workflow's work_orders.
     from(r in Run,
       join: wo in assoc(r, :work_order),
       join: d in assoc(r, :final_dataclip),
@@ -245,7 +251,7 @@ defmodule Lightning.Invocation do
       limit: 1,
       select: d
     )
-    |> Repo.one()
+    |> Repo.one(prepare: :unnamed)
   end
 
   @doc """
@@ -253,6 +259,9 @@ defmodule Lightning.Invocation do
   Used when cron_cursor_job_id is set to a specific job.
   """
   def last_successful_step_dataclip(job_id) do
+    # `prepare: :unnamed` for the same reason as `last_run_final_dataclip/1`:
+    # LIMIT 1 + ORDER BY DESC queries are vulnerable to generic-plan flips
+    # that degrade into full-table scans. Custom plans guarantee selectivity.
     from(d in Dataclip,
       join: s in Step,
       on: s.output_dataclip_id == d.id,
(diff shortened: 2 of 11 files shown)


midigofrank (approved):
Nicely done

PR #4720, opened 2026-05-07
Fix legacy workflow editor crash on AI assistant streaming events

### LINKED ISSUE #4719: Legacy workflow editor crashes when AI assistant streams
## What's broken

In the legacy workflow editor (`WorkflowLive.Edit`, served from `/w/new/legacy` and `/w/:id/legacy`), `handle_info({:ai_assistant, action, payload}, socket)` does a `case action` over `:canvas_state_changed`, `:workflow_params_changed`, `:message_status_changed`, `:register_component`, `:unregister_component`. There is no fallthrough.

When the legacy AI assistant LiveComponent registers, the parent LV subscribes to `ai_session:#{session_id}` so it can relay `:message_status_changed` back to the LiveComponent. That relay is what tells the spinner to stop and renders the finished message. The same PubSub topic also carries the four streaming events used by the collaborative editor: `:streaming_status`, `:streaming_chunk`, `:streaming_changes`, `:streaming_error`. These were added later for the React side and the legacy `case` never grew clauses for them. As soon as Apollo emits the first event (typically a `:streaming_status` "Thinking..."), the LV crashes with `CaseClauseError`.

## How it surfaced

Sentry on `v2.16.2`, 1 event so far: [LIGHTNING-1MV](https://openfn.sentry.io/issues/?query=LIGHTNING-1MV). Breadcrumbs show `:register_component` for a new AI session, then `:streaming_status` with `text: "Thinking..."`, then the crash in `LightningWeb.WorkflowLive.Edit.handle_info/2`.

The two AI assistants are wired very differently:

- The **legacy editor** uses the server-rendered `LightningWeb.AiAssistant.Component` LiveComponent. It does not display character-by-character streaming. It shows a static loading indicator and then renders the finished message once `:message_status_changed` arrives. The only PubSub event it actually consumes is `:message_status_changed`.
- The **collaborative editor** is pure React. The collab LV does not subscribe to `ai_session:*` at all. Its React side joins `LightningWeb.AiAssistantChannel`, which is where the four streaming events are consumed and rendered as the typewriter effect via a Zustand store.

Because the streaming events still flow through the shared `ai_session:#{session_id}` PubSub topic, they reach the legacy LV by accident, and that is what crashes it.

## What to fix

Minimal fix: add a catch-all clause to the case in `Edit.handle_info({:ai_assistant, ...})` so unknown actions are dropped silently. The legacy LV consumes none of the four streaming events, so dropping them is correct and does not change any user-visible behaviour. `:message_status_changed` continues to be handled and remains the relay that lets the LiveComponent finish loading.

A cleaner option, not subscribing the legacy LV to `ai_session:*` at all and routing `:message_status_changed` through a narrower channel, is more invasive than the deprecation timeline of the legacy editor warrants.

elias-ba:
## Description

This PR fixes a crash in the legacy workflow editor (`/w/new/legacy`, `/w/:id/legacy`) that fired the moment the AI assistant emitted its first streaming progress update. In production this was the "Thinking..." status message Apollo broadcasts when it starts processing. The legacy editor was listening on the same PubSub topic as the collaborative editor's AI panel but had no handler for any of the streaming events, so the first one raised a `CaseClauseError` and the LiveView reconnected. The fix drops those events in the legacy editor since it does not display any of them. The completed-message event it does need is unchanged. The collaborative editor is unaffected.

Closes #4719

## Validation steps

1. With this branch checked out, open the legacy editor for any project at `/projects/<id>/w/new/legacy?method=ai`. Ask the AI assistant any question. The editor stays connected through the "Thinking..." phase and renders the finished response. There should be no `CaseClauseError` in the server logs and no reconnect.

2. To see the bug on `main`, repeat the same flow without this branch. As soon as Apollo emits its first "Thinking..." update, the editor LiveView raises `CaseClauseError: no case clause matching: :streaming_status` and reconnects.

3. If you don't have a working Apollo connection locally, the regression test reproduces the same crash on `main` and passes on this branch:

   ```
   mix test test/lightning_web/live/workflow_live/edit_test.exs
   ```

4. Spot-check that the collaborative editor still streams normally. Open `/projects/<id>/w/new`, use the AI assistant there, and confirm the typewriter effect still renders character-by-character. The fix only touches the legacy editor's handler for the shared PubSub topic; the collaborative editor's channel is the actual consumer of those streaming events and is unchanged.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ and this project adheres to
 
 ### Fixed
 
+- Legacy workflow editor no longer crashes when the AI assistant streams
+  progress updates. The editor LiveView shares a PubSub topic with the
+  collaborative editor's React channel and was missing a clause for the four
+  streaming events, causing a `CaseClauseError` and a reconnect on the first
+  "Thinking..." update. [#4719](https://github.com/OpenFn/lightning/issues/4719)
+
 ## [2.16.3-pre3] - 2026-05-07
 
 ### Fixed
--- a/lib/lightning_web/live/workflow_live/edit.ex
+++ b/lib/lightning_web/live/workflow_live/edit.ex
@@ -2401,6 +2401,11 @@ defmodule LightningWeb.WorkflowLive.Edit do
 
       :unregister_component ->
         handle_component_unregistration(socket, payload)
+
+      # Streaming events share the ai_session:* topic but are consumed by
+      # AiAssistantChannel, not this LiveView.
+      _ ->
+        {:noreply, socket}
     end
   end
 
--- a/test/lightning_web/live/workflow_live/edit_test.exs
+++ b/test/lightning_web/live/workflow_live/edit_test.exs
@@ -622,6 +622,39 @@ defmodule LightningWeb.WorkflowLive.EditTest do
 
       assert html =~ "You are not authorized to perform this action."
     end
+
+    test "ignores AI assistant streaming events on the shared ai_session topic",
+         %{conn: conn, project: project} do
+      {:ok, view, _html} =
+        live(conn, ~p"/projects/#{project.id}/w/new/legacy", on_error: :raise)
+
+      session_id = Ecto.UUID.generate()
+
+      send(
+        view.pid,
+        {:ai_assistant, :register_component,
+         %{component_id: "new-workflow-panel-assistant", session_id: session_id}}
+      )
+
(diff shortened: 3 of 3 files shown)


brandonjackson:
Closing - we are deprecating this

PR #4722, opened 2026-05-08
Lightning to use sync v2 formats

### LINKED ISSUE #4718: Standardize project and workflow export to match portability spec
In order to line up the portability spec across our tools, we need to ensure:

1. Exporting a project should download a CLI compatible stateless project file (as described in https://github.com/OpenFn/kit/issues/1398)
2. Exporting (and importing) a workflow should use the CLI's workflow format  (as described in https://github.com/OpenFn/kit/issues/1117)

If we do these things, then file formats across the CLI and lightning will _finally_ be fully aligned and interopable and compatible. And we can make the Portability Spec really mean something,

Additional: when exporting workflow and project files from the app, let users choose the format.

### LINKED ISSUE #4727: Prevent sandbox from using same Github repo:branch as parent
prevent people from linking a sandbox to its parent repo:branch

taylordowns2000:
## Description

Closes #4718 .
Closes #4727 .

Switches Lightning to the v2 (portability spec) YAML format for the four user-facing surfaces below. The wire shape is now fully aligned with the OpenFn portability spec: [`@openfn/kit/packages/lexicon/portability.d.ts`](https://github.com/OpenFn/kit/blob/5e4d65af25a6854886c15294ed4cf17f93ecbc19/packages/lexicon/portability.d.ts).

### 1. Canvas → "View workflow as code"

`CodeViewPanel` serializes the live workflow state to v2 YAML via `assets/js/yaml/format.ts` (delegates to `assets/js/yaml/v2.ts`).

#### 1a. Export button

Same code path — the Download button writes the v2 YAML the panel renders.

#### 1b. Publish as template button

`TemplatePublishPanel` writes new templates in v2. Existing v1 templates remain readable on the Pick Template path (see 2a).

### 2. Create new → "Import code as workflow"

`YAMLImportPanel` uses the format-aware façade: detects v1 vs v2 from pasted/uploaded YAML and dispatches to the matching parser. Empty-editor placeholder shows the v2 shape so new users land on the new format.

#### 2a. Pick template

`TemplatePanel` runs every template through the same façade. v1 templates load (lenient v1 parser), v2 templates parse strictly. A test asserts that the v1 and v2 versions of the same scenario produce structurally equivalent workflow state.

### 3. Settings → "Export project" button

`Projects.export_project/2` calls `Lightning.Workflows.YamlFormat.serialize_project/2`, which emits v2 YAML. The old v1 emitter (`Lightning.ExportUtils`) is removed.

### 4. GitHub sync — block sandbox/ancestor branch collisions

A sandbox project can no longer link to the same `(repo, branch)` as any of its ancestors:

- `Projects.ancestor_ids/1` walks the `parent_id` chain via a recursive CTE.
- `ProjectRepoConnection.changeset/2` validates against ancestor branches and surfaces an inline error in the Save form.
- `VersionControl.create_github_connection/2` re-checks the same condition transactionally so a concurrent insert can't race past the validation.

## Spec coverage

Fully implemented from [the spec](https://github.com/OpenFn/kit/blob/5e4d65af25a6854886c15294ed4cf17f93ecbc19/packages/lexicon/portability.d.ts):

| Type | Field | Notes |
|------|-------|-------|
| `ProjectSpec` | `id` (required) | hyphenated project name |
| `ProjectSpec` | `name?`, `description?` | |
| `ProjectSpec` | `workflows: WorkflowSpec[]` | YAML sequence |
| `ProjectSpec` | `credentials?: Credential[]` | YAML sequence; entries without resolvable owner email are dropped |
| `ProjectSpec` | `collections?: string[]` | YAML sequence of names, alphabetically sorted |
| `WorkflowSpec` | `id?`, `name?` | `id` is the hyphenated name |
| `WorkflowSpec` | `steps: Array<Job \| Trigger>` | unified array |
| `Step` | `id?`, `name?`, `next?` | |
| `StepEdge` | `boolean \| string \| ConditionalStepEdge` | parser accepts all three; emitter prefers the object form |
| `ConditionalStepEdge` | `condition?` (JS body), `label?`, `disabled?` | |
| `Trigger` | `enabled?` | |
| `Credential` | `name`, `owner` (both required) | |
| `Job` | `adaptors?: string[]` | Lightning state has a single adaptor; collapsed/wrapped to a 1-element array |
| `Job` | `expression?` | |
| `Job` | `configuration?: object \| string` | string form only (Lightning credentials are `<email>\|<name>` strings) |

### Edge condition mapping

Lightning's internal `condition_type` enum ↔ spec `condition` (a JS expression body):

| Lightning enum | Wire `condition` |
|---|---|
| `:always` | *omitted* (or `next: <target>` shortcut) |
| `:on_job_success` | `'!state.errors'` |
| `:on_job_failure` | `'!!state.errors'` |
| `:js_expression` | `<user JS body>` (literal block when multi-line) |

Parser also accepts the spec's shortcuts:
- `next: { foo: true }` → `:always`
- `next: { foo: false }` → `:js_expression` body `"false"` (Lightning has no `:never`)
- `next: { foo: "<js>" }` → `:js_expression` body verbatim

### Lightning extensions (documented)

The portability spec marks `Trigger` fields explicitly TODO (`// TODO a trigger supports many more keys which the spec must support`). Until the spec author fills these in, we keep two trigger-only extensions:

- `type:` discriminator (`webhook` / `cron` / `kafka`)
- `openfn:` block carrying cron expression, cron cursor reference, webhook reply mode, and Kafka config

These are commented as Lightning extensions in both `v2.ex` and `v2.ts`.

## What does NOT work yet (out of scope for this PR)

- **Server-side YAML parsing.** This PR only emits v2 on the server. The provisioning controller still accepts the legacy provisioner-shape JSON unchanged. Accepting v2 YAML/JSON over the API (and the matching `@openfn/cli` integration) is intentionally deferred — there is no Phase 5 import bridge in this branch.
- **Spec fields not yet mapped:** `WorkflowSpec.globals`, `WorkflowSpec.start`, `WorkflowSpec.options`, `Step.previous`, `Job.configuration` as object. Lightning has no internal equivalents today.
- **`test/integration/cli_deploy_test.exs`** is broken by the export cutover (it compares against a v1 fixture). Tagged `:integration` and disabled by default; will be updated when the @openfn/cli suite is next exercised.

## Validation steps

1. Open a workflow → Code panel → confirm `id:`, `name:`, `steps:` array, no `jobs:`/`edges:`, jobs use `adaptors: [...]`, expression bodies under `expression: |`.
2. Click Download → file matches the panel content.
3. Publish as template → the saved template's code is v2.
4. Pick Template → both a v1 (legacy) and a v2 template load into a working canvas.
5. Import code → paste v1 YAML, paste v2 YAML, both import.
6. Project settings → Export → downloaded file has top-level `id:`, `workflows:` is a sequence (`- id: …`), `credentials:` is a sequence of `{name, owner}`, `collections:` is a sequence of names.
7. Edge conditions: a workflow with on-success / on-failure edges round-trips through Code panel → reload as `!state.errors` / `!!state.errors`.
8. GitHub sync: connect a parent project to `repo:main`, create a sandbox of it, attempt to connect the sandbox to `repo:main` → inline error, Save button disabled. Switch the sandbox to `repo:dev` → Save enabled.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


taylordowns2000:
@josephjclark , @midigofrank , sorry... dug deeper and this needs a few more tweaks. will be ready for review shortly.

josephjclark:
High level QA notes for this branch:

* [x] Export workflow yaml to a file and execute through CLI
* [x] Copy a workflow yaml from production (with the old format) and import as a new workflow on this branch. Should be identical
* [x] Copy a workflow yaml from this branch (with the new format) and import as a new workflow
* [x] Export a project, paste the yaml locally, run a workflow through the CLI

I've run a smoke test and it looks great - the format looks right and I can run a workflow yaml through the latest CLI version

EDIT: QA looks good! Just one issue executing from the project, see below

josephjclark:
@taylordowns2000 I notice an issue here where when viewing workflow yaml, the credential is not included. It's fine when exporting the project.

But this affects main too! So I'm going to raise it out as a bug, No action needed here.

taylordowns2000:
> @taylordowns2000 I notice an issue here where when viewing workflow yaml, the credential is not included. It's fine when exporting the project.
> 
> But this affects main too! So I'm going to raise it out as a bug, No action needed here.

This is super interesting @josephjclark - the reason is likely that credentials are not included in _templates_ and "view" is the precursor to "publish as template". **How should we handle this?**

(a) Leave it as is
(b) Show the credential in view, strip it out in publish?
(c) Somehow _leave it_ in templates, and strip it out on create?
(d) Something else?

josephjclark:
@taylordowns2000 I'll take that convo over to https://github.com/OpenFn/lightning/issues/4731

josephjclark:
@taylordowns2000 one change I'm going to recommend on both sides: let's add something like `schema_version` to the top of the project.yaml (or, if a workflow.yaml is standalone, let's add it there too).

The CLI currently writes a `cli.version: 2` flag to help differentiate state v1 and state v2 files. But after this work that's not just a CLI thing - it's global.

This helps tools like the CLI and maybe later the provisioner understand and anticipate the structure they are about to parse.

The practical upshot of this is that the CLI is erroring if you do `openfn checkout project-spec.yaml` from a downloaded spec.That's a result of a bad assumption in the CLI code - but without a version flag I'm struggling to override that assumption.

Now, we should probably align that version with the portability spec v4. And we should probably considering semvering it (at least major and minor, maybe not patch)


So I recommend:
- Top level `schema_version` key
- Lightning needs to set this on project and workflow export
- Set the version value to 4.0
- Patch CLI to synchronise to the same version number

This will also make it much easier to coordinate the next schema change we want to make.

BTW I've made a separate issue for the "should portability spec be a JSON schema" thing. Still want to defer that for that though

josephjclark (approved):

josephjclark:
@midigofrank This is ready for you :+1:

midigofrank (changes requested):
@josephjclark @taylordowns2000 I'm worried about the provisioner api. This PR replaces the yaml content being downloaded in the endpoint with v2 format, this will break older clis (not sure which versions).

I'll be happy to approve this if there is already a plan in place to handle legacy support

josephjclark:
@midigofrank great catch on that. The provisioner API should not have broken, that should all sit on the v1 stuff, invisible to users.

I've squashed a change which means:
- v1 CLI commands (including GH sync) all work great out of the box
- The UI exports use the v2 format and are compatible with `cli checkout` and `cli exectute`

So we should have working legacy automations, but a clean and tidy default v2 path going forward.

I haven't looked closely at the GH sync stuff so I think I'll do that tomorrow

josephjclark:
@midigofrank has raised a really good flag

The etophia deployment works by:
1. export the projetct yaml (v1 format)
2. paste it into GH (I think?)
3. Run the deploy script, which uses v1 deploy 

Obviously when we export the project in v2 style, this process will break.

Are we aware of anyone else who uses the export function like this?

options:
1. Add support to the v1 deploy to read a v2 spec file (not technically that hard, but I don't want to touch v1 if I can help it). Requires ethiopia to bump CLI version.
2. Add a small button to export the project in v2 format (or when you hit export, ask users what format they want). I don't love it.
3. Update ethiopia to use the v2 deploy command. Shouldn't be a big API change. But would need to coordinate the update (script and platform need updating together), and honesly v2 deploy from spec needs more testing

Option 1 feels like the most elegant here. All we've had to do in the CLI is convert formats on the fly - which we have good robust code for.

josephjclark:
Related, see this: https://github.com/OpenFn/kit/issues/1421

test/integration/cli_deploy_test.exs
```diff
@@ -296,7 +244,8 @@ defmodule Lightning.CliDeployTest do
       )
 
       # Lets use the updated spec
-      specPath = Path.expand("test/fixtures/canonical_update_project.yaml")
+      specPath =
+        Path.expand("test/fixtures/portability/v1/canonical_update_project.yaml")
```
josephjclark, line 359: TODO: 
1. update the CLI to support v2 spec in v1 sync
2. duplicate this test to use v2 spec

josephjclark:
@midigofrank CLI 1.38.2 is due to be released imminently and will allow a v2 project file to be deployed through the v1 `openfn deploy` API.

That should be enough I think to close out this issue. I'll come take a final look next week.

Note that the v2 deploy API does NOT let you deploy a spec file directly - I need to do a little something to get that working. I consider that a seperate issue but that and the delete stuff are high priority for next week.

midigofrank:
Thanks @josephjclark . Before merging we should revert back the change on the export project.

josephjclark:
> Thanks @josephjclark . Before merging we should revert back the change on the export project.

Which change is that? We want Export Project to be exporting v2 now, because it's fully compatible with CLI. But maybe Ethiopia needs a CLI update first to future proof it?

Or am I thinking of something else?

midigofrank:
@josephjclark yes you're right and we're saying the same thing. I wrote it as a precaution to verify that https://github.com/OpenFn/lightning/pull/4722/changes/f5078bd9c630f239199429394832133e16d3637d did not revert it back to v1

josephjclark:
Ah - we should be good then. The provisioner uses v1 but the export uses V2 (tested on Friday). I'll pick your brains on something at standup but we should finally be ready to merge!

josephjclark:
QA plan:
- [x] Can import a new workflow with v1 spec
- [x]  Can import a new workflow with v2 spec
- [x] View Workflow YAML shows v2 format
- [x] CLI can execute a v2 workflow yaml (ignoring collections and credentials for now - but there are systems for both)
- [x] Export Projects downloads a v2 format
- [x] ~Exported project can be executed with CLI~
- [x] ~Exported project can be checked out with VLI~
- [x] Exported project can be deployed with v1 CLI deploy

Can't test two these because we can't really checkout or run straight from a project.yaml either. V2is a bit restrictive about needing things checked out so that's not working great. But those are CLI issues, not iussues on this branch

Note that the v2 deploy CLI cannot yet deploy straight from yaml - but that's coming soon and is unrelated to this PR

test/integration/cli_deploy_test.exs
midigofrank: @josephjclark I'm guessing duplicating this test is the only remaining thing for this pr

josephjclark: Yes. I've just had a go at generating a test but it's failing. And actually I can repro locally by running the local v2 canonical project.yaml through the CLI. So something is up - I'm investigating.

Got distracted from it by a few things but I'll try and pick this up tomorrow so we can finally get this merged

josephjclark: @midigofrank I've figured this all out. It took ages but the problem is basically a broken CLI build.

So the CLI fix is here:https://github.com/OpenFn/kit/pull/1497  That'll go out tomorrow morning

And the test is here: https://github.com/OpenFn/lightning/pull/5035  Which I'll finish and fix once the CLI is updated.

josephjclark (approved):
Ok I've been through this, tested it, added the extra tests, and bumped CLI versions.

I think we're good to go.

Note that I haven't done a basic manual test since syncing with main - maybe someone has  time to do a basic check that the Workflow and Project export are still using the correct yaml?

josephjclark:
Thank you @midigofrank !

PR #4737, opened 2026-05-11
fix: better run follow up issues

### LINKED ISSUE #4736: Fix better run follow issues
**Describe the bug** 
1. Key bindings broken - Ctrl+Enter now click the secondary button "run with custom input" instead of the primary button which runs the workflow. Ensure they still match the documentation: https://docs.openfn.org/documentation/keyboard-shortcuts
2. The primary action got SWITCHED from "retry" to "create new workorder" when a run is loaded!

**With a run loaded...** 

**Before:** "Run" opens the dialog

(a) Primary is retry :white_check_mark: 
(b) Secondary is new work order :white_check_mark: 

**Now:** "Run" actually runs

(a) Primary is "new work worder" :warning: 👈 this is the bug!
(b) Secondary opens the dialog :white_check_mark: 
  (b1) Primary is retry :white_check_mark: 
  (b2) Secondary is new work order :white_check_mark:

**Version number** 
v2.16.3 (1.5.82)

- [x] Yes
- [ ] No

1. Go to a workflow diagram
2. Press Ctrl+Enter
3. The run panel shows up. (we rather expect the primary action. which is running the workflow to happen)

happen.

The primary action(running a workflow) should rather take place

doc-han:
## Description

This PR currently fixes several issues around Better Run Experience
1. Ctrl + Enter now calls the primary action "Run" instead of the secondary action "Run with cutom inplut".
2. Making the primary action when a run is loaded to be "Retry" instead of "Create new workorder".

Closes #4736 

## Validation steps

1. With no run loaded, press Cmd/Ctrl+Enter. workflow runs immediately (no custom input dialog).
2. Press Cmd/Ctrl+Shift+Enter. custom input panel opens.
3. Load a completed run (?run=<id> in URL). button label changes to "Run (Retry)". Clicking it saves and retries; a success toast appears and the URL updates to the new run ID.
4. With a retryable run loaded, press Cmd/Ctrl+Enter. retry fires instead of creating a new work order.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


taylordowns2000 (changes requested):
Hey @doc-han , this is looking good. a few things to change, and really I think it starts with a central concept: **pressing `cmd+enter` is the same as clicking the main button action; pressing `cmd+shift+enter` is the same as pressing the button's _secondary_ action**. with that in place, i hope the changes all make sense.

when a run is loaded, `cmd+shift+enter` is actually creating a new work order... it should not. it should open the custom run input. (put back the NewRunButton and make it take retry as its primary action when a run is loaded; open when it's not.)

assets/js/collaborative-editor/components/RunRetryButton.tsx
```diff
@@ -236,31 +236,24 @@ export function RunRetryButton({
...
-                ? [styles.submitting, 'cursor-not-allowed']
-                : [styles.chevronBase, styles.focus]
-            )}
-            aria-expanded={isDropdownOpen}
-            aria-haspopup="true"
-          >
-            <span className="sr-only">Open options</span>
-            <span className="hero-chevron-down w-4 h-4"></span>
-          </button>
-        </Tooltip>
```
elias-ba, line 263: Is the Tooltip wrapper around the chevron gone ? so the chevron no longer surfaces dropdownTooltip on hover, right ?

assets/js/collaborative-editor/components/NewRunButton.tsx
```diff
@@ -112,18 +112,33 @@ export function NewRunButton({
           transition data-closed:scale-95 data-closed:transform
           data-closed:opacity-0 data-enter:duration-200 data-enter:ease-out
           data-leave:duration-75 data-leave:ease-in"
+          onKeyDown={e => {
```
elias-ba, line 116: Can you add a one-line comment explaining this defeats default Enter handling so the parent's Cmd+Enter shortcut isn't swallowed. Otherwise a future dev will wonder why and may revert it.

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -236,6 +237,12 @@ export function Header({
   const [isSubmitting, setIsSubmitting] = useState(false);
   const activeRun = useActiveRun();
   const runIsProcessing = activeRun ? !isFinalState(activeRun.state) : false;
+  const followedRunId = params.run ?? null;
+  const isRetryable =
```
elias-ba, line 241: The guard here only checks that the run is in a final state and has at least one step. The version in useRunRetry is stricter: it also requires the step's input dataclip to still exist (not wiped_at). Pressing Cmd+Enter on a run whose input was wiped will hit the retry endpoint and fail server-side, where the only feedback is the generic "Retry failed" toast. I think we should replicate the wiped-dataclip check so the shortcut is silently disabled in that state, or confirm the server-side error toast is good enough and we accept it. What do you think ?

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -296,6 +303,57 @@ export function Header({
     updateSearchParams,
   ]);
 
+  const handleRetryClick = useCallback(async () => {
```
elias-ba, line 306: This function is a near-verbatim copy of `handleRetry` in `useRunRetry`. My preference would be to extract a small `retryRun({ projectId, runId, stepId })` helper alongside the dataclips API and have both `useRunRetry.handleRetry` and this function call it.

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -296,6 +303,57 @@ export function Header({
...
+    if (!followedRunId || !activeRun?.steps?.length || !projectId) return;
+
+    setIsSubmitting(true);
+    try {
+      await saveWorkflow({ silent: true });
+
+      const firstStep = activeRun.steps[0];
```
elias-ba, line 313: Picking index 0 assumes the steps array is ordered chronologically. Is that always guaranteed ?

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -326,6 +384,46 @@ export function Header({
     }
   }, [provider, projectId, workflowId, isNewWorkflow]);
 
+  useKeyboardShortcut(
```
elias-ba, line 387: The enabled flag covers permissions and panel state but not in-flight submissions. The button itself disables while `isRunning (isSubmitting || runIsProcessing)`, but the shortcut keeps firing. A user holding or quickly repeating Cmd+Enter can queue a second POST before `setIsSubmitting(true)` has propagated. Right ?

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -440,10 +538,13 @@ export function Header({
...
               {projectId && workflowId && firstTriggerId && !isNewWorkflow && (
                 <NewRunButton
-                  onClick={handleRunClick}
+                  onClick={() => {
+                    void (isRetryable ? handleRetryClick() : handleRunClick());
+                  }}
                   onRunWithCustomInputClick={handleRunWithCustomInputClick}
                   disabled={!canRun || isRunPanelOpen || isIDEOpen}
                   isRunning={isSubmitting || runIsProcessing}
+                  text={isRetryable ? 'Run (Retry)' : 'Run'}
```
elias-ba, line 549: Instead of `Run (Retry)` could we just have `Retry` ?

elias-ba (changes requested):

brandonjackson (approved):
Looks good @doc-han. 
@elias-ba's suggestion of changing all buttons to say "Retry" instead of "Run (Retry)" is good, but a nice-to-have. Do it only if it doesn't add risk to wrapping up this project this week.

assets/js/collaborative-editor/components/Header.tsx
doc-han: wouldn't want to change this now or have this blocking this PR. we have `Run (Retry)` at so many places in the app and wouldn't want to go changing everything.

assets/js/collaborative-editor/components/Header.tsx
doc-han: resolved. thanks

assets/js/collaborative-editor/components/Header.tsx
doc-han: yeah. it is

assets/js/collaborative-editor/components/Header.tsx
doc-han: I'll say this refactor is out of scope for this PR. The logic in both places have some difference and will be more costly to refactor.

assets/js/collaborative-editor/components/NewRunButton.tsx
doc-han: resolved

assets/js/collaborative-editor/components/Header.tsx
doc-han: Header.tsx doesn't have access to the actual dataclip object. to support this it'll have to pull the actual dataclip  just to check single field `wiped_at`. I think the server side flash error message is just fine.

assets/js/collaborative-editor/components/RunRetryButton.tsx
doc-han: deliberately removed.

elias-ba (approved):
Great work man 👏🏽

josephjclark (approved):

PR #4741, opened 2026-05-12
Remove real secrets from .env.example

### LINKED ISSUE #4706: security: .env.example contains real RSA private key and worker secret as example values
## Description

The .env.example file contains what appears to be a real RSA private key 
and worker secret used as example values rather than placeholder strings.

Specifically:
- Line with `WORKER_RUNS_PRIVATE_KEY` contains a base64-encoded value that 
  decodes to a valid `-----BEGIN RSA PRIVATE KEY-----` block
- Line with `WORKER_SECRET` contains a real 256-bit secret value

## Risk

Any operator who copies these values directly into their production `.env` 
without generating new keys will be running with publicly known credentials. 
An attacker with access to this repo can use these keys to authenticate as 
a worker to any Lightning instance using the example values.

## Suggested Fix

Replace the real key material with generation instructions:

```bash
# WORKER_RUNS_PRIVATE_KEY="generate with: mix lightning.gen_worker_keys"
# WORKER_SECRET="generate with: mix lightning.gen_worker_keys"
```

The DEPLOYMENT.md already documents `mix lightning.gen_worker_keys` — 
pointing to it from the `.env.example` is safer than providing real keys.

## Steps to Verify

```bash
echo "LS0tLS1CRUdJTiBSU0EgUFJJVkFURSBLRVktLS0tLQ..." | base64 -d
# Returns a valid RSA private key
```

kimjune01:
## Summary
- Replace real RSA private key and worker secret in `.env.example` with generation instructions pointing to `mix lightning.gen_worker_keys`
- Add CHANGELOG entry under Security section
- Add regression test script (`test/env_example_security_test.sh`) that fails if real secrets reappear
- Fixes #4706

## Test plan
- `bash test/env_example_security_test.sh`: PASS on fix branch, FAIL on main (confirms real secrets were present)
- Verified `.env.example` references `mix lightning.gen_worker_keys` for key generation


josephjclark:
Thank you for flagging this @kimjune01. I've gone ahead and fixed this myself on the main branch,.

We don't need the security test script or the changelog for this, so I can't merge this PR.

When raising PRs, please try and keep them as small and focused as possible so that its easy for maintainers to merge them :pray:

PR #4751, opened 2026-05-14
Epic: Single Sign On

### LINKED ISSUE #4621: Full SSO Experience
# Epic Brief: Full Single Sign-On Experience

| Tech Lead | Status | Created |
|-------|-------------------|--------|
| *TBD* | Draft | *2026-04-13* |

_An epic brief defines the problem and sets guardrails for the solution — but leaves enough room for designers and engineers to find the best approach. Writable in an afternoon, readable in five minutes._

## Key Hypothesis

If we offer Google and GitHub SSO on registration and login, **then we reduce onboarding friction and increase sign-up conversion.**

## Problem Validation

_Evidence that this problem is real and worth solving. Link to user reports, support tickets, analytics, or research. Each bullet should be a distinct signal._

- The "Roadshow" event in Ghana showed significant friction in the sign-up process for users in the room.
- Customers have requested SSO options such as Google and GitHub.
- The login screen offers an "external provider" option, but creating a profile with an external provider is not offered during initial registration. This is a critical UX bug that reduces trust in the platform from the start.

## Solution Concept

_One-paragraph summary of the chosen approach: what it is, where it lives in the product, and how a user encounters it._

**Add Google and GitHub as one-click sign-up and login options on the hosted SaaS platform**, so new users can register without filling out an email/password form and existing users can link their accounts.

---

## Prioritised Outcomes

_Owned by the PM. This section focuses effort on what's most essential. The feature must ship with solutions to the Must Solve outcomes._

### Must Solve

1. New users can register and log in via a third-party SSO provider without filling out an email/password form.
2. When an existing email/password user attempts SSO login, they are prompted to link their account (requiring password confirmation).
3. Google is supported as an SSO provider.
4. GitHub is supported as an SSO provider.

### Should Solve

1. Users can link or unlink SSO providers from their account settings page. Unlinking is disabled when it is the user's only auth method.
2. SSO-only users can set a password via the existing "Forgot password" flow.
3. SSO registration reduces time-to-complete for the onboarding flow compared to email/password.
4. The registration experience is visually comparable to competitors (e.g. Zapier's "Sign up with Google" flow).

### Could Solve

1. SSO-only users see a dedicated "Set password" form in account settings.
2. SSO support for self-hosted deployments.

### Won't Solve Now

1. Additional providers beyond Google and GitHub.
2. Organisation-enforced SSO-only authentication for their members.
3. Enterprise SAML/OIDC federation with customer identity providers.

---

## Acceptance Criteria

_Owned by the designer and tech lead. Do not complete until priorities are finalised and design process is finished._

TBD

---

## Open Questions

_Anything unresolved that could affect scope or implementation. Remove this section once all questions are answered._

- [ ] Is the existing "external provider" login flow live in production with active users? If so, what's the migration path? — Eng

doc-han:
## Description

This PR Implements the Full SSO Experience epic. 
- Github & Google sign-in/sign-up (Microsoft deferred)
- SSO sign-ups go through the same `AccountHook` as password sign-ups. ***

<table>
  <tr>
    <td><strong>Login page with SSO buttons</strong></td>
    <td><strong>Register page with SSO buttons</strong></td>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
  <tr>
    <td><strong>Signup confirmation page (SSO signup)</strong></td>
    <td><strong>Email collision flash on login</strong></td>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
  <tr>
    <td><strong>SSO-only user tries password login</strong></td>
    <td><strong>Profile page. link/unlink provider</strong></td>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

Closes #4621

### Setup
- Configure OAuth clients/app on a provider(Google or Github). 
- Set the redirect URL to `<host>/authenticate/<provider>/callback` where <provider> is `google` or `github` for now.
- Set your CLIENT_ID and CLIENT_SECRET as env variables. eg. `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET`.

### Validation
**Sign up via SSO**

| # | Step | Expected |
| --- | --- | --- |
| 1 | New User → "Sign up with GitHub" | Confirmation screen appears |
| 2 | Click **Cancel** | No user created |
| 3 | Click **Confirm** | Account created & Logged in, email has no password |
| 4 | Repeat with Google | Same result |

**Sign in via SSO**

| # | Step | Expected |
| --- | --- | --- |
| 1 | Existing SSO user → "Sign in with GitHub" | Straight in, no confirmation screen |
| 2 | MFA user → SSO sign-in | Routes to MFA validation page |

**Email collision**

| # | Step | Expected |
| --- | --- | --- |
| 1 | Existing password account at `you@domain.com` + SSO sign-up with same email | Redirected to login with "an account already exists, link from profile" message |
| 2 | Check DB (if you can) | No `user_identities` row created |

**Link from `/profile`**

| # | Step | Expected |
| --- | --- | --- |
| 1 | Click Link next to an unlinked provider → authorise | Flash confirms link |
| 2 | Log out, sign in via that provider | Succeeds |

**Unlink from `/profile`**

| # | Step | Expected |
| --- | --- | --- |
| 1 | Account with password + SSO → click Unlink | Identity removed |
| 2 | SSO-only account, unlink last identity | Refused with "set a password first" |
| 3 | Forgot-password → set password → Unlink | Succeeds |

**Forgot password (SSO-only)**

| # | Step | Expected |
| --- | --- | --- |
| 1 | SSO-only account tries password login | Flash "This account uses single sign-on" |
| 2 | Forgot password → set password | Both password and SSO sign-in work |

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


midigofrank (changes requested):
## Review — Epic: Single Sign On

Really nice work on this. A few change requests below before merge:

### 1. GitHub `/user` endpoint often returns no email — many GitHub logins will fail (high)

`GithubHandler` points `userinfo_endpoint` at `https://api.github.com/user`. That endpoint's `email` field is the user's **public** profile email, which is `null` for the majority of GitHub users (those who haven't set a public email) — *even though* the `user:email` scope is granted. `extract_email/1` then returns `nil` → `{:error, :no_email}` → "Could not retrieve your email."

To reliably obtain a (verified) email, GitHub requires a second call to `https://api.github.com/user/emails` and selecting the `primary && verified` entry. As written, GitHub SSO will silently fail for a large fraction of real users. (Confirmed reproducible against a test account with no public email set.)

### 2. Deployment documentation & env-var disambiguation (high)

`DEPLOYMENT.md` has no instructions for the new SSO env vars, and the variable names collide/overlap with existing ones — this will confuse operators:

- **`GITHUB_CLIENT_ID` / `GITHUB_CLIENT_SECRET` (new, SSO login)** vs **`GITHUB_APP_ID` / `GITHUB_APP_CLIENT_ID` / `GITHUB_APP_CLIENT_SECRET` (existing, DEPLOYMENT.md L97-101)**. The latter is the **GitHub App** used for repo/version-control sync — a completely different feature with confusingly similar names. The docs must clearly separate "GitHub App (project version control)" from "GitHub SSO (sign-in)".
- **`GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`** are *already documented* (DEPLOYMENT.md L363-377, "Google Oauth2") for the existing DB-backed auth-provider path, whose callback is `/authenticate/callback`. The new `GoogleHandler` reuses the **same env var names** but with a different callback (`/authenticate/google/callback`). Please clarify whether these are intentionally shared or a collision, and document the redirect URI difference.

Please add a dedicated "Single Sign-On (SSO)" section to `DEPLOYMENT.md` covering: which provider vars enable SSO, the exact redirect/callback URLs (`/authenticate/<provider>/callback`), and an explicit note distinguishing these from the GitHub App and the older Google OAuth2 settings. The `.env.example` comments are a good start but operators rely on `DEPLOYMENT.md`.

### 3. Migration is irreversible — `mix ecto.rollback` will raise (medium)

`20260515133933_allow_null_hashed_password.exs`:

```elixir
def change do
  alter table(:users) do
    modify :hashed_password, :string, null: true
  end
end
```

`modify/3` inside `change/0` is only reversible when the `:from` option is given. Without it, a rollback raises `Ecto.MigrationError`. Use `modify :hashed_password, :string, null: true, from: {:string, null: false}` (or split into `up`/`down`).

### 4. Provider emails are trusted without a verification check (medium)

New-account provisioning trusts `userinfo["email"]` as a proven-owned address. Google's `userinfo` includes an `email_verified` claim; GitHub's `/user/emails` includes `verified`. Neither is checked. An attacker controlling a provider account with an *unverified* email could provision a Lightning account under an address they don't own. (The collision path protects *existing* accounts, so this is limited to new-account creation, but it's a cheap, standard hardening step.)

### 5. SSO button click target is too small — only the inner link is clickable (medium)

On both the login and register pages, the SSO button is a `<.button>` wrapping an inner `<a href={provider.url}>`. Only the anchor (which wraps the `inline-flex` icon+text) is clickable — clicking the button's padding does nothing, so the user has to aim at the link text in the middle. The full button should be the click target. Make the link itself the styled element (e.g. `<.button_link href={provider.url}>` / `<.link>` styled as a button) rather than nesting an `<a>` inside `<.button>`. Note `IdentitiesComponent` already uses `<.button_link>` for its "Link" action — the same pattern would fix this.

### 6. Email-existence oracle on the login form (low / product decision)

The `{:error, :sso_account}` flash ("This account uses single sign-on...") and the collision message ("An account already exists for {email}...") both confirm to an unauthenticated visitor whether a given email is registered, and how. This is a deliberate UX tradeoff and common, but flagging it as a behavior change from the previous generic "Invalid email or password".

### 7. `on_conflict: :nothing` masks cross-account identity collisions (low)

`Accounts.link_user_identity/3` and the duplicate `AccountHook.link_identity/3` insert with `on_conflict: :nothing, conflict_target: [:provider, :uid]`. If `(provider, uid)` already belongs to a *different* user, the insert silently no-ops and returns `{:ok, identity}` with a `nil` id, so the caller believes the link succeeded. The controller's pre-checks make this a narrow race, but `:nothing` is hiding a real conflict — consider distinguishing "already linked to me" from "claimed by someone else" at the data layer.

### 8. Minor

- `defp display_name/1` is defined identically in `OidcController` and `IdentitiesComponent`; `link_identity` in `AccountHook` duplicates `Accounts.link_user_identity/3`. Consider consolidating.
- `CacheWarmer.execute/1`'s `try/rescue _ -> []` is very broad and will swallow genuine config errors silently.
- `GithubHandler` has no test file (`GoogleHandler` does) — given #1, a GitHub-specific userinfo test would be valuable.
- Changelog not updated (PR checklist acknowledges this).

doc-han:
> ## Review — Epic: Single Sign On
> Really nice work on this. A few change requests below before merge:
> 
> ### 1. GitHub `/user` endpoint often returns no email — many GitHub logins will fail (high)
> `GithubHandler` points `userinfo_endpoint` at `https://api.github.com/user`. That endpoint's `email` field is the user's **public** profile email, which is `null` for the majority of GitHub users (those who haven't set a public email) — _even though_ the `user:email` scope is granted. `extract_email/1` then returns `nil` → `{:error, :no_email}` → "Could not retrieve your email."
> 
> To reliably obtain a (verified) email, GitHub requires a second call to `https://api.github.com/user/emails` and selecting the `primary && verified` entry. As written, GitHub SSO will silently fail for a large fraction of real users. (Confirmed reproducible against a test account with no public email set.)
> 
> ### 2. Deployment documentation & env-var disambiguation (high)
> `DEPLOYMENT.md` has no instructions for the new SSO env vars, and the variable names collide/overlap with existing ones — this will confuse operators:
> 
> * **`GITHUB_CLIENT_ID` / `GITHUB_CLIENT_SECRET` (new, SSO login)** vs **`GITHUB_APP_ID` / `GITHUB_APP_CLIENT_ID` / `GITHUB_APP_CLIENT_SECRET` (existing, DEPLOYMENT.md L97-101)**. The latter is the **GitHub App** used for repo/version-control sync — a completely different feature with confusingly similar names. The docs must clearly separate "GitHub App (project version control)" from "GitHub SSO (sign-in)".
> * **`GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`** are _already documented_ (DEPLOYMENT.md L363-377, "Google Oauth2") for the existing DB-backed auth-provider path, whose callback is `/authenticate/callback`. The new `GoogleHandler` reuses the **same env var names** but with a different callback (`/authenticate/google/callback`). Please clarify whether these are intentionally shared or a collision, and document the redirect URI difference.
> 
> Please add a dedicated "Single Sign-On (SSO)" section to `DEPLOYMENT.md` covering: which provider vars enable SSO, the exact redirect/callback URLs (`/authenticate/<provider>/callback`), and an explicit note distinguishing these from the GitHub App and the older Google OAuth2 settings. The `.env.example` comments are a good start but operators rely on `DEPLOYMENT.md`.
> 
> ### 3. Migration is irreversible — `mix ecto.rollback` will raise (medium)
> `20260515133933_allow_null_hashed_password.exs`:
> 
> ```elixir
> def change do
>   alter table(:users) do
>     modify :hashed_password, :string, null: true
>   end
> end
> ```
> 
> `modify/3` inside `change/0` is only reversible when the `:from` option is given. Without it, a rollback raises `Ecto.MigrationError`. Use `modify :hashed_password, :string, null: true, from: {:string, null: false}` (or split into `up`/`down`).
> 
> ### 4. Provider emails are trusted without a verification check (medium)
> New-account provisioning trusts `userinfo["email"]` as a proven-owned address. Google's `userinfo` includes an `email_verified` claim; GitHub's `/user/emails` includes `verified`. Neither is checked. An attacker controlling a provider account with an _unverified_ email could provision a Lightning account under an address they don't own. (The collision path protects _existing_ accounts, so this is limited to new-account creation, but it's a cheap, standard hardening step.)
> 
> ### 5. SSO button click target is too small — only the inner link is clickable (medium)
> On both the login and register pages, the SSO button is a `<.button>` wrapping an inner `<a href={provider.url}>`. Only the anchor (which wraps the `inline-flex` icon+text) is clickable — clicking the button's padding does nothing, so the user has to aim at the link text in the middle. The full button should be the click target. Make the link itself the styled element (e.g. `<.button_link href={provider.url}>` / `<.link>` styled as a button) rather than nesting an `<a>` inside `<.button>`. Note `IdentitiesComponent` already uses `<.button_link>` for its "Link" action — the same pattern would fix this.
> 
> ### 6. Email-existence oracle on the login form (low / product decision)
> The `{:error, :sso_account}` flash ("This account uses single sign-on...") and the collision message ("An account already exists for {email}...") both confirm to an unauthenticated visitor whether a given email is registered, and how. This is a deliberate UX tradeoff and common, but flagging it as a behavior change from the previous generic "Invalid email or password".
> 
> ### 7. `on_conflict: :nothing` masks cross-account identity collisions (low)
> `Accounts.link_user_identity/3` and the duplicate `AccountHook.link_identity/3` insert with `on_conflict: :nothing, conflict_target: [:provider, :uid]`. If `(provider, uid)` already belongs to a _different_ user, the insert silently no-ops and returns `{:ok, identity}` with a `nil` id, so the caller believes the link succeeded. The controller's pre-checks make this a narrow race, but `:nothing` is hiding a real conflict — consider distinguishing "already linked to me" from "claimed by someone else" at the data layer.
> 
> ### 8. Minor
> * `defp display_name/1` is defined identically in `OidcController` and `IdentitiesComponent`; `link_identity` in `AccountHook` duplicates `Accounts.link_user_identity/3`. Consider consolidating.
> * `CacheWarmer.execute/1`'s `try/rescue _ -> []` is very broad and will swallow genuine config errors silently.
> * `GithubHandler` has no test file (`GoogleHandler` does) — given #1, a GitHub-specific userinfo test would be valuable.
> * Changelog not updated (PR checklist acknowledges this).

Resolved. Thanks. this was a very useful review

midigofrank (changes requested):
@doc-han now this is ready for merge. Just a few things:

1. Can we prefix the SSO to these envs to avoid confusion? `SSO_GITHUB_CLIENT_ID` looks more explicit
2. Why can't we reuse `/oauth/github/callback`? 
3. I don't see the need for listing all the envs for the version control here. I found the callback url in there quite useful though 👏 
4. Your changelog entry looks incomplete
5. Please use `button_link` component in the signup/signin page.  It already has all the styling you need
6. You haven't addressed the email-existence oracle

doc-han:
> @doc-han now this is ready for merge. Just a few things:
> 
> 1. Can we prefix the SSO to these envs to avoid confusion? `SSO_GITHUB_CLIENT_ID` looks more explicit
> 2. Why can't we reuse `/oauth/github/callback`?
> 3. I don't see the need for listing all the envs for the version control here. I found the callback url in there quite useful though 👏
> 4. Your changelog entry looks incomplete
> 5. Please use `button_link` component in the signup/signin page.  It already has all the styling you need
> 6. You haven't addressed the email-existence oracle

1. I don't there's a collision and also didn't want to lengthen the env name. 
2. The existing route for connecting a project to a github repo is a bit different from this and kept different. 
3. Record keeping? you mean we don't have to list all the supported env vars?

stuartc (changes requested):
Great work on this. The overall shape is solid: routing SSO sign-ups through the
same `AccountHook` as password sign-ups is the right call, the hot paths that now
have to cope with a null password are handled correctly, sessions are rotated on
every login so there's no fixation, there's no open redirect through `return_to`,
and provider tokens and secrets are never persisted or logged.

The gaps I found sit at the security boundary and a few failure edges, not in the
core flow.

I've grouped the feedback into **blocking**, **concerns worth addressing**, and
**notes and nits**. Two items are blocking and I'd want them resolved before merge.

## Blocking

**1. The OAuth callbacks have no `state` parameter.**

`Handler.authorize_url/1` (`handler.ex:89-92`) passes only `scope:`, and the callback
`OidcController.new/2` (`oidc_controller.ex:54-81`) never sends or verifies a `state`
value. Nothing ties a callback back to the browser that started the flow.

Because the callback is a GET (`router.ex:72`), an attacker can run their own provider
login, capture the resulting `code`, and force a victim's browser to hit
`/authenticate/:provider/callback?code=…`. The victim ends up logged into the
attacker's account. On the link flow (`oidc_controller.ex:38-48`) it's worse: a forged
callback grafts an attacker-controlled identity onto the victim's authenticated
session, which is a persistent backdoor.

This is a regression. The deleted `Common.authorize_url/4` used to take and pass
state, and the OAuth2 library generates none of its own, so right now it's
unmitigated. Note that the framework CSRF protection on the POST signup confirmation
form is unaffected; this is specifically the OAuth authorize request and the GET
callback.

Fix: we already have most of the plumbing. The credential OAuth flow passes `state`
around as an encrypted, self-contained string via `Lightning.SafetyString`
(`OauthCredentialHelper`), so nothing needs persisting. The SSO flow wants the same
treatment plus a nonce: mint one when redirecting, carry it in `state`, and verify it
on callback (bound to the current user on the link flow), rejecting anything missing
or mismatched.

**2. Nothing enforces one identity per provider, but the code assumes it.**

The migration adds a unique index on `(provider, uid)` and a plain index on
`user_id`, but there's no constraint stopping a user having two identities for the
same provider - even though the moduledoc asserts there's only one
(`user_identity.ex:5`). `link_user_identity` (`accounts.ex:211`) only checks whether
*this* uid is claimed elsewhere; it never checks whether the current user already has
a different uid for the same provider.

So if someone links github/uid-A and then authorises a second GitHub account (uid-B),
they get a second row. `get_user_identity/2` (`accounts.ex:248`, a `Repo.get_by`) then
raises `MultipleResultsError`, which 500s both `/profile` rendering and unlink
(`identities_component.ex:60-61`). It's directly reachable, because the link route
`GET /authenticate/:provider/link` can be hit directly.

There's a security angle too. `can_remove_identity?` / `unlink_user_identity` is the
lockout protection, and once there are two identities for a provider the lookup
crashes instead of evaluating the "would lock out" guard - so that guard is only as
sound as this missing constraint.

We've decided on one identity per provider per user, rejecting the second link. Fix:
add a `(user_id, provider)` unique index and `unique_constraint`, and/or an explicit
guard in `link_user_identity` that rejects the second link. As a belt-and-braces
measure, key the profile component's unlink by identity id rather than by provider.

## Concerns worth addressing

**3. Two concrete crashes, for SSO-only users and for duplicate providers.**

- An SSO-only user (no password) crashes the change-email form. It routes through the
  unsafe `Accounts.validate_current_password/2` (`form_component.ex:28,113`), which
  runs Bcrypt over a nil hash and raises `ArgumentError`. The password-change form
  handles nil safely; only change-email is exposed.
- The duplicate-identity case from item 2 also 500s the profile component on unlink
  (`identities_component.ex:59-61`). The duplicates collapse to the last one written
  in the UI, and the unlink button carries only `phx-value-provider`, so the lookup
  hits `MultipleResultsError`. Fixing item 2 covers this; keying the UI by identity id
  is the belt-and-braces part.

**4. A GitHub public-profile email is trusted as verified without a re-check.**

`handler.ex:125-126` short-circuits on any binary `userinfo["email"]` and stamps
`email_verified=true`, never reaching the careful selection of a primary, verified
email in `select_primary_verified_email`. For GitHub that's the `/user` public email
your own test flags as unreliable (`github_handler_test.exs:33-34`).

Since `email_verified` gates the collision and auto-link safety property
(`oidc_controller.ex:297-300`), an unverified email can be treated as verified. It's
not blocking - the collision path doesn't link automatically, it shows a "sign in and
link" flash - but the fast path is worth tightening so it consults
`select_primary_verified_email` rather than trusting the raw field.

**5. Disabling a provider locks out users who only have SSO.**

Setting and unsetting the provider env vars is already the opt-in/opt-out, so there's
no need for a separate instance-level gate. But a user who only ever signed in through
that provider, and never set a password, is locked out if an operator disables it.
Worth a deliberate decision, and probably a note in `DEPLOYMENT.md`.

**6. The email-collision check reads, then acts (a TOCTOU race).**

The collision handling reads, then acts, without a `unique_constraint(:email)` on the
SSO registration changeset, so a race surfaces as a raw 500 rather than a clean
changeset error. The `users.email` citext unique index does exist, so this is about
the shape of the error rather than data integrity. Adding `unique_constraint(:email)`
to `sso_registration_changeset` turns the 500 into a proper changeset error.

**7. Authorization for link/unlink/callback is inline only.**

These rely on inline `current_user` / `socket.assigns.user` checks, and the OIDC
routes sit in the bare `:browser` pipeline with no formal policy. The inline checks do
work, so this isn't an open hole, but we'd like it moved into the Bodyguard policy
pattern for consistency with the rest of the app. (The real defect in the link flow is
the missing state binding from item 1, not this.)

**8. Other confirmed concerns,** lower priority, grouped for brevity:

- Handler discovery depends entirely on the cache warmer, so a cold start - or the
  warmer not having run - is a single point of failure.
- OAuth error and deny callbacks, and a raising `get_userinfo`, currently 500 rather
  than redirecting with a friendly message.
- The `can_remove_identity?` lockout check counts identities without a lock, so two
  concurrent unlinks of the last two identities can both pass. It's unlikely (it's
  self-inflicted and needs two sessions) and there's no concurrency test, but a row
  lock, or running the guard inside a transaction, would close it.
- The GitHub-specific email resolution lives in the generic `Handler`
  (`handler.ex:119-164`). Fine for two providers; worth factoring into a provider
  strategy before a third lands.

## One product decision flagged (not a request to change code)

SSO-only login returns a distinct "this account uses single sign-on" message
(`user_session_controller.ex:50-58`), which lets someone enumerate which emails are
registered and SSO-only. This is intended UX per the forgot-password validation table,
so it's a conscious tradeoff rather than a bug. We're accepting it, and it shouldn't
block the PR.

I'm only flagging it because Lightning has no login rate-limiting, which is the usual
safety net here. We'll track a possible identifier-first login flow as separate work.
(Google, Microsoft, Auth0 and Okta all leak SSO routing by design, so we're in good
company, and OWASP has no enforceable control mandating generic responses.)

## Notes and nits

- `complete_signup` clears the pending blob on error, forcing a full SSO restart
  instead of a retry (`oidc_controller.ex`).
- `display_name` uses `String.capitalize`, which mangles cased provider names
  ("GitHub" becomes "Github").
- `UserIdentity.provider` is free-form, with no inclusion validation or uid
  normalisation (`user_identity.ex:19`).
- The `show/2` action looks effectively dead and isn't hardened (`oidc_controller.ex`).
- Partial provider config is silently disabled with no boot warning, which is easy to
  misconfigure quietly.
- A few test-coverage gaps worth filling: no SSO-only (nil-password) user factory; no
  test for the link callback when no link intent is set; no coverage for the SSO
  bootstrap config blocks or `sso_redirect_uri`; no `would_lock_out` concurrency test.
  (A test for a callback without state will fall out of fixing item 1.)

lib/lightning_web/controllers/user_registration_html/new.html.heex
```diff
@@ -89,6 +89,26 @@
               <.button type="submit" theme="primary">
                 Register
               </.button>
+              <%= if @auth_providers != [] do %>
+                <div class="text-xs text-secondary-700 border-b text-center leading-[0]">
+                  <span class="px-2 bg-white">or</span>
+                </div>
+                <%= for provider <- @auth_providers do %>
+                  <.button_link theme="secondary" href={provider.url}>
```
lmac-1, line 97: @doc-han I tweaked the code here so that it was horizontally aligned (previous code had a bit too much space underneath)

lib/lightning_web/controllers/user_session_html/new.html.heex
```diff
@@ -53,23 +53,25 @@
...
                 <div class="text-xs text-secondary-700 border-b text-center leading-[0]">
                   <span class="px-2 bg-white">or</span>
                 </div>
-                <.button theme="secondary">
-                  <a href={@auth_handler_url} class="w-full">
-                    <div class="-ml-1 inline-flex items-center">
-                      <.icon
-                        name="hero-identification"
-                        class="h-4 w-4 inline-block"
+                <%= for provider <- @auth_providers do %>
```
lmac-1: @doc-han I tweaked the code here so that it was horizontally aligned (previous code had a bit too much space underneath)

lib/lightning_web/controllers/user_session_html/new.html.heex
```diff
@@ -53,23 +53,25 @@
...
+                  <.button_link theme="secondary" href={provider.url}>
+                    <div class="flex items-center justify-center gap-1.5">
+                      <LightningWeb.Components.SsoIcons.provider_icon
+                        name={provider.name}
+                        class="h-4 w-4"
                       />
-                      <span class="ml-1 inline-block align-middle">
-                        via external provider
+                      <span>
+                        Sign in with {Lightning.AuthProviders.display_name(
```
lmac-1, line 68: @doc-han I updated this to use `display_name`, we don't want to use `capitalize` as it would mean it would show as "Github" rather than "GitHub"

lib/lightning_web/controllers/user_registration_html/new.html.heex
```diff
@@ -89,6 +89,26 @@
...
+                    <div class="flex items-center justify-center gap-1.5">
+                      <LightningWeb.Components.SsoIcons.provider_icon
+                        name={provider.name}
+                        class="h-4 w-4"
+                      />
+                      <span>
+                        Sign up with {Lightning.AuthProviders.display_name(
```
lmac-1, line 104: @doc-han I updated this to use display_name, we don't want to use capitalize as it would mean it would show as "Github" rather than "GitHub"

lib/lightning/auth_providers.ex
```diff
@@ -10,6 +10,15 @@ defmodule Lightning.AuthProviders do
   alias Lightning.AuthProviders.WellKnown
   alias Lightning.Repo
 
+  @doc """
+  Returns a human-friendly name for a provider, e.g. `"github"` -> `"GitHub"`.
+  """
+  @spec display_name(provider :: String.t()) :: String.t()
+  # Explicit clauses needed for providers whose correct name has mid-word
+  # capitalisation that String.capitalize/1 can't produce.
+  def display_name("github"), do: "GitHub"
```
lmac-1, line 19: Added this comment and defined "GitHub" explicitly

lib/lightning_web/controllers/oidc_html/confirm_signup.html.heex
```diff
@@ -0,0 +1,68 @@
...
+                {error}
+              </div>
+            <% end %>
+
+            <p class="text-sm text-gray-700 mb-4">
+              No account exists for <strong>{@pending["email"]}</strong>
+              yet. We'll create a new account using your
+              <strong>
+                {Lightning.AuthProviders.display_name(@pending["provider"])}
+              </strong>
```
lmac-1, line 29: Does not use `capitalize` as we handle this in `display_name` (so "GitHub" can work too)

lib/lightning_web/controllers/oidc_html/confirm_signup.html.heex
```diff
@@ -0,0 +1,68 @@
...
+                <div class="flex">
+                  <dt class="w-28 text-gray-500">Name</dt>
+                  <dd>{@pending["first_name"]} {@pending["last_name"]}</dd>
+                </div>
+              <% end %>
+              <div class="flex">
+                <dt class="w-28 text-gray-500">Provider</dt>
+                <dd>
+                  {Lightning.AuthProviders.display_name(@pending["provider"])}
+                </dd>
```
lmac-1, line 48: @doc-han updated this to use `display_name` rather than `capitalize`

lmac-1 (commented):
Finished my UX review, looking good! I directly made a couple of changes to improve the following: 
1. Improved horizontal alignment of the "Sign in with X" buttons
2. Handle capitalisation logic directly in `display_name` function in `lib/lightning/auth_providers`. `capitalize` is used there and we explicitly define "GitHub" so that it isn't displayed as "Github" 

------

A slight nitpick with the following UX, but shouldn't block the PR as it's an existing issue: 
1. User has already logged in with SSO with GitHub or Google (or email..)
2. User tries to register with the same email
3. They see an error that says "This value should be unique" 
It would be better to have a more informational message here, such as "An account already exists with this email address". But this is a product decision and should be picked up in a separate issue.

elias-ba (changes requested):
Hey @doc-han, morning! Went through this one properly and honestly it's really solid work, the SSO flow is well thought out. The way you handle the login state is genuinely better than what I usually see, and I like that you check the email is verified before creating any account, and that a matching email doesn't get silently linked. Nice.

I spent a couple of focused hours on it, mostly checking whether any of this touches our OAuth credentials, the Google Sheets and Salesforce connections that jobs use, since that was my main worry going in. Good news, it doesn't. Those connections use the OAuth clients we set up in the UI and a different callback URL, and none of that is touched here. I also checked the modules you removed and confirmed they were already dead code with no live references on main, so that's a clean cleanup.

Couple of things I'd like us to sort before this goes live though, mostly so we start on the strongest footing:

1. The Google section in `DEPLOYMENT.md` is a bit misleading. It says `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are also used by the older Google setup and tells people to register both callback URLs on the same client. But I couldn't find anywhere in the code on main that actually reads those two env vars, so that section looks like it's documenting something that isn't wired up anymore. On top of that the doc never really separates the credential connections from SSO, so someone who has a Google app for Sheets might read this and think they need to go change it, when they really don't. Could you either fix or drop that stale section, and make it clear that the credential connections are set up in the UI and aren't affected, and that the only new callback to add is the SSO one, and only if you turn SSO on?

2. There's a behavior change for anyone already using SSO that needs a note in the changelog. Login used to match people by their email, now it needs a linked identity record, and there's no migration to backfill those. So anyone who used to sign in through their provider won't get logged in straight through anymore, they'll have to re-link from their profile once. If that's intended, and I think it's the right call for security, let's just call it out so self hosted folks aren't caught off guard.

Two smaller ones, not blockers:

3. When you unlink a provider, the check for "do they still have another way to log in" and the actual delete aren't done as one operation. In theory a password-less user clicking unlink twice at once could remove both and lock themselves out. Probably worth tightening that up.

4. In the email resolution code, when GitHub gives you back an email you mark it as verified without checking. It's fine because GitHub only exposes verified emails publicly, but a one line comment saying that would save the next person a double take.

Happy to jump on a call if any of it's unclear.

doc-han:
> Hey @doc-han, morning! Went through this one properly and honestly it's really solid work, the SSO flow is well thought out. The way you handle the login state is genuinely better than what I usually see, and I like that you check the email is verified before creating any account, and that a matching email doesn't get silently linked. Nice.
> 
> I spent a couple of focused hours on it, mostly checking whether any of this touches our OAuth credentials, the Google Sheets and Salesforce connections that jobs use, since that was my main worry going in. Good news, it doesn't. Those connections use the OAuth clients we set up in the UI and a different callback URL, and none of that is touched here. I also checked the modules you removed and confirmed they were already dead code with no live references on main, so that's a clean cleanup.
> 
> Couple of things I'd like us to sort before this goes live though, mostly so we start on the strongest footing:
> 
> 1. The Google section in `DEPLOYMENT.md` is a bit misleading. It says `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are also used by the older Google setup and tells people to register both callback URLs on the same client. But I couldn't find anywhere in the code on main that actually reads those two env vars, so that section looks like it's documenting something that isn't wired up anymore. On top of that the doc never really separates the credential connections from SSO, so someone who has a Google app for Sheets might read this and think they need to go change it, when they really don't. Could you either fix or drop that stale section, and make it clear that the credential connections are set up in the UI and aren't affected, and that the only new callback to add is the SSO one, and only if you turn SSO on?
> 2. There's a behavior change for anyone already using SSO that needs a note in the changelog. Login used to match people by their email, now it needs a linked identity record, and there's no migration to backfill those. So anyone who used to sign in through their provider won't get logged in straight through anymore, they'll have to re-link from their profile once. If that's intended, and I think it's the right call for security, let's just call it out so self hosted folks aren't caught off guard.
> 
> Two smaller ones, not blockers:
> 
> 3. When you unlink a provider, the check for "do they still have another way to log in" and the actual delete aren't done as one operation. In theory a password-less user clicking unlink twice at once could remove both and lock themselves out. Probably worth tightening that up.
> 4. In the email resolution code, when GitHub gives you back an email you mark it as verified without checking. It's fine because GitHub only exposes verified emails publicly, but a one line comment saying that would save the next person a double take.
> 
> Happy to jump on a call if any of it's unclear.

Thanks @elias-ba,

1. the SSO vars are now SSO_ prefixed (e.g. SSO_GOOGLE_CLIENT_ID) so there's no collision.
2. I don't think there's any backfilling needed at this point. The old SSO was configured from inside the app without env vars. and there was no table being kept for their identities because we were doing a simple email match. if a user(which we approx. have none using this method of sign-in) is still using the old method, their simple email match should still work. our identity stuff only takes place when SSO envs are set.
3. I don't know how practical this is but I've added a db lock so that concurrent unlinks wouldn't lock a user out.
4. GitHub email. I actually do not just take the email that will be attached to the user profile, we actually make a request to github to the verified email attached to the profile (this was already flagged by @midigofrank and already taken care of)

doc-han:
> Hey @doc-han, morning! Went through this one properly and honestly it's really solid work, the SSO flow is well thought out. The way you handle the login state is genuinely better than what I usually see, and I like that you check the email is verified before creating any account, and that a matching email doesn't get silently linked. Nice.
> 
> I spent a couple of focused hours on it, mostly checking whether any of this touches our OAuth credentials, the Google Sheets and Salesforce connections that jobs use, since that was my main worry going in. Good news, it doesn't. Those connections use the OAuth clients we set up in the UI and a different callback URL, and none of that is touched here. I also checked the modules you removed and confirmed they were already dead code with no live references on main, so that's a clean cleanup.
> 
> Couple of things I'd like us to sort before this goes live though, mostly so we start on the strongest footing:
> 
> 1. The Google section in `DEPLOYMENT.md` is a bit misleading. It says `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are also used by the older Google setup and tells people to register both callback URLs on the same client. But I couldn't find anywhere in the code on main that actually reads those two env vars, so that section looks like it's documenting something that isn't wired up anymore. On top of that the doc never really separates the credential connections from SSO, so someone who has a Google app for Sheets might read this and think they need to go change it, when they really don't. Could you either fix or drop that stale section, and make it clear that the credential connections are set up in the UI and aren't affected, and that the only new callback to add is the SSO one, and only if you turn SSO on?
> 2. There's a behavior change for anyone already using SSO that needs a note in the changelog. Login used to match people by their email, now it needs a linked identity record, and there's no migration to backfill those. So anyone who used to sign in through their provider won't get logged in straight through anymore, they'll have to re-link from their profile once. If that's intended, and I think it's the right call for security, let's just call it out so self hosted folks aren't caught off guard.
> 
> Two smaller ones, not blockers:
> 
> 3. When you unlink a provider, the check for "do they still have another way to log in" and the actual delete aren't done as one operation. In theory a password-less user clicking unlink twice at once could remove both and lock themselves out. Probably worth tightening that up.
> 4. In the email resolution code, when GitHub gives you back an email you mark it as verified without checking. It's fine because GitHub only exposes verified emails publicly, but a one line comment saying that would save the next person a double take.
> 
> Happy to jump on a call if any of it's unclear.

Thanks @elias-ba,

1. the SSO vars are now SSO_ prefixed (e.g. SSO_GOOGLE_CLIENT_ID) so there's no collision.
2. I don't think there's any backfilling needed at this point. The old SSO was configured from inside the app without env vars. and there was no table being kept for their identities because we were doing a simple email match. if a user(which we approx. have none using this method of sign-in) is still using the old method, their simple email match should still work. our identity stuff only takes place when SSO envs are set.
3. I don't know how practical this is but I've added a db lock so that concurrent unlinks wouldn't lock a user out.
4. GitHub email. I actually do not just take the email that will be attached to the user profile, we actually make a request to github to the verified email attached to the profile (this was already flagged by @midigofrank and already taken care of)

doc-han:
This PR adds Single Sign-On sign-in to Lightning, letting users authenticate with GitHub or Google instead of (or alongside) an email and password. It also lets users link and unlink providers from their profile. SSO sign-ups flow through the same account-creation path as password sign-ups, so provisioning stays consistent across both methods.

### User-facing features
1. Sign in with GitHub / Google options on both the login and register pages.
2. Link/unlink - providers a new identities panel in profile settings where a user can attach or remove providers on their account.
3. Guided account handling - if someone signs in with a provider whose email already belongs to an existing account, they're told to sign in with their existing method and link the provider, rather than being silently merged.
4. Password-less accounts - users created through SSO can exist without a password.

### Design
- We now have one generic handler plus some small Github and Google config builders. 
- Providers are enabled by setting SSO_GITHUB_CLIENT_ID / SSO_GITHUB_CLIENT_SECRET and SSO_GOOGLE_CLIENT_ID / SSO_GOOGLE_CLIENT_SECRET and redirect/callback URLs are derived automatically from the host configuration.
- Github email resolution. because GitHub's user endpoint frequently returns no public email, the primary verified address is fetched from the dedicated emails endpoint.

### DB Changes
- A new user_identities table (provider, uid, user) with a unique index on provider + uid.
- The user password column made nullable, to support SSO-only users. (**important**)
- A unique constraint enforcing at most one identity per provider per user.

### Security
- CSRF protection on the OAuth flow using a single-use nonce.
- Ensuring verified emails. account creation only proceeds with a verified email from the provider
- No automatic account linking. no more email matching accounting linking
- Lockout protection. a password-less user cannot remove their last remaining identity


cc: @stuartc

elias-ba (approved):
Hey @doc-han really huge work here. Thanks for your responses on my last review items. I have tested this and it works well for me. Approving now and really amazing work here

midigofrank:
Thanks @doc-han . I have tested the legacy provider sign in using Keycloak and it works!! Great job man

midigofrank (changes requested):
Image

@doc-han I was doing my final checks when I noticed I could link Keycloak (the legacy provider) in my profile. This shouldn't be possible, right?

midigofrank (approved):
Nicely done @doc-han . I have created a follow up issue to move the legacy auth config to env vars: https://github.com/OpenFn/lightning/issues/4904

PR #4754, opened 2026-05-14
Apply zero-persistence policy to channel requests (#4622)

### LINKED ISSUE #4622: Support Zero-Peristence Mode
* Channels must support the project's zero-persistence setting.
* Apply the same principles about what's retained / what's not as with workflows (ie users can see the status information but not the request/response body)

midigofrank:
## Description
When a project's `retention_policy is :erase_all`, the channel proxy now suppresses PII at insert time: `request/response headers, body previews and hashes, query strings, paths, and the inbound client IP` are never written. `Status, timing, size, and classified-error` fields stay populated so the channel still has the observability data it needs.

A new `channel_requests.is_wiped` boolean records the per-request decision; the channel request detail page renders a single combined `"No Request/Response Data here!"` card in place of the headers/body sections when set. The channel history index keeps its Request Path column and shows a dash for suppressed values.

Also generalises `wiped_dataclip_viewer` (now `wiped_data_viewer`) so it can describe either workflow steps or channel requests, and updates the project settings copy to mention channel requests alongside workflow runs.

Closes #4622

## Validation steps

1. Go to your project settings and enable zero-pesistence
2. Try making request.
3. In the history page, you'll see that the row has no value for "Request path", a hyphen is displayed
4. If you open the request to see the details, you should see there is no IP address and the request and response section has been replaced by a wiped data viewer, as seen in the screenshot below:
Screenshot 2026-05-14 at 13 54 56

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


stuartc (changes requested):
Nice one, this is in good shape. A few things before merging:

**1. `Map.put(:is_wiped, true)` in `wipe_event_attrs` is dead code.** `is_wiped` only exists on `ChannelRequest`, not `ChannelEvent`, so `cast/3` silently drops it. The moduledoc says "the resulting ChannelEvent is marked `is_wiped: true`" — that's also wrong, the flag lives on the request. Just remove the `Map.put` and tweak the doc. Worth adding a `refute Map.has_key?(result, :is_wiped)` in the policy test to lock it in, since none of the current tests caught it (they all read `is_wiped` off the request row).

**2. `record_credential_error/3` feels misplaced.** The plug is reaching into `Repo`, `ChannelRequest`, `ChannelEvent`, and now `PersistencePolicy` directly — that's context work happening in the web layer. And we now have two places that do "build attrs → apply wipe policy → insert request + event" (here and in `Handler`), which means the next policy tweak has to be remembered in two spots.

Could this move into `Lightning.Channels` as something like `Channels.record_credential_error(req, conn_meta, error_message)`? The plug then just calls it and handles the HTTP response. As a bonus, the `with` block tidies up — the two `<-` for plain map-building are weird, not sure what the best way to write that is 🤔 .

**3. Missing role test on the channels show page.** The wiped-viewer describe block only tags `role: :admin`, so the `can_edit_data_retention: false` branch in `wiped_data_viewer` (the "Contact one of your project admins" copy + admin email tooltip) is uncovered for channels. The dataclip viewer has equivalent coverage — worth mirroring it here so we know the permission gate actually fires.

**4. Heads-up not a blocker:** the auth checkbox in the PR template should be ticked — you've wired up `:edit_data_retention` in show.ex, so it you should probably check it off when you have the tests.

The rest looks good, thanks dude!

midigofrank:
@stuartc I have made the changes. Let me know if you spot anything else

stuartc (approved):

PR #4760, opened 2026-05-15
Strip flag_outdated_runtime and test_runtime_latest CI jobs

elias-ba:
## Description

Delete the `flag_outdated_runtime` and `test_runtime_latest` CI jobs.

They compared our pinned `@openfn/cli` and `@openfn/ws-worker` versions to npm latest. Every time either package shipped a release, every open PR went red — even though the PR had nothing to do with the pin. The checks aren't blocking, but they polluted the CI signal and trained us to either ignore red or chase a runtime bump on every PR.

Per the [#devs Slack discussion](#): the existing `web_and_worker_test` integration test already covers the worker-version regression signal on every PR, and the deeper coverage `test_runtime_latest` was trying to provide belongs in a dedicated integration repo, not Lightning's CI. No replacement in this PR.

## Validation steps

CI on this PR is the validation — the two runtime jobs no longer exist, so they won't run here or on any future PR.

## AI Usage

- [x] I have used Claude Code
- [ ] I have used another model
- [ ] I have not used AI

You can read more details in our
[Responsible AI Policy](https://www.openfn.org/ai#pull-request-templates)

## Pre-submission checklist

- [x] I have performed an AI review of my code (we recommend using `/review`
      with Claude Code)
- [x] I have implemented and tested all related **authorization policies**.
      (e.g., `:owner`, `:admin`, `:editor`, `:viewer`)
- [ ] I have updated the **changelog**.
- [x] I have ticked a box in "AI usage" in this PR

PR #4761, opened 2026-05-15
Auto-bump @openfn/cli and @openfn/ws-worker via Dependabot

elias-ba:
## Description

Three orthogonal hygiene fixes around the `@openfn/cli` and `@openfn/ws-worker` runtime pins:

1. **Fix `.github/dependabot.yml`.** The existing `/assets` entry was set to `ecosystem: "mix"`, which silently matched nothing. Change it to `"npm"` and scope it to `@openfn/*` so Dependabot opens a daily auto-PR when either package ships a release.
2. **Move the `@openfn/cli` pin into `assets/package.json`.** The version was previously a hardcoded Elixir constant (`@cli_version` in `lib/mix/tasks/install_runtime.ex`), which Dependabot can't track. The Mix task now reads the pin at compile time from `package.json` via `File.read!` + `Jason.decode!`, with `@external_resource` so recompiles trip when the JSON changes. Single source of truth for both pins.
3. **Loosen brittle version assertions in `web_and_worker_test`.** The test pinned exact builds for `language-http`, `worker`, `node.js`, and `language-common`. Each upstream release turned the test red for reasons unrelated to Lightning. Asserts each version key is present and looks like a version (`\d+\.\d+`) instead.

Pairs with #4760 (which removes the per-PR runtime checks entirely). These three changes stand on their own — they pay off whether or not those checks ever return.

## Validation steps

- Run `mix lightning.install_runtime` after changing `@openfn/cli` in `assets/package.json` and confirm the new version is installed under `priv/openfn`.
- After merge, Dependabot will open its first bump PR within 24 hours of the next `@openfn/*` release.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -69,6 +69,7 @@
   "devDependencies": {
     "@eslint-community/eslint-plugin-eslint-comments": "^4.4.1",
     "@eslint/js": "^9.21.0",
+    "@openfn/cli": "1.35.3",
     "@openfn/ws-worker": "^1.25.0",
     "@playwright/test": "^1.55.0",
     "@redux-devtools/extension": "^3.3.0",
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,19 +5,19 @@
 
 version: 2
 updates:
-  - package-ecosystem: "mix" # See documentation for possible values
-    directory: "/" # Location of package manifests
+  - package-ecosystem: "mix"
+    directory: "/"
     schedule:
       interval: "weekly"
-      
-  - package-ecosystem: "mix"
-    # Workflow files stored in the default location of `.github/workflows`. (You don't need to specify `/.github/workflows` for `directory`. You can use `directory: "/"`.)
+
+  - package-ecosystem: "npm"
     directory: "/assets"
     schedule:
-      interval: "weekly"  
-      
+      interval: "daily"
+    allow:
+      - dependency-name: "@openfn/*"
+
   - package-ecosystem: "github-actions"
-    # Workflow files stored in the default location of `.github/workflows`. (You don't need to specify `/.github/workflows` for `directory`. You can use `directory: "/"`.)
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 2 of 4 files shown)


josephjclark:
Thank you for raising this @elias-ba ! The dependabot update is cool.

Looking at this stuff makes me want to rethink install_runtime and Lightning's relationship to the CLI.

I think this script is a hangover from long ago, before we had the worker, when the CLI was the runtime engine that executed jobs.

I _think_ that Lightning's only dependencies on the CLI now are:
1) Integration tests 
2) Metadata loading

The fact is that this quite complex install_runtime script is just installing 2 npm packages (and I'm not even sure what `common` is used for).

The script is also installing the CLI globally, which I don't think I love.

This is all too big to address here, but I'm feeling a compulsion to do _something_

assets/package.json
```diff
@@ -69,6 +69,7 @@
   "devDependencies": {
     "@eslint-community/eslint-plugin-eslint-comments": "^4.4.1",
     "@eslint/js": "^9.21.0",
+    "@openfn/cli": "1.35.3",
```
josephjclark, line 72: hmm. I liked this at first, but these are dependencies of the app frontend. The CLI isn't a dependency of that at all. So I'm not sure this is the right place.

And actually, given that lightning  depends on the CLI in production, shouldn't it be a prod dependency?

josephjclark: Oh wait - also, `install_runtime` will install the CLI globally. But `npm install --prefix assets` will install a local CLI in this folder. I'm not actually sure which bin will get used when running `openfn` (probably the global one).

midigofrank (changes requested):
@elias-ba I don't see the need to install the cli in the assets just to get the version so that `install_runtime` can again run `npm install` .

Also, now that we no longer check the worker version in the ci, do you still think this change is needed?

midigofrank:
@elias-ba I'm closing this. The worker version is no longer getting checked at the ci and there is more work needed to properly install the cli. 

Feel free to reopen if you think we still need it

PR #4763, opened 2026-05-15
Sandboxes: scope visibility, tighten merge gate, drop superuser bypass

### LINKED ISSUE #4745: Hide sandboxes the current user can't access from the Sandboxes list
`Lightning.Projects.list_workspace_projects/2` walks the `parent_id` tree from the workspace root and returns every descendant. It takes no `current_user` parameter. `SandboxLive.Index.load_workspace_projects/1` assigns the result to `@sandboxes` and the page renders it unfiltered.

The result is that the Sandboxes page on a parent project shows every descendant in the tree, including sandboxes the current user is not a `project_users` row on. Clicking those entries 404s. Worse, the same list backs the merge / edit / delete actions, so a user can trigger those on sandboxes they have no access to (gated only by their role on the parent).

This is inconsistent with the projects list, which only shows projects the current user has a `project_users` row on.

## How it surfaced

[Loom](https://www.loom.com/share/d2cd3f99195640e0972d0fc42dcf152a). Visible during UAT on staging where legacy sandboxes have only their creator as a member (audit: 12 on staging, 5 on prod with significant gaps, all created before sandbox provision auto-copied parent users).

## What to fix

Minimal fix: `list_workspace_projects/2` grows a `current_user` parameter and filters descendants to those where the user has a `project_users` row. `SandboxLive.Index.load_workspace_projects/1` passes `socket.assigns.current_user` through. Matches projects-list behaviour today.

Cleaner option, out of scope here: a richer access model where users are assigned to "everything in workspace / all sandboxes / specific project," controlled from the parent's collaboration list.

### LINKED ISSUE #4762: Hide non-accessible sandboxes from the sandboxes list
**Describe the bug**

When a user is added to a parent project after one of its sandboxes already exists, they don't get access to that pre-existing sandbox (project users aren't synced down to existing sandboxes, which is intentional). But the sandbox still appears in their sandboxes list, and depending on their role on the parent project they can also merge, edit, or delete it from the UI.

Projects already follow the right rule: users only see projects they have access to. Sandboxes should do the same.

Loom showing the current behaviour: https://www.loom.com/share/d2cd3f99195640e0972d0fc42dcf152a

**To Reproduce**

1. As user A, create a parent project and provision one sandbox under it.
2. As user A, add user B to the parent project as `:admin` or `:editor`. Do not add user B to the sandbox.
3. Sign in as user B and open the sandboxes list for the parent project.
4. Observe that the sandbox created in step 1 is listed, even though user B is not a project user on it.
5. Observe that user B can also act on it (merge / edit / delete) based on their parent-project role.

**Expected behavior**

The sandboxes list should only include sandboxes the current user is a project user on. Sandboxes the user cannot access should be hidden, and the corresponding actions should not be reachable. This matches how the projects list already behaves.

**Additional context**

This is the cheap short-term fix. The deeper question is the access model itself: today, parent-project access does not flow to sandboxes, and sandbox access does not flow to siblings. A future access model could let collaborators be added at the workspace level, at the "all sandboxes" level, or to a specific project/sandbox, and would change how visibility is computed. That redesign is out of scope here; this issue is just the visibility fix to make the current behaviour consistent with projects before the public release of sandboxes.

elias-ba:
### Problem

Today, having any role on a parent project automatically surfaces every sandbox underneath, in the Sandboxes page, the global project picker, and the breadcrumb. The actual entry policy is already direct-membership-only on `main`, so users see sandboxes they can't actually open. The mismatch is confusing and leaks names of sandboxes a user has no business seeing.

### Fix

Sandboxes only show up for users with direct membership on that sandbox (or for support users when the sandbox's own support-access flag is on). Visibility now matches what the entry policy already enforces.

- Sandboxes only appear for users with direct membership, in three places: the Sandboxes page, the global project picker, and when visiting the sandbox URL directly.
- If you can see a sandbox but not its parent, the breadcrumb at the top-left starts at the highest project you can see; ancestors you can't reach are hidden silently.
- If you can see a sandbox but not the sandboxes between it and the root, the picker collapses the gap and nests the sandbox under the highest visible ancestor.

### Two alignments bundled in

Both are pre-existing bugs/inconsistencies rather than new policy decisions, so I bundled them here:

- **Merge button.** Now requires admin/owner on the source sandbox (or admin/owner on the root). On `main`, an editor on the root could click Merge, but the cleanup that retires the source requires admin/owner, so the merge silently leaves the source in place. The button now matches what the action can complete.
- **Superuser bypass.** Removed from sandbox policies. `Lightning.Policies.ProjectUsers` already ignores `User.role` for regular projects, so this aligns sandboxes with the rest of Lightning. Superuser is a type of user, not a project role.

The bigger access-rule decisions (parent-admin floor, root cascade) live in #4768 for separate discussion.

Closes #4762.
Closes #4745.

### Setup

Seed the demo users. Either set `IS_RESETTABLE_DEMO=yes` and run `mix run priv/repo/demo.exs`, or from IEx:

```elixir
iex -S mix
iex> Lightning.SetupUtils.setup_demo(create_super: true)
```

Either gives you four users (all password `welcome12345`): **Amy Admin** (`demo@openfn.org`, owner of DHIS2), **Esther Editor**, **Vikram Viewer**, and **Sizwe Super** (superuser).

Sign in as Amy, open DHIS2's **Sandboxes** page (left sidebar), and click **+ New sandbox** to provision one. To add or remove members on a project, open its **Settings** in the left sidebar and pick the **Collaboration** tab. Reset Esther's and Vikram's memberships between scenarios as needed.

### 1. A user without a sandbox membership cannot see the sandbox

Sign in as Amy and add Esther to **DHIS2** as an Admin using DHIS2's **Collaboration** tab. Do **not** open the sandbox's Collaboration tab; Esther stays absent from the sandbox itself.

1. Sign out. Sign in as Esther.
2. Open DHIS2's **Sandboxes** page.
3. Click the project name in the top-left breadcrumb to open the project picker.

The sandbox is absent from the Sandboxes list, and the picker shows only DHIS2 with no nested entries. Esther's Admin role on DHIS2 does not surface the sandbox. Visiting the sandbox URL directly (`/projects/<sandbox_id>/w`) redirects her away.

### 2. Adding the user directly to the sandbox makes it visible

Still signed in as Amy, open the sandbox's **Collaboration** tab and add Esther as a Viewer.

1. Sign in as Esther.
2. Reload DHIS2's **Sandboxes** page.
3. Re-open the project picker.

The sandbox now appears in both the Sandboxes list and the picker. Visiting the sandbox URL directly now loads it.

### 3. Support-user access follows the sandbox's own support-access toggle

Sign in as Sizwe, open `/settings/users`, edit a fresh user with **no `project_users` row anywhere**, tick **Support user**, and save.

1. Sign in as that support user.
2. Open the sandbox URL directly (`/projects/<sandbox_id>/w`).

The sandbox inherits DHIS2's `allow_support_access` at provisioning. While the sandbox's flag is `true`, the support user can open it. As Amy, open the sandbox's **Settings** and toggle `allow_support_access` to `false`:

3. Reload as the support user. Access is denied. DHIS2's own flag value is irrelevant; only the sandbox's flag controls support access to the sandbox.

### 4. The breadcrumb and picker hide ancestors the user can't reach

Two related observations on the same underlying rule (ancestors you can't access aren't shown).

**Breadcrumb truncation at the access root.** As Amy, provision sandbox-A under DHIS2 (if you haven't) and add Esther to **sandbox-A** as an Editor (no role on DHIS2). As Esther, open sandbox-A, then its **Sandboxes** page, and provision a nested sandbox-B (Esther becomes its Owner at creation). Open any page inside sandbox-B. The top-left breadcrumb shows `sandbox-A/sandbox-B`, not `DHIS2/sandbox-A/sandbox-B`.

**Picker collapses hidden intermediates.** As Amy, reset Esther's memberships, then re-add her: Editor on DHIS2, Viewer on sandbox-B, no role on sandbox-A. As Esther, open the project picker. It shows DHIS2 at the top level with sandbox-B nested directly under it (one level of indentation); sandbox-A is absent.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


brandonjackson (approved):

josephjclark:
@elias-ba the diff here is kinda crazy which makes it a bit harder to read what's going on assess risk. Does it just need a rebase?

josephjclark:
@elias-ba my instinct is that the behaviours here are all correct. I'd like to test a bit deeper but I think this is a good change.

josephjclark:
This will fix https://github.com/OpenFn/lightning/issues/4745 right?

assets/js/picker/Picker.tsx
```diff
@@ -203,7 +210,15 @@ export function Picker(props: PickerProps) {
   }, [openPicker, openEvent]);
 
   const go = (href: string, sameSection = false) => {
-    window.location.href = href + (sameSection ? window.location.hash : '');
+    const fullHref = href + (sameSection ? window.location.hash : '');
```
josephjclark, line 213: What problem is this change addressing?

lib/lightning/policies/sandboxes.ex
```diff
@@ -2,12 +2,30 @@ defmodule Lightning.Policies.Sandboxes do
...
-  - Editors (and above) on the root project can merge sandboxes
-  - Editors (and above) on the parent project can provision sandboxes
-  - Superusers can manage any sandbox anywhere
+  A sandbox is an independent project with its own `project_users`
+  membership. Authority is decided by the actor's role on the project
+  involved in the action, never inherited from an ancestor and never
+  granted by user type (the `:user`/`:superuser` enum has no bearing
+  on public-surface authorization).
+
+  ## Two-sided merge gate
```
josephjclark: Are all these comments valuable? Feels like a lot of information (too much?) in the wrong place

lib/lightning/policies/sandboxes.ex
```diff
@@ -22,24 +40,20 @@ defmodule Lightning.Policies.Sandboxes do
...
-  User can create sandboxes if they are:
-  - Editor/admin/owner of the parent project they're creating the sandbox under
+  User must be `:editor`, `:admin`, or `:owner` of the parent project
+  they are creating the sandbox under.
 
   ### `:merge_sandbox`
-  User can merge a sandbox into a target project if they are:
-  - Editor/admin/owner on the target project
-  - Superuser
+  User must be `:editor`, `:admin`, or `:owner` on the target project.
```
josephjclark: But must also be admin or owner in the source project right? We should document this here

lib/lightning/policies/sandboxes.ex
```diff
@@ -50,94 +64,63 @@ defmodule Lightning.Policies.Sandboxes do
...
         Enum.any?(
-          root_project.project_users,
-          &(&1.user_id == user.id and &1.role == :editor)
+          sandbox.project_users,
+          &(&1.user_id == user.id and &1.role in [:owner, :admin])
         )
 
-    has_full_privileges = is_superuser or is_root_owner_or_admin
-
-    if has_full_privileges do
```
josephjclark, line 115: So this is one of our changes, where previously OG owners and superusers retain control of sandboxes.

This is being dropped in favour or making the sandbox fully independent and relying only on the database rules. Which feels clean. But doesn't it allow a sandbox admin to lock out the superuser or owner from the sandbox, effectively stealing the project?

assets/js/picker/Picker.tsx
elias-ba: The picker dropdown used to call `window.location.href = ...` when you selected a project, which is a full browser reload. The `go` helper here triggers a soft Phoenix LiveView navigation instead, so the sidebar and the picker button stay mounted across the transition (no flicker). It's UX polish rather than a bug fix.

lib/lightning/policies/sandboxes.ex
elias-ba: The superuser bypass is now gone, but the root-cascade still applies: an owner/admin on the workspace root (the OG owner) keeps manage rights on every descendant sandbox via `has_root_project_permission?`. A sandbox admin can't lock them out. So the original concern doesn't bite: the workspace owner retains control of any sandbox in their workspace.

lib/lightning/policies/sandboxes.ex
elias-ba: Trimmed the moduledoc to the rule list. Kept a short note explaining why destructive actions are admin/owner scoped (matches the rest of Lightning) and why `User.role` is ignored (it's a user type, not a project role - same as `Lightning.Policies.ProjectUsers`).

lib/lightning/policies/sandboxes.ex
elias-ba: Good call - kept this in the PR after all. You're right that the merge button on `main` is gated by `is_root_editor_plus`, but the cleanup that retires the source requires admin/owner (via `:delete_sandbox`). So an editor on the root could click Merge, the data sync would run, but the source sandbox would silently stay in place. That mismatch is a real bug. The fix tightens `check_manage_permissions/3` so the merge button matches the action: admin/owner on the source sandbox (or admin/owner on the root). Updated the moduledoc to document the rule explicitly. This also aligns merge with the rest of Lightning, where destructive actions are admin/owner scoped.

elias-ba:
> This will fix https://github.com/OpenFn/lightning/issues/4745 right?

Yes, #4745 and #4762 describe the same visibility bug.

elias-ba:
> @elias-ba the diff here is kinda crazy which makes it a bit harder to read what's going on assess risk. Does it just need a rebase?

Scoped the PR down. The four access-rule changes (cascade, parent-admin floor, merge gate, superuser bypass) were originally bundled together. After your review, only the visibility fix plus the two access changes that I'd call straight bug fixes survived: the merge button now matches the action it triggers, and `User.role: :superuser` is no longer treated as a project-access bypass (consistent with `Lightning.Policies.ProjectUsers`). Everything else lives in #4768 for separate discussion. Also retargeted at `release/2.16.4`.

assets/js/picker/Picker.tsx
josephjclark: Got it. And do we use this technique elsewhere? Is it always so painful or do we have some library support for it?

just feels like a lot of effort just to navigate!

lib/lightning/policies/sandboxes.ex
```diff
@@ -22,24 +33,22 @@ defmodule Lightning.Policies.Sandboxes do
...
-  User can create sandboxes if they are:
-  - Editor/admin/owner of the parent project they're creating the sandbox under
+  User must be editor/admin/owner of the parent project they're creating
+  the sandbox under.
 
   ### `:merge_sandbox`
-  User can merge a sandbox into a target project if they are:
-  - Editor/admin/owner on the target project
-  - Superuser
+  User must be editor/admin/owner on the target project.
```
josephjclark: And admin/owner on the sandbox itself right? Don't want to be too picky but it's an important detail which I think should be surfaced here

lib/lightning/policies/sandboxes.ex
josephjclark: Just tested and the owner cannot be removed from a sandbox, so an admin can't block them out. Thanks!

assets/js/picker/Picker.tsx
elias-ba: The UX was really bad, so I think the effort is worth it. Now I am not sure whether this is the right way to fix honestly. It was the easier for me. Any idea or suggestion ?

lib/lightning_web/live/project_live/settings.html.heex
josephjclark, line 939: are these diffs expected here? Look unrelated

josephjclark (approved):
Looks great ,diff is clean and changes look like proper bug  fixes, with the behavioral questions factored out.

Thanks @elias-ba

lib/lightning/policies/sandboxes.ex
elias-ba: Done. Module doc now mentions admin/owner on the source sandbox alongside the target-side rule

PR #4776, opened 2026-05-19
Bump worker to 1.25.1

josephjclark:
Bump worker version to 1.25.1

Integration tests not updated for a patch


josephjclark:
I've raised this for integration tests but there is a release pending. Suggest we retest and merge after release

rorymckinley:
> I've raised this for integration tests but there is a release pending. Suggest we retest and merge after release

I think we usually update the CHANGELOG as well.

PR #4801, opened 2026-05-27
Update adaptors on the fly

### LINKED ISSUE #220: Credential types list should be pre-populated (caching?)

### LINKED ISSUE #325: Build service to pull credential schemas from npm/github

### LINKED ISSUE #1996: Adaptor registry is out of sync when the app is running in a cluster

### LINKED ISSUE #2209: Better Adaptor List Refreshing

### LINKED ISSUE #3114: Update adaptor icons list for new adaptors

stuartc:
## Description

Lightning now keeps its own adaptor catalogue in Postgres and refreshes it while
running, so a new adaptor or a new adaptor version shows up in the editor within
the hour with no rebuild and no redeploy.

Adaptor icons and credential schemas come from that catalogue too, served over
HTTP at request time rather than downloaded into the Docker image at build time.

In a cluster, one node does the refreshing and the others learn about it over
PubSub. Deprecated adaptors and deprecated versions are hidden from the pickers.

The design is strategy based:

### `:npm`

Similar to the original implementation, it calls out to NPM to get a list of
adaptors for the `@openfn` organisation. It filters for `language-` package
names, and has (currently) a hard-coded list of adaptors to exclude.

It fetches adaptor schemas via jsDelivr and icons from GitHub.

### `:local`

Similar to the NPM strategy, but all local - having parity with the original
`LOCAL_ADAPTORS` env variable. It retrieves the icons and schemas directly from
the local checkout, and supports overlay directories.

Both strategies are used by a centralized store and caching layer, where adaptor
versions, listings, schemas and icons are persisted in the database.

Another new feature is the ability to (more easily) dump and import all adaptor
information, including icons & schemas, into a tarballs and you can import them
onto another (in most cases) offline instance.

This replaces `Lightning.AdaptorRegistry`, which fetched the npm listing once at
boot and held it in memory, and the four build-time Mix tasks that baked schemas
and icons into the image (`install_schemas`, `install_adaptor_icons`,
`download_adaptor_registry_cache`). All of that is _deleted_.

These is a new [ADAPTORS.md](ADAPTORS.md) doc that explains how to set things up
for local repos, offline setups and some basic troubleshooting.

Closes #4473  
Closes #3114  
Closes #2209  
Closes #325  
Closes #1996  
Closes #220

## Validation steps

1. **The default path, from nothing.**
   `mix ecto.reset && iex -S mix phx.server`. The catalogue starts empty and the
   scheduler fills it from npm on boot; until it does, saving a job is refused
   with "adaptor catalogue is not ready yet". Open a workflow, confirm the
   adaptor picker fills and a job saves.
2. **A new version, no restart.** In the running server's iex:
   `Lightning.Adaptors.refresh(await: true)`, or click **Refresh Adaptor
   Registry** via the Admin Menu > Maintenance page. An editor left open should
   pick up the change without a reload.
3. **Icons.** Open a workflow. Every job node in the diagram, the minimap and
   the adaptor picker should show its adaptor's icon rather than a text label.
   These come from the running instance now, not from the image, so if they
   render at all the new path works. To see the caching, open DevTools' network
   tab and reload: the icon requests should be served from cache, and each URL
   carries a content hash, so a changed icon gets a new URL rather than a stale
   one.
4. **Credential schemas.** Go to `/credentials` and click **New credential**.
   The type list is built from the catalogue now, instead of the `priv/schemas`
   files that used to ship in the image, so it should include every
   non-deprecated adaptor. Pick one and check the form fields match that
   adaptor's configuration schema.
5. **A local adaptors checkout.**
   `ADAPTORS_STRATEGY=local ADAPTORS_LOCAL_REPO=/path/to/adaptors`, restart, and
   the picker should list the monorepo packages. Comma-separate two roots and
   check the log names each shadowed package.
6. **Offline.** Set `ADAPTORS_ICONS_PATH=/tmp/adaptor-icons` first and restart.
   By default, the icon cache lands under `System.tmp_dir!/0`, which can be
   random depending on your OS.

   ```sh
   mix lightning.adaptors.dump --path snapshot.json
   tar czf icons.tar.gz -C /tmp/adaptor-icons .
   ```

   Now reset the DB, empty the icon directory, and import with the network off:

   ```sh
   mix ecto.reset
   rm -rf /tmp/adaptor-icons && mkdir -p /tmp/adaptor-icons
   tar xzf icons.tar.gz -C /tmp/adaptor-icons
   mix lightning.adaptors.import --path snapshot.json --replace
   ```

   Start the server with no network. The picker and the icons should both work,
   and the hourly refresh should log a warning and leave the imported rows alone
   rather than emptying them. `mix lightning.adaptors.snapshot` builds the same
   file straight from npm if you have no populated instance to dump from, though
   it carries no icons.

7. **Refreshing switched off.** `ADAPTORS_REFRESH_INTERVAL_SECONDS=0` stops the
   scheduled hourly refresh, for anyone who wants to control it themselves.
   Confirm the two manual routes from step 2 still work with it set:
   `Lightning.Adaptors.refresh(await: true)` in iex, and the Maintenance page's
   refresh buttons.

**On upgrade.** Three migrations run (`create_adaptors`,
`add_adaptor_catalogue_indexes`, and widening `credentials.schema` from 40 to
100 characters). For a connected instance that's all: the first refresh fills
the catalogue from npm. Four things worth knowing:

- **The network moved from build time to runtime.** Image builds no longer hit
  npm, so they're faster and can't fail on a registry hiccup, but a fresh
  instance now needs npm reachable when it starts. An airgapped instance must be
  seeded before use.
- `SCHEMAS_PATH` and `ADAPTORS_REGISTRY_JSON_PATH` are now ignored, with no
  warning. Anyone setting them should drop them.
- `LOCAL_ADAPTORS=true` and `OPENFN_ADAPTORS_REPO` still work and do warn at
  boot, pointing at `ADAPTORS_STRATEGY=local` and `ADAPTORS_LOCAL_REPO`.
- `ADAPTORS_ICONS_PATH` is set to `/app/priv/adaptor_icons` in the release image, and `docker-compose.yml` mounts a named volume there. Outside the image it still defaults under the system temp dir. Either way, persistent storage has to be mounted at that path or icons refetch after every restart.

**New dependency.** `highlander_pg ~> 1.0` for the advisory-lock singleton,
which needed `{:postgrex, override: true}`.

**Where to look hardest**

- `lib/lightning/adaptors/scheduler.ex` and `supervisor.ex` — the refresher is a
  Postgres advisory-lock singleton. When it goes wrong it goes wrong quietly:
  either no node refreshes, or all of them do. There's a leader-handover
  integration test, but only one.
- `lib/lightning/adaptors/icon_cache.ex` and
  `lib/lightning_web/controllers/adaptor_icon_controller.ex` — filesystem writes
  keyed on package names that came from npm, served on an unauthenticated route.
  The last three commits are path-traversal hardening, so this is where the
  churn was.
- `lib/lightning/workflows/job.ex` — if the adaptors store fails to get
  populated, this turns into refused workflow saves. Worth checking "not found"
  and "not ready" stay distinguishable.
- `lib/lightning/credentials/schema_reconciler.ex` plus the `credentials.schema`
  widening from 40 to 100 characters, we now store full npm package names
  (instead of the short handed `http`, `dhis2` etc) and this process rewrites
  existing rows from short adaptor names to full npm names. It would have been
  much easier to use a migration but we need an adaptor listing. I imagine we
  can remove this in a subseqent release.

**Two new routes.** `/adaptors/icons/:name/:filename` is
public and unauthenticated, deliberately, so icons cache. `/adaptors/catalogue`
needs a session but isn't project-scoped, which is right for global reference
data. `/settings/maintenance` is superuser-only, checked in `mount/3` and again
in each `handle_event`. No policy module changed.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


midigofrank (approved):

lmac-1:
Being nosy here as I switched branches and noticed something.. did this PR mean to change Claude commands to skills as part of this PR? @stuartc

stuartc:
> Being nosy here as I switched branches and noticed something.. did this PR mean to change Claude commands to skills as part of this PR? @stuartc 

Why yes, it did! The branch was in flight for like 3-4 weeks, and I did some housekeeping while working on things in different models/sandbox envs. 
I can't remember the exact reason for the commands to skills, it might have been around invocable vs non-invocable via other prompts or subagents; can't recall.

PR #4813, opened 2026-05-29
don't write no-ops as new URL history items

### LINKED ISSUE #4812: Must click "back" 3x to go back after navigating to canvas
If you click into a workflow from the workflow list and try to go back to the list by either clicking "back" or pressing _delete/backspace_ in your browser, you must execute that "back" action 3 times to actually get back.

This is probably because sometimes (for reasons?) we call `updateSearchParams` a few times while loading a complex page with many possible param, but many of the updates are no-ops!

Image

If we compare, and return early, hopefully this won't happen anymore.

Version `v2.16.6 (1.5.86)`

## Video of bug

https://www.loom.com/share/9c45aa33db714c4daa982417e1cba6e7

## Video of fix with diff above applied

https://www.loom.com/share/17f8c56f0d8d41a58fc71df6960e2732

taylordowns2000:
## Description

This pr closes #4812 by preventing no-ops (where previous URL == new URL) from being added as new URL history items.

## Validation steps

1. Go to workflow list
2. Click into workflow
3. Click back
4. See that it works after a single click, not 3 clicks

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to
 
 ### Fixed
 
+- Fix issue where back button must be pressed 3 times to go back once from the
+  Workflow canvas [#4812](https://github.com/OpenFn/lightning/issues/4812)
 - Fix `purge_deleted` Oban job crashing when a soft-deleted project has
   associated OAuth clients. The `project_oauth_clients` join rows are now
   cleaned up alongside the other project-scoped deletes in
--- a/assets/js/react/lib/use-url-state.ts
+++ b/assets/js/react/lib/use-url-state.ts
@@ -117,13 +117,16 @@ class URLStore {
     newParams: Record<string, string | number | boolean | null>
   ) => {
     const newURL = new URL(window.location.pathname, window.location.origin);
-    // Only set params with non-null values (clears all existing params)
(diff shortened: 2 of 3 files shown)


lmac-1 (approved):
I added an extra commit to extend this fix to `replaceSearchParams` for housekeeping. Double checked and this shouldn't have any extra implications. Approved

stuartc (approved):
@taylordowns2000 I made a few tweaks here:

Two gaps with the guard/is the url the same. `updateHash` didn't have it, although it's not actually being called by anything right now which is weird but I'm hesitant to remove it because it's not impossible we will use hash state urls again and if used it'd still push a history entry on a no-op and could bring the stacking bug back. Routed that through a shared function guard too (nothing calls it yet, so just being safe).

The check was order-sensitive - it compared full URL strings, so if `replaceSearchParams` got the same params in a different order it didn't match and pushed a dupe anyway. Made the comparison ignore param order so a reorder counts as a no-op.

Added a little `currentURL` helper the three writers seed from - so there's one correct guard instead of two copies and one missing. Ended up being slightly less code overall. Added tests for the new cases too.

Issue #4800, opened 2026-05-27
Dataclip insert times out building the search vector, which loses the run


stuartc:
**Describe the bug**

When a run finishes, the worker sends each step's output dataclip to Lightning to
be saved. Saving a dataclip runs a database trigger that builds its search vector
(`update_dataclip_search_vector`, using `jsonb_to_tsvector`). For large dataclips
this is slow — slow enough that the insert can sit on a database connection for
over 70 seconds and get cancelled. When that happens the whole insert is rolled
back, so the dataclip is never saved.

The problem is what comes next. The worker carries on sending the rest of the
run's events, and every one of them now points at something that was never
saved: the next step can't find its input dataclip, log lines can't be tied to a
step, and finally `run:complete` references a `final_dataclip_id` that doesn't
exist. Lightning rejects all of these and the run ends up lost.

This is one of the failure shapes behind #4794 (Shapes A, B and X). We already
have a guard for dataclips that are too big for a tsvector
(`program_limit_exceeded`), but that doesn't help here — the dataclip isn't too
big, the insert is just too slow.

**Version number**

Seen on production. Worker was on `1.24.1` for the traced run; the same path
still exists on current `main`.

I have reproduced this locally on main:

- [ ] Yes
- [ ] No

**To Reproduce**

1. Run a workflow that produces a large dataclip (big JSON body).
2. Under load, the dataclip insert blocks on the search-vector trigger long
   enough to hit the connection timeout (~70s) and roll back.
3. The dataclip is never saved.
4. Following events fail with `input_dataclip does not exist`,
   `step_id must be associated with the run`, and finally
   `final_dataclip_id does not exist` — the run is lost.

**Expected behavior**

A slow search-vector build shouldn't be able to roll back the dataclip insert or
lose the run. The dataclip should still be saved even if its search vector can't
be built in time.

**Additional context**

Likely worth moving the search-vector work out of the synchronous insert path —
e.g. build it in a background job, or put a statement timeout on the trigger and
skip the vector (saving the dataclip anyway) when it's exceeded.

There's a separate worker-side problem that turns this into a *lost* run rather
than a recoverable hiccup: the worker treats the failed dataclip send as success
and keeps going. That will be tracked in a `kit` issue. This issue is just about
the database side.

Found while investigating #4794.

PR #4818, opened 2026-05-30
Defer log_lines search_vector indexing off the insert path

### LINKED ISSUE #4425: Run channel timeouts for `run:log` and `step:complete` events
The worker is reporting a fair number of `step:complete` and `run:log` timeouts every single day. We set `WORKER_MESSAGE_TIMEOUT_SECONDS=80` in the worker which reduced the `run:log` timeouts but you can still see a few of them

Here's the sentry link: https://openfn.sentry.io/issues/6509590273/events/c2ee36d85fbc48e7b4552a321b82e58a/

stuartc:
## Description

This changes how `log_lines` full-text search gets indexed — taking it off the
log-insert hot path to cut down the `run:log` channel timeouts we've been seeing
under heavy log volume.

The old `AFTER INSERT` trigger computed `to_tsvector` synchronously inside the
insert transaction (so it never actually deferred anything) and double-wrote
every row via a self-`UPDATE`. It also had no `program_limit_exceeded` guard, so
a single oversized message would abort the insert and roll back the whole batch
on the `insert_all` path.

Now inserts leave `search_vector` NULL and a background Oban worker
(`Lightning.LogLines.SearchVectorWorker`) backfills it out-of-band on a dedicated
`search_indexing` queue. There's a guarded `safe_to_tsvector` SQL function and a
partial index (`WHERE search_vector IS NULL`) so finding pending rows stays
cheap. Live log streaming is unaffected — it's push-based over PubSub and never
reads `search_vector`; only full-text log *search* now lags ingestion slightly,
typically under a minute.

Three migrations, ordered 1a→1b→1c: add `safe_to_tsvector`, add the partial index
(built per-partition `CONCURRENTLY` then attached to the parent), then drop the
trigger.

Closes #4425

## Caveat

This is one thing about this approach worth mentioning:

In the situation of an idle instace, with a few logs being inserted. 
Indexing is poll-driven by the 1-minute cron - a lone insert can't enqueue the worker itself. 
So with ~zero load, a new row waits for the next cron tick: **up to ~60s, ~30s on average.** 
This is a floor set by the cron cadence, not by load.

## Validation steps

1. Run the migrations (`mix ecto.migrate`). Confirm the `log_lines`
   `set_search_vector` trigger is gone, `safe_to_tsvector` exists, and the
   partial index is VALID across the parent + all 100 partitions. The
   `dataclips` trigger should be untouched.
2. Insert a log line — its `search_vector` is NULL and it isn't matched by log
   search yet.
3. Run `Lightning.LogLines.SearchVectorWorker` — the row's `search_vector` is
   populated and now matches `to_tsquery('english_nostop', …)`.
4. Insert an oversized (>1MB) message — it drains to an empty vector rather than
   erroring, and doesn't get stuck retrying.
5. `mix test test/lightning/log_lines/search_vector_worker_test.exs test/lightning/runs_test.exs`

1. The worker drains newest-first and "snowballs" an immediate follow-up while
   there's backlog, falling back to a 1-minute cron heartbeat. Concurrency is 1
   on the dedicated queue (SKIP LOCKED makes bumping it safe later).
2. Two subtle bits worth a close look: the snowball's Oban `unique` states are
   restricted to `[:available, :scheduled]` on purpose — the default includes
   `:executing`/`:completed`, which makes a running job dedup its own successor
   and kills the chain — and the index migration drops any INVALID leftover
   before re-creating, so a failed `CONCURRENTLY` build can be re-run.
3. `safe_to_tsvector` is also meant as the template for the dataclip trigger fix
   (#4800), which is separate.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [ ] I have performed an AI review of my code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ and this project adheres to
 
 ### Fixed
 
+- Reduce `run:log` channel timeouts under heavy log volume by moving `log_lines`
+  search indexing off the insert path. The full-text search vector is now
+  backfilled by a background worker rather than computed synchronously on every
+  insert, so log search is eventually-consistent (typically within a minute).
+  [#4425](https://github.com/OpenFn/lightning/issues/4425)
 - Channel join crashes when multiple users open the same workflow concurrently
   [#4802](https://github.com/OpenFn/lightning/issues/4802)
 - Fix `purge_deleted` Oban job crashing when a soft-deleted project has
--- a/config/config.exs
+++ b/config/config.exs
@@ -187,6 +187,8 @@ config :lightning, :is_resettable_demo, false
 config :lightning, :default_retention_period, nil
 config :lightning, :claim_work_mem, nil
 
+config :lightning, :log_lines_search_indexing, batch_size: 2_500, max_batches: 10
+
 config :lightning, Lightning.Runtime.RuntimeManager, start: false
 
 config :lightning, LightningWeb.CollectionsController,
--- a/config/test.exs
+++ b/config/test.exs
@@ -133,6 +133,10 @@ config :lightning, CLI, child_process_mod: FakeRambo
 
 config :lightning, :is_resettable_demo, true
 
+# Tiny budget so a handful of pending rows fills multiple full batches and trips
+# the per-run budget guard, exercising the snowball follow-up path.
+config :lightning, :log_lines_search_indexing, batch_size: 2, max_batches: 2
+
 config :lightning, :github_app,
   app_id: "111111",
   app_name: "test-github",
--- a/lib/lightning/config.ex
+++ b/lib/lightning/config.ex
@@ -98,6 +98,20 @@ defmodule Lightning.Config do
     end
 
     @impl true
+    def log_lines_search_indexing_batch_size do
+      log_lines_search_indexing_config() |> Keyword.fetch!(:batch_size)
+    end
+
+    @impl true
+    def log_lines_search_indexing_max_batches do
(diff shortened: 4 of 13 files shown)


taylordowns2000:
@stuartc , fantastic find here. I noticed that the description seemed to reference inserting multiple log lines at once. Wanted to double check: batch log lines insert is turned off on app.openfn.org, right?

I think we tried it a few months back and when the crashes started occurring we switched it off. I believe Joe has been tracking a bug related to log line ordering on the bulk log line feature too.

(Just double checking that we're all on the same page here.)

stuartc:
@taylordowns2000 similar, but it's not related to batch log lines, with this we update the tsvector in batches (and asynchronously). should be a bit faster and way less stress on the index.

And no, you're correct; we don't use batch logging on production right now. I can't remember what the issue was, but worth following up on this - but yeah it's independent of this.

There's a complimentary PR https://github.com/OpenFn/lightning/pull/4821 over here for dataclips. Holding both of them as draft just for a sec, the current approach on this relies on a 'snowballing' oban job, which is nice because it disconnects the insertion of loglines and dataclips from indexing (nice). But there is a small downside in that there is a chance that in a quiet system that has something inserted at like 2s after the minute would have to wait for the next minute to roll over to be indexed/searchable. So it's really designed right now for busy systems. I want to evaluate what something a bit more eager would look like before calling this done done.

midigofrank (approved):
Nicely done dude

Issue #4828, opened 2026-06-03
AI assistant emits placeholder ids/code (`__ID_JOB_…__`, `__CODE_BLOCK_…__`) instead of UUIDs/real code in generated workflows


midigofrank:
## Summary

The AI assistant (workflow template generation, via the Apollo service) sometimes produces a workflow whose nodes carry **unsubstituted template placeholders** instead of real values — symbolic ids like `__ID_JOB_<name>__` / `__ID_EDGE_<name>__` and code bodies like `__CODE_BLOCK_<name>__`. When such a workflow is applied to the collaborative Y.Doc and saved, the placeholder ids reach the DB as `:binary_id` values and fail.

Until recently this **crashed the collaborative Session** (see #4816). With the changeset validation added there, it no longer crashes — the malformed id now surfaces as a **toast/validation error** to the user. **This issue exists to track the actual root cause: why the AI assistant emits placeholders rather than substituting them, and to confirm when that is fixed.** The crash fix is a safety net, not a fix for the AI behaviour.

## Observed placeholder format

From production crash payloads (the full `save_workflow` payload was captured in the Session crash):

```
jobs:    "id" => "__ID_JOB_Envoyer-dans-DHIS2__", "body" => "__CODE_BLOCK_Envoyer-dans-DHIS2__"
edges:   "id" => "__ID_EDGE_Collecter-les-donnees->Envoyer-dans-DHIS2__",
         "source_job_id" => "__ID_JOB_Collecter-les-donnees__"
```

Crucially the payload is a **mix** of placeholders and real UUIDs — some jobs/edges/triggers already had valid ids (e.g. job `77438a5a-…`, trigger `3ae0637e-…`) alongside the placeholders. This suggests a **partial substitution / partial apply**, not a wholesale failure.

## Where the placeholders come from

The tokens `__ID_JOB_`, `__ID_EDGE_`, `__CODE_BLOCK_` appear **nowhere in the Lightning codebase** (backend or frontend). They originate from the **Apollo AI service** output for `session_type: "workflow_template"` (`Lightning.AiAssistant` → `Lightning.ApolloClient`). Lightning is the consumer: the generated workflow is stored on the chat message `code` and applied to the Y.Doc, then saved via `LightningWeb.WorkflowChannel` → `save_workflow`. Substitution of these placeholders into real UUIDs/code is expected to happen before the workflow is applied/saved; in these cases it didn't (fully).

[...]

PR #4829, opened 2026-06-03
Dont crash collaborative session on Ecto errors

### LINKED ISSUE #4816: Collaborative save_workflow crashes Session GenServer on Ecto constraint and ChangeError
**Describe the bug**

When a user saves a workflow in the collaborative editor, the call chain is `WorkflowChannel.handle_in("save_workflow", …)` (`lib/lightning_web/channels/workflow_channel.ex:355`) → `Session.save_workflow/2` → `Lightning.Workflows.save_workflow/3`.

In several cases the changeset / Multi raises an exception instead of returning `{:error, changeset}`, which propagates out of `Session.handle_call({:save_workflow, …})` (`lib/lightning/collaboration/session.ex:309`), terminates the Session
GenServer, and crashes the Phoenix channel. The user sees the collaborative editor disconnect mid-save.

Tracked in Sentry as [LIGHTNING-1AF](https://openfn.sentry.io/issues/7182438175/) (71
events, first seen 2026-01-12, last 2026-05-25) and the Ecto subset of [LIGHTNING-1D5](https://openfn.sentry.io/issues/7219642618/) (22 events total; cross-node timeout subset of 1D5 will be filed
separately).

Three sub-cases across the events of both issues. Operation (insert vs. update) is from the `Ecto.ConstraintError` text in each event:

| Sub-case | Events | Op | Notes |
|---|---|---|---|
| `Ecto.ConstraintError` on `triggers_cron_cursor_job_id_fkey` (foreign_key_constraint) | 47 | UPDATE | The trigger changeset (`lib/lightning/workflows/trigger.ex:119-125`) does not declare `foreign_key_constraint(:cron_cursor_job_id, …)`, so a stale reference raises rather than producing a changeset error. |
| `Ecto.ChangeError` on `Lightning.Workflows.Job.id` not matching `:binary_id` | 6 | INSERT/UPDATE (raised during dump) | Observed values: `"__ID_JOB_Envoyer-dans-DHIS2__"`, `"__ID_JOB_Fetch-new-birth-events-from-OpenCRVS__"`. These look like unsubstituted template placeholders — possibly from AI Assistant–generated YAML, but the origin needs investigation. **AI Assistant origin spun off to #4828.** |
| `Ecto.ConstraintError` on `workflows_pkey` (unique_constraint) | 4 | INSERT | The Workflow changeset (`lib/lightning/workflows/workflow.ex:86-95`) does not declare `unique_constraint(:id, name: "workflows_pkey")`. **Spun off to #4830** — to be fixed by preventing the duplicate INSERT (reconcile insert-vs-update by id), **not** by adding a changeset constraint. |

## Spin-off issues

This issue covers making `save_workflow` fail safely (return a structured error instead of crashing the Session). Two aspects have been extracted into their own issues:

- **#4828 — AI Assistant placeholder ids.** Follow-up on *why* the AI Assistant emits unsubstituted template placeholders (`__ID_JOB_…__`, `__CODE_BLOCK_…__`) instead of real UUIDs/code — the upstream cause of the `Ecto.ChangeError` sub-case. The crash safety net does not fix the AI behaviour.
- **#4830 — `workflows_pkey` duplicate INSERT.** Extracts the `workflows_pkey` sub-case. The fix must prevent the collaborative editor from INSERTing a workflow id that already exists (it is the same user's own workflow, not a UUID collision), **not** mask it with a changeset `unique_constraint(:id)`.

One workflow topic (`workflow:collaborate:137bde75-0dc5-4af5-88bb-2c1d7f1f703c`) accounts for ~47/66 of the events, all in the FK class. A first-pass lookup did not find this workflow in prod or staging at the time of writing. Possible explanations include hard-delete (e.g. project purge), soft-delete that's being filtered out of the query used,
or a query mismatch — needs investigation to confirm which.

For the FK sub-case, the constraint fires on UPDATE, which is consistent with the trigger row already existing in the database (loaded via `Session.fetch_workflow/1`, `lib/lightning/collaboration/session.ex:468-509`) and its `cron_cursor_job_id` no longer matching any job. The mechanism that produced this state — whether the referenced job was removed by a collaborator, by an AI-assisted regeneration, by a separate code path, or by something else — is not yet identified.

**Version number** Observed on v2.15.16, v2.16.0, v2.16.1, v2.16.2,
v2.16.3, v2.16.4, v2.16.5, v2.16.6.

- [ ] Yes
- [ ] No

**To Reproduce**

Not reproduced locally. Sentry shows the call site, the failing constraint / value, and the inbound `save_workflow` Phoenix message payload for each event.

**Expected behaviour**

`Session.save_workflow/2` returns a structured error for any data shape the changeset can't accept; the Session GenServer and Phoenix channel stay up.

**Additional context**

- Cross-node `unobserve` timeouts and node disconnects during `save_workflow` are a separate failure class and will be filed separately.
- The `__ID_JOB_<name>__` value shape in the `Ecto.ChangeError` sub-case suggests a templated origin (e.g. AI Assistant–generated YAML where placeholders weren't substituted), but this is a hypothesis from the value pattern alone and has not been confirmed. See #4828.

midigofrank:
## Description

Fixes the collaborative `save_workflow` crash classes  so the Session GenServer returns a changeset/validation error (surfaced as a toast) instead of raising and terminating the channel:

1. **Stale `cron_cursor_job_id`** — a cron trigger whose cursor points at a job that no longer exists.
   - Server: `Trigger.validate/1` now declares
     `foreign_key_constraint(:cron_cursor_job_id, …)`, so the violation is a
     changeset error instead of an `Ecto.ConstraintError`.
   - Client: `removeJob` clears any trigger's `cron_cursor_job_id` that points
     at the job being removed, so the dangling reference isn't created in the
     first place.

2. **Malformed UUID ids** — job/trigger/edge ids that aren't valid UUIDs (e.g. unsubstituted template placeholders like `__ID_JOB_…__`). `cast/3` does not format-check `:binary_id`, so these previously raised `Ecto.ChangeError` during dump on insert. A reusable `Lightning.Validators.validate_uuid/2` is now applied in the Job, Edge, and Trigger changesets, returning a changeset error instead.

Two concerns originally noted under #4816 were extracted into their own issues
and are therefore **out of scope** for this PR:

- **#4830** — the `workflows_pkey` duplicate INSERT
- **#4828** — the upstream cause of the placeholder ids (AI Assistant).

Closes #4816.

## Validation steps

1. Backend changeset tests:
   `mix test test/lightning/workflows/trigger_test.exs test/lightning/workflows/job_test.exs test/lightning/workflows/edge_test.exs`
   — a malformed UUID and a non-existent `cron_cursor_job_id` each return a
   changeset error rather than raising.
2. Frontend store test:
   `cd assets && npx vitest run test/collaborative-editor/stores/createWorkflowStore.test.ts -t removeJob`
   — `removeJob` clears a cron cursor pointing at the deleted job.
3. Manual (cron cursor, the reproducing path): open a workflow in the collaborative editor, make the trigger a cron trigger, add a new job and set it as the trigger's **Cron Input Source**, then delete that job before saving, and save. The save now succeeds with no session disconnect (previously the Session crashed with an `Ecto.ConstraintError`).

1. The two fixes are independent; each converts a raise into a changeset error.
   The client-side `removeJob` cleanup additionally prevents the dangling cron
   cursor from ever being saved.
2. For the malformed-UUID fix, scope was kept to `:id` plus the cross-node
   reference ids (edge source/target/source_trigger, trigger cron cursor);
   `validate_uuid` only runs on a non-nil change, so unset optional FKs are
   unaffected. The `Trigger` check runs after `validate_by_type` so a cursor
   that gets nilified for webhook/kafka triggers isn't falsely flagged.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


stuartc (approved):
Hey @midigofrank, I built a fair bit on top, wanted to give you the quick
rundown of what I added:

Your two fixes (stale cron cursor + malformed UUIDs) were 👌 :ok I mostly
generalised and hardened around them so the whole save path is solid:

**Widened the input validation**

Your UUID rejection was spot on, so I extended the existing validate_uuid
changeset validator to cover the FK columns too
(`workflow_id`/`project_id`/`credential` ids across
`job`/`trigger`/`edge`/`workflow` - it was only guarding `:id` before), and
de-duped a copy of it that had drifted into SearchParams. A malformed FK now
surfaces as a field error instead of an Ecto.ChangeError at insert. Also wrapped
the save transaction in a rescue so any unexpected Ecto error becomes a
changeset error + toast rather than crashing the session GenServer.

**Moved cron-cursor integrity into the DB**

Made the cron_cursor_job_id FK compound (enforces same-workflow) with ON DELETE
SET NULL. Since the DB now guarantees this, I dropped the client-side
dangling-reference reconciler from the save path - it couldn't fix the
concurrent-editor race anyway and could actually null a cursor a collaborator
had just set.

Server's authoritative now, client's just "advisory".

**Scoped the rescue to the transaction only**

So post-commit side effects (kafka/telemetry/reconcile) run outside it. Once a
save is durably committed we never want to downgrade it into a misleading "save
failed" toast; better to crash honestly and let the session re-sync.

**Tidy-up**

Split save_workflow into helpers to kill the Credo complexity (that I
introduced), swapped the logging closure for an explicit level/message, added
test coverage for the CastError branch, and de-duped the JS reconcile tests.

Issue #4834, opened 2026-06-04
sandboxes don't merge credentials?


josephjclark:
Maybe I'm losing my mind here.

When I merge a sandbox, I expect that any credentials I assign in the sandbox get merged over to main.

The two project instances should point to different environments of the same credential. But it's still the same credential.

And yet, if I create a step in a sandbox, and add a credential to it, and merge it back into main - that credential is dropped!

Something feels really off here.

josephjclark:
I can see that the project credentials in a project and its sandbox use different UUIDs.

So when merging projects, we need to try and map credentials if we're to merge them.

I think the credential-environment abstraction isn't quite where I thought it was 🤔 

This looks like it's been forever thus but I'm sure I've tested this? maybe not?

PR #4835, opened 2026-06-04
Sandboxes: handle credentials on merge

### LINKED ISSUE #4831: ValidationError when merging sandboxes
When trying to merge a sandbox, Users are seeing a toast error which says "unable to merge workflow: validation error".

No more details than that, so I'm not even sure what's failing validation

Aisha's loom: https://www.loom.com/share/5b999ff40b284206982978f452a495f9

josephjclark:
## EDIT

This PR has been updated to include https://github.com/OpenFn/lightning/pull/4842

Instead of attaching credentials in the UI at create time, @elias-ba's work now attached credentials at _merge_ time, and tells the user about it. Big improvement which removes The Ick.

## Description

This PR ensures that when merging a sandbox, any credentials added to the sandbox get properly mapped to a matching credential on the parent.

When adding a new credential, the UI will auto-fill to also add the credential to the parent.

Closes #4831 #4834 

## Discussion

I'm _sure_ I've tested this back in the early days of sandboxes. I can visualise doing it. Did we break it? Am I misremembering? Most likely I didn't test properly after the environments work.

~The original fix is to map credentials from sandbox -> parent on merge, which fixes most of the problems. But that still leaves an issue when you create a whole new credential on the sandbox: after merging, the credential is not linked to the parent project. What happens then? Do we auto-create the credential? Probably but it feels a little bit icky.~

~My solution is to handle this in the UI: when creating a new credential within a sandbox, automatically add it to the parent project. Now we ensure it'll survive on merge.~

## Validation steps

Run the following tests on a new sandbox and merge
* Add an existing credential to an existing step
* Add an existing credential to a new step
* Create a new workflow and add an existing credential to the step
* Add a new credential to an existing step (expected behaviour)?

Also need to ensure existing production behaviour isn't broken

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,13 @@ and this project adheres to
 - Ensure that credentials are properly transferred when merging a sandbox. This
   fixes a validation error which can occur on merge
   [#4831](https://github.com/OpenFn/lightning/issues/4831)
+- Free up a workflow's name when it is deleted by a merge, so a later merge can
+  reuse that name [#4831](https://github.com/OpenFn/lightning/issues/4831)
+- Replace the generic "validation error" on a failed sandbox merge with a clear
+  message, naming the conflicting workflow when there is one
+  [#4831](https://github.com/OpenFn/lightning/issues/4831)
+- Add a credential created in a sandbox to its full ancestor chain, so it
+  survives a merge into any ancestor
 
 ## [2.16.7] - 2026-06-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,7 @@ and this project adheres to
 - Consolidated run and work order state definitions into single source of truth
   by adding `Run.active_states/0`, `WorkOrder.states/0`, and
   `WorkOrder.active_states/0` and replacing all hardcoded state lists across the
-  codebase
-  [#4589](https://github.com/OpenFn/lightning/issues/4589)
+  codebase [#4589](https://github.com/OpenFn/lightning/issues/4589)
 
 ### Fixed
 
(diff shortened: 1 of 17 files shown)


josephjclark:
I've pushed a clauded fix. Haven't look at details yet, but it works in manual testing.

Two limitations suggested by Claude (1. is on my list to test anyway):

1. Credential the parent doesn't own: if the sandbox references a credential the parent has no project_credential for, the reference maps to nil (dropped) rather than creating one in the parent.
2. Keychain credentials: keychain_credential_id is not remapped (keychains are per-project clones, not shareable by credential_id). No merge test covers keychains; a new workflow using a keychain would still fail import. Worth a separate ticket.

I think we can leave 2 to later

josephjclark:
I suppose we should also be thinking about: when this fails, can we show a better error?

elias-ba (approved):
Hey @josephjclark nice work man, thanks a lot for taking this one. The credential remap reads cleanly, and matching on the shared credential and rewriting it in the merged document is a sound approach.

I've opened #4842 on top to cover a couple of cases this doesn't reach yet: freeing a deleted workflow's name on merge, which is the other cause behind the "validation error", and extending the new-credential pre-fill to the full ancestor chain so a credential also survives a merge from a nested sandbox. Together I think the two PRs close #4831 and #4834.

josephjclark:
@elias-ba changes are merged in

@theroinaochieng this ought to get released to prod as soon as possible, it fixes a big blocking user bug. It ought to have another review but it's quite isolated. Maybe @midigofrank can give it a real quick look

midigofrank (approved):
Nicely @josephjclark

PR #4839, opened 2026-06-08
Consolidate email format validation onto a single canonical validator

### LINKED ISSUE #4765: Collaborator add form silently rejects emails accepted by user registration
## Summary

The "Save Collaborator" button in the project collaboration tab stays disabled for emails that are accepted by user registration, and no error message is shown to explain why. Users hit a dead end with no signal that anything is wrong.

## How to reproduce

1. Register a user with a valid-but-non-standard email like `alice@local` (passes registration).
2. Sign in as a project owner/admin. Open **Settings → Collaboration → Add new collaborator(s)**.
3. Type `alice@local` and pick a role.
4. The Save button stays grayed out. No visible error.

## Root cause

Two different email validators with two different regexes:

- `Lightning.Accounts.User.validate_email_format/1` (`lib/lightning/accounts/user.ex:174`):
  `~r/^[^\\s]+@[^\\s]+$/` — accepts anything with an `@` and no spaces.
- `LightningWeb.ProjectLive.Collaborators.collaborators_changeset/2` (`lib/lightning_web/live/project_live/collaborators.ex:32`):
  `~r/^[\\w-\\.]+@([\\w-]+\\.)+[\\w-]{2,}$/` — requires a TLD-style domain.

So an email like `alice@local` passes registration but fails the collaborator form. The mismatch can produce users who exist in the system but cannot be added to any project as collaborators.

On top of that, the form swallows the validation error: the Save button binds to `disabled={!@changeset.valid?}` but the email field is rendered with `display_errors={false}` (`lib/lightning_web/live/project_live/new_collaborator_component.html.heex:66`). The user has no way to know the email is the problem.

## Two issues, one form

1. **Inconsistent email validation across the app.** Whatever validator User uses should be the canonical one. If a user can be created with an email, an admin should be able to add that user as a collaborator. Consolidate on a single shared email validator (e.g., factor `User.validate_email_format/1` into a shared helper and use it from `Collaborators` too).
2. **Silent disabled submit.** When the changeset is invalid, the user should see an error message next to the offending field. Either remove `display_errors={false}` on the email input, or surface the per-field errors via the existing `error_field/1` helper that's already imported in the component.

## Expected behaviour

- Adding a collaborator whose email passed registration should succeed.
- If validation fails for any reason, an inline error explains which field is invalid and why, so the user knows what to fix.

## Suggested scope

A single PR can do both: extract a shared email format validator (Accounts side is the source of truth), have `Collaborators` use it, and surface the field-level errors in the collaborator add modal.

lmac-1:
The collaborator add/invite forms used a different email regex to the rest of the app, causing a silent dead end: a user could be created with an address the collaborator forms then rejected. This PR fixes that by consolidating onto a single validator.

Previously the collaborator forms used `~r/^[\w-\.]+@([\w-]+\.)+[\w-]{2,}$/` — stricter than everywhere else, and notably rejecting plus-addressing. The new `Lightning.Validators.validate_email_format/1` uses the Zod v4 `z.regexes.email regex` — consistent with the marketing site — and is now the single source of truth across user creation, credential transfer, and both collaborator flows.

The new regex is stricter than the old loose validators were: `alice@local`-style addresses that previously passed will now be rejected. See the additional notes before shipping.

Closes #4765

## Validation steps

1. Add a user+tag@example.com address as a collaborator — should succeed (previously rejected by the form)
2. Invite a new user with a plus-addressed email via the invite form — should succeed
3. Attempt a credential transfer to a valid email — should succeed
4. (Optional) Create a user with user+tag@example.com via the superuser admin panel — should succeed; alice@local should now fail

1. **Is there any reason why we might have allowed emails like `lucy@local`? Perhaps on local deployments?**
2. Stu has run a query on production to check that there are no users with emails blocked by the new regex. 
3. There is an additional PR (#4839) that handles the validation not appearing bug reported in the same issue

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lmac-1:
@stuartc are you able to run the query on prod for the regex? Then I'll change this to a ready PR 🥳

lmac-1:
@stuartc has confirmed that there are no email addresses on prod that fall outside of this regex. Ready for review.

midigofrank (approved):
Nicely done @lmac-1

elias-ba (approved):
Nice work @lmac-1

PR #4840, opened 2026-06-08
Run the collaboration test suite async, with deterministic document shutdown

stuartc:
## Description

This PR fixes a class of intermittent failures in the collaborative editing test suites and converts them from serial to async. Two changes make that possible: documents can now be shut down deterministically, and a collaboration supervision tree is no longer tied to one set of global process names.

Collaborative documents start under a global DynamicSupervisor and had no clean way to be stopped — they outlived the process that started them. In tests that meant a document's processes leaked past the test that created them and kept writing to the database after its SQL sandbox had checked in, crashing the *next* test with an `owner ... exited` error. Run order made it look random. The suites also had to run `async: false`, because every test shared one registry, one `:pg` scope and one dynamic supervisor.

**Deterministic shutdown**

- `Lightning.Collaborate.stop_document/1` — synchronous, idempotent teardown that runs the final persistence flush. The symmetric partner to `start_document/2`.
- `start_document/3` takes an optional `owner:` pid. The document tree monitors it and stops itself (`:normal`, so the flush runs and the transient child isn't restarted) when the owner exits. Defaults to `nil` — no monitor — so production documents still outlive the LiveView that starts them.

**Instance injection**

- `Lightning.Collaboration.Instance` describes the three names one tree owns: its registry, its dynamic supervisor and its `:pg` scope. `Instance.default/0` returns the existing production atoms, so production wiring is byte-identical.
- `Supervisor`, `Registry`, `DocumentSupervisor`, `Session`, `Persistence` and `PersistenceWriter` accept those names as defaulted leading arguments, so every existing call shape still means what it meant.
- Tests get an isolated tree per test via `start_collaboration_document/2,3` and a `start_collaboration_instance/0` helper, plus a `:collaboration_process_allow` test-only config callback that grants sandbox and Mox access to children the supervisor spawns.

Nine collaboration test files now run `async: true`, taking the repo from 50 serial test files to 43.

**A production bug this surfaced**

`Collaborate.start/2` decided whether to tear a document down on failure by checking `:pg` membership *before* starting it. Membership is registered inside `DocumentSupervisor.init/1`, so a caller racing another's start saw an absent document, took the create branch, and `start_document/4` masked the resulting `{:already_started, pid}` as a plain `{:ok, pid}`. Both callers then believed they owned the document, and a failed session start would tear down a document the other was editing. The created-or-found answer now comes from the start attempt itself, and other `start_child/2` errors pass through instead of raising `CaseClauseError`.

It also adds a developer guideline, `.claude/guidelines/testable-supervision-trees.md`, on writing supervision trees that stay addressable in tests without forcing the suite to run serially.

## Validation steps

1. `mix test test/lightning/collaboration/ test/lightning/collaborate_test.exs` — 162 tests, 0 failures, and the timing line should read `1.0s async, 0.00s sync`. These were previously serial and intermittently red.
2. `mix test test/lightning_web/channels/workflow_channel_test.exs test/lightning_web/channels/workflow_channel_broadcast_test.exs` — 149 tests, 0 failures, 1 skipped. These still use the global instance and were deliberately left serial.
3. Confirm production is untouched by default: `start_document/2` with no `owner` registers no monitor and the document behaves exactly as before.

1. Production behaviour is unchanged by default. Owner-monitoring is opt-in, nothing in `lib/` passes `owner:`, and `Instance.derive/1` special-cases the production supervisor name so the three atoms are identical to before.
2. Original flake evidence: a repeat harness gave 0/40 normal runs and 0/30 under simulated CI timing (`+S 4:4`, tight `assert_receive` timeout), against a pre-fix baseline of 11/40.
3. Since merging `main`, five full-suite runs at fixed seeds (5170 tests each) attributed zero failures to this branch. The failures those runs did surface are pre-existing on `main` and unrelated — mostly `capture_log` returning empty across six files, which looks like one shared log-capture problem worth its own issue.
4. Two limitations left in deliberately, both unreachable in production because nothing in `lib/` passes an owner: a document has a single owner slot, so a second caller's `owner` is silently discarded; and cross-node, two nodes can each start a SharedDoc for the same name, since the registry and dynamic supervisor are node-local.
5. `WorkflowReconciler` is not instance-aware — it resolves SharedDocs through the global `:pg` scope like the production save path — so `workflow_reconciler_test.exs` deliberately uses the default global instance. Its comments explain the constraint that keeps that safe.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [ ] I have implemented and tested all related **authorization policies** (N/A — no auth surface changed)


lib/lightning/collaboration.ex
```diff
@@ -52,45 +52,160 @@ defmodule Lightning.Collaborate do
...
+          instance :: Instance.t(),
+          workflow :: Lightning.Workflows.Workflow.t(),
+          document_name :: String.t(),
+          opts :: Keyword.t()
+        ) :: {:ok, pid()}
+  def start_document(%Lightning.Workflows.Workflow{} = workflow, document_name) do
+    start_document(Instance.default(), workflow, document_name, [])
+  end
+
+  def start_document(%Instance{} = instance, workflow, document_name) do
```
elias-ba, line 186: I think this is what dialyzer is unhappy about, though correct me if I'm reading it wrong. `start_document/3` has two clauses that differ by first arg, this instance-first one and the `(workflow, document_name, opts)` one below it, and there's a single arity-3 `@spec` (the workflow-first form), so the `start_document(instance, workflow, document_name)` call in `start/2` looks like it breaks the contract. One option would be to drop this instance-first clause and have the three callers pass opts explicitly as arity-4 (`start_document(instance, workflow, document_name, [])`); adding a fourth `@spec` would work too if you'd rather keep the clause. Curious which you'd prefer.

elias-ba (commented):
Hey @stuartc, I went through this and it looks really great 👏🏽 . I think I found what's behind the dialyzer failure and left a note inline, though you know this area far better than I do, so tell me if I've misread it.

elias-ba (approved):

elias-ba:
I went ahead and pushed a couple of commits here @stuartc: one adds the missing `@spec` so dialyzer is happy, and one merges main to clear the conflicts. Merging main meant your newer session tests (`reset_workflow/2` and the reconnect one) needed porting onto the owner-anchored allow setup you built, so I did that, and the collaboration suite runs green. If you'd rather handle any of that differently, just shout and I'll back it out.

lib/lightning/collaboration/document_supervisor.ex
```diff
@@ -11,32 +11,71 @@ defmodule Lightning.Collaboration.DocumentSupervisor do
...
+  child spec uses `restart: :transient`, a `:normal` stop is not restarted by
+  the DynamicSupervisor.
+
+  The default timeout must stay above what `terminate/2` can take, or a slow
+  flush makes this exit `:timeout` while the tree is still running and
+  unflushed — and `Collaborate.stop_document/2` reports that as `:ok`.
+  `terminate/2` bounds each child stop at 5s, so 15s leaves headroom over the
+  10s worst case.
+  """
+  def stop(pid, timeout \\ 15_000) when is_pid(pid) do
```
lmac-1, line 49: @stuartc I tweaked this to to 15s because `terminate/2` can take up to 10s (5s per child stop), so the old default could time out mid-flush while `stop_document/2` still reported `:ok.`

lmac-1 (approved):
I did a light /code-review and I'm happy with it. It came up with some unrelated improvements which aren't important here. I only paid attention to an improvement related to changes added in this branch and made the update myself directly (see inline comment). Happy for this to be merged if you're ok with that @stuartc

PR #4841, opened 2026-06-08
Fix workflows_pkey duplicate INSERT on collab reconnect, and remove its root cause (#4830)

### LINKED ISSUE #4830: Collaborative editor can INSERT an already-existing workflow id (workflows_pkey crash)
## Summary

Saving a workflow from the collaborative editor can attempt to **INSERT a workflow whose id already exists**, raising `Ecto.ConstraintError` on `workflows_pkey` and **crashing the collaborative Session GenServer** (`LightningWeb.WorkflowChannel`, `save_workflow`). It is **not** a UUID-generation collision — the pre-existing rows belong to the **same user** who hit the crash.

> **Fix direction (important):** the fix must **prevent the duplicate INSERT** — reconcile insert-vs-update by workflow id so an already-persisted workflow is UPDATEd, never re-INSERTed. **Do _not_ "fix" this by adding `unique_constraint(:id, name: "workflows_pkey")` to the `Workflow` changeset.** That only converts the crash into a validation/toast error while leaving the real bug in place.

## What we have verified

- **From the Sentry crash payloads (observed):** the crash is `Ecto.ConstraintError` on `"workflows_pkey" (unique_constraint)` "when attempting to insert struct", raised inside `save_workflow` on `LightningWeb.WorkflowChannel` (`workflow_channel.ex:355`), terminating the Session GenServer.
- **Same user, not a UUID collision (confirmed in prod):** the already-existing `workflows` rows whose ids collide belong to the **same user** who triggered the crashing save. So this is a workflow being re-INSERTed, not two independently-generated UUIDs colliding.
- **The changeset raises rather than erroring (code + runtime check):** `Workflow.validate/1` (`lib/lightning/workflows/workflow.ex`) declares `unique_constraint([:name, :project_id])` but **not** `unique_constraint(:id, ...)`. Confirmed via `project_eval`: running `save_workflow` with a `:built` `%Workflow{}` that carries an already-persisted id raises `Ecto.ConstraintError` on `workflows_pkey` (instead of returning a changeset error). This is a property of the save path given such a struct — it does not establish how production produced that struct.

## What we have NOT verified (hypotheses / code-level theory)

- **The code path that produces the INSERT.** Reading the code, the only way `save_workflow` INSERTs is when `state.workflow` is a `:built` struct with `lock_version` nil/0; `load_workflow("new", id, ...)` (`workflow_channel.ex`) constructs exactly such a struct, and `fetch_workflow/1` (`lib/lightning/collaboration/session.ex`) passes it through to an INSERT. **This is inference from the code, not confirmed as the production path.**
- **That `state.workflow` was `built+nil` in the incidents** — inferred from the fact that an INSERT occurred, not directly observed.
- **How a session would join `action: "new"` for an id that already exists** — completely open. Candidate triggers (all unconfirmed): socket reconnect / lost save reply leaving the client on `/w/new`; cross-node failover (cf. `LIGHTNING-1D5` `nodedown`/`unobserve`); the new-workflow **Import** path carrying an explicit `id`; stale client state (bfcache / y-indexeddb).
- **Reproduction.** We have **not reproduced** the bug. Manual UI testing only ruled out some *non-triggers*: the happy path (create → navigate to `/w/<id>`, `action: "edit"`), a single-session double-click (the GenServer serializes the two saves; after the first INSERT `state.workflow` becomes `:loaded`, so the second is an UPDATE), and browser-back (history is replaced).

## Production evidence (Sentry `LIGHTNING-1AF` + `LIGHTNING-1D5`, observed payloads)

4 events total (`workflows_pkey` on INSERT), all non-AI (no template placeholders). 1AF/1D5 appear to be duplicate groupings of the same crash site.

| Date (UTC) | Release | User  | Workflow | Payload doc state |
|---|---|---|---|---|
| 2026-05-25 10:21:21 | v2.16.5 | External User A`9feb5c9a-7141-4d5e-a8c9-e0c041f2b2cb`  | `c80ebf63-7b05-455e-a2d7-222baacd5704` | "Untitled workflow", empty, `lock_version: nil` (join_ref 464) |
| 2026-05-25 10:21:09 | v2.16.5 | External User A `9feb5c9a-7141-4d5e-a8c9-e0c041f2b2cb` | `c80ebf63-…` (same id) | "Transform data", populated, `lock_version: 3` (join_ref 327) |
| 2026-05-11 13:18:24 | v2.16.3 | External User B `e6de35ea-effb-4b19-a4eb-5a13ccff370a`  | `ddeef56a-e864-4a67-8da5-555d19739348` | "Get data", `lock_version: 2` |
| 2026-05-07 14:37:05 | v2.16.2 | mtuchi dev `4baaf510-2594-4a3a-aada-db4b64bd9a34`  | `d1f30777-594a-410f-bb0e-c3ddeaac387a` | "Transform data", `lock_version: 5` |

Observed notes (not yet explained):
- For `c80ebf63`, two events 12s apart with **different `join_ref`** → two separate channel joins for one id (one with an empty doc, one populated). A double-save within a single session is therefore not sufficient on its own.
- One event's payload was a fully populated doc (`lock_version: 3`) yet the crash was still an INSERT — interesting and unexplained; consistent with (but not proof of) the session holding a `built+nil` struct independent of doc contents.

## Suggested fix area (behavioural, not a changeset constraint)

- `load_workflow("new", id, ...)` should detect an existing row for `id` and load it (switch to the edit/update path) instead of constructing a `built+nil` struct; and/or
- `fetch_workflow/1` should reconcile by id against the DB so `insert_or_update` never INSERTs an id that already exists.

## Related

- #4816 — parent (Session crashes on `save_workflow`).
- #4828 — AI placeholder ids (separate, unrelated cause).

stuartc:
## Description

This fixes the `workflows_pkey` duplicate-key crash you'd sometimes hit when reconnecting to the collaborative editor after a save — and then goes after the thing that was actually causing it.

The root cause was that we were resolving "which workflow am I editing, and is it new or existing?" in two completely separate places: once when the channel joins (`load_workflow/5`) and again on the session save/reset path (`fetch_workflow/1`). Those two could quietly disagree about whether an id should map to a `:built` struct (→ INSERT) or a `:loaded` one (→ UPDATE). So a reconnect that still thought it was "new" would try to INSERT a row that already existed → boom.

So there are two parts here:

1. The immediate fix — the reconnect path heals the join action and reconciles to the saved row, so the next save UPDATEs instead of colliding.
2. The real fix — I pulled all of that resolution logic into one place, `Lightning.Collaboration.WorkflowResolver`. Both callers now go through it, so they physically can't disagree any more. It also hands back an explicit `kind` (`:new | :existing | :version`) so the channel stops trying to guess "is this new?" from the shape of the struct. `fetch_workflow/1` and the old `NotLoaded` poking are gone.

The nice property that closes the bug: a `"new"` join for an id that already has a row resolves to the existing `:loaded` row (`kind: :existing`), so the save routes to UPDATE. No more duplicate INSERT.

Closes #4830

## Validation steps

1. Open a workflow in the collab editor, make a change, save it.
2. Force a reconnect (drop the socket / refresh the tab) and save again. Before this, that could blow up with a `workflows_pkey` duplicate-key error — now it should just save cleanly.
3. Sanity-check the three normal flows still work: creating a brand-new workflow, editing an existing one, and viewing an older version.
4. `mix test test/lightning/collaboration/ test/lightning_web/channels/workflow_channel_test.exs` — should be green (273 tests here).

A few things worth your eyes:

- The bit that actually closes the bug is the reconcile-by-id behaviour — a `"new"` action on an id that already has a row comes back as `:existing`. There's a resolver test pinning this specifically, since getting it wrong would bring the duplicate-INSERT straight back.
- Auth ordering is unchanged: `"new"` still authorises first, `"edit"`/version still resolve first. Authorisation stays in the channel; the resolver only does the project-ownership check, and only when it's handed a `:project`.
- One small tightening: a `"new"` rejoin for an id that belongs to *another* project now returns a wrong-project error instead of silently building a fresh struct. Felt like the safer behaviour.
- I kept the `#4829` `Ecto.ConstraintError` rescue in `workflows.ex` as a backstop — didn't want to remove the belt while adding the braces.
- While moving things around, the reset path briefly lost its "this workflow was deleted" error — I put that back (resetting a soft-deleted workflow returns `workflow_deleted` again, same as before), with a test.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@ and this project adheres to
 
 ### Fixed
 
+- Fix a `workflows_pkey` duplicate-key crash when reconnecting to the
+  collaborative editor after a save. Workflow resolution is now centralised in a
+  single `Lightning.Collaboration.WorkflowResolver`, so the channel join and the
+  session save path can no longer disagree on whether an id should INSERT or
+  UPDATE. [#4830](https://github.com/OpenFn/lightning/issues/4830)
+
 ## [2.16.7] - 2026-06-04
 
 ### Changed
--- a/assets/js/collaborative-editor/contexts/SessionProvider.tsx
+++ b/assets/js/collaborative-editor/contexts/SessionProvider.tsx
@@ -13,6 +13,7 @@ import {
   useCallback,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from 'react';
 
@@ -34,6 +35,17 @@ const logger = _logger.ns('SessionProvider').seal();
 interface SessionContextValue {
   sessionStore: SessionStoreInstance;
   isNewWorkflow: boolean;
+  /**
+   * Reports the live "new workflow" status up to the SessionProvider so the
+   * channel-join `action` stays honest across in-place reconnects.
+   *
+   * The SessionContextStore (the source of truth, cleared by
+   * `clearIsNewWorkflow()` after the first successful save) is created in
+   * StoreProvider, which is a *child* of SessionProvider — so SessionProvider
+   * cannot read it directly. A small bridge inside StoreProvider subscribes to
+   * `useIsNewWorkflow()` and calls this to keep the join-param action current.
+   */
+  setIsNewWorkflow?: (isNewWorkflow: boolean) => void;
   initialRunData?: string; // JSON-encoded RunStepsData from server
 }
 
@@ -79,12 +91,27 @@ export const SessionProvider = ({
     [version, workflowId]
   );
 
-  const joinParams = useMemo(
+  // Track the live "new workflow" status in a ref so the channel-join `action`
+  // is read at connect/reconnect time rather than frozen at mount.
+  //
+  // Seeded from the `isNewWorkflow` prop (LiveView `data-is-new-workflow`) so
+  // the very first join is correct before any save. After the first successful
+  // save, the SessionContextStore flag is cleared and `setIsNewWorkflow(false)`
+  // is reported up via the SessionContext bridge, so a subsequent in-place
+  // reconnect rejoins with `action: "edit"` instead of the stale "new".
+  const isNewWorkflowRef = useRef(isNewWorkflow);
+
+  const setIsNewWorkflow = useCallback((next: boolean) => {
+    isNewWorkflowRef.current = next;
+  }, []);
+
+  // Stable getter: read lazily so reconnects pick up the current action.
+  const getJoinParams = useCallback(
     () => ({
       project_id: projectId,
-      action: isNewWorkflow ? 'new' : 'edit',
+      action: isNewWorkflowRef.current ? 'new' : 'edit',
     }),
-    [projectId, isNewWorkflow]
+    [projectId]
   );
 
   // Handle roomname changes (version switching)
@@ -138,7 +165,7 @@ export const SessionProvider = ({
     isConnected,
     sessionStore,
     roomname,
-    joinParams,
+    getJoinParams,
     onError: handleProviderError,
     onProviderReady: handleProviderReady,
     onProviderReconnected: handleProviderReconnected,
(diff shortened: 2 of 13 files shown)


midigofrank (approved):
Nicely done 🙌

Issue #4163, opened 2025-12-09
Why Are Workflows Not Permanently Deleted?


taylordowns2000:
# Why Are Workflows Not Permanently Deleted?

## Summary

Workflows use a different deletion pattern than users, credentials, projects,
and webhook auth methods. When a workflow is "deleted," it's soft-deleted
immediately (`deleted_at` set to now), but **never permanently purged**. This
contrasts with other entities that use a `scheduled_deletion` field and are
eventually hard-deleted by cron jobs.

🚨⭐️ **Unless there is some special reason that this was done differently, we should do it the standard way.** ⭐️🚨

## Current Behavior

Related to #511 (implemented in https://github.com/OpenFn/lightning/pull/605) and #1670 

### Workflows: Soft Delete Only

When you delete a workflow (`Workflows.mark_for_deletion/3` in
`lib/lightning/workflows.ex:459-472`):

```elixir
Workflow.request_deletion_changeset(workflow, %{
  "deleted_at" => DateTime.utc_now()
})
```

- Sets `deleted_at` to current timestamp **immediately**
- Disables all triggers on the workflow
- Creates an audit record
- **No scheduled purge** - the workflow stays in the database forever

The workflow is filtered out of queries via `where: is_nil(w.deleted_at)`:

- `Workflows.get_workflows_for/2` (line 389)
- `Provisioner.preload_dependencies/2` (line 297)
- Project schema association:
  `has_many :workflows, Workflow, where: [deleted_at: nil]` (line 50)

### Other Entities: Scheduled Deletion + Purge

Users, credentials, projects, and webhook auth methods use a two-phase approach:

1. **Schedule deletion**: Set `scheduled_deletion` to a future date (based on
   `PURGE_DELETED_AFTER_DAYS` env var)
2. **Cron purge**: Daily jobs permanently delete records past their scheduled
   date

From `lib/lightning/config/bootstrap.ex:251-260`:

```elixir
cleanup_cron =
  if Application.get_env(:lightning, :purge_deleted_after_days) > 0,
    do: [
      {"15 2 * * *", Lightning.Accounts, args: %{"type" => "purge_deleted"}},
      {"30 2 * * *", Lightning.Credentials, args: %{"type" => "purge_deleted"}},
      {"45 2 * * *", Lightning.Projects, args: %{"type" => "purge_deleted"}},
      {"0 3 * * *", Lightning.WebhookAuthMethods, args: %{"type" => "purge_deleted"}}
    ],
    else: []
[...]

PR #4842, opened 2026-06-08
Sandboxes: address remaining merge failures

elias-ba:
## Description

This builds on #4835 and covers the cases where merging a sandbox could still fail or quietly drop something. Three changes:

- **Free up a deleted workflow's name on merge.** When a merge deletes a workflow, it now renames it (the same way deleting from the workflows list already does) so the name is free to reuse. A leftover name was the remaining cause of the "validation error" people saw on merge.
- **Stop swallowing merge failures.** A failed merge used to show a bare "validation error" and leave no trace, so we only learned about one when someone reported it. Now the user sees "Couldn't merge this sandbox. Please try again.", and the full reason is logged and sent to Sentry, so we find out about failures as they happen and can fix the cause instead of waiting on a report. The one exception is hitting a plan limit, which the user sees directly and which we don't treat as an error.
- **Choose which sandbox credentials to carry over.** The merge dialog now lists the credentials that exist only in the sandbox under "Credentials to add", with per-credential toggles and a select-all. The ones left selected are attached to the target as part of the merge, so they survive instead of being silently dropped; the ones you deselect are left behind. See screenshot
Screenshot 2026-06-10 at 16 52 00

## Validation steps

1. In a sandbox, create a credential and attach it to a step. Open the merge dialog and confirm it appears under "Credentials to add". Merge with it selected and confirm it is attached to the target; on a fresh sandbox, deselect it and confirm it is not.
2. Delete a workflow on the parent through a merge, then merge again recreating a workflow with the same name. It should succeed (before, this failed).
3. Make a merge fail and confirm the flash shows the try-again message rather than a raw "validation error".

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning_web/live/sandbox_live/index.ex
+++ b/lib/lightning_web/live/sandbox_live/index.ex
@@ -990,15 +990,47 @@ defmodule LightningWeb.SandboxLive.Index do
   end
 
   defp format_merge_error(%Ecto.Changeset{} = changeset) do
-    changeset.errors
-    |> List.first()
-    |> case do
-      {field, {message, _}} -> "#{field}: #{message}"
-      _ -> "Failed to merge: validation error"
+    case first_changeset_error(changeset) do
+      {path, message} -> "Failed to merge: #{path}: #{message}"
+      nil -> "Failed to merge: validation error"
     end
   end
 
   defp format_merge_error(%{text: text}), do: text
   defp format_merge_error(reason) when is_binary(reason), do: reason
   defp format_merge_error(reason), do: "Failed to merge: #{inspect(reason)}"
+
+  # Finds the first error anywhere in the changeset, including errors nested in
+  # associations (e.g. a workflow's name or a job's credential). Without this,
+  # a nested error has no top-level entry and the merge screen would only show a
+  # generic "validation error" with no indication of the real cause.
+  defp first_changeset_error(%Ecto.Changeset{} = changeset) do
+    changeset
+    |> Ecto.Changeset.traverse_errors(&interpolate_error_message/1)
+    |> first_nested_error([])
+  end
+
+  defp first_nested_error(errors, path) when is_map(errors) do
+    Enum.find_value(errors, fn {field, value} ->
+      first_nested_error(value, [to_string(field) | path])
+    end)
+  end
+
+  defp first_nested_error(values, path) when is_list(values) do
+    Enum.find_value(values, &first_nested_error(&1, path))
+  end
+
+  defp first_nested_error(message, path) when is_binary(message) do
+    {path |> Enum.reverse() |> Enum.join("."), message}
+  end
+
+  defp first_nested_error(_other, _path), do: nil
+
+  defp interpolate_error_message({message, opts}) do
+    Regex.replace(~r/%\{(\w+)\}/, message, fn _whole, key ->
+      opts
+      |> Keyword.get(String.to_existing_atom(key), key)
+      |> to_string()
+    end)
+  end
 end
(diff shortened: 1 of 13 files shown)


lib/lightning_web/live/sandbox_live/index.ex
```diff
@@ -989,16 +989,17 @@ defmodule LightningWeb.SandboxLive.Index do
...
 
-  defp format_merge_error(%Ecto.Changeset{} = changeset) do
-    changeset.errors
-    |> List.first()
-    |> case do
-      {field, {message, _}} -> "#{field}: #{message}"
-      _ -> "Failed to merge: validation error"
-    end
+  defp format_merge_error({:workflow_name_conflict, name}) do
+    ~s(Couldn't merge: a workflow named "#{name}" already exists in the target project. Rename it and merge again.)
```
josephjclark: huh. Why is this an error? Sandboxes have identically named workflows all the time right? I am merging workflow A into workflow A.

Is this just a wording problem?

josephjclark:
> Keep a sandbox credential through a merge. A credential created in a sandbox is now added to every ancestor project, not just the root, so it survives a merge into any of them.

So this was something I wanted to avoid on the base PR. It felt a little bit icky to auto-assign credentials right up the hierarchy? Probably fine but I didn't want to do it without more thought. That's why I fixed it in the UI and only added the root project. It gives the user more control.

I think we should spin this out into a clean issue. If we're going to auto-assign to the whole stack, we ought to consider:
- if we add the credential, then remove it before merging, it'll still be attached further up the stack. How and when would we purge the credential?
- We should drop the "fix it in the UI" of the base project. It's done more thoroughly here, but also the UI isn't entirely accurate any more because it doesn't includes middle ancestors
- Are there any privacy concerns here? If I add a dev credential to the sandbox - maybe something with oauth - do I really want to assign that to the merge target? Again, the UI gives users control. This is a wider question over credentials and I think there's an epic to get into it. I wanted to sidestep all that.

josephjclark (changes requested):
@elias-ba I like `Free up a deleted workflow's name on merge`. I like the idea of the better validation errors but in practice I don't think we're adding a lot of value? And I'm concerned about  `Keep a sandbox credential through a merge` - I really don't think that's the right play :(

lib/lightning_web/live/sandbox_live/index.ex
elias-ba: This message is misleading, sorry about that. It was actually trying to capture the soft-deleted workflow problem, where a workflow deleted by an earlier merge kept its name, so the next merge that reused that name tripped the constraint. Merging A into A was never the issue.

After our chat yesterday this is no longer a concern. We decided not to patch the soft-deleted workflows already sitting on prod, and to fix the problem at the root instead by implementing #4163.

So this is now changed. In this PR, a workflow deleted during a merge gets renamed the same way the delete path in the workflows list already does, which frees the name and fixes the cause going forward. The failure message is now just a generic "Couldn't merge this sandbox. Please try again.", and every merge failure going through that path is also logged in Sentry so we can observe what's actually happening.

elias-ba:
@josephjclark ready for another look when you get a chance. What changed since your review:

- Dropped the misleading name-conflict message. Merge failures now show a generic try-again message and are logged to Sentry. Root cause is fixed by renaming workflows on merge-delete (broader lifecycle fix tracked in #4163, and we're not patching the existing prod data).
- Credentials are no longer auto-assigned up the hierarchy. The merge dialog now has a "Credentials to add" picklist, so the user chooses which sandbox-only credentials carry over.

josephjclark (approved):
Love it, thanks Elias!

Small tweaks to make the credential merge text look like this

Image


I've tested that credentials merge OK _and_ that you can deselect credentials to not merge them

PR #4845, opened 2026-06-10
Tolerate slow or unreachable cross-node SharedDoc during session cleanup

### LINKED ISSUE #4817: Collaborative editor: cross-node SharedDoc unobserve crashes Session/channel on node disconnect or timeout (LIGHTNING-1D5 + LIGHTNING-1A5)
**Describe the bug**

In the collaborative editor, the Session GenServer makes a cross-node
`GenServer.call(…, {:unobserve, …}, 5000)` against the SharedDoc on
another BEAM node (via `SharedDoc.unobserve/1`, which expands to
`GenServer.call(server, {:unobserve, self()})` with the default 5s
timeout). When that node is briefly unreachable (call times out) or
has disconnected (`nodedown`), the call crashes the Session GenServer.
Depending on which process is observing the crash, this surfaces as
either the Session terminating during cleanup, or the Phoenix channel
coming down mid-save and the client seeing the editor disconnect.

This is a separate failure class from #4816, which covers Ecto
constraint / `ChangeError` crashes in the save flow.

**Tracked in Sentry as two groupings of the same root cause:**

- [LIGHTNING-1D5](https://openfn.sentry.io/issues/7219642618/) — the
  **channel-side** view. The Phoenix channel is blocked in
  `GenServer.call(Session, {:save_workflow, …}, 10000)` at
  `lib/lightning_web/channels/workflow_channel.ex:355` when the Session
  it called dies on the nested `unobserve`, so the EXIT propagates up
  and takes the channel down. ~22 events, mostly `(EXIT) time out`,
  some `no connection`. (One event also grouped into
  [LIGHTNING-1AF](https://openfn.sentry.io/issues/7182438175/).)

- [LIGHTNING-1A5](https://openfn.sentry.io/issues/7132840920/) — the
  **Session-side** view. The Session crashes in `terminate/2` while
  running the same cross-node `unobserve`. 44 events (first seen
  2025-12-22, last 2026-05-21). Of 27 events sampled, **27/27 crash at
  `lib/lightning/collaboration/session.ex:144` in
  `Session.terminate/2`**, and **27/27 are `(EXIT) no connection`
  (nodedown), zero timeouts**. The remote PID's first segment is always
  a non-zero node id (e.g. `#PID<79283.237392.0>`), confirming the
  SharedDoc lives on another node. Only 4/27 events have a
  `Last message: {:save_workflow, …}`; the rest have none, i.e. the
  Session was terminating for ordinary reasons (parent/channel went
  away) and crashed during cleanup, not mid-save.

The two are the same termination seen from different processes: while a
`save_workflow` call is open, the Session runs `terminate/2` →
`unobserve` → times out or hits a dead node and crashes; 1A5 logs it
from the Session, 1D5 logs the channel that inherited the exit.

**Confirmed `unobserve` call sites**

There are exactly two `SharedDoc.unobserve/1` sites in
`lib/lightning/collaboration/session.ex`:

- **`session.ex:144`** — `terminate/2` (the site LIGHTNING-1A5 pins
  down empirically across all sampled events).
- **`session.ex:429`** — the `{:DOWN, parent_ref, …}` handler.

**Sub-shapes observed across both groupings**

| Sub-case | Where | Shape |
|---|---|---|
| `(EXIT) no connection to lightning@10.x.x.x` | `terminate/2` @ `session.ex:144` (1A5) | Remote SharedDoc node disconnected; all 27 sampled 1A5 events. |
| `(EXIT) time out` on `GenServer.call(#PID<remote>, {:unobserve, …}, 5000)` | nested under the blocked `save_workflow` call (1D5) | Remote SharedDoc slow to reply within 5s; majority of 1D5 events. |
| `(EXIT) no connection` under the blocked `save_workflow` call | (1D5) | Remote node disappeared mid-call; minority of 1D5 events. |

**Version number** Observed on v2.15.16, v2.16.0–v2.16.6.

- [ ] Yes
- [x] No

**To Reproduce**

Not reproduced locally. Sentry shows the failing cross-node
`GenServer.call(…, {:unobserve, …}, 5000)` for each event, and (for the
1D5 grouping) the inbound `save_workflow` payload.

**Expected behaviour**

A slow or briefly-unreachable SharedDoc must not crash the Session
GenServer or the Phoenix channel. The cross-node `unobserve` during
termination/cleanup should tolerate the remote being slow or gone
(node down should be a no-op, not a crash), and the save flow should
surface a structured error or trigger a clean reset/rejoin rather than
a crash.

**Additional context**

- The Ecto-constraint failure class in the save flow is tracked in #4816.
- `.claude/guidelines/yex-guidelines.md §Transaction Deadlock Rules`
  may be relevant — Y.Doc transactions on the BEAM have non-obvious
  deadlock hazards.

midigofrank:
## Description

The collaborative editor Session calls SharedDoc.unobserve/1 during cleanup, which issues a cross-node GenServer.call to a SharedDoc that may live on another node. When that node is unreachable (:noconnection) or slow to reply (:timeout), the call exited and crashed the Session; a Phoenix channel blocked in save_workflow then inherited the exit and dropped the client's editor connection.

Wrap both unobserve call sites (terminate/2 and the parent :DOWN handler) in safe_unobserve/1, which catches the exit and logs at warning instead of crashing. Skipping the unobserve is safe: the SharedDoc monitors each observer and runs the same cleanup from its own :DOWN handler when the Session dies.

Closes #4817

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,16 @@ and this project adheres to
 - Consolidated run and work order state definitions into single source of truth
   by adding `Run.active_states/0`, `WorkOrder.states/0`, and
   `WorkOrder.active_states/0` and replacing all hardcoded state lists across the
-  codebase
-  [#4589](https://github.com/OpenFn/lightning/issues/4589)
+  codebase [#4589](https://github.com/OpenFn/lightning/issues/4589)
 
 ### Fixed
+
+- Stop the collaborative editor's Session (and the Phoenix channel calling it)
+  from crashing when the cross-node `SharedDoc.unobserve/1` during cleanup hits
+  a SharedDoc on a node that is unreachable (`:noconnection`) or slow to reply
+  (`:timeout`); the failed unobserve is now tolerated as a no-op since the
+  SharedDoc cleans up observers via its own monitor.
+  [#4817](https://github.com/OpenFn/lightning/issues/4817)
 
 ## [2.16.7] - 2026-06-04
 
--- a/lib/lightning/collaboration/session.ex
+++ b/lib/lightning/collaboration/session.ex
@@ -137,11 +137,11 @@ defmodule Lightning.Collaboration.Session do
       Process.demonitor(state.parent_ref)
     end
 
-    # Don't check Process.alive? - it only works for local PIDs
-    # and shared_doc_pid can be on another node in a distributed cluster.
-    # Sending to a dead process is safe (message is discarded).
+    # Don't check Process.alive? - it only works for local PIDs and
+    # shared_doc_pid can be on another node in a distributed cluster.
+    # safe_unobserve/1 tolerates the remote being slow or gone (see #4817).
     if shared_doc_pid do
-      SharedDoc.unobserve(shared_doc_pid)
+      safe_unobserve(shared_doc_pid)
     end
 
     Presence.untrack_user_presence(
@@ -151,6 +151,17 @@ defmodule Lightning.Collaboration.Session do
     )
 
     :ok
+  end
+
+  defp safe_unobserve(shared_doc_pid) do
+    SharedDoc.unobserve(shared_doc_pid)
+  catch
+    :exit, reason ->
+      Logger.warning(
+        "SharedDoc.unobserve skipped during session cleanup: #{inspect(reason)}"
+      )
+
+      :ok
   end
 
   def lookup_shared_doc(document_name) do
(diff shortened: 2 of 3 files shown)


midigofrank:
@stuartc , I'm merging this in. There is a follow up issue to handle the cause of the crash https://github.com/OpenFn/lightning/issues/4877 . Also, the failing lint check isn't related to this PR:  https://github.com/OpenFn/lightning/issues/4878

PR #4855, opened 2026-06-15
Add socket-level DB connection resilience

stuartc:
## Description

This PR _changes_ the `Lightning.Repo` database connection so that, on Linux, a
connection whose network path has died fails fast instead of blocking on a dead
socket.

The (default) 15s DBConnection `timeout` guards a slow-but-live query on a healthy
connection, but it does **not** cover a dead network path: if the route to the
database is silently severed (e.g. a TCP reset during a Kubernetes node-pool
upgrade), the BEAM can block on the dead socket until the kernel's
retransmission budget expires (~15 min). We saw this take down an Oban
Scheduler job that was stuck in `executing` after a node in k8s was drained; taking the network down underneath a running instance.

On Linux, this sets the `TCP_USER_TIMEOUT` socket option, which caps how long
transmitted data may go unacknowledged before the socket is dropped:

- **`DATABASE_TCP_USER_TIMEOUT`** (ms) defaults to `DATABASE_TIMEOUT + 5s` (20s),
  so the application query timeout fires first on a connection that is merely
  slow. Set it to override, or `0` to restore the previous behaviour. A value
  below `DATABASE_TIMEOUT` logs a warning at boot. Ignored on non-Linux
  platforms.

It does not abort a live query: `tcp_user_timeout` measures unacknowledged data,
not query duration, so a long read whose bytes were acked at send time never
accumulates against it. A dead idle pooled connection is caught lazily on its
next use within the same budget.

This is on by default because the behavior it replaces — blocking ~15 min on a
dead socket — is a latent failure mode every Linux deployment inherits, and the
deployments most exposed (rolling node upgrades, cloud NAT/LB resets) are the
common shape. The `0` escape hatch and the Linux gate keep it safe.

A server-side `statement_timeout` was considered and dropped: it shares the
single Repo pool with migrations and the retention purge (which already requests
up to 150s), so a global ceiling is unsafe without per-transaction overrides we
don't yet have.

## Validation steps

1. `mix test test/lightning/config/bootstrap_test.exs` — covers the
   `tcp_user_timeout` socket option, the `ECTO_IPV6` coexistence case (the prod
   block appends rather than clobbers `socket_options`), the
   below-`DATABASE_TIMEOUT` boot warning, and the SSL toggle.
2. Boot with `DATABASE_TCP_USER_TIMEOUT=20000` on Linux and confirm the Repo
   `socket_options` carry the raw `{:raw, 6, 18, ...}` entry; on macOS confirm it
   is omitted.

1. The prod block in `configure/0` re-sets `socket_options` for the `ECTO_IPV6`
   toggle; it now **appends** the resilience option (`maybe_ipv6 ++
   db_socket_options`) rather than clobbering it. A unit test guards this.
2. Default-on is a deliberate behavior change for Linux deployments, hence the
   CHANGELOG entry sits under **Changed**, not **Added**. Deploy-side tuning is a
   separate decision and lives in the deployment config, not this repo.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)
- [x] I have implemented and tested all related **authorization policies**
      (n/a — config-only change, no new data access)


CHANGELOG.md
```diff
@@ -25,6 +25,12 @@ and this project adheres to
 
 ### Changed
 
+- On Linux, database connections now fail fast when their network path dies
+  (e.g. a TCP reset during a node-pool upgrade) instead of blocking on a dead
+  socket for up to ~15 minutes. `DATABASE_TCP_USER_TIMEOUT` (ms) defaults to
+  `DATABASE_TIMEOUT + 5s`; set it to `0` to restore the previous behaviour. No
```
rorymckinley, line 31: @stuartc The wording here and in DEPLOYMENT.md can be interpreted as "it is ok for this to be set on non-Linux systems as it is a no-op". I have not found any documentation to confirm this, so perhaps it may be best to reword it to indicate that Lightning will not apply this setting on non-Linux hosts, to better align with the code?

rorymckinley:
@stuartc I blew my timebox trying to find non-clanker sources on setting raw socket values. TCP socket programming is a deep rabbit hole and I have no prior knowledge to lean on - so chasing it further is probably not cost-effective. The rest of the PR looks fine - I made a comment relating to a cosmetic documentation change.

rorymckinley (approved):

stuartc:
> @stuartc I blew my timebox trying to find non-clanker sources on setting raw socket values. TCP socket programming is a deep rabbit hole and I have no prior knowledge to lean on - so chasing it further is probably not cost-effective. The rest of the PR looks fine - I made a comment relating to a cosmetic documentation change.

Yeah I'm going to make a test harness to put this through it's paces on a linux vm before merging.

So **let not merge** until I've done that. And yeah that language needs another pass. It shouldn't read "previous" either - that ages fast.

midigofrank (approved):
Nice. Great learnings here

PR #4881, opened 2026-06-18
Reply cleanly on unhandled credential errors in run channel

### LINKED ISSUE #4853: RunChannel.handle_credential_error/5 has no fallback clause for unexpected OAuth errors
## Summary

`LightningWeb.RunChannel.handle_credential_error/5` (`lib/lightning_web/channels/run_channel.ex:482-552`) pattern-matches a fixed set of error tags but has no catch-all clause. An unexpected error term raises `FunctionClauseError`, which crashes the run channel during `fetch:credential` instead of returning a clean `{:error, ...}` reply to the worker.

## Handled cases

The function only has clauses for these five tags:

- `:environment_not_configured`
- `:project_not_found`
- `:environment_mismatch`
- `:reauthorization_required`
- `:temporary_failure`

## The unhandled case

`Lightning.Credentials.Resolver.resolve_credential/2` can return an open-ended error shape that none of those clauses match.

In `lib/lightning/credentials.ex:1205-1213`, when an OAuth token refresh fails with anything **other than**:

- `invalid_grant` / HTTP 400-401 → mapped to `:reauthorization_required`
- HTTP 429-503 → mapped to `:temporary_failure`

…it falls through to a raw pass-through:

```elixir
{:error, error} ->
  # audit ...
  {:error, error}   # raw, untyped error term
```

That raw `error` is wrapped by the resolver (`lib/lightning/credentials/resolver.ex:114`) into `{:error, {error, credential}}` and handed to `handle_credential_error/5` as the error tuple. The first element is then an arbitrary term — e.g. `%{status: 500}`, `:timeout`, a `Mint.TransportError`, connection-refused, etc.

Since none of the five clauses match and there is no fallback clause, this raises `FunctionClauseError` and crashes the channel.

## Impact

Any OAuth-provider failure that is **not** a 429/503 or an auth-revocation — most notably 5xx responses, network/transport errors, and timeouts — crashes the run channel mid-`fetch:credential`. The worker loses the connection rather than receiving a clean error reply.

## Code references

- `lib/lightning_web/channels/run_channel.ex:482-552` — `handle_credential_error/5` clauses
- `lib/lightning_web/channels/run_channel.ex:154-155` — call site forwarding `{:error, error_tuple}`
- `lib/lightning/credentials/resolver.ex:114` — wraps reason into `{reason, credential}`
- `lib/lightning/credentials.ex:1205-1213` — raw pass-through of unknown OAuth refresh errors

## Note on a non-issue

The resolver's `%Credential{}` clause can surface a raw `:environment_not_found` (`resolver.ex:99`), but this channel only ever calls `resolve_credential(run, id)` with a `%Run{}` (`run_channel.ex:144`). On the `%Run{}` path that tag is remapped to `:environment_mismatch` (`resolver.ex:110`), so it never reaches `handle_credential_error/5` here.

midigofrank:
## Description

RunChannel.handle_credential_error/5 pattern-matched a fixed set of error tags with no fallback, so an unexpected error term — most notably an OAuth provider timeout during token refresh, which surfaces as an untyped transport error — raised FunctionClauseError and crashed the channel mid-fetch:credential, dropping the worker's connection.

Add a fallback clause that logs the unexpected term and replies with a clean error so the worker receives a response instead of losing the connection.

Closes #4853

## Validation steps

1. _(How can a reviewer validate your work?)_

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ and this project adheres to
 
 ### Fixed
 
+- Stop the run channel from crashing during `fetch:credential` when an OAuth
+  provider times out while refreshing a token.
+  [#4853](https://github.com/OpenFn/lightning/issues/4853)
 - Stop the collaborative editor's Session (and the Phoenix channel calling it)
   from crashing when the cross-node `SharedDoc.unobserve/1` during cleanup hits
   a SharedDoc on a node that is unreachable (`:noconnection`) or slow to reply
--- a/lib/lightning/credentials/resolver.ex
+++ b/lib/lightning/credentials/resolver.ex
@@ -55,6 +55,7 @@ defmodule Lightning.Credentials.Resolver do
           | :project_not_found
           | :environment_mismatch
           | Credentials.oauth_refresh_error()
+          | term()
 
   @type resolve_error :: {error_reason(), Credential.t() | nil}
 
--- a/lib/lightning_web/channels/run_channel.ex
+++ b/lib/lightning_web/channels/run_channel.ex
@@ -549,4 +549,15 @@ defmodule LightningWeb.RunChannel do
     {:reply, {:error, "Could not reach the OAuth provider. Try again later"},
      socket}
   end
+
+  # Fallback for unexpected error terms
+  defp handle_credential_error(socket, error, id, _project_id, run_id) do
+    Logger.warning(
+      "Unhandled credential resolution error for credential #{id} " <>
+        "(run #{run_id}): #{inspect(error)}"
+    )
+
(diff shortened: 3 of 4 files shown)


midigofrank:
@elias-ba I'm merging this in. I know you were interested in some metrics? I haven't added any, this change simply stops the run channel from crashing when it receives an unknown error from the `Credentials.resolve_credential_body/2`. There are some audits in place for tracking these errors: https://github.com/OpenFn/lightning/blob/main/lib/lightning/credentials/audit.ex#L80

PR #4888, opened 2026-06-23
Global Chat: Receive job code for multiple steps

### LINKED ISSUE #4890: Global assistant: Receive several job code step changes
Lightning should only receive full YAMLs with job code embedded from the Global Assistant on Apollo. This will enable viewing and applying several job code step changes. More details here https://github.com/OpenFn/apollo/issues/507

hanna-paasivirta:
## Description

This PR lets Lightning receive job code for multiple steps at once from the Apollo AI server when using global chat. Before, the backend tried to extract a single job's code from the response. Now global chat always receives the full workflow YAML with every step body embedded. That means the assistant can change several steps and Lightning can apply all of them in one go.

Only global chat is affected. Job chat and workflow chat are unchanged.

Closes #4890 

**This PR will break the Global Assistant. The corresponding changes on Apollo need to be released soon after this one** See: https://github.com/OpenFn/apollo/pull/547 

### Implementation details
- On the backend, `build_global_message` no longer tries to pick out one job's code. It now always reads the `workflow_yaml` attachment and stores the full YAML on the message. Global messages are tagged with `meta: %{"from_global" => true}` and never carry a `job_id`, even when the session was started from a job step. The old single-job extraction code is removed. This includes `extract_global_code_and_job`, `resolve_job_from_key`, and the name normalization helpers. We previously matched steps by name. We now match by step UUID instead, because the assistant might rename a step and name matching would break on rename. This relies on the AI server preserving the step IDs in the YAML it returns.

- `prepare_message_attrs` now skips the job association logic for global messages. Non-global messages still go through `maybe_put_job_id_from_session` and `maybe_put_unsaved_job_meta` exactly as before. The channel serializer adds a `from_global` flag to each message and drops the legacy `from_global_job_code` handling.

- On the frontend, a new `from_global` field is added to the `Message` type. Global messages carry the full workflow YAML in `code` and no `job_id`. A new `handlePreviewGlobalStep` handler mirrors `handlePreviewJobCode`, but extracts the open step's body from the workflow YAML by matching the open step's UUID. It shows a diff only when that step's body actually changed and clears any stale diff otherwise. `handleApplyWorkflow` is relaxed so global messages can apply the full workflow even while a job is open. Non-global workflow chat keeps the workflow_template-only guard, so its Apply stays a no-op when a job is open. Auto-preview routing picks `handlePreviewGlobalStep` for global messages and `handlePreviewJobCode` for job-code messages, and `useAutoPreview` now allows global messages that have no `job_id`.

- When a step preview fails, we log the error and surface it to the user with a toast. Invalid workflow YAML shows an alert. A step whose ID is missing from the YAML (likely because the server did not preserve the ID) shows a warning.

- The diff preview only shows for the single step that is currently open, and only when the user has a job-code step open. A follow-up PR will make it easier to view several step changes and diffs at once. This PR focuses on the basic functionality: receiving and applying changes correctly across several steps from Apollo.

## Validation steps
Run the Apollo server from this branch with the required corresponding changes from Apollo: https://github.com/OpenFn/apollo/pull/547 

1. On the OpenFn app, open an existing workflow, click into a job code step, and open the AI assistant. Make sure to tick the box for "Global assistant - experimental feature" to talk to the Global assistant.
2. Ask the assistant to make an edit on several steps, including the one you are viewing (e.g. add a log to steps a, b that says "hello")
3. Verify the diff appears as expected in the current step.
4. Click apply, and verify that the changes were implemented in the other parts of the workflow as expected.

+ Verify that generate with AI from scratch (with job code filled in) still works like before.

Changes should only affect Global assistant, but please verify that this is true.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


elias-ba (approved):
Really nice work @hanna-paasivirta 👏🏽 🎉. I think this breaks the AI Assistant (at least the Global Chat), right ? So until PR 547 of Apollo be merged and deployed we can't really ship this guy. Maybe communicate that very well with the team so that they won't cut a release with this. Again great work

josephjclark:
So I've run some testing against this and it seems to work just fine on main apollo? Because apollo already sends the full yaml back, and this PR just uses it and ignores the other stuff.

So long as we release this before the apollo update, users won;'t see any breakage. This PR is fully compatible with both apollo releases.

The preview and UI are broken, btu we know that and work that is coming up soon.

I think that means this PR is ready to go

josephjclark (approved):

PR #4901, opened 2026-06-26
Edit in sandbox: create or join a sandbox from a live workflow

### LINKED ISSUE #4859: Edit in sandbox (create + picker)
**As a** workflow builder on a live workflow, **I want** to branch it into a sandbox and edit there instead of touching main, **so that** I can make and test changes safely while production keeps running on the current live version.

Part of #4852 (Sandbox DevX epic).

## Outcome

From a live workflow, Edit in sandbox opens a picker to create a new sandbox or join an active one, and lands you in the sandbox editor with the workflow's trigger live.

## Scope

- Edit in sandbox picker modal (new component): "Create new sandbox (branch from the current live version)" plus a list of active sandboxes to join, sorted by last-edited time, each identifying the sandbox's creator by name and avatar.
- A sandbox name is required at creation; Create stays disabled until a name is entered. Auto-generated and optional names are deferred to a later iteration.
- "Create new" goes through `Lightning.Projects.Sandboxes.provision/3` (full-clone, no workflow subsetting at creation). "Join" opens the workflow in the sandbox editor. New `WorkflowChannel` events.
- On opening a workflow for editing in a sandbox, enable that workflow's trigger in the sandbox (deliberate divergence from `clone_triggers_with_disabled_state`).
- `sandbox` badge in the editor header.

## Acceptance criteria

- [ ] Given a live workflow, when I click Edit in sandbox, then a picker opens with Create new and a list of joinable active sandboxes.
- [ ] Given Create new, when it completes, then a full-clone sandbox is provisioned, I land in the sandbox editor, the workflow is editable, and a `sandbox` badge shows.
- [ ] Creating a sandbox requires a non-empty name; Create is disabled until one is provided.
- [ ] The workflow's trigger has its own `/i/...` endpoint and is live in the sandbox, and main is unchanged.
- [ ] When I reopen the picker, the active sandbox is listed to join, sorted by last-edited, each row identifying the sandbox's creator.

## Key surfaces

New picker component, `Sandboxes.provision/3`, `WorkflowChannel`, editor header, awareness/presence (`useAwareness`, `ActiveCollaborators`).

## Dependencies

Builds on the lifecycle foundation. Phase M2 (sandbox round trip). In review as PR #4901.

## Estimate

8 story points.

### LINKED ISSUE #4989: Switch-to-draft modal: close on Esc and warn clearly it takes production offline
## What's off

The switch-to-draft confirmation does not close on Esc, and its message does not make clear that switching a live workflow to draft takes the production workflow offline.

## What to do

Let Esc dismiss the modal, and reword the copy so it is explicit that going to draft takes the workflow out of production.

Surfaced during the Edit in sandbox review (#4901).

### LINKED ISSUE #4990: Live workflow: show and allow copying the full webhook URL
## What's off

On a live (read-only) workflow you cannot see or copy the full webhook URL.

## Why it matters

People need the endpoint even when the workflow is locked, for example to configure the system that calls it.

## What to do

Make the full webhook URL visible and copyable on a live workflow. The read-only lock should not hide it.

Surfaced during the Edit in sandbox review (#4901).

### LINKED ISSUE #4991: Normalize sandbox and lifecycle button styling into design tokens
## What

The new sandbox and lifecycle buttons use one-off styling (box shadows, sizes like `text-[10px]`) that does not match the rest of the app.

## What to do

Fold these into shared design tokens as part of the design-system pass, so the buttons look consistent with everything else.

Surfaced during the Edit in sandbox review (#4901).

elias-ba:
## Description

This PR adds "Edit in sandbox". From a live workflow, the header gains an "Edit in sandbox" action that opens a picker to either branch a new sandbox from the current live version or join an active one, then lands the user in the sandbox editor with that workflow's trigger live and its own endpoint. The parent project is untouched; changes come back through promotion in a later slice of the epic.

Creating clones the parent project (the existing full-clone provisioning), then promotes the edited workflow to live inside the sandbox. The other cloned workflows stay draft, so a workflow's state stays coherent with its triggers. The sandbox copy remains editable because the live read-only lock only applies outside a sandbox. The picker lists only the parent's active sandboxes that already contain this workflow (last edited first), each showing the sandbox's creator and creation time; a name is required to create a new one. A sandbox badge is shown in the editor header. Server-side errors (permission, usage limit) surface their real message in the picker, including the limit upsell, rather than a generic toast.

Closes #4859
Closes #4989
Closes #4990
Closes #4991

## Validation steps

1. On a live workflow, click Edit in sandbox to open the picker.
2. Create new provisions a full-clone sandbox, lands in the sandbox editor with the workflow editable and a sandbox badge.
3. Confirm the workflow's trigger is live in the sandbox with its own endpoint, and the parent is unchanged.
4. Reopen the picker from another live workflow and confirm the active sandbox is listed to join. Sandboxes that do not contain that workflow are not listed.

1. Two things are deliberately deferred rather than half-built here: refining what the header shows inside a sandbox (suppressing the draft/live controls and adding Promote) belongs with the promote work (#4929); and hiding the action client-side for users who cannot provision (today the server enforces it and the picker surfaces the reason) needs a permission flag on the session context (#4920).
2. The picker sorts sandboxes by last-edited time and shows each sandbox's creator (the person who provisioned it).

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


assets/js/collaborative-editor/components/EditInSandboxPicker.tsx
```diff
@@ -0,0 +1,346 @@
...
+          className="inline-flex h-6 w-6 items-center justify-center rounded-full
+            bg-gray-200 text-[10px] font-medium text-gray-700 ring-2 ring-white"
+        >
+          {collaboratorInitials(collaborator)}
+        </span>
+      ))}
+      {overflow > 0 && (
+        <span
+          className="inline-flex h-6 w-6 items-center justify-center rounded-full
+            bg-gray-100 text-[10px] font-medium text-gray-500 ring-2 ring-white"
```
lmac-1: I don't love `text-[10px]`, but this is more for us to tidy up these sorts of sizes and design tokens when we generalise the design system more comprehensively later on. No block to merging.

assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -567,6 +581,21 @@ export function Header({
...
               )}
+              {lifecycleState === 'live' && !isSandbox && !isNewWorkflow && (
+                <button
+                  type="button"
+                  data-testid="edit-in-sandbox-button"
+                  onClick={() => {
+                    setShowEditInSandboxPicker(true);
+                  }}
+                  className={cn(
+                    'inline-flex items-center gap-1 rounded-md bg-white px-3 py-2 text-sm font-semibold text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 hover:bg-gray-50 disabled:cursor-not-allowed disabled:opacity-50'
```
lmac-1: here we can just use `className` as it is one long static string

lmac-1:
@elias-ba I noticed that you haven't checked the box that you've done an AI self review of the code here. Have you done a Claude review?

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -383,6 +386,44 @@ defmodule LightningWeb.WorkflowChannel do
...
+             name: name,
+             env: "dev",
+             color: random_sandbox_color()
+           }),
+         {:ok, cloned_workflow} <-
+           promote_cloned_workflow(sandbox, workflow.name, user) do
+      {:reply, {:ok, %{project_id: sandbox.id, workflow_id: cloned_workflow.id}},
+       socket}
+    else
+      error -> workflow_error_reply(socket, error)
```
lmac-1, line 416: `workflow_error_reply` in `workflow_channel.ex` doesn't have a clause for `{:error, :nesting_too_deep}`, which is one of the return values from `Projects.provision_sandbox` (see `sandboxes.ex:585`). If that error reaches `workflow_error_reply`, the channel process will crash and drop the user's WebSocket connection. Needs a handler clause added, similar to the existing ones at line 1045-1069.

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -383,6 +386,44 @@ defmodule LightningWeb.WorkflowChannel do
...
+           Projects.provision_sandbox(parent, user, %{
+             name: name,
+             env: "dev",
+             color: random_sandbox_color()
+           }),
+         {:ok, cloned_workflow} <-
+           promote_cloned_workflow(sandbox, workflow.name, user) do
+      {:reply, {:ok, %{project_id: sandbox.id, workflow_id: cloned_workflow.id}},
+       socket}
+    else
```
lmac-1, line 415: If `provision_sandbox` succeeds but `promote_cloned_workflow` fails, the sandbox is already committed to the database and this `else` branch doesn't do anything to clean it up. 

The user gets an error but the sandbox persists and consume quota. Not blocking but could add a compensating delete of the sandbox here, or both steps wrapped in a single transaction to make it a better user experience.

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1149,6 +1190,93 @@ defmodule LightningWeb.WorkflowChannel do
...
+  end
+
+  defp limit_new_sandbox(parent) do
+    case ProjectLimiter.limit_new_sandbox(parent.id) do
+      :ok -> :ok
+      {:error, _reason, message} -> {:error, message}
+    end
+  end
+
+  defp promote_cloned_workflow(sandbox, workflow_name, user) do
```
lmac-1: This function queries the database directly instead of going through the `Workflows` context. It would be better to extract a `Workflows.get_workflow_by_name(project_id, name)` function in `lib/lightning/workflows.ex`

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1149,6 +1190,93 @@ defmodule LightningWeb.WorkflowChannel do
...
+
+    Lightning.Helpers.url_safe_name(raw)
+  end
+
+  defp default_sandbox_name(workflow, parent) do
+    base = workflow.name || parent.name || "sandbox"
+    "#{base}-sandbox"
+  end
+
+  defp random_sandbox_color do
```
lmac-1: This is a duplicate of `get_random_color/0` in `sandbox_live/form_component_ex`. 

Since `color_palette_hex_colors/0` already lives in `LightningWeb.SandboxLive.Components` we could extract this into a shared public `random_color/0` function there and call it from both places.

lib/lightning_web/channels/workflow_channel.ex
```diff
@@ -1149,6 +1190,93 @@ defmodule LightningWeb.WorkflowChannel do
...
+
+  defp render_collaborator(%{user: user}) do
+    %{
+      id: user.id,
+      name: collaborator_name(user),
+      email: user.email
+    }
+  end
+
+  defp collaborator_name(user) do
```
lmac-1, line 1523: Non-blocking. Claude mentioned how this function is duplicating the first/last name logic that already appears in `mailer.ex`, `workflow_json.ex` and elsewhere. There isn't a canonical helper yet. But we could extract this out to a shared `Accounts.display_name/1` function (not sure if that's the right place) and replace all the copies. This is out of scope of this PR but flagging for follow-up.

lib/lightning/projects.ex
```diff
@@ -1896,6 +1896,47 @@ defmodule Lightning.Projects do
...
+        preload: [project_users: :user]
+      )
+      |> Repo.all()
+
+    sandbox_ids = Enum.map(sandboxes, & &1.id)
+
+    joinable_workflow_ids =
+      from(w in Workflow,
+        where:
+          w.project_id in ^sandbox_ids and w.name == ^workflow_name and
```
lmac-1: If the parent project has no sandboxes, `sandbox_ids` will be `[]` and the workflow lookup query still runs against the database unnecessarily, always returning nothing. Suggest adding an early return guard here: `if sandbox_ids == [], do: []`

lmac-1 (changes requested):
This is looking exciting @elias-ba - it's a big review so strap in 🤩! I've left a few comments in the code that I think are worth addressing within this PR. One final code comment: 

`assets/js/collaborative-editor/components/EditInSandboxPicker.tsx:1` and test files — multi-paragraph block comments violate CLAUDE.md

>   CLAUDE.md: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."

Four blocks violate this:
 - EditInSandboxPicker.tsx lines 1–14 (14-line file header with 3 paragraph breaks)
 - EditInSandboxPicker.tsx lines 36–39 (serverMessage function comment)
 - EditInSandboxPicker.test.tsx lines 1–12
 - Header.sandbox.test.tsx lines 1–12

---

My feedback from a UX/UI perspective on **"Edit in sandbox"** experience (not blocking but observational):

1. It was a bit confusing seeing "Join an active sandbox" and then both sandboxes having disabled "Join" buttons. Perhaps it shouldn't show this section if there is no sandbox to join?

Image

2. When hovering over the disabled "Join" button in the above scenario, there was no explanatory tooltip message to explain why. Sometimes it appears but after a long pause, not instantly. 

3. The "Join an active sandbox" wording was a bit confusing.. it made me feel like I wasn't a part of that sandbox before. Not sure what would be better wording here. 

4.  What exactly are the collaborator circles for? What do they represent? 

5. If entering a sandbox name is optional, our styling in the header should support long workflow/sandbox names more elegantly. Suggestions: 
(1) Make the Save button grow 
(2) Add ellipsis for the sandbox/workflow name when it's a certain length:

Image

6. I think we should improve the toast message when you try to add a sandbox with a name that already exists rather than the generic toast message. It should tell the user that a sandbox already exists with that name so they know how to fix it. 

7. It is confusing for me when I switch to sandbox and see "Live" -> Switch to draft. I know what it means because of the extra context from meetings. But it's a bit strange to come from somewhere where I can't edit a live workflow and then it is marked as "Live" in the sandbox. 

------
Some general comments outside of the scope of this PR of things I noticed for this epic while I'm here: 

1. Switch to draft modal - should Esc close this modal?

2. Switch to draft modal: I'm not sure the message is clear enough that it will turn off the production workflow. Not sure of the wording but perhaps needs to warn more explicitly.

3. At one point I forgot I was reviewing this PR, went to rename my workflow and got confused why I couldn't edit anything. Perhaps this tooltip should be extended to mention why you don't have permission to edit and how to edit? (however, you might not be able to mention sandboxes as the experience is different for free vs paid users)

Image

4. The box shadows of the new buttons feel very different stylistically compared to the other parts of the application. Are they part of the new design system?
Image

5. You cannot copy or see the full webhook URL for a live workflow

6. Go live button is the wrong colour when disabled (should be the same as Save)
Image

7. I see the old trigger picker.. you should merge from main to avoid any nasty conflicts while you can 🤭

8. The top bar feels very hectic.. there is a LOT of things going on up there now. I'm sure this will get better as other epics come in but it feels like a lot.

assets/js/collaborative-editor/components/EditInSandboxPicker.tsx
elias-ba: Actually done in this PR, not deferred, sorry for the confusing wording. The new lifecycle/sandbox buttons now use the app's shared `shadow-xs` instead of the heavier `shadow-sm`, and this `text-[10px]` went away when I redesigned the picker rows. #4991 tracks it and closes with this.

assets/js/collaborative-editor/components/Header.tsx
elias-ba: Nice catch. Done, it's a plain `className` now.

lib/lightning_web/channels/workflow_channel.ex
elias-ba: Good catch, this would have dropped the socket. Added a `:nesting_too_deep` clause plus a catch-all so an unexpected reason can't crash the channel.

lib/lightning_web/channels/workflow_channel.ex
elias-ba: You're right. Provision and promote now live in `Projects.provision_editing_sandbox/4`, which deletes the half-built sandbox if the promote fails, so it can't linger and eat quota.

lib/lightning_web/channels/workflow_channel.ex
elias-ba: Done, extracted `Workflows.get_workflow_by_name/2` and the channel goes through it now.

lib/lightning_web/channels/workflow_channel.ex
elias-ba: Good call. Pulled a shared `random_color/0` into `SandboxLive.Components` and both places call it now.

lib/lightning_web/channels/workflow_channel.ex
elias-ba: Agreed it's worth doing, but it touches `mailer.ex`, `workflow_json.ex` and a few others that have nothing to do with this PR, so I've left it as a follow-up: #4988.

lib/lightning/projects.ex
elias-ba: Done, added the early return so we skip the query when there are no sandboxes.

elias-ba:
Thanks for the super thorough review @lmac-1, this was really helpful. I've replied to the code comments inline. On the rest, I've restated each of your points so they're easy to follow:

**Multi-paragraph comments:** collapsed all four blocks to one-liners.

**UX/UI**

- **"Join" section showing with all-disabled buttons:** Done, the picker only lists sandboxes that actually contain this workflow now, and the whole section is hidden when there's nothing to join.
- **No instant tooltip on the disabled Join:** Moot now, since non-joinable sandboxes aren't listed there's no disabled Join anymore.
- **"Join an active sandbox" wording is confusing:** You're right, and honestly I'm not sure of the best wording either. It's part of the broader design pass, so I've left the copy to that, tracked in #4992.
- **What are the collaborator circles for:** Fair, they were unclear. I've replaced them with the sandbox's creator and creation time, plus a colour accent per sandbox, so each row tells you who made it and when.
- **Long names need more elegant styling:** Done, long names truncate in the header now, with a tooltip showing the full name on hover.
- **Better toast for a duplicate sandbox name:** Done, a duplicate name now surfaces the real "has already been taken" error instead of the generic toast.
- **"Live" then Switch to draft inside a sandbox is confusing:** Agreed it was odd. Inside a sandbox I've dropped the draft/live badge and the Go live / Switch to draft controls, and replaced them with a Promote button (disabled, "coming soon" until the promote work lands in #4929), so you no longer see "Live" there.

**Out-of-scope epic notes**

- **Should Esc close the switch-to-draft modal:** You were right, it wasn't, a higher-priority editor shortcut was swallowing Esc. Fixed now, so this modal and the Edit-in-sandbox picker both close on Esc.
- **Switch-to-draft modal should warn it turns off production:** Done, the modal now says it takes the workflow out of production and disables its triggers (#4989).
- **The read-only tooltip should explain why you can't edit:** Improved, a live workflow now reads "This workflow is live. Switch to draft or edit in a sandbox to make changes" for people who can act on it, and keeps the plain "you don't have permission" for viewers. The plan-specific wording you flagged is the tricky part, so I've kept it simple for now.
- **Box shadows feel different from the rest of the app:** Good catch, they were using a heavier shadow. Aligned the new buttons to the shared button shadow.
- **Can't copy or see the webhook URL on a live workflow:** Done, the copy button was being disabled by the live read-only lock. It's a read action so it's enabled again (#4990).
- **Go live is the wrong colour when disabled:** Done, the disabled Go live now uses the same styling as Save (primary-300).
- **Old trigger picker / merge from main:** Yep, the branch is on a stale base. I'll merge main before this goes up to catch those conflicts.
- **Top bar feels hectic:** Agreed. I've trimmed some already (dropped the sandbox badge, and hide the lifecycle badge inside a sandbox), and I expect the design pass will tidy the rest.

Ready for another look when you get a chance. The follow-ups I've split out are #4988 (display-name helper) and #4992 (picker wording). Could you take another look?

lmac-1:
Hey @elias-ba, I am doing a more detailed review, but I came across this bug while testing. Not sure if it's on your radar already: 
https://www.loom.com/share/18a54061d12e432c9baf43e090d4af67

I suspect things might sort themselves out when added in with the new 'create' workflow. Maybe. Who knows.

lmac-1 (commented):
Hey Elias, great work on cleaning up all the feedback from before. I really love the way you organise your epics and raise issues for smaller things so they don't get lost. 

Looking great except for this part (shouldn't block merge though, would just be nice to clean it up a bit):

> _Better toast for a duplicate sandbox name: Done, a duplicate name now surfaces the real "has already been taken" error instead of the generic toast._

I see "- has already been taken" rather than "Name has already been taken" or "{sandboxname} has already been taken". This matches a lot of current error messages in the app so it's not terrible, but if it's an easy win, we should clean it up to make it nicer. This is because there might be some sandbox with a name that we don't know about (it doesn't appear in 'active sandbox' list) so the user might be confused to know what has already been taken.

Image

While we're in there, the same error surface has a related problem. If I type `My_Sandbox` I get "- has invalid format", and if I type something like `!!!` I get "- can't be blank" for a name I definitely did type. That's because `url_safe_name` keeps underscores, dots and accented letters, but the project name validation only allows lowercase letters, numbers and hyphens — so the two disagree about what's valid. Same root cause as the message above, so possibly one fix for both. (I think this mismatch predates the PR — the sandbox creation form has it too — but the picker shows it with less context, since there's no derived-name preview to explain what happened.). Not blocking in this PR but flagging as a general UX clean up for later. 

---

I found the following bugs. I know that the scope of this PR might not include all of these, so I will leave it to you to decide what should be covered as part of this PR, and what should be a follow-up:

1. The flow breaks in this scenario: 
    - In draft status, I edit the name of the workflow
    - I click "Go live"
    - I click "Edit in sandbox"
    - I try to create a sandbox and I get an error "An internal error has occurred" without much information

2. Another bug:
     - I have a workflow called "Workflow A" that has webhook trigger and common job.
     - I click "Edit in sandbox" from workflow A in a sandbox called "Sandbox A"
     - I delete the original workflow A at the project level
     - I create another workflow called "Workflow A" that has a different structure, let's say cron job and chatgpt job. Let's refer to this as Workflow B
     - I click "Edit in sandbox"
     - The sandbox I created in the previous step (Sandbox A) appears as an active sandbox
     - I click on that sandbox and it shows me the original Workflow A in that sandbox rather than the new workflow we created with the chatgpt job.

3. Renaming the workflow doesn't seem to update the "active sandboxes" list - is the search looking up workflows inside the sandbox by name?
    - Precondition: I already have a sandbox with some workflows from the default event-based workflow template with name "Event-based workflow"
    - I rename my workflow from "Event-based workflow" to something else
    - I click Go live and then Edit in sandbox
    - The active sandboxes shows all the sandboxed with "Event-based workflow" inside rather than this new workflow I have renamed. 
    - If I refresh the page and click "Edit in sandbox" I correctly get no active sandboxes

4. Escaping out of sandbox creation still takes me into the sandbox:
    - I click "Create sandbox"
    - While it's creating, I press Escape (or click the ✕)
    - The modal closes, but the request is still in flight, and when it succeeds I get navigated into the new sandbox anyway
    - The other confirm dialogs in t he editor block closing while a request is in flight, so it'd be consistent to do the same here

Claude says: 

> I think bugs 1, 2 and 3 are all the same root cause, and it might be worth fixing at that level rather than three times.
> 
> The picker works out "which sandboxes contain a copy of this workflow" by matching on **workflow name**. That's why renaming breaks it (1 and 3), and why deleting and recreating a workflow with the same name matches an unrelated clone (2).
> 
> But there's already a lineage link we could use instead: when a sandbox is provisioned, `copy_workflow_version_history/2` copies the parent workflow's latest `workflow_versions` hash onto the clone, so the two share a version hash. The schema docs note that the same hash existing across workflows is expected, and the rows are append-only, so that shared ancestor survives even after the sandbox diverges. Matching on shared lineage rather than name would fix all three at once — and 1 would disappear entirely, since the name would stop being involved.
> 
> One caveat: two genuinely unrelated workflows with identical content would share a hash, so if that's a concern the cleanest version is probably an explicit "cloned from" column on the sandbox workflow. Your call which is proportionate — but either way I don't think name should be the identity here.
> 

Please advise whether these bugs should be handled in this PR or follow-up work.

---

A question rather than a change request, I don't know too much about Kafka — but I'd like to understand it. Creating the sandbox makes the cloned workflow live, which switches its trigger on. That makes sense to me for webhook and cron: the sandbox gets a URL nobody knows yet, or runs its own copy on a schedule. Either way it only affects itself, which is what we want so you can see it working.

Kafka feels different though, because the sandbox reaches *out* and attaches to the same production broker and topics. Those connection settings (hosts, topics, sasl, username/password) live on the trigger config rather than in the credential system, so unlike normal credentials they aren't scoped by the sandbox's `dev` environment. The clone gets a fresh `group_id`, so production still receives everything — but every message would get processed twice, once by production and once by the sandbox.

Is that OK for how we use Kafka, or should Kafka triggers stay switched off when the sandbox is created?

Issue #4848, opened 2026-06-10
Epic: AI-First Starting UX


theroinaochieng:
## Key Hypothesis
If we make AI the default way to start a workflow, then users will reach their first "a-ha" moment faster.

## Problem Validation

1. The current "new workflow" screen presents multiple competing options (Templates, Generate with AI, Import YAML) with no clear default — users face a decision before they've even started.
2. Hackathon user testing (2026-04-14): users found the templates menu confusing — "Making a new workflow gives me a ton of options, 'Event Based' doesn't really (quickly) tell me that uses a webhook to trigger."
3. Hackathon user testing (2026-04-14): users had difficulties generating and saving workflows from AI — when generation fails, the system doesn't communicate the failure clearly and continues working off a broken workflow.

## Solution Concept

The "New" experience should funnel users straight into AI-assisted workflow creation, with other paths (templates, import) clearly available but secondary. 

When a user navigates to `/new`, they see a landing screen with a structured choice of creation paths — each path results in a workflow being created in the database. That first DB write is the central event this epic is designed around.

## Core Concept: The Pink Toolbar

**The pink toolbar is a design term for the state where no workflow yet exists in the database.**

<img width="1268" height="842" alt="Image" src="https://github.com/user-attachments/assets/445be924-c92a-434f-80a9-bd0681259974" />

It is a concept, not a visual colour. It means:
- The workflow has not been saved to the DB yet — no record exists, no ID, no version history
- No toolbar 
- The URL is still `/new`

**The pink toolbar ends the moment a workflow is written to the database for the first time.** After that:
- Top toolbar appears with Save button enabled
- The URL updates to `/w/{id}`
- The workflow appears in the user's workflow list — if they navigate away, they will find it

## Must Solve

- Clicking "New" places the user inside the editor immediately.
- AI is the default and most prominent starting path.
- Workflows that are started editing save automatically — no explicit "Create" button required.
- A "Build From Scratch" option is clearly available.
- Templates remain visible and accessible.
- Templates system removed from critical path / untangled.

## Decisions

- There will be a new Build from Scratch card available which will create a workflow with a webhook and job with trigger inspector component visible
- Template option will open a modal with cards with title and description with ellipsis. Clicking a template will create a workflow instantly with that template. No browsing possible without creation.
- Import with YAML will allow users to import a YAML document or paste YAML code
[...]

PR #4902, opened 2026-06-26
AI-First Starting UX: Build with AI path

### LINKED ISSUE #4868: AI-First Starting UX: Build with AI path - landing screen transition, auto-create, and panel lock
**Parent epic:** #4848 AI-First Starting UX
**Depends on:** #4856

## What this issue builds

Four concrete pieces of new behaviour:

1. **Landing screen → canvas transition**: "Build it →" opens the canvas, opens the AI panel,
   and auto-sends the user's prompt — no re-typing required
2. **Auto-save on first YAML**: once the AI applies a valid workflow, it is saved to the DB
   automatically — no "Create" button exists in this design
3. **Panel lock**: the AI panel cannot be closed while the workflow has not yet been saved —
   closing it would leave the user on a blank canvas with no path forward
4. **Validation error routing**: if pre-validation fails, the error goes to the AI chat panel
   (not a toast) so the user can ask the AI to fix it

**What is not changing:** the AI chat panel itself, auto-apply behaviour (already works), the
`page === 'workflow_template'` mode (already set correctly on `/new`), the `saveWorkflow` silent
option (already supported), save failure toasts for non-validation errors.

  ## Decisions Made
  
- **Auto-save.** When the AI returns valid YAML, the workflow is saved
    to the DB automatically. There is no "Create" button in this flow.

- **No top toolbar until save**: There is no top toolbar until the workflow is saved to the database.
  
- **Panel lock.** The AI panel cannot be closed while no workflow exists in the DB. Closing
    it would leave the user on a blank canvas with no path forward. 

- **Errors route to chat, not toasts.** If pre-validation fails, the error appears in the
    AI chat panel so the user can ask the AI to fix it.

- **Landing → canvas is a UI state change, not navigation.** No pushState call is made.
    This keeps browser history clean: back from the canvas goes to the workflow list, not
    back to the landing screen.
    
- **Canvas stays blank until YAML arrives.** The AI may send conversational messages before
    returning YAML. This is expected — the blank canvas persists throughout.

## User flow

1. User types a prompt and clicks "Build it →" on the landing screen
2. Landing screen is replaced by a blank canvas. AI panel opens on the right with visual blank canvas.
3. User's prompt appears as a sent message. AI begins processing.
4. App is in pink toolbar state: URL is `/new`, no top toolbar, no DB record exists.
5. AI panel is locked — it cannot be closed until a workflow exists.
6. AI may send conversational messages before returning YAML. Canvas stays blank throughout.
7. When AI returns YAML:
   - Pre-validation runs (existing: `parseWorkflowYAML` → `validateIds` → `convertWorkflowSpecToState`)
   - **Valid YAML**: applied to canvas, auto-saved to DB, pink toolbar ends
   - **Invalid YAML**: error injected into AI chat panel, canvas stays blank, user stays in
     pink toolbar state, panel stays locked
8. On successful creation:
   - URL updates to `/projects/{project_id}/w/{workflow_id}` without a page reload
   - Top toolbar visible. Save becomes enabled. Run is disabled. 
   - Workflow is named by the AI.
   - AI panel unlocks — conversation may continue to refine the workflow.

If the AI returns a message with no YAML (e.g. asks a clarifying question): canvas stays blank,
panel stays locked. User continues the conversation. Navigating away is safe — no DB record
exists.

### Piece 1 — Landing screen → canvas transition

`createUIStore.ts` already has `openAIAssistantPanel(initialMessage?: string)` at line 190.
Calling it with the user's prompt sets `aiAssistantPanelOpen = true` and stores
`aiAssistantInitialMessage`. The existing `useAIInitialMessage` hook (`hooks/useAIInitialMessage.ts`)
picks this up and auto-sends it when the panel mounts. No new hook code needed.

The landing screen component should:
1. Call `openAIAssistantPanel(prompt)` (via `useUICommands()` from `hooks/useUI.ts:44`)
2. Switch the view from the landing screen to the canvas via React state — do NOT call
   `pushState` or add a new history entry

URL stays at `/new`. The landing screen → canvas switch is a pure UI state change, not a
navigation event. This keeps browser history clean:
- Back from landing screen → wherever the user came from (e.g. workflow list)
- Back from canvas (pre-creation) → also goes straight to workflow list, not back to the
  landing screen (which is fine — user can navigate to `/new` again to start fresh)
- Back from canvas (post-creation) → workflow list, because `replaceState` on first save
  swaps `/new` for `/w/{id}` without adding a history entry

### Piece 2 — Auto-save on first YAML

In `handleApplyWorkflow` (`hooks/useAIWorkflowApplications.ts:171-228`), after
`importWorkflow()` succeeds, add:

```ts
if (isNewWorkflow) {
  await saveWorkflow({ silent: true })
}
```

`saveWorkflow({ silent: true })` suppresses the success toast but still runs the URL transition
(`window.history.replaceState` `/new` → `/w/{id}`) and calls `clearIsNewWorkflow()`.

**`isNewWorkflow` and `saveWorkflow` are not currently in scope here.** Thread them through:

- Add `saveWorkflow` to the `workflowActions` parameter interface at
  `hooks/useAIWorkflowApplications.ts:125-132`
- Add `isNewWorkflow: boolean` as a new top-level parameter to the hook interface
- In `AIAssistantPanelWrapper.tsx`, pass both when calling the hook:
  - `isNewWorkflow` — already at line 163 via `useIsNewWorkflow()`
  - `saveWorkflow` — already in `useWorkflowActions()` destructure at line 544

### Piece 3 — Panel lock

Two close paths need to be blocked when `isNewWorkflow` is true:

**Keyboard shortcut** — `$mod+k` handler at `AIAssistantPanelWrapper.tsx:121-132`:
```ts
// before
{ enabled: !isPinnedVersion && aiAssistantEnabled }

// after
{ enabled: !isPinnedVersion && aiAssistantEnabled && !isNewWorkflow }
```

**Close button** — `AIAssistantPanel.tsx:393-413` calls `onClose` → `handleClosePanel` in
`AIAssistantPanelWrapper`. Wrap before passing as prop:
```ts
const onClose = isNewWorkflow ? undefined : handleClosePanel
```
When `onClose` is undefined, the close button in `AIAssistantPanel` should be hidden or
disabled. Check the component interface to confirm the right approach.

**AI assistant toolbar button on the landing screen**: hide or disable it. On the landing
screen the text input is the entry point — a second route to open a blank panel would
bypass the designed flow.

### Piece 4 — Validation error routing

Pre-validation already runs before `importWorkflow` is called (existing code):
- `parseWorkflowYAML` (`assets/js/yaml/util.ts:291-332`): YAML syntax + JSON schema against
  `workflow-spec.json` + duplicate job name check
- `validateIds` (`hooks/useAIWorkflowApplications.ts:32-80`): guards against object-typed IDs
- `convertWorkflowSpecToState` (`assets/js/yaml/util.ts:157-265`): edge target/source cross-
  reference checks

No new validation logic is needed. The change is **how errors are surfaced**: when
`isNewWorkflow` is true and validation throws, inject the error as a message into the AI chat
panel instead of showing a toast. The user can then ask the AI to fix the problem.

**Known server-side validation not covered client-side** (job name format/length, cron
expression validity, edge `condition_type` enum, `condition_expression` requirement for
`js_expression` edges): these are edge cases the AI is unlikely to produce. If one slips
through and the server rejects the save, the existing "Save failed: invalid workflow" toast is
an acceptable last resort.

## Files to modify

| File | Change |
|---|---|
| `stores/createUIStore.ts` | No change — `openAIAssistantPanel(initialMessage?)` already exists |
| `hooks/useAIInitialMessage.ts` | No change — auto-send pattern already in place |
| `components/AIAssistantPanelWrapper.tsx:121-132` | Add `&& !isNewWorkflow` to shortcut `enabled` |
| `components/AIAssistantPanelWrapper.tsx` | Pass `isNewWorkflow` + `saveWorkflow` into hook; wrap `handleClosePanel` |
| `components/AIAssistantPanel.tsx:393-413` | Hide/disable close button when `onClose` is undefined |
| `hooks/useAIWorkflowApplications.ts:125-132` | Add `saveWorkflow` + `isNewWorkflow` to interface |
| `hooks/useAIWorkflowApplications.ts:171-228` | Add auto-save after `importWorkflow()` |
| Landing screen component (Issue 1) | Call `openAIAssistantPanel(prompt)` on submit |

## Out of scope

- Changes to the AI chat panel (Fn Assistant) itself
- The trigger picker or Build from scratch
- Auto-apply or auto-save on existing workflows — only the new workflow creation path changes

### Transition from landing screen to canvas

- [ ] "Build it →" (or Enter) replaces the landing screen with a blank canvas
- [ ] AI chat panel opens automatically on the right
- [ ] User's prompt appears as a sent message without re-typing
- [ ] AI begins processing immediately
- [ ] URL remains `/new`, Save is disabled, no DB record exists

### Back navigation

- [ ] Clicking back from the landing screen returns the user to wherever they came from
- [ ] Clicking back from the canvas (before workflow is created) goes to the workflow list —
      not back to the landing screen
- [ ] Clicking back from the canvas (after workflow is created) goes to the workflow list
- [ ] No `pushState` calls are made during the landing screen → canvas transition

### Panel lock

- [ ] AI panel cannot be closed while in pink toolbar state on the canvas
- [ ] `$mod+k` does not close the panel while in pink toolbar state
- [ ] Close button on the AI panel is hidden or disabled while in pink toolbar state
- [ ] AI assistant toolbar button is hidden or disabled on the landing screen
- [ ] After workflow is created, panel can be freely opened and closed

### Auto-save on valid YAML

- [ ] Canvas stays blank while AI is processing
- [ ] When valid YAML is applied, workflow is saved to DB automatically
- [ ] No "Workflow saved" toast fires (silent save)
- [ ] URL updates from `/new` to `/projects/{project_id}/w/{workflow_id}` without a page reload
- [ ] Save becomes enabled after creation
- [ ] Run is disabled after creation
- [ ] Workflow is named by the AI, not "Untitled workflow"
- [ ] AI panel stays open and unlocked — conversation may continue

### Validation failure

- [ ] If pre-validation fails, nothing is written to the canvas or DB
- [ ] Error is surfaced in the AI chat panel, not as a toast
- [ ] User remains in pink toolbar state and can ask AI to fix the problem
- [ ] Panel remains locked

### Save failure (non-validation)

- [ ] Connection error: red toast "Failed to save workflow" with Retry; canvas content kept
- [ ] Server error: red toast "Failed to save workflow" with Retry; canvas content kept
- [ ] Permission error: red toast "Permission Denied"

lmac-1:
**What I need in this review:** Focus on the "Build with AI" path of the epic. Any big code organisation problems will be fixed in the clean up issue. 

### Description

This PR adds the "Build with AI path" of the #4848 AI-First Starting UX epic. 

How it works:
1. User goes to /new and types a prompt in the "Build with AI" textarea
2. Pressing Enter or clicking "Build it →" dismisses the landing screen and opens the AI panel with the prompt pre
sent
3. The canvas stays blank while the AI processes
4. When the AI returns YAML, it is auto-applied to the canvas and saved to the DB automatically — no "Create" button
5. The URL transitions from /new to /projects/{id}/w/{id} without a page reload

### Key decisions:

- **Auto-save is silent.** No "Workflow saved" toast fires on first creation — it would be noise in a flow the user didn't manually trigger.
- **Panel lock while unsaved.** The AI panel cannot be closed (⌘K disabled, close button hidden) while the workflow hasn't been saved yet. Closing it would leave the user on a blank canvas with no path forward.
- **Validation errors route to chat, not toasts.** If the AI returns invalid YAML, the error is injected as a message into the chat panel so the user can ask the AI to fix it, rather than hitting a dead-end toast.
- **Save failure shows a Retry button.** If the DB save fails, a toast fires with a Retry action that retries only the save — the YAML has already been applied to the canvas so re-importing it is unnecessary. Fit-view is skipped on failure so the canvas doesn't zoom into an unpersisted workflow.
-  **Streaming dedup via `appliedViaStreamingRef`**. The streaming `streamingChanges` event and the final `new_message` both carry the same YAML. The ref is set before the streaming apply and cleared by the auto-apply effect after skipping, preventing a double import. The ref is also reset on apply or save failure so the `new_message` path can retry if streaming failed.
- **URL params are blocked on /new**. Params like `?panel=run`, `?panel=editor`, `?panel=settings`, and `?chat=true` could open panels or bypass the landing screen before the user has done anything. The store ignores all URL params at init time when `isNewWorkflow=true`, and `WorkflowEditor` guards its URL→panel sync effect with the same flag. This prevents e.g. someone sharing a `/new?panel=run` link that skips straight to the run panel on a blank workflow.

Closes #4868

## Validation steps

1. Go to `/new`, type a prompt, press Enter — confirm landing screen dismisses, AI panel opens with prompt pre-sent
2. Wait for AI to return YAML — confirm workflow appears on canvas and URL updates to `/w/{id}` without a reload
3. While AI is processing, ⌘K  — should do nothing
4. After creation, confirm panel can be freely opened and closed
5. Test with a prompt that produces invalid YAML (might be hard to do - might have to hack it with Claude) — confirm error appears in the chat panel, canvas stays blank, panel stays locked

1. What do you think about the toolbar appearing? Is the effect very jumpy and jarring? What could be the alternative? A completely read-only toolbar before save?
2. We are still working out how the disclaimer will work in this new flow. To be addressed in follow-up work.
3. There are still details to be ironed out. For example, you can still "View conversations" in AI tab before you have created a workflow. Still need to iron out other areas that you shouldn't be able to "get to" before a workflow is created. This can be done in follow-up work. 
4. I am not satisfied with the error handling flow, and might need help to tidy this up in some follow-on work for the AI assistant in general - from a Claude review: _the long tail of fix(useAIWorkflowApplications): ... commits (9+) is real evidence of implicit state churn, not just normal bug fixing. appliedViaStreamingRef alone is now reset from four different call sites for four different failure paths, alongside appliedMessageIdsRef and hasLoadedSessionRef tracking overlapping concerns. This is the classic shape of state that wants to be an explicit state machine (idle | streaming-applied | applying | save-failed) instead of booleans mutated ad hoc — worth a follow-up ticket before more logic piles on top._
5. When reviewing, make it clear that there is still some legacy code from the old "Create a new workflow" route using the side panel. This will be cleaned up in a follow up issue. 
7. **Landing screen gate is reactive, not static.** `createUIStore` is created once inside `useState` — on a slow connection the channel hasn't joined yet, so isNewWorkflow is false at init time. Initialising `showLandingScreen: true` always and moving the isNewWorkflow guard into `useShowLandingScreen` (which reads from the reactive `SessionContextStore`) means the landing screen correctly appears once the channel joins, even on slow connections.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -103,6 +103,9 @@ export function AIAssistantPanelWrapper({
   const isPinnedVersion =
     currentVersion !== undefined && currentVersion !== null;
 
+  const { isReadOnly } = useWorkflowReadOnly();
+  const isNewWorkflow = useIsNewWorkflow();
+
   // Track IDE state changes to re-focus chat input when IDE closes
   const isIDEOpen = params.panel === 'editor';
   const [focusTrigger, setFocusTrigger] = useState(0);
@@ -128,7 +131,7 @@ export function AIAssistantPanelWrapper({
       toggleAIAssistantPanel();
     },
     0,
-    { enabled: !isPinnedVersion && aiAssistantEnabled }
+    { enabled: !isPinnedVersion && aiAssistantEnabled && !isNewWorkflow }
   );
 
   const aiStore = useAIStore();
@@ -157,10 +160,7 @@ export function AIAssistantPanelWrapper({
   const workflow = useWorkflowState(state => state.workflow);
   const limits = useLimits();
 
-  // Check readonly state and new workflow status
   // AI can apply changes if: not readonly OR is a new workflow (being created)
-  const { isReadOnly } = useWorkflowReadOnly();
-  const isNewWorkflow = useIsNewWorkflow();
   const canApplyChanges = !isReadOnly || isNewWorkflow;
   const isWriteDisabled = !canApplyChanges;
 
@@ -683,7 +683,7 @@ export function AIAssistantPanelWrapper({
           <div className="flex-1 overflow-hidden">
             <AIAssistantPanel
               isOpen={isAIAssistantPanelOpen}
-              onClose={handleClosePanel}
+              onClose={isNewWorkflow ? undefined : handleClosePanel}
               onNewConversation={handleNewConversation}
               onSessionSelect={handleSessionSelect}
               onShowSessions={handleShowSessions}
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -14,6 +14,8 @@ const getRandomStatus = () =>
     Math.floor(Math.random() * INITIAL_LOADING_STATUSES.length)
   ];
 
+import { randomUUID } from '#/common';
+
 import { useURLState } from '../../react/lib/use-url-state';
 import {
   useMonacoRef,
@@ -564,7 +566,7 @@ export function AIAssistantPanelWrapper({
   const onValidationError = useCallback(
     (errorMessage: string) => {
       const message: Message = {
-        id: crypto.randomUUID(),
+        id: randomUUID(),
         role: 'assistant',
         content: errorMessage,
         status: 'error',
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -574,6 +574,10 @@ export function AIAssistantPanelWrapper({
     [aiStore]
   );
 
+  // Track whether we applied via streaming so the auto-apply effect in the
+  // hook can skip the duplicate when the final new_message arrives
+  const appliedViaStreamingRef = useRef(false);
+
   // Hook to handle workflow/job code application logic
   const { handleApplyWorkflow, handlePreviewJobCode, handleApplyJobCode } =
     useAIWorkflowApplications({
@@ -607,6 +611,7 @@ export function AIAssistantPanelWrapper({
       previewingMessageId,
       setApplyingMessageId,
       appliedMessageIdsRef,
+      appliedViaStreamingRef,
     });
 
   // Auto-preview job code when AI responds with code
@@ -626,9 +631,6 @@ export function AIAssistantPanelWrapper({
   const appliedStreamingChangesRef = useRef<Record<string, unknown> | null>(
     null
   );
-  // Track whether we applied via streaming so we can skip the duplicate
-  // auto-apply when the final new_message arrives
-  const appliedViaStreamingRef = useRef(false);
   useEffect(() => {
     if (!streamingChanges || !canApplyChanges) return;
     // Avoid re-applying the same streaming changes object
@@ -644,7 +646,6 @@ export function AIAssistantPanelWrapper({
     } else if (aiMode?.page === 'job_code' && 'code' in streamingChanges) {
       const code = streamingChanges['code'] as string;
       if (code) {
-        appliedViaStreamingRef.current = true;
         handlePreviewJobCode(code, '__streaming__');
       }
     }
@@ -655,22 +656,6 @@ export function AIAssistantPanelWrapper({
     handleApplyWorkflow,
     handlePreviewJobCode,
   ]);
-
-  // When a new assistant message with code arrives after we already applied
-  // via streaming, mark it as already applied to prevent duplicate auto-apply
-  // and update previewingMessageId to the real ID to prevent diff flicker
-  useEffect(() => {
-    if (!appliedViaStreamingRef.current) return;
-
-    const latestAssistantMessage = [...messages]
-      .reverse()
-      .find(m => m.role === 'assistant' && m.code && m.status === 'success');
-
-    if (latestAssistantMessage) {
-      appliedMessageIdsRef.current.add(latestAssistantMessage.id);
-      appliedViaStreamingRef.current = false;
-    }
-  }, [messages, appliedMessageIdsRef]);
 
   return (
     <div
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 const INITIAL_LOADING_STATUSES = [
   'Thinking about the question...',
@@ -26,6 +26,7 @@ import {
   useAISessionId,
   useAISessionType,
   useAIStore,
+  useAIStreamingApply,
   useAIStreamingChanges,
   useAIStreamingContent,
   useAIStreamingStatus,
@@ -574,9 +575,18 @@ export function AIAssistantPanelWrapper({
     [aiStore]
   );
 
-  // Track whether we applied via streaming so the auto-apply effect in the
-  // hook can skip the duplicate when the final new_message arrives
-  const appliedViaStreamingRef = useRef(false);
+  // Pending streaming apply record (YAML already imported during streaming),
+  // read by the hook's auto-apply effect to skip the duplicate import when
+  // the final new_message arrives with the same YAML
+  const streamingApply = useAIStreamingApply();
+  const streamingApplyActions = useMemo(
+    () => ({
+      set: aiStore._setStreamingApply,
+      setSaveFailed: aiStore._setStreamingApplySaveFailed,
+      clear: aiStore._clearStreamingApply,
+    }),
+    [aiStore]
+  );
 
   // Hook to handle workflow/job code application logic
   const { handleApplyWorkflow, handlePreviewJobCode, handleApplyJobCode } =
@@ -611,7 +621,8 @@ export function AIAssistantPanelWrapper({
       previewingMessageId,
       setApplyingMessageId,
       appliedMessageIdsRef,
-      appliedViaStreamingRef,
+      streamingApply,
+      streamingApplyActions,
     });
 
   // Auto-preview job code when AI responds with code
@@ -640,7 +651,8 @@ export function AIAssistantPanelWrapper({
     if (aiMode?.page === 'workflow_template' && 'yaml' in streamingChanges) {
       const yaml = streamingChanges['yaml'] as string;
       if (yaml) {
-        appliedViaStreamingRef.current = true;
+        // handleApplyWorkflow records the streaming apply in the store
+        // (after a successful import) so the final new_message can skip it
         void handleApplyWorkflow(yaml, '__streaming__');
       }
     } else if (aiMode?.page === 'job_code' && 'code' in streamingChanges) {
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -61,7 +61,7 @@ import {
   useWorkflowState,
 } from '../hooks/useWorkflow';
 import { useKeyboardShortcut } from '../keyboard';
-import type { JobCodeContext } from '../types/ai-assistant';
+import type { JobCodeContext, Message } from '../types/ai-assistant';
 import { Z_INDEX } from '../utils/constants';
 import {
   prepareWorkflowForSerialization,
@@ -548,6 +548,7 @@ export function AIAssistantPanelWrapper({
     startApplyingJobCode,
     doneApplyingJobCode,
     updateJob,
+    saveWorkflow,
   } = useWorkflowActions();
 
   // Get applying state from workflow store for disabling Apply button across all users
@@ -557,6 +558,20 @@ export function AIAssistantPanelWrapper({
   const isApplyingJobCode = useWorkflowState(state => state.isApplyingJobCode);
   const applyingJobCodeMessageId = useWorkflowState(
     state => state.applyingJobCodeMessageId
+  );
+
+  const onValidationError = useCallback(
+    (errorMessage: string) => {
+      const message: Message = {
+        id: crypto.randomUUID(),
+        role: 'assistant',
+        content: errorMessage,
+        status: 'error',
+        inserted_at: new Date().toISOString(),
+      };
+      aiStore._addMessage(message);
+    },
+    [aiStore]
   );
 
   // Hook to handle workflow/job code application logic
@@ -573,6 +588,8 @@ export function AIAssistantPanelWrapper({
           : null,
       currentUserId: user?.id,
       aiMode,
+      isNewWorkflow,
+      onValidationError,
       workflowActions: {
         importWorkflow,
         startApplyingWorkflow,
@@ -580,6 +597,7 @@ export function AIAssistantPanelWrapper({
         startApplyingJobCode,
         doneApplyingJobCode,
         updateJob,
+        saveWorkflow,
       },
       monacoRef,
       jobs,
--- a/assets/js/collaborative-editor/components/MessageList.tsx
+++ b/assets/js/collaborative-editor/components/MessageList.tsx
@@ -545,18 +545,34 @@ export function MessageList({
                 }
               >
                 <div className="space-y-3">
-                  <MarkdownContent
-                    content={
-                      isStreaming(message)
-                        ? message.content.replace(/\n+$/, '')
-                        : message.content
-                    }
-                    showAddButtons={
-                      !isStreaming(message) && showAddButtons && !message.code
-                    }
-                    isWriteDisabled={isWriteDisabled}
-                    className={PROSE_CLASSES}
-                  />
+                  {message.status === 'error' &&
+                  !isStreaming(message) &&
+                  message.content.trim() ? (
+                    <div
+                      className="rounded-lg border border-red-200 bg-red-50 px-3 py-2"
+                      data-testid="ai-validation-error"
+                    >
+                      <div className="flex items-start gap-2">
+                        <span className="hero-exclamation-circle h-4 w-4 text-red-600 flex-shrink-0 mt-0.5" />
+                        <p className="text-sm text-red-700 leading-relaxed">
+                          {message.content}
+                        </p>
+                      </div>
+                    </div>
+                  ) : (
+                    <MarkdownContent
+                      content={
+                        isStreaming(message)
+                          ? message.content.replace(/\n+$/, '')
+                          : message.content
+                      }
+                      showAddButtons={
+                        !isStreaming(message) && showAddButtons && !message.code
+                      }
+                      isWriteDisabled={isWriteDisabled}
+                      className={PROSE_CLASSES}
+                    />
+                  )}
 
                   {/* Status (e.g. "Generating code...") Apollo may stream
                       after the text answer, while we wait for code. Same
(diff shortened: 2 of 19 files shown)


nintyboy:
Hey @lmac-1 popped together some looms for design review: 

No Ai : https://www.loom.com/share/6d76a5aea6e54a0bab8a577b6d432d26
With AI: https://www.loom.com/share/3c53229a24324705aa288b6135bd9e8f

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -199,21 +215,88 @@ export function useAIWorkflowApplications({
...
+                          retryError instanceof Error
+                            ? retryError.message
+                            : 'Unknown error occurred',
+                      });
+                    }
+                  })();
+                },
+              },
+            });
+          }
```
elias-ba: Ah, I think we get two toasts here for a single failure. The `saveWorkflow` we call is the wrapped one from `useWorkflowActions`, and that wrapper already pops its own "Failed to save" alert (with a Retry) and then re-throws - so this `catch` lands a second one on top. The `{ silent: true }` only mutes the success toast sadly, not the error path. The disconnect case is actually fine, the wrapper returns `null` and we throw the connection-lost message ourselves. So maybe we keep just the `!saved` branch here and let the wrapper own the real errors ? Side note while I was in here: the new test stubs `saveWorkflow` as a bare reject, which skips the wrapper, so it never catches this.

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
```diff
@@ -559,6 +560,24 @@ export function AIAssistantPanelWrapper({
     state => state.applyingJobCodeMessageId
   );
 
+  const onValidationError = useCallback(
+    (errorMessage: string) => {
+      const message: Message = {
+        id: crypto.randomUUID(),
```
elias-ba: This one caught my eye. I'm fairly sure `crypto.randomUUID` will throw on a self-hosted HTTP box - it's only defined in a secure context (HTTPS or localhost), so on a plain LAN deploy it's `undefined`, and here it throws before the message ever gets added. So when the AI returns invalid YAML the user just sees... nothing. We already dodge this everywhere else with the `randomUUID()` helper in `common.js` (Math.random based, see `WorkflowDiagram.tsx` / `useConnect.ts`). Could we reuse that here ?

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
```diff
@@ -683,7 +686,7 @@ export function AIAssistantPanelWrapper({
           <div className="flex-1 overflow-hidden">
             <AIAssistantPanel
               isOpen={isAIAssistantPanelOpen}
-              onClose={handleClosePanel}
+              onClose={isNewWorkflow ? undefined : handleClosePanel}
```
elias-ba, line 703: Something I'd like to think through with you: I think a user can get trapped here if the auto-save fails. `isNewWorkflow` only flips to false on a successful save (`clearIsNewWorkflow`, `useWorkflow.tsx:436`), so while it keeps failing the panel has no close button (`onClose` is undefined), Cmd+K is off, and the canvas is gated behind the new-workflow guards. Their only escape is the Retry toast or a full reload. Is that what we want ? Feels like they should at least be able to close the panel or drop back to the landing screen when the save hasn't gone through. I honestly don't have a clean fix in mind - curious what you think.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -199,21 +215,88 @@ export function useAIWorkflowApplications({
...
+        applySucceeded = true;
+
+        if (isNewWorkflow) {
+          try {
+            const saved = await saveWorkflow({ silent: true });
+            if (!saved) {
+              throw new Error(
+                'Your connection was lost. Reconnect and try again.'
+              );
+            }
```
elias-ba: More of a question than a problem. Once the first save happens, `clearIsNewWorkflow()` runs and `isNewWorkflow` is false from then on. So the very first AI message auto-saves, but every refinement after it ("add a second job") imports onto the canvas and just leaves the unsaved dot - nothing persists until they hit Save. It matches the normal editor so it's defensible, but inside this AI-first flow it feels a little inconsistent, and I'm not sure an AI-first user would realise they need to save. Intended for now, or should refinements save too ? Probably a product call.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -415,6 +507,16 @@ export function useAIWorkflowApplications({
...
     ) {
+      appliedMessageIdsRef.current.add(latestMessage.id);
+
+      // Streaming already applied this YAML — skip the re-import to avoid
+      // a transient dirty state (Y.Doc write → unsaved red dot) between
+      // the import and save that would otherwise follow.
+      if (appliedViaStreamingRef?.current) {
+        appliedViaStreamingRef.current = false;
+        return;
+      }
```
elias-ba: Non-blocking, and honestly more of a nagging worry than a bug I could reproduce. This skip depends on the ref being cleared on every path. If streaming applies the YAML (ref = true) but the matching `new_message` never comes back as `success` + `code`, the ref stays true and the next unrelated response gets skipped right here, silently - no import, no error. It's basically the churn you called out in point 4, four reset sites is a lot to keep straight. A little state machine (or having `handleApplyWorkflow` return success and setting the ref from that) would make the worry go away. Totally happy to leave it for the follow-up though.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -199,21 +215,88 @@ export function useAIWorkflowApplications({
...
+          error instanceof Error ? error.message : 'Invalid workflow YAML';
+
+        if (isNewWorkflow && onValidationError) {
+          onValidationError(errorMessage);
+        } else {
+          notifications.alert({
+            title: 'Failed to apply workflow',
+            description: errorMessage,
+          });
+        }
```
elias-ba: I think invalid YAML ends up posting two identical error bubbles. Streaming tries it first, fails, and this catch resets the ref and fires `onValidationError` (bubble 1). Then the final `new_message` shows up with the same YAML, but the ref is false again so the skip guard at line 515 lets it through - it re-applies, fails, bubble 2. In the happy path that guard is exactly what stops the double apply, we just lose it on the failure path. Did you bump into this when you were testing the invalid-YAML case ?

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
```diff
@@ -224,9 +307,15 @@ export function useAIWorkflowApplications({
...
       jobs,
       setApplyingMessageId,
+      isNewWorkflow,
+      onValidationError,
+      saveWorkflow,
     ]
   );
 
+  // Keep ref pointing at the latest callback so Retry toast closures never go stale
+  saveWorkflowRef.current = saveWorkflow;
```
elias-ba: Tiny perf nit, non-blocking. `saveWorkflow` from `useWorkflowActions` is a fresh function every render (it's an IIFE at `useWorkflow.tsx:393` and the hook isn't memoized), so pulling it into these deps makes `handleApplyWorkflow` new each render, which makes the auto-apply effect tear down and re-run every render - even on every streaming token. No correctness issue since `appliedMessageIdsRef` dedupes, it's just churn. Since you already keep `saveWorkflowRef` around, maybe the callback reads `saveWorkflowRef.current(...)` and we drop `saveWorkflow` from the deps ?

assets/js/collaborative-editor/components/MessageList.tsx
```diff
@@ -545,18 +545,34 @@ export function MessageList({
...
+                    <div
+                      className="rounded-lg border border-red-200 bg-red-50 px-3 py-2"
+                      data-testid="ai-validation-error"
+                    >
+                      <div className="flex items-start gap-2">
+                        <span className="hero-exclamation-circle h-4 w-4 text-red-600 flex-shrink-0 mt-0.5" />
+                        <p className="text-sm text-red-700 leading-relaxed">
+                          {message.content}
+                        </p>
+                      </div>
```
elias-ba, line 560: I think this quietly drops the Retry button for genuine send failures that happen to have text. Before, any assistant message with `status: 'error'` got the "Failed to send / Retry" banner. Now anything with content lands in this read-only red box, and the banner underneath is gated on `!content.trim()`. So a generation that dies after streaming a bit of text (it gets marked error over in `AIChannelRegistry.ts:682`) shows up here with no way to retry. I wonder if an explicit field like `errorKind: 'validation' | 'transport'` would be safer than leaning on `content.trim()` to tell them apart ? Not a blocker for this PR though.

assets/test/collaborative-editor/hooks/useAIWorkflowApplications.workflow.test.ts
elias-ba, line 1: Last one, promise 😄 I only noticed because I was scrolling for a while - this file has crept up to 700 lines. That's quite big. Nothing for now, but maybe in the cleanup pass we peel the save / streaming cases out into their own file ?

elias-ba (changes requested):
Really nice work Lucy 🙏 The Build with AI flow feels good to use and the tests are genuinely thorough. I left a fair few inline comments but please don't be scared by the count - most are questions or non-blocking. Three I'd love us to sort before this merges:

- **Double toast on save failure** - the wrapped `saveWorkflow` already shows its own error toast and re-throws, so we get two for one failure.
- **`crypto.randomUUID`** - I think this can throw on a self-hosted HTTP deploy. We already have a `randomUUID()` helper in `common.js` for exactly this.
- **Getting stuck in the panel** - if the auto-save keeps failing, there's no close button and no way back to the canvas. Not sure of the best fix, keen to hear your thoughts.

Everything else is either a question for you / product (auto-saving refinements, the panel escape) or fits nicely into the cleanup issue you already planned, so no rush there.

And your point 4 about the ref churn is spot on - I came at it from a few different angles and kept landing on the same thing. I think a small state machine would kill a whole class of these quietly-skipped bugs. Really good instinct flagging it.

Not blocking the epic branch in my view once those three are handled. Thanks again, I really enjoyed reviewing this one 🙌

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
lmac-1: Hey @elias-ba , thanks for raising this. I double checked with product and they are happy with a generic response "Failed to save. Please retry or try again later. [Retry]" with no way to get out. Brandon said :

> "report this as a critical error in our telemetry system since if the AI assistant returned a response and this failed, that suggests there isn’t a general network problem - there is a problem on our end which requires our urgent attention" 

but I'm not sure we can trigger Sentry errors from the frontend? What do you think? I think for now it's something that we can work on to improve but shouldn't block things.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
lmac-1: Yeah that's intentional. We only auto-save when the workflow is returned so that the workflow is persisted in the database. Any changes after that will need to be manually saved. Autosave in general is going to be handled in a follow-up big piece of work.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
lmac-1: I have committed a fix to make this a lot cleaner and stem from the store. So I think this is okay now.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
lmac-1: Yes. I have it noted as a "nice-to-have" to fix. I spent some time today trying to improve this but it seems to be quite a gnarly problem and not straightforward. I think I will have to note it for a future improvement. Hopefully it won't be visible very often, and the user will 'just' see an extra toast message. Noisy but not the end of the world. I would rather that than rushing a fix that could have implications for other parts of the app using these silent saves. 

I have added a success toast though if the Retry passes, which I think is a better experience.

assets/js/collaborative-editor/hooks/useAIWorkflowApplications.ts
lmac-1: This should be fixed now. We are getting this logic from the store now to keep track of when code is applying.

assets/test/collaborative-editor/hooks/useAIWorkflowApplications.workflow.test.ts
lmac-1: Yes will note it to clean up in the cleanup PR, thanks!

lmac-1:
Hey @elias-ba thank you so much for your timely and thorough review. This was so helpful. 

Here's where I got to with your 3 items:
1. `crypto.randomUUID` — fixed, swapped to the `randomUUID()` helper from common.js (same one used in WorkflowDiagram.tsx / useConnect.ts)
2. **Double toast on save failure**: fixed for the single-failure case you flagged. It's still not 100% and I'd like to look at things in a bit more depth but I think the fix might affect more areas of the app so would rather not rush that. Will note it down to consider after the AI first stuff is done.
3. Getting stuck in the panel (see the comment for more detail), product is happy with the current no-escape behaviour for now (generic "Failed to save, Retry", no way out), with a note to add telemetry so we notice when it happens.

Also I cleaned up the `ref` stuff for applying with the yaml and it's much cleaner now. 

I think we are good to merge for now and work out the other bits later on. I'll come back to your comments when I'm doing a more general sweep. 

Thanks again.

elias-ba (approved):
Thanks Lucy, all three are sorted on my end. Approving 🙌

assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
elias-ba: On the Sentry question: we can report from the frontend. We already run `Sentry.init()` in `app.js` through `@sentry/react`, so a `Sentry.captureException` / `captureMessage` call is available client side. That means Brandon's idea of flagging this as a critical error is doable whenever we pick up the follow-up. Agreed it shouldn't block this one.

PR #4910, opened 2026-06-30
fix: handle NUL byte in output_dataclip on step:complete

### LINKED ISSUE #4893: step:complete crashes with Postgrex 22P05 when output_dataclip body contains a NUL byte
**Describe the bug**

When a worker reports `step:complete` with an `output_dataclip` whose JSON contains a NUL byte (`\u0000`), the dataclip insert crashes with an **unhandled** `Postgrex.Error` and the step never completes:

```
Postgrex.Error: ERROR 22P05 (untranslatable_character)
unsupported Unicode escape sequence
\u0000 cannot be converted to text.
```

Sentry: **LIGHTNING-13C** (48 events, first seen 2025-08-07, still occurring as of 2026-06-24; `prod`, `app.openfn.org`).

**Version number** `v2.16.6 (1.5.86)` (from the Sentry event); reproduced on current `main`.

- [x] Yes
- [ ] No

**Root cause**

The crash is in `Lightning.Runs.Handlers.CompleteStep`, inserting the `Dataclip` body (`lib/lightning/runs/handlers.ex:467-482`):

```elixir
Dataclip.new(%{
  id: dataclip_id,
  project_id: project_id,
  body: output_dataclip |> Jason.decode!() |> ensure_map(),  # <- NUL byte ends up here
  type: :step_result
})
|> Repo.insert()
```

Stacktrace (confirmed via Sentry):

```
CompleteStep.call/2:398
  -> update_step/2:410          (maybe_save_dataclip)
    -> Ecto.Repo.Schema.do_insert/4
      -> Ecto.Adapters.SQL.raise_sql_call_error/1
```

The worker sends literal JSON text containing the 6-character escape `\u0000` (valid JSON — this is *not* a JSON-validity problem). `Jason.decode!` turns that escape into a real NUL byte in the Elixir binary. PostgreSQL `jsonb`/`text` physically cannot store a NUL byte (0x00), so the insert raises. Because the error is unhandled, the whole `step:complete` reply fails and the step/run is left without a clean completion.

**To Reproduce**

A `step:complete` payload of the form (note `output_dataclip` carries the escape):

```json
{
  "step_id": "...",
  "run_id": "...",
  "project_id": "...",
  "reason": "success",
  "output_dataclip_id": "...",
  "output_dataclip": "{\"value\": \"foo\\u0000bar\"}"
}
```

A reproduction test (currently asserts the crash):

```elixir
test "reproduces LIGHTNING-13C: output_dataclip with a NUL byte crashes the insert" do
  # ... set up workflow/run/step ...
  output_dataclip = ~s({"value": "foo\\u0000bar"})
  assert Jason.decode!(output_dataclip) == %{"value" => "foo\0bar"}

  assert_raise Postgrex.Error, ~r/unsupported Unicode escape sequence/, fn ->
    Runs.complete_step(%{
      step_id: step.id, reason: "success",
      output_dataclip: output_dataclip,
      output_dataclip_id: Ecto.UUID.generate(),
      run_id: run.id, project_id: workflow.project_id
    })
  end
end
```

This passes — confirming the exact error class, code path, and message match Sentry.

**Expected behavior**

The step/run should complete successfully without crashing. NUL bytes (and other PostgreSQL-incompatible control chars) in the dataclip body should be handled gracefully rather than raising an unhandled error.

**Proposed handling**

There's already a precedent: `Lightning.LogMessage` (`lib/lightning/ecto_types.ex:63`) sanitizes exactly this for the `run:log` path — stripping NUL bytes, control chars, and literal `\u0000` escapes, replacing them with the Unicode replacement character. The dataclip body never got the same protection.

Options:
- **A — targeted:** sanitize the raw `output_dataclip` string before `Jason.decode!` in `maybe_save_dataclip`, reusing the `LogMessage` regex logic.
- **B — boundary (recommended):** a custom jsonb Ecto type (the map analogue of `LogMessage`) that recursively sanitizes string keys/values, applied to `Dataclip.body`. This protects **all** dataclip write paths (webhook-received dataclips, run input dataclips, step results), not just `CompleteStep`, which can hit the same NUL crash elsewhere.

Open product question: **sanitize-and-keep** (replace NUL, run completes — consistent with logs) **vs reject** (return an error reply; avoids silently mutating output but loses the whole output and risks stuck runs). Recommendation: sanitize, matching existing log behaviour.

**Additional context**

- `error_message` is commented out of the Step schema (`step.ex:55`) and not persisted, so it can't trigger this; the dataclip `body` (jsonb) is the only crashing path observed. `error_type`/`exit_reason` are short `:string` codes — same class, low risk.
- Retained Sentry events cluster suspiciously: 4 within ~12 min on 2026-06-23 and 5 within ~17 min on 2026-04-15, suggesting individual workflows repeatedly emitting NUL-containing output.

mvanhorn:
## Description

This PR fixes a crash on `step:complete`. When a worker reports an
`output_dataclip` whose decoded JSON contains a NUL byte (`0x00`),
`Lightning.Runs.Handlers.CompleteStep` raised an unhandled `Postgrex.Error`
(code `22P05`, `untranslatable_character`) while inserting the `Dataclip`,
because PostgreSQL cannot store `0x00` in `text`/`jsonb`. The raise propagated
out of `maybe_save_dataclip/2` and failed the whole reply, leaving the step/run
without a clean completion (Sentry LIGHTNING-13C).

The fix detects the unstorable NUL byte before the insert in the third
`maybe_save_dataclip/2` clause and returns `{:error, changeset}` instead of
letting `Repo.insert/1` raise. Detecting before the insert (rather than rescuing
the `Postgrex.Error`) keeps the surrounding `Repo.transact/1` transaction from
being poisoned, and `CompleteStep.call/2`'s existing `with` already propagates
the error tuple to the caller. Worker-side handling of the error response lives
in a separate repo and is out of scope.

Two details worth calling out:

- The check runs on the body that is actually persisted. `Dataclip.new/1`
  strips the unpersisted top-level `configuration` key via
  `remove_configuration/1`, so the guard reads
  `Ecto.Changeset.get_field(changeset, :body)` after that stripping. A NUL byte
  that lives only under `configuration` never reaches PostgreSQL and is allowed
  through.
- The returned error changeset is built without the dataclip body.
  `LightningWeb.ChannelHelpers.reply_with/2` sends `inspect(error)` to Sentry,
  so a changeset carrying `changes.body` would leak the decoded job output into
  Sentry. The error changeset carries only the `:body` error message.

Closes #4893

## Validation steps

1. `mix test test/lightning/runs_test.exs` - three new tests in the
   `describe "complete_step/2"` block cover: a NUL byte at the top level returns
   `{:error, %Ecto.Changeset{}}` (no raise), a NUL byte nested in an array
   element is also caught, and a NUL byte present only under top-level
   `configuration` still completes with `{:ok, _step}` (since that key is not
   persisted).
2. The existing `complete_step/2` happy-path tests still pass unchanged (normal
   output dataclips insert and complete as before).

1. The NUL detection (`contains_null_byte?/1`) walks decoded JSON maps, lists,
   and binaries, checking both keys and values, so it catches NUL bytes at any
   depth.
2. This is the Lightning-side half discussed on the issue thread (reject rather
   than mutate user data). The worker not yet respecting an error response on
   `step:complete` is tracked separately.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)
      (e.g., `:owner`, `:admin`, `:editor`, `:viewer`) - n/a, this is an
      internal data-persistence guard with no authorization surface


theroinaochieng:
Hey @mvanhorn thank you so much for your contribution! Our team is a little swamped at the moment but we'll check out your PR as soon as we can.

brandonjackson:
heya @mvanhorn, great catch. We've actually just spotted this on our end too. Could you tell us a bit about the user story / scenario that caused the issue for you? We're seeing it when working with spreadsheet. Curious to learn if other behaviours might trigger too.

mvanhorn:
Honest answer: I came at this from the Sentry error (LIGHTNING-13C) rather than a specific workflow I was debugging, so I don't have one clean repro to point at. The trigger is anything that lands a raw `0x00` byte in the decoded `output_dataclip` JSON, which Postgres then rejects for `text`/`jsonb`. Spreadsheet data is a very plausible source — some CSV/XLSX cells carry embedded NULs, and adaptors that stringify binary-ish fields can too. Other paths I'd expect to hit it: outputs from external APIs/DBs that embed NUL in string fields, and adaptors that base64-decode or read file bytes into state. Since the fix handles the NUL before the insert, it should cover all of those regardless of the exact source. Happy to add a targeted test around the spreadsheet case if that's the one you're seeing most.

midigofrank:
This is being blocked by https://github.com/OpenFn/kit/issues/1478 . Lets wait for the worker implementation

lmac-1:
@theroinaochieng does this still require my review or will this be handled by Frank? Seems quite Elixir focused. Happy to steer a Claude review as an extra pair of eyes but not sure if @doc-han might be able to add more value.

lmac-1:
@midigofrank is this one still blocked by https://github.com/OpenFn/kit/issues/1478?  

Thanks for merging @mvanhorn and thank you for your patience.

mvanhorn:
Rebased onto current main; the branch was conflicting on `CHANGELOG.md` only. Kept both sets of entries. The fix and its tests are unchanged.

lmac-1:
Thanks @mvanhorn ☺️

Issue #4876, opened 2026-06-17
AI-First Starting UX: Cleanup — remove old canvas entry points


lmac-1:
Depends on: #4856 #4868 #4870 #4871

## Phase 1 — Remove old entry points

No logic changes, UI cleanup only.

- [x] Remove the empty-canvas placeholder in `WorkflowEditor.tsx` (the
      "Create your workflow" state with Browse/AI/Import buttons) — the
      landing screen (#4856) always covers this state now
- [x] Remove the `?method=`-based URL sync and its supporting state in the UI
      store, now that the left-panel create flow is gone
- [x] Resolve all remaining `TODO-AI-FIRST` comments in code
- [x] Confirm no other left-panel entry points are left duplicating the
      landing screen
- [x] Canvas renders cleanly when empty

## Phase 2 — Known follow-up bugs from prior PRs

Small bugs/gaps surfaced while building this epic, not yet fixed:

- [ ] Failed auto-saves after a successful AI response aren't logged/reported
      anywhere — add error telemetry **deferred to #4981**
- [x] Save-retry toast can show stale/duplicate state when a save is retried
      automatically vs. by the user 
- [ ] The workflows REST API controller can crash instead of returning a
      proper error response for certain save failures **deferred to #4982**
- [x] Break up the oversized AI-workflow-applications test file
- [x] Test the cleanup against the global assistant, not just new-workflow

## Phase 3 — Design polish (stretch, may be pushed back)

Address Tyrell's design feedback:

- No AI: https://www.loom.com/share/6d76a5aea6e54a0bab8a577b6d432d26
- With AI: https://www.loom.com/share/3c53229a24324705aa288b6135bd9e8f
- Templates: https://www.loom.com/share/60adb69177b344fbbb73b0c6e1dbc909

Template modal notes: title weight, spacing top/bottom on cards, search bar
styling, whether the modal should reach the bottom of the page, and adding a
"Create" button (may be pushed back). Consolidating modals into a shared
component is a maybe — only pursue if it doesn't expand scope.

## Acceptance Criteria

- [ ] The empty-canvas placeholder in `WorkflowEditor.tsx` is removed
- [ ] No duplicate import entry points exist outside the landing screen
- [ ] Canvas renders cleanly when empty
- [ ] Decide whether phase 2 items are fixed or split into their own linked issues

PR #4912, opened 2026-06-30
AI-First Starting UX: Templates path

### LINKED ISSUE #4870: AI First Starting UX: Browse templates path - modal entry point, portability, and workflow creation
**Parent epic:** #4848 AI-First Starting UX
**Depends on:** #4856

## What this issue builds

Three concrete pieces:

1. **A portable template modal component** — accepts a templates array as a prop and an
   `onSelect` callback; knows nothing about workflow creation logic
2. **Template data wiring** — fetches templates (base + user-created) via the existing Phoenix
   Channel and passes them into the modal component
3. **Workflow creation on card click** — selecting a template applies it to the canvas and
   saves the workflow to the DB, ending pink toolbar state

**What is not changing:** the template data source (`list_templates` channel message +
`WorkflowTemplates.list_templates/0`), the `importWorkflow` function in `createWorkflowStore.ts`,
the `validate_workflow_name` channel message, or the template YAML content itself.

**TBC: The designs for the modal may change**

## User flow

1. User clicks "Browse templates" on the landing screen
2. A modal opens: "Browse Workflow Templates" — search bar, flat list of template cards
3. User can search/filter the list
4. Clicking a card:
   - Modal closes
   - Canvas is populated with the template's structure
   - Workflow is saved to the DB (first DB write, same as AI path auto-save)
   - Pink toolbar ends: URL updates to `/projects/{id}/w/{workflow_id}`, Save becomes enabled
   - Workflow is named after the template with uniqueness enforced (existing behaviour)
5. Closing the modal via the X button returns the user to the landing screen — no navigation
6. Pressing browser back while the modal is open navigates away from `/new` entirely (standard
   modal behaviour — the modal does not push a history entry)

After creation, pressing browser back goes to wherever the user came from (workflow list), not
back to `/new`. This is consistent with the AI path: `replaceState` swaps `/new` → `/w/{id}`
without adding a history entry.

### Component design: portability first

The modal component should be a **pure presentational component** — it receives data as props
and fires callbacks. It has no knowledge of channels, workflow state, or creation logic.

```ts
interface WorkflowTemplateBrowserProps {
  templates: TemplateItem[]
  onSelect: (template: TemplateItem) => void
  onClose: () => void
}
```

Where `TemplateItem` is a minimal type:

```ts
interface TemplateItem {
  id: string
  name: string
  description?: string
  code: string  // raw YAML string
}
```

This matches the shape already returned by `fetchTemplates` in
`assets/js/collaborative-editor/api/templates.ts` (which calls `'list_templates'` over the
channel) and the `BaseTemplate` type in
`assets/js/collaborative-editor/constants/baseTemplates.ts`.

**Do not pass workflow store references, save functions, or creation callbacks into this
component.** The `onSelect` callback receives the raw template — the parent handles everything
else. This is the portability contract: the component can be dropped anywhere and only needs
a templates array and two callbacks.

### Template data source (existing, reused)

Templates are already fetched via the Phoenix Channel:

- `fetchTemplates` in `assets/js/collaborative-editor/api/templates.ts` sends
  `'list_templates'` over the channel
- The channel handler in `lib/lightning_web/channels/workflow_channel.ex:608` calls
  `Lightning.WorkflowTemplates.list_templates/0` and returns rendered templates
- Base templates (webhook + cron) live in
  `assets/js/collaborative-editor/constants/baseTemplates.ts`
- `TemplatePanel.tsx:36-277` shows the existing merge pattern: base templates are prepended
  to user templates before rendering

The parent of the new modal should fetch templates on mount using the same `fetchTemplates`
call and merge them with the base templates in the same order. No new channel messages or
backend changes are needed.

### Workflow creation on card select

When `onSelect` fires with a template, the parent component should:

1. Call `parseWorkflowYAML(template.code)` (from `assets/js/yaml/util.ts:291`) to parse YAML
   into a `WorkflowSpec`
2. Call `convertWorkflowSpecToState(spec)` (`assets/js/yaml/util.ts:157`) to convert to a
   `YAMLWorkflowState`
3. Call `workflowStore.importWorkflow(workflowState)` — this already:
   - Sends `'validate_workflow_name'` over the channel (`workflow_channel.ex:451`) to enforce
     uniqueness (appending " 1", " 2" etc. if needed)
   - Calls `YAMLStateToYDoc.applyToYDoc(ydoc, validatedState)` to write into Y.Doc
   - Fires `flowEvents.dispatch('fit-view')` to frame the canvas
4. After `importWorkflow` succeeds, call `saveWorkflow({ silent: true })` — same auto-save
   pattern as the AI path (Issue 2, `hooks/useAIWorkflowApplications.ts:171-228`)

This is the same sequence the existing `TemplatePanel.tsx` uses for steps 1-3, and the same
auto-save pattern Issue 2 adds for step 4. No new logic is needed — only the wiring.

### Naming behaviour

The workflow name comes from the YAML `name:` field in the template's `code`. The
`validate_workflow_name` channel message enforces uniqueness within the project (existing
behaviour). **The modal component does not touch naming** — it passes the raw template to
`onSelect` and the existing `importWorkflow` + channel validation handles the rest.

Do not add any rename event listeners, name-update callbacks, or name-watching logic to the
modal component. The name should not be connected to the modal in any observable way.

### Modal open/close state

The modal is controlled by a boolean state variable in the landing screen component (Issue 1).

- "Browse templates" card click → sets `isTemplateBrowserOpen = true`
- X button in modal → sets `isTemplateBrowserOpen = false` (returns to landing screen)
- Card click → triggers `onSelect` → after workflow creation, navigates to canvas (landing
  screen is no longer rendered)

The modal does **not** call `pushState` or add a history entry. Pressing browser back while
the modal is open will navigate away from `/new` by natural browser behaviour — this is
intentional and consistent with standard modal behaviour anywhere else in the app.

### No canvas preview on hover

The existing `TemplatePanel` previews templates on the canvas as the user browses. **This
behaviour is not being built.** Clicking a card is the only action. Do not wire up
`uiStore.selectTemplate`, `TemplateDetailsCard`, or any preview mechanism.

## Files to modify / create

| File | Change |
|---|---|
| New: `components/WorkflowTemplateBrowser.tsx` | Portable modal component — accepts `templates`, `onSelect`, `onClose` props |
| `assets/js/collaborative-editor/api/templates.ts` | No change — reuse `fetchTemplates` as-is |
| `assets/js/collaborative-editor/constants/baseTemplates.ts` | No change — reuse base templates as-is |
| Landing screen component (Issue 1) | Fetch templates on mount; control modal open state; wire `onSelect` to `importWorkflow` + `saveWorkflow` |
| `assets/js/yaml/util.ts` | No change — reuse `parseWorkflowYAML` (line 291) and `convertWorkflowSpecToState` (line 157) |

**Note:** `TemplatePanel.tsx`, `TemplateCard.tsx`, `TemplateSearchInput.tsx`, and
`TemplateDetailsCard.tsx` are not modified. The new component is written fresh and is
intentionally not a fork of `TemplatePanel.tsx` — that component carries canvas preview
logic, selection state, and left-panel-specific concerns that do not belong here.

## Out of scope

- Template category filters (Event-based, Scheduled) — flat list only; categories may come later
- Canvas preview on hover
- "Build with AI" fallback from the search bar (present in `TemplateSearchInput.tsx` — not
  replicated here)
- Any changes to the template data model or backend
- `TemplatePanel.tsx` and the existing left-panel browsing flow — untouched

### Modal opens from landing screen

- [x] Clicking "Browse templates" on the landing screen opens the template modal
- [x] The modal is visible over the landing screen
- [x] The user is still in pink toolbar state while the modal is open (URL `/new`, Save
      disabled, no DB record)
- [ ] The modal renders a search bar and a flat list of template cards (base templates +
      user-created templates, same set as existing template browser)

### Search

- [ ] Typing in the search bar filters the visible template cards
- [ ] Clearing the search bar restores the full list
- [ ] Search is case-insensitive

### Card selection creates the workflow

- [x] Clicking a template card closes the modal
- [x] The canvas is populated with the template's workflow structure
- [x] The workflow is saved to the DB automatically (no "Create" button, no manual save)
- [x] Pink toolbar ends: URL updates to `/projects/{id}/w/{workflow_id}` without a page reload
- [x] Save becomes enabled after creation
- [x] The workflow is named after the template (with uniqueness enforced — e.g. "Event-based
      workflow 1" if the name already exists in the project)
- [x] The workflow name is not updated or watched by the modal at any point

### Closing the modal

- [x] Clicking the X button (or equivalent close action) closes the modal and returns the user
      to the landing screen
- [x] No workflow is created when the modal is closed without selecting a template
- [x] The user remains in pink toolbar state after closing

### Back navigation

- [ ] Pressing browser back while the modal is open navigates away from `/new` entirely —
      consistent with standard modal behaviour
- [ ] After workflow creation, pressing browser back goes to wherever the user came from
      (e.g. workflow list) — not back to `/new` or the modal

### Portability

- [x] The modal component receives templates as a prop array — it does not fetch data itself
- [x] The modal component has no direct dependency on workflow store, channel, or save logic
- [ ] `onSelect` and `onClose` are the only callbacks — no other workflow-state coupling

lmac-1:
## Description

This PR wires up the "Browse templates" flow for the AI-first epic. When a user clicks "Browse templates" on the landing screen, a modal opens showing base templates and any saved user templates. Clicking a template card closes the modal, navigates to the canvas and creates a new workflow in the database - unlike the existing templates flow which only populates the editor client-side to preview.

The modal is built as two components by design: `TemplateBrowserModal` (pure presentational, plain props) and `TemplateBrowserModalWrapper` (Phoenix channel + store wiring). This keeps the UI easy to iterate on without needing the full Lightning infrastructure. Product has requested to make this part of the app as decoupled as possible so that they can easily iterate on templates and improve the UI very easily. 

### Key decisions
- Base templates are always shown regardless of search query (same as existing behaviour)
- Search behaviour follows the same pattern as the existing flow
- No debounce - filtering is in-memory so live-on-keystroke is fine

Closes #4870

## Validation steps

1. Go to Create a new workflow
2. Click Browse templates
3. Opens modal with templates
4. Search for a template and leave something written in search
5. Check that you can close the modal with X and Esc
6. Reopen modal and make sure nothing is in search bar
7. Searching filters templates on title/description/tags
8. Clicking on a template will close the modal, create a new workflow in the database automatically and show you the workflow on the canvas

1. `TemplateBrowserModal` has no Lightning imports — it can be rendered in isolation (Storybook, tests) without a Phoenix channel or Y.Doc store. `TemplateBrowserModalWrapper` is the only place that touches hooks/store, following the `AIAssistantPanel` / `AIAssistantPanelWrapper` pattern already in the codebase.
2. We will be cleaning up the dead code in #4876. I've kept this PR tight to make it easier to review. 
3. Known issue: if `saveWorkflow` fails after `importWorkflow` has already run (e.g. a network error mid-select), the canvas is left with the template's nodes loaded but no database record behind them and creates a weird landing page state. This affects both the template browser and the existing YAML import flow. Will fix in a follow-up.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lmac-1:
@stuartc Hey Stu, this is ready for review, hopefully should be straightforward. I've done a couple of Claude reviews. Don't merge when/if you approve as there's an order to the PRs (this is a child of the Build with AI stuff). Thanks!

nintyboy:
Hey @lmac-1  I've popped together a loom for the design review: https://www.loom.com/share/60adb69177b344fbbb73b0c6e1dbc909

stuartc:
**On the `historyPatch` switch in `useWorkflow.tsx` (nit about the seam, not the behaviour)**

Good catch moving off raw `replaceState` here. Just to make the reasoning explicit for anyone reading later: `replaceState` left `liveSocket`'s internally-tracked href stale at `/w/new`, so on a socket reconnect the LiveView rejoined against the old URL, remounted `Collaborate.mount` with no `id`, and `workflow_assigns/2` minted a fresh ephemeral workflow UUID, silently detaching the user from the workflow they'd just saved. So keeping the tracked href in sync is the right fix.

One approach nit: this is now the only `window.liveSocket` reference in the whole collaborative editor, which we've deliberately kept LiveView-unaware (it's our one all-React island). We already have the seam for exactly this: the `ReactComponent` hook injects a `navigate` callback (`assets/js/react/hooks/react-component.tsx:73`) that's exposed through `LiveViewActionsContext` next to `pushEvent`/`handleEvent`. It just currently hardcodes `"replace":false`.

Could we route this through that instead? Two small changes:

- Extend the injected `navigate` to take `navigate(path, { replace?: boolean })` (parametrise the hardcoded flag in `react-component.tsx`, and widen the type at `LiveViewActionsContext.tsx:14`).
- Call `navigate(newUrl, { replace: true })` via `useLiveViewActions()` here instead of `window.liveSocket`.

That keeps the editor `liveSocket`-free and the save flow testable without a live socket. One behavioural note to decide on: the injected `navigate` uses `execJS`'s `patch`, which re-fires `handle_params` (harmless for `Collaborate` — it only loads run data), whereas your current `historyPatch` call is client-only. If you'd rather preserve the exact current behaviour, the injected callback can just call `liveSocket.historyPatch(path, replace ? 'replace' : 'push')` internally instead of `execJS`. Either way the `liveSocket` reference ends up in `assets/js/react/hooks/`, where it belongs, rather than in the editor.

Not blocking, more of an architecture-consistency thing since we're early in the epic.

lmac-1:
Thanks @stuartc, this is really great feedback. Will take a look and make things more consistent.

stuartc:
**The double-submit handling here is better than our other modals — worth lifting into a shared hook**

Quick positive callout: `handleSelect`'s `if (isSaving) return` plus the `isSaving` disabling of the cards is more robust than what most of our create/save handlers do. `YAMLImportModal.handleSave`, for instance, has no re-entrancy check at all, just a disabled button via its `importState` machine. Rather than let these drift, I think your approach is the one to standardise on.

I'd suggest extracting a small hook, `useActionLock`:

```ts
import { useCallback, useRef, useState } from 'react';

/**
 * Guards an async action against re-entrant calls: `run` invokes `fn` unless a
 * previous invocation is still in flight, and `isPending` mirrors the in-flight
 * state for disabling UI. `fn` is expected to handle its own errors; rejections
 * pass through to the caller, and the lock is always released.
 *
 * Not for locks that outlive the promise (e.g. useRunRetry's
 * WebSocket-confirmed submit) — this releases when `fn` settles.
 */
export function useActionLock<A extends unknown[], T = void>(
  fn: (...args: A) => Promise<T>
): { run: (...args: A) => Promise<T | undefined>; isPending: boolean } {
  const fnRef = useRef(fn);
  fnRef.current = fn;

  const pendingRef = useRef(false);
  const [isPending, setIsPending] = useState(false);

  const run = useCallback(async (...args: A): Promise<T | undefined> => {
    if (pendingRef.current) return undefined;
    pendingRef.current = true;
    setIsPending(true);
    try {
      return await fnRef.current(...args);
    } finally {
      pendingRef.current = false;
      setIsPending(false);
    }
  }, []);

  return { run, isPending };
}
```

The wrapper then loses its manual flag entirely:

```ts
const { run: handleSelect, isPending: isSaving } = useActionLock(async (template: Template) => {
  const spec = parseWorkflowYAML(template.code);
  const state = convertWorkflowSpecToState(spec);
  await importWorkflow(state);
  const saved = await saveWorkflow({ silent: true });
  if (!saved) {
    notifications.alert({ title: 'Not connected', description: 'Connect to the server before creating a workflow.' });
    return;
  }
  closeTemplateBrowserModal();
  dismissLandingScreen();
});
```

A few things it bakes in: the check is a synchronous `useRef` so a fast double-click can't slip through between render flushes (a plain state read can); the reset lives in one `finally` instead of your three `setIsSaving(false)` calls; `run` is `useCallback`-stable so it's safe to hand to `useKeyboardShortcut`; and it passes the result through so callers that want it later aren't blocked.

Deliberately keeping this at the component layer, not in `useWorkflowActions`/the store: the server already serialises saves through the session GenServer and upserts by the fixed client-held UUID, so a double-click can't create two workflows. What we're preventing is interleaved `importWorkflow` Y.Doc writes and duplicate toasts, which is a caller concern. I'd also keep `useRunRetry` and the AI apply path away from it (their locks aren't promise-scoped / solve a different cross-user problem), which the doc comment notes.

Totally fine as a follow-up rather than something you block this PR on.

---

**Opportunities to bring things in line**

This generalises well beyond this modal, so once the hook exists there are a few spots we could bring in line as we touch them:

- `components/GitHubSyncModal.tsx:73-78` — the one worth prioritising: Cmd/Ctrl+Enter calls `handleSaveAndSync` with no guard and the textarea is never disabled, so holding the shortcut fires concurrent save-and-syncs. User-reachable today.
- `components/inspector/TemplatePublishPanel.tsx:78-140` — disable-only, no early-return guard; in create mode a double-fire of `publish_template` could create duplicate templates (nothing upserts it the way workflow saves are protected).
- `components/inspector/trigger/TriggerEditWizard.tsx:89-95` — `finish` has no guard and the button isn't disabled during `commit()`, so a double-click double-commits.
- `SelectedDataclipView.tsx:39-58` and `WorkflowSettings.tsx:81-99` — lower risk, but near-zero-cost to adopt.

None of these are yours to fix here — just noting them so the hook lands as a shared tool rather than a one-off, and we chip away at the rest opportunistically.

stuartc (approved):

lmac-1:
Thanks @stuartc for this. I applied your feedback and got things working and raised https://github.com/OpenFn/lightning/issues/4917 as a follow up to do after AI first on a rainy day.

lmac-1:
@stuartc can you just put your eyes on this commit https://github.com/OpenFn/lightning/pull/4912/changes/5c4a9f57bce3a46835eacf41b26c0e25220ff47f ? 

It required me having to wrap a number of tests in `LiveViewActionsProvider` and mocking some stuff and just wanted to make sure that's okay. Alternative would be to have the `saveWorkflow` in a separate hook but I didn't like that.. I feel like everything should be inside `useWorkflow`

PR #4918, opened 2026-07-02
AI-First Starting UX: Build from scratch

### LINKED ISSUE #4895: AI-First Starting UX: Build from scratch
This is part of the #4848 epic

Clicking Build from Scratch card on /new will:
1. Create a workflow in the database with a webhook trigger and one job (the same workflow that the user gets when they currently select "Event-based template" in the old templates flow in "Create a workflow" 
2. This will open the canvas with the trigger picker component visible. The workflow will not be in pink toolbar state and the Save button will be visible.
3. Ideally, the trigger picker component will be on the state of "What triggers this workflow?" state but this is not necessary if it makes it architecturally messy

lmac-1:
Adds the "Build from scratch" path from the new-workflow landing screen (part of the #4848 AI-first starting UX epic).

Clicking the "Build from scratch" card:

1. Seeds a blank workflow (a webhook trigger, one "Untitled step" job on `@openfn/language-common@latest`, and an always-on edge) directly into the already-open collaborative session's Y.Doc, via the same `importWorkflow` → `saveWorkflow` client path that "Browse templates" and "Import YAML" already use on this screen. There's no server-side creation endpoint.
2. `saveWorkflow`'s own success handler rewrites the URL from `/w/new` to the real `/w/<workflowId>` in place (`replace: true`, no full-page navigation).
3. Because the workflow is already persisted by the time the landing screen dismisses, the canvas opens in the normal saved state: no pink unsaved toolbar, Save button visible.
4. The trigger inspector opens on the normal read-only show panel, same as any other trigger.

Workflows are named "Untitled workflow", "Untitled workflow 1", etc. via `Workflows.unique_workflow_name/2`, invoked through the client's existing `validate_workflow_name` channel event (the same one templates/YAML import already trigger inside `importWorkflow`). This also fixes a bug where clicking the card a second time in the same project failed on the name unique constraint.

Double-clicks are prevented client-side: `useActionLock` disables the card while a save is pending.

Closes #4895

1. **This PR changed approach since Frank's review:** the original version created the workflow server-side (`Workflows.create_webhook_workflow/2`) and navigated to the canvas with a `trigger_view=picker` param to auto-open the trigger inspector. Frank pointed out this didn't need its own server endpoint when the client-side `workflowStore` path already covers creation for templates and YAML import — this description reflects that rewrite, not the original implementation.
1. Creation goes through the same collaborative-session `save_workflow` pipeline templates/YAML import use, so snapshots, `lock_version`, and audit behave like any other workflow save.
2. The workflow-activation limiter needs no special handling here: `Session.handle_call({:save_workflow, ...})` already runs `maybe_disable_triggers_on_limit/1` on every collaborative save, so a project at quota gets its new trigger auto-disabled after save, the same as any other save on this path.
3. Failure to save shows a "Failed to create workflow" flash and keeps the user on the landing screen.
4. **Known gap, not introduced by this PR:** a viewer's channel join for a new workflow is rejected server-side (`WorkflowChannel.join/3` requiring `:create_workflow`), so a viewer can never persist a workflow — but the rejection currently produces no visible error. The vendored `y-phoenix-channel` provider only registers an `"ok"` receiver on `channel.join()`, not an `"error"` one, so the join rejection never reaches the client. The landing screen isn't gated on join success, so a viewer sees a normal, clickable card that silently does nothing on click. This is shared with templates/YAML import (same join, same missing receiver) — flagging for a product/UX call rather than fixing here.

## Validation steps

1. Open the new workflow screen in a project — the landing screen appears.
2. Click **Build from scratch**.
3. Land on the canvas of a saved "Untitled workflow" with a webhook trigger connected to a job, toolbar in the normal saved state. The trigger inspector should not be open — clicking the trigger node opens the normal read-only show panel.
4. Repeat in the same project: the second workflow should be created as "Untitled workflow 1" instead of failing.
5. Delete one and build from scratch again: the freed name is reused.
6. Double-click the card rapidly: only one workflow should be created, and the card should grey out while pending.
7. As a project **viewer**: the card is still visible and clickable, but clicking it should never actually create a workflow. There is currently no visible error message for this case (see note 4 above).
8. Press Back from the canvas: you should return to wherever you were before the landing screen, not back to it.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lib/lightning/workflows.ex
```diff
@@ -135,8 +136,7 @@ defmodule Lightning.Workflows do
           keyword()
         ) ::
           {:ok, Workflow.t()}
-          | {:error, Ecto.Changeset.t(Workflow.t())}
-          | {:error, :workflow_deleted}
+          | {:error, term()}
```
lmac-1: This came out of a personal Claude review saying that this definition wasn't correct, but not sure.. so please shout if this is wrong.

assets/js/collaborative-editor/CollaborativeEditor.tsx
```diff
@@ -190,6 +195,27 @@ function LandingScreenWrapper({
...
     openAIAssistantPanel,
   } = useUICommands();
+  const { pushEventTo } = useLiveViewActions();
+  const { run: runBuildFromScratch, isPending: isBuildingFromScratch } =
+    useActionLock(async () => {
+      try {
+        await new Promise<void>((resolve, reject) => {
+          const timeout = setTimeout(
+            () => reject(new Error('build_from_scratch timed out')),
+            10_000
```
lmac-1: `useActionLock` requires a promise, so we create a timeout in order to only build from scratch if successful.

assets/js/collaborative-editor/CollaborativeEditor.tsx
```diff
@@ -190,6 +195,27 @@ function LandingScreenWrapper({
...
+          );
+          pushEventTo('build_from_scratch', {}, () => {
```
lmac-1: Claude tells me it's okay to use `pushEventTo` here in collaborative editor but would like confirmation that it's the right pattern to do stuff like that here.

lib/lightning/workflows.ex
midigofrank: Hmm, did it say specifically which definition was wrong? `term()` means any type, which includes `:workflow_deleted` and `Ecto.Changeset`

lmac-1: So according to Claude, the old spec was wrong, `save_workflow` has three paths that don't return Workflow changesets: 
- `{:error, :workflow_deleted}` (line 331)
- `{:error, false}` on snapshot failure (line 373)
-  a generic Multi catch-all that can return some other schema's changeset (line 376-381).

So `Ecto.Changeset.t(Workflow.t())` alone never covered all cases.

The real fix would be spelling out `Ecto.Changeset.t(Workflow.t()) | Ecto.Changeset.t() | :workflow_deleted | false`. Claude went with `term()` instead since Dialyzer treats that union no differently and no caller pattern-matches on the specific shape — but happy to change it to the precise union if you'd rather document it explicitly. What's your preference?

elias-ba (approved):
Amazing work @lmac-1, you are a rockstar 🎸

midigofrank (changes requested):
Hey @lmac-1 , this looks great. Two things from the elixirland:

1. Instead of blanking all the errors paths as `term()` maybe we could just add the missing ones? So something like `{:error, Ecto.Changeset.t() | :workflow_deleted | false}`. We should change the `false` into something readable like `:snapshot_failed`
2. Do we really need to define the `create_webhook_workflow` function on the server side? This is way too specific. Given we already have an api for creating workflows from the workflowStore, I would very much prefer to just have this go through that path.

lmac-1:
Thanks @midigofrank 

On 1 - noted and done. While checking every place that reads the result of `save_workflow`, I found one more spot: `workflows_controller.ex` (the REST API for creating/updating workflows). It has a big list of "if the error is X, send back this response" checks, but there's no check for "workflow was deleted" or "snapshot failed" — so if either of those ever happens through the API, the request would crash instead of returning a normal error message. This bug already existed before this PR for the "deleted" case - I didn't introduce it, and this PR doesn't touch that file. I'd rather not fix it here since it's unrelated to what this PR is doing. I've filed #4960 to keep track. 

On 2 - this had me going around in circles all day. Yes I agree it makes it simpler but it made the implementation a bit more complicated. I decided to deprioritise having the "Trigger picker" component open on "What triggers this workflow" when you build from scratch. I will try to build this in later but it was getting really messy and I need to move forward. I've applied the changes so it goes through `importWorkflow` -> `saveWorkflow` and uses a new base template in `baseTemplates` file. I've updated PR description to reflect the new approach so it can be understood what happened. I also cleaned up any code related to having trigger inspector open.

midigofrank:
Hey @lmac-1 , great catch. Opening an issue is the right call here. 

Regarding 2, I think that's fine

midigofrank (approved):

lmac-1:
Thanks @midigofrank! Appreciate you reviewing this one quickly. Merging 🚀

Issue #4938, opened 2026-07-06
Record sandbox workflow lineage across a fork


elias-ba:
**As a** maintainer of the sandbox merge, **I want** each sandbox workflow to carry a stable link to the parent workflow it was cloned from, **so that** merges, deletions, and renames are resolved by identity instead of guessed from names.

Follow-up to #4919 (the fix there is a safe interim) and part of #4852 (Sandbox DevX epic).

## The gap

When a sandbox is provisioned, `create_sandbox_workflows` builds a map of parent workflow id to new sandbox workflow id, uses it only to wire up the clone, and discards it. Each sandbox workflow ends up with a fresh id and no stored link to its origin, so at merge time `MergeProjects.merge_project` re-matches the two projects by workflow name.

The lineage we need exists at fork time and we throw it away. Working from names alone, the merge cannot tell a workflow added to the parent after the fork from one deleted in the sandbox (the #4919 bug), cannot follow a workflow through a rename, and falls back to name-and-structure heuristics for node and divergence matching. Persisting the origin link fixes these at the root.

## Approaches

Ways to persist the link, lightest to most capable:

- **Option A — a column on `workflows` (recommended).** A nullable `source_workflow_id` referencing the origin workflow, set in `create_sandbox_workflows` from the map it already builds. Smallest change, and it gives per-workflow identity, so it covers both the deletion distinction and rename-following.
- **Option B — a `workflow_lineages` table.** One row per clone (sandbox workflow, origin workflow, forked-at). Same per-workflow identity as A, with room for multi-hop history later, but heavier than this needs today.
- **Option C — a fork manifest on the sandbox.** Just the set of origin workflow ids captured at fork. Lightest, and enough to tell added-after-fork from deleted-in-sandbox, but it carries no per-workflow identity, so it does not help renames or node matching.

Recommendation: Option A. A and B meet the full goal; C is a lighter alternative if only the deletion distinction is wanted.

## How the merge uses it

With per-workflow identity (A or B), the merge resolves the parent/sandbox correspondence by the stored link first and falls back to name matching only for sandboxes created before this ships. That retires the `workflow_added_after_fork?` timestamp guess and lets a rename be matched in place instead of read as a delete plus a create.

## Open decisions

- **Existing sandboxes** have no stored lineage. Decide between a one-time best-effort backfill (by name and fork time) and keeping the name fallback for old sandboxes while new ones get exact lineage.
[...]

PR #4940, opened 2026-07-06
Sandbox merge no longer deletes a parent workflow added after the fork

### LINKED ISSUE #4919: Sandbox merge deletes a parent workflow added after the sandbox was forked
**As a** user merging a sandbox back into its parent project, **I want** a workflow that was added to the parent after I forked the sandbox to be kept by default, **so that** a normal select-all merge does not silently delete a workflow the sandbox never touched.

Related to #4852 (Sandbox DevX epic), but this is a live data-loss bug in the already-shipped sandbox merge UI and can ship to main on its own.

## What's broken

A workflow that exists in the parent project but not in the sandbox is treated as "deleted in the sandbox". In `build_merge_workflow_list` (in the sandbox live index), any target workflow whose name is absent from the sandbox is labeled `is_deleted: true`, tagged "Deleted in sandbox", and marked `is_changed: true`, which makes it pre-checked in the merge modal by default. A workflow added to the parent after the sandbox was forked is indistinguishable, in this code, from one genuinely deleted in the sandbox. So a default select-all merge soft-deletes a legitimately new parent workflow.

## How it surfaces

Fork a sandbox, then add a new workflow directly to the parent project. Open the merge modal from the sandbox. The parent-added workflow appears checked with the label "Deleted in sandbox". Proceeding with the default selection deletes it from the parent.

## What to fix

- Default target-only workflows to unchecked (keep), with a neutral label such as "Exists in the project but not this sandbox. Check to delete it from the project."
- The per-workflow checkbox already controls keep versus delete (a deselected target-only workflow is passed through, not deleted), so no fork-time lineage logic is required, only the default and the label.
- Guard the select-all default path so an untouched target-only workflow is not swept into deletion.
- Update `merge_projects_test` to cover keep-by-default for a parent-added workflow.

## Acceptance criteria

- Given a workflow added to the parent after forking, when I open the merge summary, then it is listed unchecked with a neutral label (not "Deleted in sandbox").
- Given I proceed with the default selection, then that workflow is not deleted.
- Given I explicitly check it, then deliberate deletion still works.

elias-ba:
## Description

This fixes a data-loss bug in the sandbox merge. A workflow added to a project after a sandbox was branched was flagged for deletion and pre-selected on the merge screen, so a routine merge could silently delete it.

The merge screen now excludes those workflows entirely: a workflow added to the project after the fork was never part of the sandbox, so it does not appear and the merge never touches it. Workflows deleted inside the sandbox still appear, default to unchecked, and are removed only if you explicitly select them, so deletion is opt-in and anything left unchecked is kept.

Closes #4919

## Validation steps

**Delete a workflow in the sandbox**

1. Create a project with one workflow, then create a sandbox from it.
2. In the sandbox, delete the workflow.
3. Open the sandbox's merge screen and select the project as the target.
4. Check that the workflow is listed as "Deleted in sandbox" and unchecked.
5. Merge without checking it. Check that the workflow still exists in the project.
6. Open the merge screen again, tick the workflow, and merge. Check that the workflow is now gone from the project.

**Add a workflow to the project after branching**

1. Create a project, then create a sandbox from it.
2. In the project, create a new workflow.
3. Open the sandbox's merge screen and select the project as the target.
4. Check that the new workflow is not listed.
5. Merge. Check that the new workflow still exists in the project.

**Merge normally**

1. In a sandbox, create a workflow and edit an existing one.
2. Open the merge screen and select the project as the target.
3. Check that both workflows are ticked.
4. Merge without changing the selection. Check that the new workflow is created in the project, the edit is applied, and no workflow was deleted.

Telling a workflow added to the project after the fork apart from one deleted in the sandbox is inferred from the workflow's creation time relative to the sandbox, because fork lineage is not persisted yet. That inference only decides which workflows are shown; it never deletes anything on its own, so the fix is safe regardless of its accuracy. The deterministic follow-up, persisting fork lineage, is tracked in #4938.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have implemented and tested all related **authorization policies**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,12 @@ and this project adheres to
 
 ### Fixed
 
+- Sandbox merge no longer deletes a workflow that was added to the project after
+  the sandbox was branched. Such workflows were never part of the sandbox, so
+  they are excluded from the merge screen entirely. Workflows deleted inside the
+  sandbox still appear and now default to kept, so removing them from the project
+  is opt-in. [#4919](https://github.com/OpenFn/lightning/issues/4919)
+
 ## [2.16.8] - 2026-07-01
 
 ## [2.16.8-pre] - 2026-06-18
--- a/lib/lightning_web/live/sandbox_live/components.ex
+++ b/lib/lightning_web/live/sandbox_live/components.ex
@@ -370,30 +370,38 @@ defmodule LightningWeb.SandboxLive.Components do
                 </span>
                 <span
                   :if={wf.is_changed && !wf.is_new && !wf.is_deleted}
+                  id={"merge-wf-changed-#{wf.id}"}
                   class="flex items-center gap-1 text-xs font-medium text-green-700"
-                  title="This workflow has been modified in the sandbox"
+                  phx-hook="Tooltip"
+                  aria-label="This workflow has been modified in the sandbox"
                 >
                   Changed
                 </span>
                 <span
                   :if={wf.is_diverged}
+                  id={"merge-wf-diverged-#{wf.id}"}
                   class="flex items-center gap-1 text-xs font-medium text-amber-700"
-                  title="This workflow was modified in the target project - this change will be lost"
+                  phx-hook="Tooltip"
+                  aria-label="This workflow was modified in the target project - this change will be lost"
                 >
                   <.icon name="hero-exclamation-triangle-mini" class="h-3.5 w-3.5" />
                   Diverged
                 </span>
                 <span
                   :if={wf.is_new}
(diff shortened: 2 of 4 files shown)


josephjclark (approved):
Works great for me Elias - I think the solution of just hiding workflows that aren't in the sandbox works super well. It's just clean.

Tested around a few things and behaviour looks good

You should speak to @theroinaochieng about whether we put this out as a patch.

My feeling is that as no-one has complained it can probably wait for the next release

midigofrank (approved):
Nicely done @elias-ba

PR #4942, opened 2026-07-06
Fix AI assistant crash when the AI server errors mid-stream

### LINKED ISSUE #4941: AI assistant error handler crashes on non-2xx streaming responses
## What's broken

The AI assistant error handler crashes when an Apollo **streaming** request returns a non-success HTTP status. `handle_error_response/2` in `Lightning.AiAssistant` was written for the non-streaming path, where the Tesla response body is a decoded JSON map, and it reads `body["message"]` directly. On the streaming endpoints the body is always a lazy `Stream` (the streaming client is configured with `adapter: [response: :stream]`), so doing map access on it raises:

```
** (FunctionClauseError) no function clause matching in Access.get/3
    (elixir) lib/access.ex: Access.get/3
    (lightning) Lightning.AiAssistant.handle_error_response/2
    (lightning) Lightning.AiAssistant.MessageProcessor.process_message/1
```

The crash replaces the real Apollo error (whatever non-2xx it returned) with an unhandled exception, so the background job is discarded and the underlying cause is masked.

## How it surfaces

The streaming query functions (`job_chat_stream`, `query_workflow_stream`, `query_global_stream`) match a 2xx response and stream it; anything else falls through to `handle_error_response/2`:

```elixir
case ApolloClient.job_chat_stream(...) do
  {:ok, %Tesla.Env{status: status, body: body}} when status in 200..299 ->
    process_stream(session, body, ...)

  error ->
    handle_error_response(error, session)   # body here is still a Stream on non-2xx
end
```

The first clause of `handle_error_response/2` matches any non-success `Tesla.Env` and calls `body["message"]`, which only works when `body` is a map. For streaming responses it is a `Stream.unfold/2` function, so `Access.get/3` blows up.

Observed in production as a discarded `Lightning.AiAssistant.MessageProcessor` job in the `ai_assistant` queue (lightning 2.16.6). Low frequency, since it requires Apollo to return a non-2xx specifically on a streaming call.

## What to fix

Make `handle_error_response/2` tolerate a non-map (streaming) body instead of assuming `body["message"]` is safe. Options: guard the non-success `Tesla.Env` clause with `is_map(body)`, and/or add a clause for streaming bodies that returns a generic error message without draining or map-accessing the stream. Either way a non-2xx streaming response should return a clean `{:error, message}` and log the real status, not raise.

Regression coverage should include a non-2xx streaming response reaching `handle_error_response/2`.

elias-ba:
## Description

The AI assistant talks to the Apollo AI server over a streaming connection. When that server replied with an error status, the code that handles errors assumed it was dealing with an ordinary JSON response and tried to read a field off it. Over a streaming connection the response body isn't a plain object, so this crashed, the background job was discarded, and the real error was hidden from the user.

This fixes the error handler so an error response over a streaming connection now returns a clean error message instead of crashing. The failing HTTP status is also included in the log so the underlying cause is easier to trace.

Closes #4941

## Validation steps

1. With the AI assistant enabled, open the assistant on a job (or the workflow chat) in a project you can edit.
2. Put the AI server into a failing state so a request comes back unsuccessful, for example point the Apollo endpoint at an unreachable or misconfigured URL.
3. Send a message to the assistant.
4. Confirm the assistant shows a clean error message and the chat stays usable: the message should not hang on a loading state and the editor session should not drop. Before this fix, an error returned mid-stream crashed the handler and the message silently failed.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


midigofrank (approved):
Nicely done

elias-ba:
Closing this. The fix is already on main by another route: the non-success branch guards the body shape before reading a message off it, so the crash this was written for can no longer happen. This sat approved since July and got overtaken.

PR #4946, opened 2026-07-08
AI-First Starting UX: Remove AI disclaimer

### LINKED ISSUE #4911: AI-First Starting UX: Remove AI disclaimer
Remove the AI disclaimer in the side panel and update the footer in the chat panel to : 

Image

Also update the Build with AI text area to have a footer with disclaimer: 

Image

Also tidy up anything to do with the current AI disclaimer and delete code related to that.

lmac-1:
## Description

This PR removes the AI disclaimer view in the AI assistant. The disclaimer is now in the footer of the AI assistant, linking to https://openfn.org/ai and also in the text area in the first "Build with AI" section.

Also makes it so that when you're in the "Build with AI" flow, you cannot view "Conversations" in the three dot menu in the AI assistant until a workflow exists. This is because otherwise the user might get stuck and have a weird experience if they went into other conversations. 

Closes #4911

## Validation steps

1. Start a new workflow via "Build with AI" — confirm the disclaimer footer text is visible under the chat input and in the landing screen's text area, and that the old full-screen disclaimer/gate no longer appears.
2. While still in the "Build with AI" flow (before a workflow exists), open the three-dot menu in the AI assistant — confirm "Conversations" is hidden.
3. Once a workflow has been created, reopen the three-dot menu — confirm "Conversations" is now visible again.
4. Run the migration (`mix ecto.migrate`) against a user who previously accepted the old disclaimer and confirm their `ai_assistant.disclaimer_read_at` preference key is removed without affecting other stored preferences.

1. Shouldn't pay too much attention to the legacy editor AI assistant as it's being removed in the next release — though I did also strip its disclaimer flow (onboarding screen, `mark_disclaimer_read` handler, `has_read_disclaimer` state) so we're not leaving half-removed code behind in the meantime.
2. Heads up that this isn't just a UI change — I also removed the `mark_ai_disclaimer_read` and `mark_disclaimer_read` socket events, plus the `has_read_ai_disclaimer` / `has_read_disclaimer` fields from the channel payloads, since nothing needs them anymore. The frontend change ships in the same PR so this should be invisible in normal use, but during a deploy, a browser with the old JS still cached could send one of those events to the new backend and just get silence back instead of a reply, rather than an error. Small edge case, but flagging it since it's an API surface change and not purely additive.
3. Deleted `lib/lightning_web/live/ai_assistant/quotes.ex` outright (the AI/responsibility quotes shown on the old onboarding screen) since none of that content is reused in the new footer. Shout if it should live on somewhere else.
4. Swapped the blocking disclaimer for a persistent footer nudge instead. Worth noting this actually lines up with the Responsible AI Policy's own language, it describes AI-usage indicators as "a nudge, not a warning," so a passive footer feels like a reasonable fit, but flagging in case anyone wants a harder gate for it.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


elias-ba (approved):
Nicely done @lmac-1 👏🏽 🧹

PR #4953, opened 2026-07-08
Show reauthorize prompt when an OAuth token is revoked

### LINKED ISSUE #4947: Gmail OAuth refresh failure logs at error level and shows generic message instead of "reauthorize"
## What's broken

When a user opens an OAuth credential whose refresh token is no longer valid on the provider's side (Gmail returning `invalid_grant` / "Token has been expired or revoked"), two things go wrong in how we handle it. Neither blocks the user, but both are worth tidying.

First, the failure is logged at `error` level in the token-refresh async handler in `LightningWeb.CredentialLive.GenericOauthComponent`. Sentry's `LoggerHandler` captures every `Logger.error` as an event, so an expected "please reauthorize" condition ends up in the alert stream as if it were an application fault. Our own logging guideline (`.claude/guidelines/logging.md`) puts "a user must re-authorise a credential" squarely in the `info`/`warning` bucket.

Second, the user sees the generic fallback error message instead of the tailored one we already wrote for this exact case. When a refresh fails, the async handler tags the result as `{:http_error, ...}`. But `LightningWeb.CredentialLive.OauthErrorFormatter` has a purpose-built clause keyed on `{:refresh_failed, %{error: "invalid_grant"}}` ("Refresh Token Revoked" / "Reauthorize"). Because of the tag mismatch, that clause is never reached for this path. A `400` with `invalid_grant` matches none of the `{:http_error, ...}` clauses (401, 403, 429, >=500, network_error) and falls through to the catch-all, which renders "OAuth Error / An error occurred / Try Again". The action button still relaunches authorization regardless of its label, so recovery works, but the guidance is vague enough that a user might contact support instead of just clicking through.

## How it surfaced

A single production Sentry event (`LIGHTNING-20D`) fired when a user opened a Gmail credential to use it. The stored token was expired, we attempted a background refresh, and Gmail rejected it with `invalid_grant`. No user was blocked and the credential was not lost.

## What to fix

- Downgrade the refresh-failure log for revoked/expired tokens to `warning` so it stays in the logs but out of Sentry. Keep genuine faults (unexpected 5xx, malformed responses) at `error`.
- Route refresh failures through the `{:refresh_failed, ...}` shape (or otherwise make the `invalid_grant` formatter clause reachable) so the user gets the "Refresh Token Revoked / Reauthorize" message instead of the generic fallback.
- While in the handler, sanity-check the sibling clauses (the async `:exit` crash path and the userinfo `401` path) for the same log-level pattern.

## Notes

Google returns `invalid_grant` identically for several underlying causes (user revoked access, password change, token unused past its lifetime, and others), so we can't and shouldn't try to disambiguate the reason from the error. "Your access is no longer valid, please reauthorize" is the right message for all of them.

elias-ba:
## Description

This PR fixes how the credential editor handles an OAuth token that the provider reports as expired or revoked.

Previously, when refreshing a stored token failed because the user needed to reauthorize (for example, a Gmail credential whose access was revoked), the editor showed a generic "an error occurred, try again" message and logged the condition at error level. The generic message could send a user to support when all they needed to do was sign in again, and the error-level log turned an expected, user-actionable situation into noise in our error monitoring.

Now the editor tells apart a "you need to reauthorize" outcome from a genuine transport failure. Revoked or expired tokens surface the tailored reauthorize prompt with a button that takes the user back through sign-in, and the condition is logged as a warning. Genuine faults and transient provider errors (such as a provider outage) keep their existing, more specific messages and error-level logging.

Closes #4947

## Validation steps

1. Create an OAuth credential (for example, Google/Gmail) and connect it successfully.
2. Revoke OpenFn's access from the provider's account settings so the stored refresh token is no longer valid.
3. Open the credential in OpenFn. Confirm you see a clear message that access has been revoked along with a reauthorize button, rather than a generic error.
4. Click the reauthorize button and confirm it sends you back through the provider's sign-in and restores the credential.

1. The diff also makes one pre-existing assertion in the credential test file order-insensitive. It was asserting a fixed ordering of workflow names that the underlying query does not guarantee, and the new tests here perturbed that ordering enough to make it flaky. The change only relaxes the assertion; it does not alter behavior.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


lib/lightning_web/live/credential_live/generic_oauth_component.ex
```diff
@@ -359,6 +371,22 @@ defmodule LightningWeb.CredentialLive.GenericOauthComponent do
      |> assign(authorize_url: authorize_url)}
   end
 
+  defp log_token_error(true, error, socket) do
```
midigofrank: Could you assign a variable to the booleans just to make it easier to know what `true` and `false` here mean. Or maybe you could define a function head:

```elixir
defp log_token_error(reauthorization_required, error, socket)
```

midigofrank (changes requested):
Nicely done @elias-ba . I have left a little change request

lib/lightning_web/live/credential_live/generic_oauth_component.ex
elias-ba: Good call. I named the parameter and moved the branching into a guard clause, so the two heads now dispatch on `reauthorization_required` by name rather than on bare `true`/`false`. Kept the multi-clause shape so each log level stays in its own body. Pushed in 1592d18.

midigofrank (approved):
Nice. Thanks @elias-ba

PR #4955, opened 2026-07-08
Order workflows alphabetically in credential revoke dialog

### LINKED ISSUE #4954: Workflow list in credential revoke dialog has nondeterministic order
## What's broken

`Lightning.Workflows.project_workflows_using_credentials/1` builds the list of workflow names shown in the "this credential is in use by the following workflows: X, Y" confirmation dialog when a user revokes a credential's access to a project. The query selects distinct workflow names with no `order_by`.

Postgres does not guarantee row order without an explicit `ORDER BY`, and because the query uses `distinct` (which can plan as a hash aggregate), the order of the returned names is effectively arbitrary. The dialog therefore lists the workflows in a nondeterministic order that can change between requests and between environments.

## How it surfaced

The test covering that confirmation dialog asserted a fixed order of two workflow names. It passed most of the time by luck of insertion order, but it is not a guaranteed order. A recent change that added test records shifted the shared factory sequence enough to change the generated names, which flipped the hash-aggregate order and made the assertion fail on some seeds. That test assertion has since been made order-insensitive so it no longer flakes, but the underlying query is still nondeterministic.

## What to fix

Add a deterministic `order_by` (by workflow name) to `project_workflows_using_credentials/1` so the confirmation dialog lists workflows in a stable, predictable order. With ordering guaranteed, the corresponding test can assert the exact order again.

This is cosmetic and low priority. All the correct workflow names are already listed; only their order is unstable.

elias-ba:
## Description

Small cleanup, not a user-visible fix.

The query behind the credential revoke-access dialog ("this credential is in use by the following workflows: ...") selects distinct workflow names with no explicit ordering, so the order was left to the database. In practice it already comes out alphabetical as a side effect of how the distinct is executed, but that is an implementation detail, not a guarantee. This adds an explicit order-by so the alphabetical order is guaranteed rather than incidental.

It also adds unit tests for `project_workflows_using_credentials/1`, which previously had none, covering how it groups results by project, filters by credential, and returns the names sorted.

Note: a test cannot meaningfully prove the order-by is doing anything, because the database sorts on its own at test-scale data. The added tests assert the function's contract (grouping, filtering, sorted result); they are not a regression guard for the ordering specifically.

Closes #4954

## Validation steps

1. Create a credential and share it with a project that has two or more workflows using it.
2. Open the credential and start to revoke its access to that project.
3. Confirm the dialog lists the affected workflows in alphabetical order.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


midigofrank (approved):
Great catch @elias-ba . Thank you

PR #4964, opened 2026-07-10
Lint: fix credo error

josephjclark:
Our lint tests are failing for a bunch of reasons, but one of them is this credo error:

> One `Enum.reject/2` is more efficient than `Enum.reject/2 |> 
┃       Enum.reject/2`

Looking at the code in question I'm not too worried about performance, I think the code is clearer using two iterations.

Claude says this:

> our current code has a comment on line 907-911 explaining why target-only-added-after-fork workflows get dropped, and the two separate rejects read as two distinct filtering concerns ("drop ones still in the sandbox" then "drop ones added after the fork"). Merging them into a single or slightly blurs those two intents.  On a list of workflows the efficiency gain is negligible.

On that basis I've disabled credo for that line. Open to suggestions but I think this is quite reasonable.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/lib/lightning_web/live/sandbox_live/index.ex
+++ b/lib/lightning_web/live/sandbox_live/index.ex
@@ -912,6 +912,7 @@ defmodule LightningWeb.SandboxLive.Index do
     deleted_entries =
       target_workflows
       |> Enum.reject(fn wf -> MapSet.member?(source_workflow_names, wf.name) end)
+      # credo:disable-for-next-line Credo.Check.Refactor.RejectReject
       |> Enum.reject(fn wf -> workflow_added_after_fork?(wf, source) end)
       |> Enum.map(fn wf ->
         %MergeWorkflow{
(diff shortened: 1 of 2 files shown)


elias-ba (approved):
Pushed a small alternative @josephjclark: folded the two rejects into one pass with `or` so credo passes for real, rather than suppressing it.

elias-ba:
Hey @josephjclark, I pushed another commit directly here to fix a flaky JS test that was failing CI on this PR. Nothing to do with your change, just a pre-existing test in `Header.test.tsx` that leaked state between tests and tripped under certain run orders. Added the per-test teardown the sibling file already does, and ran the suite shuffled many times to confirm it's stable.

josephjclark:
@elias-ba that commit should have been a different PR really! Let me check and merge.

lib/lightning_web/live/sandbox_live/index.ex
```diff
@@ -911,8 +911,10 @@ defmodule LightningWeb.SandboxLive.Index do
     # a merge never silently deletes them, and removal is opt-in.
     deleted_entries =
       target_workflows
-      |> Enum.reject(fn wf -> MapSet.member?(source_workflow_names, wf.name) end)
-      |> Enum.reject(fn wf -> workflow_added_after_fork?(wf, source) end)
+      |> Enum.reject(fn wf ->
```
josephjclark, line 914: Is this change expected?

midigofrank (approved):

PR #4965, opened 2026-07-11
Handle deleted workflows and snapshot failures in API

### LINKED ISSUE #4960: WorkflowsController can crash on :workflow_deleted / :snapshot_failed
> Found while working on the AI-first starting UX epic (#4848), specifically
> PR #4918. That PR gave one of `save_workflow`'s failure reasons a proper
> name (`:snapshot_failed`, previously a bare `false`), which is what
> surfaced this gap — the API controller was already missing a case for it
> under the old name, it was just unreadable. No code from that epic needs
> to change for this fix; this is a separate, pre-existing bug in a file
> that epic doesn't touch.

### What's wrong

`Lightning.Workflows.save_workflow/3` can fail in a few different ways. Two of
them are: the workflow was deleted, or the workflow saved but its snapshot
(a frozen copy used for run history) failed to save.

The REST API controller for workflows
(`lib/lightning_web/controllers/api/workflows_controller.ex`) has a function,
`maybe_handle_error/3`, whose job is to turn a failure reason into a proper
HTTP response (like "422: your data is invalid" or "404: not found"). It
checks for a specific list of known failure reasons. Anything not on that
list gets returned unchanged.

The problem: a controller function in Phoenix (our web framework) is required
to return a special "response" value. If it returns something else instead —
like the raw error reason — the server crashes instead of sending back a
normal error message.

Neither "workflow was deleted" nor "snapshot failed" are on that checked
list, so if either happens through the API, the request crashes instead of
returning a clean error.

### How to hit it

- **Deleted workflow**: `PATCH /api/workflows/:id` fetches the workflow
  first, but that fetch doesn't exclude deleted workflows. If the workflow
  was soft-deleted in between, the save fails with "deleted" and the crash
  happens. This bug already existed before PR #4918 — it's not new.
- **Snapshot failed**: rarer — needs the snapshot to fail validation right
  after the workflow itself saved successfully. Possible on both create and
  update.

### Suggested fix

Add two more cases to `maybe_handle_error/3` for these two failure reasons,
returning a sensible error response (e.g. 404/409 for deleted, 500 for
snapshot failed) instead of falling through unchanged.

### Context

Found while reviewing PR #4918, which gave the "snapshot failed" case a
proper name (`:snapshot_failed`) instead of returning a bare `false`. That
review is what surfaced this gap in the API controller — a different part
of the app (the collaborative editor) already had the same gap and got fixed
in that PR.

jmtdev0:
## Description

This PR fixes the workflows REST API so failures from `save_workflow/3` are
returned as proper HTTP responses instead of falling through as raw tuples and
crashing the Phoenix controller.

The controller now:

- Returns `404 Not Found` when a workflow is soft-deleted between fetch and
  save.
- Returns `500 Internal Server Error` when saving the workflow snapshot fails.
- Accepts both the current `false` snapshot failure value and the clearer
  `:snapshot_failed` value introduced by the related save-workflow work.

It also adds regression coverage and an Unreleased changelog entry.

Closes #4960

## Validation steps

1. Run `mix test test/lightning_web/controllers/api/workflows_controller_test.exs`.
2. Confirm the workflow update tests report `57 tests, 0 failures`.
3. Review the deleted-workflow case and confirm it returns `404` with
   `errors: ["Not Found"]`.
4. Review the snapshot-failure case and confirm it returns `500` with a clear
   snapshot error instead of crashing the controller.

Validated with:

- `mix test test/lightning_web/controllers/api/workflows_controller_test.exs`
- `mix compile --warnings-as-errors`
- `git diff --check`

1. The root cause was the final tuple fallback in `maybe_handle_error/3`, which
   returned these unhandled failures unchanged. Phoenix controllers must return
   a connection or a valid response, so the raw tuple caused the request to
   crash.
2. The current `save_workflow/3` implementation returns `false` for a snapshot
   failure, while the clearer `:snapshot_failed` atom is used by the related
   change discussed in PR #4918. Handling both keeps this fix correct for the
   current branch and forward-compatible with that naming improvement.
3. There is no separate POST regression test because POST and PATCH both pass
   through the same `maybe_handle_error/3` clauses. Triggering a snapshot
   collision during creation would require artificial database setup and would
   test a different failure mechanism rather than a different controller
   contract. The two PATCH tests cover both distinct error responses directly.
4. The response choices follow the issue's suggested semantics: deleted
   resources return `404`, while a snapshot persistence failure returns `500`.

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)
  *Not applicable: this PR only changes error handling after the existing authorization checks.*
- [x] I have ticked a box in "AI usage"

*This PR was developed by Codex with jmtdev0's supervision.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ and this project adheres to
   project is opt-in. [#4919](https://github.com/OpenFn/lightning/issues/4919)
 - Fixed an issue where LOCAL_ADAPTORS is not respected by install_schemas task
   [#4943](https://github.com/OpenFn/lightning/issues/4943)
+- Return clean API errors when a workflow is deleted during save or its snapshot
+  cannot be saved, instead of allowing the workflows controller to crash.
+  [#4960](https://github.com/OpenFn/lightning/issues/4960)
 - Prevent AI Assistant channel joins from crashing when a chat references a
   deleted workflow or project. Missing records now fail authorization cleanly.
   [#4914](https://github.com/OpenFn/lightning/issues/4914)
--- a/lib/lightning_web/controllers/api/workflows_controller.ex
+++ b/lib/lightning_web/controllers/api/workflows_controller.ex
@@ -562,6 +562,19 @@ defmodule LightningWeb.API.WorkflowsController do
     |> json(%{id: workflow_id, errors: ["Not Found"]})
   end
 
+  defp maybe_handle_error(conn, {:error, :workflow_deleted}, workflow_id),
+    do: maybe_handle_error(conn, {:error, :not_found}, workflow_id)
+
+  defp maybe_handle_error(conn, {:error, reason}, workflow_id)
+       when reason in [:snapshot_failed, false] do
+    conn
+    |> put_status(:internal_server_error)
+    |> json(%{
+      id: workflow_id,
+      errors: ["Could not save the workflow snapshot."]
+    })
+  end
+
   defp maybe_handle_error(_conn, result, _workflow_id) when is_map(result),
     do: result
 
--- a/test/lightning_web/controllers/api/workflows_controller_test.exs
+++ b/test/lightning_web/controllers/api/workflows_controller_test.exs
@@ -998,6 +998,55 @@ defmodule LightningWeb.API.WorkflowsControllerTest do
                |> json_response(404)
     end
 
+    test "returns 404 when the workflow is deleted before update", %{
+      conn: conn,
+      project: project
+    } do
+      workflow = insert(:simple_workflow, project: project)
+
+      workflow =
+        workflow
+        |> Ecto.Changeset.change(
+          deleted_at: DateTime.utc_now() |> DateTime.truncate(:second)
+        )
+        |> Repo.update!()
+
+      assert %{"id" => workflow_id, "errors" => ["Not Found"]} =
+               conn
+               |> patch(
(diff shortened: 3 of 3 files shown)


jmtdev0:
Follow-up on CircleCI lint

After inspecting the failed [lint job 98087](https://circleci.com/gh/OpenFn/lightning/98087), I made a small formatting-only adjustment to the new regression test. The test logic and production behavior are unchanged.

The new pipeline [98090](https://circleci.com/gh/OpenFn/lightning/98090) confirms that the formatting issue is resolved. The other PR-specific checks also pass:

- [compile 98090](https://circleci.com/gh/OpenFn/lightning/98090)
- [test_elixir 98091](https://circleci.com/gh/OpenFn/lightning/98091)
- [check_dialyzer 98093](https://circleci.com/gh/OpenFn/lightning/98093)
- [test_javascript 98094](https://circleci.com/gh/OpenFn/lightning/98094)

The remaining [lint job 98092](https://circleci.com/gh/OpenFn/lightning/98092) fails only on `credo`, `deps.audit`, and `hex.audit`. I believe these failures are unrelated to this PR: Credo reports existing repository-wide suggestions, while the dependency audit findings concern packages already present in `mix.lock`, which this PR does not modify. The focused controller suite still passes locally with 57 tests and 0 failures.

_This PR was developed by Codex with jmtdev0's supervision._

PR #4967, opened 2026-07-13
Fix orphaned canvas on failed workflow-creation save

### LINKED ISSUE #4939: AI-First Starting UX: Orphaned canvas nodes when workflow creation save fails
## Problem

All three workflow-creation paths — templates modal, YAML import modal, and AI-assisted apply — follow an **import-then-save** pattern: nodes are written into the Y.Doc first, then `saveWorkflow` is called. If the save fails (dropped connection, server error, timeout), the canvas is left showing nodes that were never persisted to the database — an orphaned workflow with nothing behind it.

Import-before-save is structurally required: the server ignores the client's payload on `save_workflow` and persists from its own server-side copy of the doc, so there's nothing to save until the doc is populated.

Today, all three creation flows call `saveWorkflow` in a "silent" mode that suppresses the shared retry/error handling that already exists for other save paths (Cmd+S, Save button). That means a failed creation save currently fails **silently** — no toast, no retry, no indication anything went wrong. The user is left looking at a workflow that looks created but isn't.

## Why not just roll back the doc on failure?

Rejected. A save timeout is ambiguous — the server may have actually saved despite the client not hearing back. Rolling back the doc in that case would sync an emptied doc over a workflow that was, in fact, persisted — which is worse than the current orphan. Recoverable, honest failure states are safer than a rollback that can silently destroy real data.

## Proposed approach

1. **Fix the save contract** — `saveWorkflow` should never silently return null; it either succeeds or throws. Remove dead null-handling scattered across callers.
2. **Route creation-flow failures through the existing shared retry/toast handler** instead of suppressing feedback. Replace the blunt "silent" flag with a more granular notification option, and make the retry prompt persistent (non-auto-dismissing) specifically for new/unsaved workflows, since it's the only way back from an orphan.
3. **Gate creation on known connectivity** — if the editor is offline, show immediate "Not connected" feedback and never touch the doc, avoiding a ~10s timeout and the resulting orphan in the common offline case.

## Acceptance criteria

- [ ] Creating a workflow (template, YAML import, or AI apply) while offline gives immediate feedback and does not populate the canvas.
- [ ] A save failure during creation shows a clear, non-dismissing "Retry" prompt (since silently leaving is not recoverable otherwise).
- [ ] Retrying a failed creation save, once connectivity/server issues resolve, completes the workflow creation normally (URL updates, modal closes, canvas view fits).
- [ ] Existing save flows (Cmd+S, Save button, save-before-run) are unaffected — no new/duplicate toasts.
- [ ] No workflow can be left on canvas with zero corresponding save attempt having surfaced feedback to the user.

## Out of scope

- Rolling back the Y.Doc on save failure (see rationale above).
- Server-side atomic import+save.
- Pruning of abandoned document state on the backend (separate, pre-existing issue).
- Unrelated dead-code cleanup not on the path of this fix (e.g. `saveAndSyncWorkflow`'s matching null pattern).

lmac-1:
## Description

Template, YAML import, build-from-scratch, and AI apply all do import-then-save: write nodes into the Y.Doc, then save. A failed save left those nodes stranded on the canvas with no record in the database, because every creation flow suppressed save failures with `silent: true` instead of surfacing them.

- Route creation-flow saves through the shared save handler with a new granular `notify` option instead of `silent`, so failures show a persistent Retry toast (persistent because dismissing it would hide the only path back to a saved workflow).
- Gate each creation flow on known socket connectivity first, so an offline attempt fails instantly instead of hanging out a 10s push timeout before failing anyway.
- Drop the dead `| null` return type on `saveWorkflow` (it always throws or resolves, never returns null) and the unreferenced `Workflow.Actions` interface that repeated the same false contract.

Closes #4939

## Validation steps

**1. Offline gate**
Disconnect (devtools → Network → Offline, or stop the Phoenix server briefly)
and try each creation path: pick a template, import a YAML file, click
"Build from scratch", ask the AI assistant to build a workflow.
Expect: an immediate "Not connected" alert. No 10s hang, no orphaned nodes
on the canvas.

**2. Forced save failure → Retry → recovery**
There's no code hook left for this (removed before merge), so trigger a real
save failure instead: stop your local Postgres (e.g. `docker stop <postgres
container>`) right after starting one of the four creation flows, then try
to create. Expect a red "Failed to save workflow" toast with a **Retry**
action that does not auto-dismiss, and the modal/landing screen stays open.
Restart Postgres, click **Retry** → "Workflow saved" toast, URL patches to
`/projects/:id/w/:workflowId`, canvas fits.

Repeat for at least one of: template, YAML import, build-from-scratch, AI
apply — the AI apply path is the one most worth spot-checking, since its
save now goes through the shared handler for the first time.

**3. Run-flow regression**
On an already-saved workflow, use Cmd+Enter and the Retry button in the run
panel. Expect exactly one outcome toast each time — no stray "Workflow
saved" toast alongside it.

1. This is a gap found by Stu, who wrote the plan for me to implement
2. When you kill the server and click "Browse templates", the loading hangs for 10 seconds even when disconnected. This won't be fixed in this PR. 
3. EDIT: I added an extra commit to tidy up the shared logic for 3 of the 4 creation paths. This means the connectivity gate and import-then-save sequence for blank/landing, template browser, and YAML import now go through one shared useCreateWorkflowFlow hook instead of being copy-pasted per file, and all four paths (including AI assistant) show the exact same "Not connected" copy instead of three different, occasionally-inaccurate versions of it. Behaviour is unchanged — this is purely dedup, verified by the existing test suite (updated where mocks needed to catch up).

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lmac-1:
@elias-ba @stuartc added this to "Notes to reviewer" following my most recent commit (this came up in code review in a child branch of this):

> 3. EDIT: I added an extra commit to tidy up the shared logic for 3 of the 4 creation paths. This means the connectivity gate and import-then-save sequence for blank/landing, template browser, and YAML import now go through one shared useCreateWorkflowFlow hook instead of being copy-pasted per file, and all four paths (including AI assistant) show the exact same "Not connected" copy instead of three different, occasionally-inaccurate versions of it. Behaviour is unchanged — this is purely dedup, verified by the existing test suite (updated where mocks needed to catch up).

lmac-1:
Merging this into parent. Will integrate Stu's review feedback into main cleanup PR.

PR #4970, opened 2026-07-14
AI-First Starting UX: Clean up to remove old new workflow path

### LINKED ISSUE #4876: AI-First Starting UX: Cleanup — remove old canvas entry points

lmac-1:
## Description

Cleanup pass for the AI-First Starting UX epic (#4848): removes the old
canvas-based create-workflow entry points now that the landing screen
owns that job, and fixes the bugs/rough edges found along the way.
Closes #4876 (phase 1).

### Removed
- The old "Create your workflow" placeholder and the left-panel entry
  points it replaced.
- The `?method=` URL param (and its now-pointless test fixture) and
  orphaned `Z_INDEX` constants.
- Leftover `TODO-AI-FIRST` comments.

### Fixed
- Save-failure toast stuck around even after a later save succeeded — now
  clears itself. (Needed a `preventDefault()` on the Retry button too:
  Sonner auto-dismisses on click otherwise, which silently blocked
  further retries.)
- AI-apply on a new, unsaved canvas gave up permanently if it was blocked
  offline — now retries automatically on reconnect.
- "Not connected" was shown during the normal channel-join window on
  every fresh page load — now distinguishes "still connecting" from
  actually disconnected.
- AI-apply fired a duplicate fit-view dispatch on new workflows.
- A failed workflow-creation save (template/YAML/build-from-scratch/AI
  apply) could leave the canvas orphaned with nothing written — fixed the
  import-then-save sequencing (#4939).
- Template browser could show stale results if closed before its fetch
  resolved and reopened quickly.
- A migration that clears an AI-disclaimer preference had no `down` —
  `mix ecto.rollback` would've raised.
- Minor design-feedback fixes to `LandingScreen`/`YAMLImportModal`/etc.

### Cleaned up
- Deduped the create-workflow gate/save logic across entry points into
  one `useCreateWorkflowFlow` hook.
- Moved template-browser state (templates, loading, search query) from
  local `useState` into `createUIStore`, consistent with how other panel
  state is managed.
- Removed dead code: a no-op `useCallback` wrapper, an unused
  `isConnected` return value, unused import-panel state in the UI store,
  a duplicated toast-persistence spread, and an unmemoized `useSession`
  selector in the AI panel.

### Tests
- Split the oversized `useAIWorkflowApplications` test file into seven
  files by scenario, then deduped their ~700 lines of repeated mock setup
  into one shared helper.
- Added regression tests for the fit-view double-dispatch fix and the
  `{:error, :snapshot_failed}` save path (previously untested).

### Not fixed here, tracked separately
- #4981 — failed auto-saves after an AI apply aren't logged anywhere.
- #4982 — the workflows REST API 500s instead of returning a proper error
  on a malformed create request.
- #4973 — the rest of this branch's code-review findings (test coverage
  gaps, dead backend code, stale comments) are lower-priority
  maintainability debt, not bugs. Drew the line at correctness + cheap
  fixes for this PR; #4973 is next. **Reviewers: that list is already
  known and tracked — no need to re-flag it here.**

## Validation steps

1. Start a new workflow via the landing screen (AI, template, YAML
   import) — no leftover old UI or `?method=` param.
2. Fail a save, then retry successfully — toast clears instead of
   stacking.
3. Start a new workflow via AI while offline — it applies automatically
   once reconnected, no manual click needed.
4. Open/close the template browser rapidly a few times — no stale results.
5. `mix ecto.rollback` / `mix ecto.migrate` over the disclaimer-preference
   migration both work.
6. `mix test` / `cd assets && npm test` both green.

- Repeated save failures update the same toast in place rather than
  showing each attempt separately — a UX call, not a bug, left for later.
- The AI-apply path deliberately isn't forced onto the shared
  `useCreateWorkflowFlow` hook — it's a genuinely different shape
  (streaming, collaborative), not duplicated code.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


priv/repo/migrations/20260707080205_remove_ai_disclaimer_read_preference.exs
```diff
@@ -1,9 +1,12 @@
 defmodule Lightning.Repo.Migrations.RemoveAiDisclaimerReadPreference do
   use Ecto.Migration
 
-  def up do
-    execute("""
-    UPDATE users SET preferences = preferences - 'ai_assistant.disclaimer_read_at' WHERE preferences ? 'ai_assistant.disclaimer_read_at'
-    """)
+  def change do
+    execute(
```
josephjclark, line 5: I don't think this is a big deal but I would rather this diff wasn't here? Unless it's reverting a divergence on the main branch

josephjclark (commented):
I don't have deep context or authority on this. But I've run the app and scanned the code and it looks fine. There's one diff I'd question, which has a comment attached.

Regarding:

> Save-failure toast stuck around even after a later save succeeded 

Claude says:

The fix works as described — I read the code, confirmed the dismiss actually fires on success, and the tests were updated. But it only fixes one of three failure toasts. The error handler shows different toasts depending on why the save failed:

1. Generic failure ("Failed to save workflow") — fixed ✅
1. "Permission Denied" — not fixed
1. "Save failed: invalid workflow" (validation errors) — not fixed

Toasts 2 and 3 get the same "never expire on a new workflow" treatment, but have no name, so nothing ever dismisses them. Example: AI generates a workflow that fails server-side validation → permanent "invalid workflow" toast → user fixes it and saves fine → the toast stays forever anyway. That's the exact bug this PR set out to kill, still alive in two branches a few lines above the fix. And it's not on the author's "already known, don't re-flag" list (#4973 — I checked).

So my suggested review comment: give all three failure toasts the same shared id (only one save-failure toast should ever be live anyway), so the existing success-handler dismissal covers them all. It's roughly a two-line change.



I am not sure about this - don't have the context. If there really is a two line fix here I'd be tempted to do it, but there's a risk of touching somethign else that won't get properly tested or reviewed later. I think these toasts are quite hard to reproduce so tricky to test them

josephjclark:
Speaking to the epic more generally: this PR basically works but I do seem to be getting a timeout in the app, preventing the new workflow being shown. I can see from my local apollo log that the workflow was generated in 39 seconds.

The UX after hitting "build it" is also pretty awful because the user is just going to stare at an empty canvas with no feedback for 30 seconds.

There is work in progress to make sure that the assistant sends more and better feedback to the app, and that the app has a better UI to show that feedback to the user. We might want to do that stuff before merging the final AI-first PR.

I don't know if there's a plan to render a placeholder chat while the assistant is working?

priv/repo/migrations/20260707080205_remove_ai_disclaimer_read_preference.exs
lmac-1: So this migration file is new in AI first and came from the work in removing the AI disclaimer. A personal review came up saying `change` was better than `up` for rollbacks. 

The idea is 'wiping' `ai_assistant.disclaimer_read_at` from `preferences` because this is something we aren't setting anymore. What would be the best way to do this migration?

lmac-1: Agreed with Joe that this is fine

josephjclark (approved):

lmac-1:
Thanks @josephjclark for your review, I've raised #5002 for some extra follow-on work based on your comment.

PR #4971, opened 2026-07-14
Enable Hex dependency cooldown

stuartc:
## Description

This PR _adds_ Hex 2.5.0's dependency [release-age cooldown](https://hexdocs.pm/hex/Mix.Tasks.Hex.html) to Lightning.

Setting `hex: [cooldown: "1w"]` in `mix.exs` means Hex won't resolve to a package version until it has existed for at least a week. This mitigates the common supply-chain attack where a compromised release is pulled into builds before anyone notices and retires it — most such incidents are caught within days. Lightning only consumes dependencies (it doesn't publish), so this only affects what we pull in, never what we ship.

Day-to-day impact is basically nil: `mix deps.get` against an intact `mix.lock` bypasses the cooldown entirely, so normal builds and CI are unchanged. The cooldown only bites when resolving something new (`mix deps.update`, adding a dep, or `deps.get` after unlocking). Security fixes for a dep we're already locked to re-resolve past the cooldown automatically; a genuinely urgent fresh fix can be forced for a single command with `HEX_COOLDOWN=0d mix deps.update <pkg>`.

The catch: **Hex older than 2.5.0 silently ignores the config**, giving false confidence. So:
- `bin/bootstrap` now enforces Hex `>= 2.5.0` for local dev, auto-upgrading and failing fast if it can't.
- CI/Docker already installs the latest Hex via `mix local.hex --force`, so it's covered.

## Validation steps

1. On a machine with Hex `< 2.5.0`, run `bin/bootstrap` and confirm it upgrades Hex and reports `>= 2.5.0`.
2. With Hex `>= 2.5.0`, run `mix deps.get` against the committed `mix.lock` and confirm it resolves unchanged (cooldown does not bite locked versions).
3. Run `mix hex.outdated` and confirm any versions inside the 1-week window are annotated `(cooldown)`.

1. Value chosen is `1w` — covers all-but-one of the analysed supply-chain incidents while costing us almost nothing for an app that doesn't ship on every dependency bump.
2. `ensure_hex_version` in `bin/bootstrap.d/common.sh` uses `sort -V` so it treats e.g. 2.10 as newer than 2.5.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,16 @@ and this project adheres to
   than an application error.
   [#4947](https://github.com/OpenFn/lightning/issues/4947)
 
+### Security
+
+- Enabled Hex's dependency cooldown (`hex: [cooldown: "1w"]`), so newly-resolved
+  dependency versions must have existed for at least a week before Hex will pull
+  them in, mitigating supply-chain attacks that get caught and retired within
+  days. Normal builds against an intact `mix.lock` are unaffected.
+  `bin/bootstrap` now enforces Hex `>= 2.5.0`, which is required for the
+  cooldown to take effect.
+  [#4971](https://github.com/OpenFn/lightning/pull/4971)
+
 ## [2.16.8] - 2026-07-01
 
 ## [2.16.8-pre] - 2026-06-18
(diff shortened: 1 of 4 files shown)


elias-ba (approved):
Hey @stuartc I followed the validation steps and it worked as expected. Nice one man 👏🏽

PR #4984, opened 2026-07-18
Make workflow title clickable to return to root editor view

brandonjackson:
## Description

Makes the workflow title in the editor breadcrumbs clickable. Clicking it returns to the root workflow editor view: it closes the full IDE (equivalent to the IDE's close/"x" button), closes any other open panel, deselects the current node, and drops any run-viewing context, landing on the bare canvas.

## Validation steps

1. Open a workflow in the collaborative editor.
2. Open the full IDE (or select a node to open the inspector), and/or open a run.
3. Click the workflow title in the breadcrumbs.
4. Verify all panels close, the node is deselected, any run overlay is cleared, and the view returns to the bare workflow canvas.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have implemented and tested all related **authorization policies**. (N/A - no authorization changes)


brandonjackson:
@elias-ba could you take a look at this? it has always driven me absolutely crazy that i can't click the workflow title to close the IDE. i saw several developers in DR doing this during training session too. @lmac-1 i know you've got some other toolbar stuff coming in shortly so asked claude to anticipate and align with what's coming, should have limited merge conflicts to resolve. A few of the oddities in this PR are things claude did and then i realised they were in the forthcoming PR anyways so just left them in.

elias-ba (approved):
Hey @brandonjackson I took a look at this this weekend and I also 100% agree with you, it's useful and very natural to click the workflow title and hope it gets us back to the canvas. In your initial implementation, I also saw that it was not only closing the IDE but would also close open panels like job panel, edge panel, etc. I generalized it to close other panels it was not closing like the run panel. It's looking great to me. I am approving it but because I added few commits to it I think it would be great if someone else looks at the code. I see that @lmac-1 is added as a reviewer, so @lmac-1 if you can look at this tiny one today, it would be really great to merge it. Thanks again @brandonjackson for this little helpful one.

lmac-1:
Brandon confirmed that although the PR desc says all panels should be closed, AI assistant panel should remain open. 
I will continue reviewing this after reviewing sandbox stuff and filing some follow-on AI first issues.

lmac-1 (approved):
Thanks @elias-ba. I was half way through this review yesterday and just applied one small improvement directly: the title-click handler was calling `useFollowRun(null)` just to get hold of its `clearRun` teardown. `useFollowRun` is a "follow this run" lifecycle hook — calling it with `null` subscribes the always-mounted header to active-run and channel state for nothing, and reads like a mistake at the call site. I added a dedicated `closeRunViewer` command to `useHistoryCommands` (which subscribes to nothing) and call that instead.

**Follow-up (optional separate PR)**: `useFollowRun` still returns a now-redundant `clearRun`. The cleaner end state is to drop `clearRun` from `useFollowRun` entirely — so it owns only the connect side of the lifecycle — and move the other two callers, `FullScreenIDE` and `CollaborativeWorkflowDiagram`, onto `closeRunViewer` too. I left that out here to keep this PR scoped to the breadcrumb; those two components aren't otherwise touched by this change, and the run-following refactor deserves its own review and revert boundary. I'll leave it up to you whether this is something you'd like to implement or not (and raise an issue).

elias-ba (approved):

elias-ba:
Thanks @lmac-1 — good catch on the run viewer teardown, and thanks for just fixing it directly rather than bouncing it back. I've opened #5045 for the follow-up cleanup you suggested so it doesn't get lost.

PR #4985, opened 2026-07-18
Return 422 instead of 500 when workflow create body omits edges, jobs, or triggers

### LINKED ISSUE #4982: Workflows REST API crashes (500) instead of returning 422 when jobs/triggers/edges are omitted
## Summary

`POST /api/projects/:project_id/workflows` (and the equivalent non-nested
route) raises an uncaught `FunctionClauseError` — a 500 — instead of a
validation error, if the request body omits `jobs`, `triggers`, or `edges`
entirely (as opposed to sending them as empty lists).

## Repro

```bash
curl -X POST http://localhost:4000/api/projects/$PROJECT_ID/workflows \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"name":"My Workflow"}'
```

Expected: `422 Unprocessable Entity` with a validation error (e.g. missing
edges/triggers).

Actual: `500 Internal Server Error`.

## Root cause

`LightningWeb.API.WorkflowsController.validate_workflow/2`
(`lib/lightning_web/controllers/api/workflows_controller.ex`) only has a
clause that matches when the params map already contains `"edges"`,
`"jobs"`, and `"triggers"` keys:

```elixir
defp validate_workflow(
       %{
         "edges" => edges,
         "jobs" => jobs,
         "triggers" => triggers
       },
       ids_map
     ),
     do: validate_workflow(edges, jobs, triggers, ids_map)
```

If any of those keys is missing from the request body, no clause matches
and Elixir raises `FunctionClauseError`, which isn't handled by the
controller's `with`/`maybe_handle_error` error-mapping and surfaces as an
unhandled 500.

The `update` path isn't affected — it builds an `Ecto.Changeset` first
(via `Workflows.change_workflow/2`), which always has `:edges`/`:jobs`/
`:triggers` fields via schema defaults, so it hits the changeset clause
instead. Only `create`, which validates the raw params map directly, can
hit the crash.

## Fix

Confirmed with a passing regression test locally. The minimal fix is to
make the second clause match any map and default the missing keys to
`[]`:

```elixir
defp validate_workflow(%{} = params, ids_map) do
  edges = Map.get(params, "edges", [])
  jobs = Map.get(params, "jobs", [])
  triggers = Map.get(params, "triggers", [])

  validate_workflow(edges, jobs, triggers, ids_map)
end
```

This routes the request through the existing validation logic, which
already returns a proper 422 (e.g. `"Missing edge with
source_trigger_id."`) for a workflow with no edges.

## Provenance

Found incidentally while reviewing the AI-First Starting UX epic (#4848),
not caused by it — this crash predates that work (introduced in #2792,
Jan 2025). Not part of that epic's scope; tracking here as its own fix.

SAY-5:
## Description

This PR fixes a crash in the workflows REST API. `POST /api/projects/:project_id/workflows` raised an unhandled `FunctionClauseError` (a 500) instead of a validation error when the request body left out `edges`, `jobs`, or `triggers` entirely, rather than sending them as empty lists.

`validate_workflow/2` only had a map clause matching when all three keys were present, so a create body missing any of them matched no clause. The update path is unaffected because it builds a changeset first, which always has those fields via schema defaults. Only create validates the raw params map directly.

The fix adds a fallback clause that defaults the missing keys to `[]` and routes through the existing validation, which already returns a proper 422 (`"Missing edge with source_trigger_id."`) for a workflow with no edges.

Closes #4982

## Validation steps

1. `POST /api/projects/:project_id/workflows` with a body of just `{"name": "..."}` (no `edges`/`jobs`/`triggers`).
2. Before this change the response is a 500; after it, a 422 with the edges validation error.
3. The added test `returns 422 when edges, jobs and triggers are omitted` covers this in `workflows_controller_test.exs`.

1. I couldn't run the full ExUnit suite here (it needs Postgres and full deps compilation), so the regression test is written to mirror the neighbouring `returns 422 when edges misses a source trigger` case and the empty-list path was traced by hand to confirm it lands on `{:error, :edges_misses_a_trigger}` and the existing `reply_422`. Please run it in CI.

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,10 @@ and this project adheres to
 
 ### Fixed
 
+- The workflows REST API now returns a 422 validation error instead of a 500
+  when a create request omits `edges`, `jobs`, or `triggers` from the body
+  rather than sending them as empty lists.
+  [#4982](https://github.com/OpenFn/lightning/issues/4982)
 - Sandbox merge no longer deletes a workflow that was added to the project after
   the sandbox was branched. Such workflows were never part of the sandbox, so
   they are excluded from the merge screen entirely. Workflows deleted inside the
--- a/lib/lightning_web/controllers/api/workflows_controller.ex
+++ b/lib/lightning_web/controllers/api/workflows_controller.ex
@@ -413,6 +413,14 @@ defmodule LightningWeb.API.WorkflowsController do
        ),
        do: validate_workflow(edges, jobs, triggers, ids_map)
 
+  defp validate_workflow(%{} = params, ids_map) do
+    edges = Map.get(params, "edges", [])
+    jobs = Map.get(params, "jobs", [])
+    triggers = Map.get(params, "triggers", [])
+
+    validate_workflow(edges, jobs, triggers, ids_map)
+  end
+
   defp validate_workflow(edges, jobs, triggers, ids_map) do
     # {:ok, _ids} <- validate_ids(edges),
     with {:ok, triggers_ids} <- validate_ids(triggers),
--- a/test/lightning_web/controllers/api/workflows_controller_test.exs
+++ b/test/lightning_web/controllers/api/workflows_controller_test.exs
@@ -745,6 +745,25 @@ defmodule LightningWeb.API.WorkflowsControllerTest do
              }
     end
 
+    test "returns 422 when edges, jobs and triggers are omitted", %{
+      conn: conn,
+      project: project
+    } do
+      conn =
+        post(
+          conn,
+          ~p"/api/projects/#{project.id}/workflows",
+          Jason.encode!(%{name: "workflow without children"})
+        )
+
+      assert json_response(conn, 422) == %{
+               "id" => nil,
+               "errors" => %{
+                 "edges" => ["Missing edge with source_trigger_id."]
+               }
+             }
+    end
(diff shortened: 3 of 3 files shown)


lmac-1 (approved):
Hiya @SAY-5, thank you so much for your contribution! This is a clean, well-scoped fix, and the write-up in the description is accurate. Thanks also for ticking the AI usage box 🙏

You mentioned you couldn't run the suite locally, so I ran it here: 56 tests pass, format and Credo clean. I also checked the regression test actually pins the bug. I've verified that removing your fallback clause reproduces the `FunctionClauseError`, restoring it goes green.

**One thing before this can merge:** could you merge the latest `main` into your branch? There's a conflict in `CHANGELOG.md`.

**Optional while you're there:** a test for only *one* key missing (jobs and triggers present, no `edges`) that crashed the same way and is probably the more common shape. Happy to add it after merge if you'd rather not.

The lint failure looks like pre-existing repo drift rather than your diff so merging `main` may clear it.

Approving so you're not blocked by me.

lmac-1:
Thanks @SAY-5 ! Merging :)

Issue #4897, opened 2026-06-26
Carry workflow state in the wire format (project.yaml + CLI)


elias-ba:
**As a** developer deploying workflows through the CLI, GitHub sync, or YAML, **I want** a workflow's lifecycle state to travel in the wire format, **so that** exporting and re-importing a live workflow keeps it live instead of quietly reverting to a draft.

Part of #4852 (Sandbox DevX epic). Cross-repo. Blocked on the CLI and provisioner push rules (see below).

## What's broken

The workflow lifecycle field `state` (`draft | live`) is the user-facing contract for whether a workflow is processing data, but it does not travel on the wire. The provision API JSON, the `project.yaml` export, and GitHub sync all round-trip `trigger.enabled` as a first-class field, while none of them carry `workflow.state`. The CLI (in the `kit` repo) round-trips and hashes `trigger.enabled` for its diff engine and has no lifecycle concept at all.

## How it surfaced

While completing the lifecycle epic we found round-trips are lossy for lifecycle: export a live workflow, re-import it via CLI deploy, GitHub sync, or YAML import, and it comes back as a draft because `state` is absent and resets to the schema default. As an interim, non-breaking fix, the importer now infers `state` from the resulting trigger-enabled set (any enabled trigger means live, otherwise draft), with an explicit `state` in the spec always winning. That keeps imports coherent, but it is an inference, not the real contract.

## Verified safe to defer, unsafe to half-build (2026-07-05)

Checked against prod and both repos. Phase 1 ships with zero wire-format regression WITHOUT this issue. The Lightning export never emits `state` (export code is identical on main and the epic branch) and still emits `trigger.enabled`, and import is additive (infers `state` from enabled when absent). A prod check found no workflow with mixed enabled/disabled triggers (0 of 2,857 active), so the inference is exact for every workflow and there are no phantom CLI diffs.

The danger is doing the EXPORT half of this issue first. If Lightning starts emitting `state` before `lexicon` and the CLI accept it:

- An older or not-yet-upgraded provisioner rejects any document containing `state` (`validate_extraneous_params` returns a 422), breaking GitHub sync and CLI deploy during rollout and against older instances.
- Both CLI paths silently cache and round-trip an unknown `state` back on the next deploy, and the epic-branch provisioner applies an explicit incoming `state` (explicit wins over inference), so a stale cached value can silently flip a workflow's draft/live state on an unrelated deploy. That is silent data loss.

[...]

PR #5005, opened 2026-07-21
Promote from the editor

### LINKED ISSUE #4929: Promote from the editor (MVP: reuse the existing sandbox merge)
**As a** workflow builder editing in a sandbox, **I want** a Promote action that merges my changes back into the parent project using the sandbox merge we already have, with the same authorization, **so that** I can complete the round trip now without waiting on a review-and-approval system we have not built yet.

## Context

For the first release, promotion is simply merging. It reuses the existing sandbox merge (`Sandboxes.merge/4` into `MergeProjects`) and the existing `merge_sandbox` authorization policy (editor, admin, or owner on the target project, the same gate as today's sandbox merge). There is no promotion request, no pending or withdraw, and no approval gate. Those belong to the later branch-protection phase (#4861, #4862).

## Acceptance criteria

- From the collaborative editor while in a sandbox, a Promote action merges the edited workflow back into the parent using the existing merge, scoped to that workflow.
- Promote runs from a double-disclosure confirmation modal with two actions, Cancel and Promote. The modal states what will happen: the change merges into the parent's live workflow, which stays live. There is no non-publishing promote; promoting never takes the workflow off live.
- Authorization matches today's sandbox merge (`merge_sandbox`: editor, admin, or owner on the target project). No new policy is introduced. The post-merge archive still requires admin/owner on the source sandbox, unchanged.
- After a successful merge the parent workflow reflects the change and a new version is recorded (#4866).
- The promote flow archives the sandbox by calling `schedule_sandbox_deletion` after the merge succeeds. Note: `Sandboxes.merge/4` does not archive on its own, the existing sandbox-merge LiveView does it as a separate call, so this flow must replicate that (see #4649). Archiving is not atomic with the merge, so the merged-but-not-archived case is handled gracefully.
- Diverged workflows are surfaced in the merge summary, informational and non-blocking (see #4863).
- No promotion-request object is created. There is nothing to approve or withdraw.
- Covered by tests (the promote/merge path and the header behavior).

## In-sandbox header (deferred here from #4901)

The collaborative editor header must be coherent inside a sandbox: show Promote, and suppress the go-live / switch-to-draft controls that only apply to a workflow on its own project. After a successful promote the sandbox is archived, so the header should reflect the archived, read-only state rather than leaving stale controls.

## Scope and key surfaces

Collaborative editor header (the in-sandbox Promote action and the control suppression above), `Sandboxes.merge/4`, the `merge_sandbox` policy, `schedule_sandbox_deletion`, and the existing merge summary UI in the sandbox LiveView (reuse rather than rebuild where practical).

## Dependencies

Builds on Edit in sandbox (#4859). Pairs with the diverged surfacing (#4863) and the parent-added-workflow fix (#4919). Version recording (#4866) hangs a release off this promote, so this path must create a snapshot of the merged parent workflow and record who promoted (to be confirmed against the merge/provisioner path).

## Note

Promotion requests (pending and withdraw) and the approval-and-validation step are deliberately deferred to the branch-protection phase (#4861, #4862). This issue is the now-path.

Part of #4852.

elias-ba:
## Description

This PR adds promoting a workflow from the sandbox editor back into its parent project, along with the lifecycle controls needed to make that safe. Promote merges the workflow you are editing into the parent's live workflow, then offers to archive the sandbox or keep it, so you can promote several related workflows before retiring it. To support editing and testing safely, a non-live workflow (a draft, or anything in a sandbox) can have a trigger turned on for testing behind a warning, and read-only (live) workflows now consistently hide run and edit actions instead of showing them disabled. It also includes a pass over the header, the confirmation dialogs, and the edit-in-sandbox picker.

Closes #4929

## Validation steps

1. Open a live workflow on the main project. The header shows a "Live" badge (with an explanatory tooltip) next to the version picker, no Save or Run buttons, and "Edit in sandbox" / "Switch to draft" actions. You can still read the workflow's configuration and copy the webhook URL, but the canvas and inspectors are read-only and undo/redo does nothing.
2. Click "Edit in sandbox". The picker opens with a "Create a new sandbox" section (pressing Enter creates it; a duplicate name shows an inline error) and, when any exist, a "Join an active sandbox" list. Creating one lands you in the sandbox with the workflow as an editable draft and its trigger off.
3. In the sandbox, select the trigger and turn it on. A warning appears about receiving real events (with a "don't show again" option); confirming enables it. Turning it off does not warn.
4. Make a change and click "Promote". Confirm, then either "Keep sandbox" (you stay in the sandbox and can switch to another workflow and promote it too) or "Archive sandbox" (you return to the parent, now live with your changes, and the sandbox is scheduled for deletion). An editor who cannot archive sees only "Done".
5. Confirm the parent workflow reflects the promoted changes and its version badge is current.

1. This stacks on #4901 (edit in sandbox) and targets that branch, so it is best reviewed after #4901. The user-facing copy in the new dialogs, the badge tooltip, and the picker is still being finalised, which is why this is a draft.
2. On import, a brand-new workflow is labelled live or draft based on whether any trigger is enabled, while an existing workflow keeps its stored state on a round-trip, so a GitHub or CLI sync cannot silently flip a draft whose trigger you turned on for testing. Carrying draft/live explicitly in the wire format is tracked separately (#4897).

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


assets/js/collaborative-editor/components/Header.tsx
```diff
@@ -463,7 +580,40 @@ export function Header({
...
             <Breadcrumbs>{children}</Breadcrumbs>
           </div>
-          <ReadOnlyWarning className="ml-3" />
+          {/* The Live badge already implies read-only, so suppress the
+              redundant "Read-only" pill whenever the Live badge is shown for
+              the current live version. Still show it for a pinned/deleted
+              read-only view, where "Live" (the current state) doesn't explain
+              why this view is read-only. */}
+          {!(
+            lifecycleState === 'live' &&
```
lmac-1, line 553: Is `lifecycleState === 'live' && !isNewWorkflow && !isSandbox` the condition for the live badge being visible? If so, I would group this together into a variable like `const showsLiveBadge =
  lifecycleState === 'live' && !isNewWorkflow && !isSandbox;` so it's clearer for someone else reading the code later.

lib/lightning/projects/provisioner.ex
```diff
@@ -75,51 +83,90 @@ defmodule Lightning.Projects.Provisioner do
...
+        error
+    end
+  end
+
+  @doc """
+  Extracts the workflow ids carried by an import/merge document that a live
+  collaborative document should reconcile against, excluding workflows the
+  document marks for deletion. Only workflows present in the document are
+  returned, so sibling workflows on the target are never touched.
+  """
```
lmac-1, line 161: This promise doesn't quite hold on the scoped promote path:

> Only workflows present in the document are returned, so sibling workflows on the target are never touched.

When `selected_workflow_ids` is set, `build_merged_workflows/6` emits `build_passthrough_workflow/1` for every unselected target workflow (`merge_projects.ex:853-859`). Those entries carry a plain "id" and no "delete", so they survive the filter here and each gets a `ReconcileRequested`.

The net effect is that promoting workflow A resets the live document for every other workflow in the parent project. `reset_shared_doc/2` clears jobs, edges, triggers and positions, then re-serialises from the database, so anyone editing an untouched sibling workflow loses their unsaved canvas changes.

Nothing in the database is affected—IDs, runs and logs are all fine—but it's a genuine "my work just vanished" experience for anyone with the parent project open during a promote.

To reproduce: 
1. Have a project with 2 workflows (one live, one draft)
2. Open workflow A in one tab and make a change, keep the tab open
3. Open workflow B in a second tab and click "Edit in sandbox". 
4. Make a change in the sandbox and click Promote
5. The changes in step 2 are wiped even though it's a different workflow

lib/lightning/collaboration/workflow_reconciler.ex
```diff
@@ -50,6 +61,133 @@ defmodule Lightning.Collaboration.WorkflowReconciler do
...
+    Lightning.Workflows.get_workflow(workflow_id,
+      include: [:jobs, :edges, :triggers]
+    )
+  end
+
+  # Full reset from the database, mirroring the lifecycle reset
+  # (Session.clear_and_reset_doc / Persistence.clear_and_reset_workflow). All
+  # Yex collections are retrieved before the clear transaction to avoid a VM
+  # deadlock, and serialize_to_ydoc runs its own transaction afterwards.
+  defp reset_shared_doc(shared_doc_pid, workflow) do
```
lmac-1, line 157: From Claude (non-blocking but nice to tidy up): 

> `reset_shared_doc/2` plus its private `clear_array/1` and `clear_map/1` are a verbatim third copy of `Session.clear_and_reset_doc/2` and its helpers (`session.ex:598-639`) — same pre-transaction handle retrieval, same clear transaction, same `WorkflowSerializer.serialize_to_ydoc/2` call. `Persistence.clear_and_reset_workflow/2` (`persistence.ex:130-160`) is a partial second copy with its own `clear_array/1`.

lmac-1 (commented):
Hey @elias-ba, thanks for this and sorry for the delay. 

These are some issues that I think should be resolved, I'll let you decide if it should be follow on work:
1. Clicking "Promote" on a sandbox workflow resets the live editor for every other workflow in the parent project, discarding anyone's unsaved changes (more details inline on `provisioner.ex`). I'd expect unsaved changes on the live promoted workflow to be discarded, but not other workflows that weren't promoted. 
2. Renaming a workflow via "Edit in sandbox" and then clicking "Promote" creates a new workflow rather than updating the existing one. This splits its history into a new workflow, and the new copy arrives as a draft with its trigger disabled - so it doesn't run. A new workflow also means new job and trigger ids.
3. Question on `maybe_infer_workflow_state/2`: The new `:loaded` clause means an existing workflow always keeps its stored state. If a sync or CLI deploy pushes a document with every trigger disabled, does the workflow stay `:live`? If so, it'd be read-only in the editor but not processing anything. The draft-preservation goal in the description only seems to require guarding the `draft → live` direction. Is blocking `live → draft` intentional? 
Everything else outsideof this 

I would also like @midigofrank to review this to double check I haven't missed anything from the provisioner side of things.

**Low priority issues:**
These shouldn't block merging and I am documenting them for completeness. 
1. Something strange happens when I enable/disable the trigger in a sandbox: it seems like it increases the version numbers in the version dropdown but the version number does not update. Example: 
   1. Enable the trigger in a sandbox
   2. Version dropdown updates to v1 (Save button has no red unsaved indicator)
   3. Disable the trigger 
   4. Version dropdown stays at v1
   5. If you enable/disable, you will see that version dropdown stays at v1 but if you open it you will see lots of different version numbers higher than 1
   6. Clicking Save updates to latest 
2. Promote button is active for a collaborator who does not have permission to promote - should it be read only? Example - a collaborator only with access to the sandbox but not production project can click Promote. They get a toast message telling them they don't have permission for that action, but wondering if the button should be read only with tooltip message?

PR #4887, opened 2026-06-22
AI-First Starting UX: Landing screen and import from YAML

lmac-1:
## Context

Part of epic #4848 — the goal is to get users to use the AI assistant primarily and to make sure we don't lose their progress. The landing screen is the entry point to that flow; the design decisions below are shaped by the constraint that the header (which lets users navigate away) must stay hidden until a workflow actually exists in the DB. This is also known as the "pink toolbar" * state in the issues of this epic. Whenever that is referenced there is no pink toolbar, it's just a concept to describe a state whereby no workflow exists in the database.

We are taking away the 'preview' state and the 'Create' button in the top toolbar. Instead, workflows will be created in the database earlier. In this PR, you will see that when you import your workflow with YAML and click "Create" in the modal, it will create the workflow in the database instantly.

## What this adds

A React overlay that replaces the bare `/w/new` canvas, plus the first wired-up entry point: Import from YAML.

When a user creates a new workflow they see: a Build with AI input, a Build from Scratch card, a Browse Templates card, and an "or import a YAML file manually" text link. Clicking the YAML link opens a modal with a file dropzone and import flow. All other paths (Build with AI, Build from Scratch, Browse Templates) are stubbed — callback props are wired, nothing is connected yet.

Closes #4856, #4871

### Key files

  - `LandingScreen.tsx`: the overlay component
  - `YAMLImportModal.tsx`: Headless UI Dialog with dropzone and import flow
  - `createUIStore.ts`: `showLandingScreen`, `dismissLandingScreen`, `showYAMLImportModal`, `openYAMLImportModal`
  - `useUI.ts`: `useShowLandingScreen`, `useShowYAMLImportModal`, hooks exposed via `useUICommands`
  - `CollaborativeEditor.tsx`: renders `LandingScreenWrapper`, gates the header

### Design decisions

1. **Header gated on `isNewWorkflow`, not `showLandingScreen`.** There's an interim state on the Build with AI path where the landing screen is dismissed but no workflow exists in the DB yet — the user is still talking with AI before receiving a workflow from AI. Gating on `showLandingScreen` would expose the header at that point. `BreadcrumbContent` reads `useIsNewWorkflow()` from the store and self-hides until the first DB write.
2. **Modal state in UIStore, not local state.** `WorkflowEditor`'s `Ctrl+\` shortcut needs to open the YAML modal but is a sibling of `LandingScreenWrapper` — local state would require prop-drilling across the tree.
[...]

PR #5006, opened 2026-07-22
AI-First Starting UX

### LINKED ISSUE #4848: Epic: AI-First Starting UX

lmac-1:
This work replaces the old create a workflow flow with a landing screen offering four ways in, and moves the moment a workflow is written to the database from "user presses Create" to "user chooses how to start". Built as eight reviewed sub-PRs plus a handful of commits made directly on this branch (listed below).

Closes #4848

### Why

The goal of the epic is to get users into the AI assistant as the primary way of building a workflow, and to stop losing their work along the way. The old flow put users on a bare canvas with a left-hand "Create your workflow" panel and a Create button in the toolbar; nothing existed in the database until they pressed it, so anything they did before that could evaporate.

The new flow inverts that: choosing a path *is* the act of creation. The row is written immediately, and from then on it is an ordinary saved workflow.

### What a user sees

`/projects/:id/w/new` now opens a "Where would you like to start today?" screen with four options:

| Path | What it does | Workflow row is created |
|---|---|---|
| Build with AI | Prompt goes straight to the assistant; canvas stays blank while it thinks | When the AI's first YAML is applied (including mid-stream) |
| Build from scratch | Seeds a webhook trigger + one job + an edge | On click |
| Browse templates | Modal of base and user templates | On clicking a template card |
| Import from YAML | Modal with a dropzone and validation | On pressing Create in the modal |

All four converge on the same client path — `importWorkflow` → `saveWorkflow` over the collaborative session — so snapshots, `lock_version`, audit and the trigger-activation limiter behave exactly as they do for any other save. There is no server-side creation endpoint.

Once the row exists, the URL is rewritten from `/w/new` to `/w/{id}` in place (no reload) and the editor header appears.

### Key architectural decisions

**The row is created late, and deliberately.** Nothing is written when `/w/new` mounts — the id is generated client-side and the workflow only reaches the database on first save.

**The header is hidden for the whole pre-save state**, not just while the landing screen is up. Gating on `!isNewWorkflow` rather than `!showLandingScreen` keeps it hidden through the AI's interim state, where the landing screen is already dismissed but no workflow exists. Rationale: a header gives the user a way to navigate away from something that isn't saved.

**URL params are ignored on `/w/new`.** `?panel=run`, `?panel=settings` and `?chat=true` could otherwise open panels or skip the landing screen on a workflow that doesn't exist. Three mechanisms cover this: the UI store's initialisation and `WorkflowEditor`'s URL→panel sync are both gated while new, and `useAIPanelURLSync` strips `job`, `trigger`, `edge`, `w-chat` and `j-chat` from the URL on mount while new. `useAIMode` does read `?panel=editor` without checking `isNewWorkflow`, but it needs a `job` param too, and that one is stripped — verified by hand.

**The AI panel can't be closed until the workflow is saved.** That's a trade-off, not a safety guarantee. Work isn't lost — the workflow is written as soon as the AI's first YAML applies. But until then there's no way back to the other options. A "back to options" button was rejected: it brings back the state where users lose what they started. Hitting the AI quota used to leave people stuck here; fixed in `487078fe43`. Losing connection doesn't — the prompt is saved and re-sent.

**Auto-save on creation is silent, but failures are loud.** No "saved" toast for a save the user didn't trigger; a failed creation save raises a persistent Retry toast, since dismissing it would remove the only route back to a saved workflow.

**AI validation errors go to the chat, not a toast**, so the user can ask the assistant to fix its own YAML instead of hitting a dead end.

### What was removed

- The old canvas-based create entry points and the "Create your workflow" left-panel.
- The Create button and the preview state it belonged to.
- The `?method=` URL param.
- The full-screen AI disclaimer gate, replaced by a persistent footer link. This also removed the `mark_ai_disclaimer_read` / `mark_disclaimer_read` socket events and the `has_read_ai_disclaimer` / `has_read_disclaimer` payload fields, plus `lib/lightning_web/live/ai_assistant/quotes.ex`.

### How the branch was built

Each of these was reviewed when it merged. **You do not need to re-read them** — this is a map for navigating the diff and for finding the discussion behind a given decision.

| # | What it added |
|---|---|
| #4887 | Landing screen overlay + import from YAML; other paths stubbed |
| #4902 | Build with AI: auto-apply, streaming dedup, panel lock, URL param blocking |
| #4912 | Templates path; presentational/wrapper split |
| #4946 | Removed the AI disclaimer gate; channel payload changes + migration |
| #4918 | Build from scratch: client-side seeding, unique naming, double-click lock |
| #4967 | Failed-save recovery: `notify` option, offline gates, shared `useCreateWorkflowFlow` |
| #4970 | Cleanup phase 1: removed the old create path |
| #5003 | Cleanup phase 2: dead code, `filterTemplates` extraction, channel `workflow_kind` fix |

Several commits are **not** in any sub-PR and were made directly on this branch:

- `e057c0fc91` — redirect users without `:create_workflow` away from `/w/new`. **This fixes a pre-existing bug, not an epic regression:** `main`'s `Collaborate` mount has never checked the permission either. The epic only made it more visible, by replacing the old panel with four inviting cards.
- `f67cfe9873` — changelog entries for the epic.
- `4532209594` — key the `:create_workflow` mount guard on `live_action`, not on params, so `/w/new?id=<uuid>` can no longer skip the permission check.
- `40aa7ea6a3` — format channel errors at construction, so `notify: 'none'` (Run, Retry) no longer surfaces the placeholder "Channel request failed" in place of the real message.
- `487078fe43` — disable the landing "Build with AI" card when the project's AI quota is reached, matching the assistant panel.
- `13ff0c1dad` — remove the landing-screen E2E spec (Claude mistakenly added to e2e tests that aren't used anywhere).
- `cba559f6bb` — a failed YAML parse no longer reports itself as a connection error; restore the file input's accessible name; correct the build-from-scratch card copy and the UIStore docs.
- `2b8c51e9a5` — the read-only tooltip no longer tells users to click the removed Create button. It is surfaced by the canvas and the edge/trigger inspectors, none of which sit behind the unmounted header, so it was reachable.
- `1dae82c5a5` — tidying: two test files mocked five `useUI` members deleted with the old left panel, and a comment still explained the AI initial-message flow in terms of it. No behaviour change.
- `b81933f39a` — merge `main`. Part of why the diff is large: it brings in `2489a03950`, which scopes client-supplied ids to the mounted project and enforces edit permission on the server. It also closed four findings from my own review of this branch.

### Notes for reviewers

**Please state reachability for every finding** — the concrete sequence of user actions that produces it. If you can't, say so and rank it below everything else.

This is worth insisting on here because two findings from an earlier review of this work were real mechanisms that no user can reach, and each took real effort to disprove:

- *"`save_and_sync` doesn't promote `workflow_kind`"* — true, but Save & Sync cannot be the first save: the header hosting it is unmounted until the workflow exists. The promotion doesn't cover everything, though — `validate_workflow_name` reads only `assigns.project`, which is what let a saved workflow rename itself "+1". That is being fixed in #5009 (still open at time of writing), so it is **still reproducible on this branch** until that lands.
- *"a second tab keeps a stale `workflow_kind`"* — true, but `/w/new` generates a fresh UUID on every mount and never persists it, so two tabs never share a channel topic.

"This code path is wrong" isn't actionable until we know who gets there.

### What I want checked

Each sub-PR was reviewed in isolation. What none of those reviews could see is the seams between them — the four paths were built weeks apart, and that is where I'd expect problems.

**Remnants of the old flow — my main concern.** The old flow had accumulated years of behaviour and was removed across two cleanup passes. Everything found so far was invisible until someone went looking, because none of it breaks anything — it just sits there: the `Create` button label outliving its button, `save_and_sync` still carrying first-save logic, `?method=` outliving its last caller, a read-only rule pointing at the removed Create button. All fixed, but found by pulling threads rather than reading the diff. The pattern to look for is **logic guarded by a condition the epic made permanently true or false** — gates that no longer gate, branches on a flag whose value is fixed at that point in the tree, handlers for paths that can no longer be entered.

**Do the four paths still agree?** They should differ only in how content is produced, then converge on the same import → save → navigate sequence. Divergence in error handling, save timing, navigation or read-only behaviour is more likely drift than intent.

**State that spans paths.** `isNewWorkflow` and `workflow_kind` both encode "does a row exist yet?" and are updated by unrelated mechanisms; they have already drifted once. Also worth a look: landing-screen visibility vs creation-in-progress, and the connection flags (`connectionState` / `isSessionConnected` / `isSessionConnecting`).

**First-save failure and interruption.** Every path creates the row on first save, and that state has the most branches: save fails, persistent Retry toast, retry succeeds, URL patches late. The offline-gate and reconnect-retry paths are hard to exercise by hand, so they're the least manually validated part of the branch. Same for leaving mid-AI-generation, reconnecting during creation, and two tabs on `/w/new`.

**Tests that pass without asserting anything.** Roughly 60 of the 130 files are tests, and two were found asserting nothing real — one because `waitFor` around a negative assertion resolves immediately, both because their setup made the guard under test unreachable. Both fixed. Green tests here shouldn't be read as coverage; the useful question for any load-bearing test is whether it fails when the behaviour breaks.

## Known and not fixed here

All of this is already known, so it doesn't need re-reporting — but I've split it by what it actually is, rather than calling it all "deferred".

### Open bugs

- **Browse templates hangs for ~10s when the server is unreachable**, instead of failing fast like the other three paths. Known since #4967 and not fixed there either.
- AI assistant is quite slow to come back when creating workflow and just has empty canvas and AI chat - this is something that will be improved later on

### Decisions still open

- **`workflow_kind` is cached at join and hand-patched on save.** Both its readers only ever compare it to `:new`, and both already reload from the database, so the cache buys nothing. Nothing is currently broken by it — every path that could serve a stale value turns out to be unreachable — but it means each new write path has to remember to re-patch the assign. Deriving it at point of use would remove the class. Not done here because it's a design change I'd rather make deliberately than fold into a cleanup.
- **Whether the canvas should be read-only mid-generation.** The `isUnsavedNewWorkflow` rule makes it read-only whenever the doc has content but no row exists — transiently on the AI path, persistently in the AI save-failed state, and invisibly behind the landing overlay during the other paths' import→save window. Its message used to point at the removed Create button; that copy is fixed (see commits below), but the underlying question is a product call I haven't answered.
- **Browse templates has no "Create" button.** Selecting a template card creates the workflow immediately. A confirm step ("Create" button in the template modal) was requested but not built here.

### Pre-existing, not introduced by this epic

- **Channel join errors are invisible to the client.** `y-phoenix-channel@0.1.1` registers only an `"ok"` receiver on join, so every rejection — permission, wrong project, workflow not found, snapshot not found — presents as an editor stuck on "connecting". This is true on `main` today and affects more than creation. **This got more consequential with `2489a03950`**, which added new server-side rejection paths: a viewer whose Y.Doc frames are now being dropped sees no indication at all.
- **Two policy-layer gaps survive `2489a03950`.** `:access_write` and `:access_read` both collapse to bare project membership (`policies/workflows.ex:14`), and the non-member support-user clause returns `support_user` without consulting `allow_as_support_user?/2` (`policies/project_users.ex:128`). Neither is introduced here and neither is specific to creation, but the first now backs the AI-session creation gate, so it is worth knowing.
- **`get_limits` client refresh has never worked.** `getLimits()` pushes and the server replies, but the client registers an `.on('get_limits')` handler for an event the server never pushes, and nothing awaits the reply. Identical on `main`; the three "refresh limits after run" call sites have never done anything.
- **Abandoned AI creations leave orphaned chat sessions.** `cleanup_unsaved_workflow_sessions` only runs on save, so opening `/w/new`, prompting the AI and closing the tab leaves a `workflow_template` session with `workflow_id: nil` forever. Pre-existing mechanism — but the epic makes the AI path the primary route, so it will accumulate faster.
- **The AI-apply state is four ad-hoc refs.** `appliedViaStreamingRef` was promoted to real store state (`streamingApply`) during this epic, but `appliedMessageIdsRef`, `inFlightApplyRef`, `pendingReconnectApplyRef` and `hasLoadedSessionRef` remain. A review during #4902 flagged this as wanting an explicit state machine (`idle | streaming-applied | applying | save-failed`). Worth knowing `hasLoadedSessionRef` is subtle enough that it silently made two tests vacuous.

### Already triaged — please don't re-report

Found in my own review of this branch, reachability-checked, deliberately left. **If you find something not on this list, I want to hear it.**

- The four paths don't exclude each other; "Build it" can clear the canvas under an in-flight build-from-scratch save
- `useCreateWorkflowFlow` is only ever mocked, hiding a dead `isConnecting` alert and an `isConnected`-vs-`isSynced` gate
- YAML modal: mode toggle keeps stale state; Escape destroys the draft
- Templates: a failed fetch leaves a modal that looks healthy once the toast is dismissed (`main` had "Try again")
- LiveView assigns go stale after the in-place URL rewrite — inert today

Follow-on issues filed: 
- #5023 - however Claude thinks this was a bug that already existed on main in the old set up. So it's an improvement. 
- #5024 - something that didn't fit into this epic due to complexity and time constraints.

## Validation steps

**Automated.** `assets` vitest suite green after the `main` merge (159 files, 2565 tests). Backend suite re-run against the merged branch — see CI. `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` clean on changed files. There is no E2E coverage here — Playwright isn't in CI — so the manual steps below are the real test plan.

**The four creation paths.** Each should create the row, rewrite `/w/new` → `/w/{id}` in place with no reload, and reveal the header.

1. **Build with AI** — type a prompt, press Build it. Canvas stays blank while it thinks; the row appears when the first YAML applies, including mid-stream.
2. **Build from scratch** — one click seeds a webhook trigger, one job and an edge.
3. **Browse templates** — open the modal, click a card. No confirm step, by design.
4. **Import from YAML** — open the modal, paste or drop valid YAML, press Create.

**Failure and interruption** (the least-validated area — see "What I want checked").

5. Kill the channel just before the first save on any path → persistent Retry toast, **no** success toast. Restore, press Retry → save succeeds and the URL patches late.
6. Start Build with AI, close the tab mid-generation. Reopen `/w/new` → you get a fresh workflow, not the abandoned one.
7. Go offline immediately after clicking a landing card → the flow either recovers on reconnect or fails with a clear message; it should not sit silently.
8. Open `/w/new` in two tabs → each mints its own UUID; they don't share a channel or overwrite each other.

**Permissions.** Steps 11–12 exercise behaviour that arrived with the `main` merge, on files this branch also reworked — the riskiest part of the merge.

9. Viewer opens `/projects/:id/w/new` → redirected to the workflow list, "not authorized".
10. Viewer opens `/projects/:id/w/new?id=<a workflow in that project>` → still redirected; the query param can't skip the guard.
11. Viewer opens an existing workflow and edits a job name or body → the change is rejected server-side and does not reach other collaborators.
12. Open `/projects/A/w/<id of a workflow in project B>` → "Workflow not found" redirect, not a rendered page.

**Limits and regressions.**

13. Project at its AI quota → the Build with AI input and button are disabled and the server's limit message shows in a tooltip. Other three paths still work.
14. AI disabled instance-wide → the AI card is absent; other three paths still work.
15. Browse a template whose stored YAML no longer parses → the error says the workflow couldn't be read, **not** "check your connection" (`cba559f6bb`).
16. On the YAML import dropzone, confirm the file input has an accessible name — `getByLabelText('Upload a file')`, or a screen reader (`cba559f6bb`).

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)
      (e.g., `:owner`, `:admin`, `:editor`, `:viewer`) — `/w/new` is gated on
      `:create_workflow` at mount, keyed on the route rather than on params so
      `?id=` can't skip it, with tests for both the viewer redirect and the
      `?id=` bypass. The channel independently checks `:create_workflow` before
      resolving a new workflow. The two policy-layer gaps noted under
      "Pre-existing" are not introduced here.


lmac-1:
FYI: @elias-ba @midigofrank the aim of this review is to get it approved and into main and flag anything that should block the merge to main so we can start to unite all the epics in one place. Anything that is follow-on work will be filed in an issue and tackled later.

lmac-1:
Megan will also do a QA review tomorrow but this will not block the merge. This will probably lead to other follow-on improvements.

midigofrank (changes requested):
Hey @lmac-1 , this is nicely done! Tested it out, the only UI feedback is on the YAML upload form: the upload icon isn't clickable.

Image


Additionally, I asked claude to review with maintainability in mind, here what it had to say:

## Findings

### 1. HIGH — Eager error formatting in the constructor turns some channel errors into a permanent hang
`assets/js/collaborative-editor/lib/errors.ts:27`

`ChannelRequestError`'s `super()` now calls `formatChannelErrorMessage({type, errors})`, whose first statement is `channelError.errors.base?.[0]`. Several channel handlers reply *without* an `errors` key — `lib/lightning_web/channels/workflow_channel.ex:577/583/598` reply `{:error, %{reason: ...}}`.

In `useChannel.ts:59` the constructor is evaluated as the argument to `reject(...)`, so it throws **before** `reject` runs: the promise never settles.

**Scenario:** a user without `write_webhook_auth_method` clicks Save in the webhook auth-method panel → `commitAuthMethods` (`inspector/trigger/useWebhookTrigger.ts:97`) awaits forever, no toast, spinner stuck. Previously the message was the static `'Channel request failed'` and the promise rejected cleanly.

**Fix:** default `errors` to `{}` in the constructor, or guard inside `formatChannelErrorMessage`.

---

### 2. MEDIUM — "Build with AI" is a one-way trapdoor with no way back
`assets/js/collaborative-editor/CollaborativeEditor.tsx:218`

`onBuildWithAI` calls `dismissLandingScreen()` then `openAIAssistantPanel(prompt)`. Nothing in the codebase ever sets `showLandingScreen` back to `true` (only `createUIStore.ts:107` initial and `:206` set-false). Meanwhile on `/new` the AI panel has no close affordance: `AIAssistantPanelWrapper.tsx:714` passes `onClose={isNewWorkflow ? undefined : handleClosePanel}`, and the `$mod+k` toggle is disabled by `enabled: ... && !isNewWorkflow`.

**Scenario:** a user types a prompt, hits Build it, then changes their mind (or the AI request errors out) → stuck on a blank canvas with an unclosable panel and no route back to templates / import / blank. Only a full page reload recovers.

This also makes the landing screen hard to extend: any future "back" affordance needs a new store command.

---

### 3. MEDIUM — First-save `workflow_kind` promotion is only in `save_workflow`, not `save_and_sync`
`lib/lightning_web/channels/workflow_channel.ex:392`

`:366` self-promotes `workflow_kind: :existing` after the first save. `save_and_sync` (`:392`) also persists via `Session.save_workflow/2` but never promotes, so a session that reaches it first stays `:new` forever — `request_versions` (`:476`) returns `%{versions: []}` and `get_context` (`:212`) reports `latest_snapshot_lock_version: nil` for the rest of the session.

The only thing preventing this today is a UI invariant documented in a JS comment (`useWorkflow.tsx:604`: "Save & Sync is unavailable while the workflow is still new"). A channel is a public API surface; encoding a server invariant in a client comment breaks the next time the header rules change.

**Fix:** extract the promotion into a shared helper used by both handlers.

---

### 4. MEDIUM — Stale-validation race: Create can import the *previous* YAML
`assets/js/collaborative-editor/components/YAMLImportModal.tsx:115`

`importState` and `validatedState` are only ever written inside the 300 ms `pDebounce`d validator (`:69`). `handleYAMLChange` (`:103`) updates `yamlContent` but does not eagerly move state out of `'valid'`.

**Scenario:** paste valid YAML A, wait for `'valid'`, then paste/edit to B and click **Create** within 300 ms — the button is still enabled, `validatedState` still holds A, and the workflow is created from A while the editor shows B.

**Fix:** set `importState('parsing')` and clear `validatedState` synchronously in `handleYAMLChange`.

---

### 5. MEDIUM (maintainability) — `useWorkflowActions` now hard-requires `LiveViewActionsProvider`
`assets/js/collaborative-editor/hooks/useWorkflow.tsx:377`

The new `const { navigate } = useLiveViewActions();` throws (`LiveViewActionsContext.tsx:39`) when the provider is absent. `useWorkflowActions` is consumed by ~15 components/hooks with nothing to do with navigation — `EdgeForm`, `JobForm`, `WorkflowSettings`, `EdgeInspector`, `JobInspector`, `useTriggerDraft`, `useWebhookTrigger`, `GitHubSyncModal`, `FullScreenIDE`, etc.

All of them now carry a hidden mount requirement for a single call site (the first-save URL patch) — which is why this PR had to touch `__helpers__/triggerInspectorHelpers.tsx` and `storeMocks.ts`. Every future component or test that uses any of the ~20 returned actions inherits the constraint.

**Fix:** read `navigate` in the one place that needs it (a small `useFirstSaveNavigation` hook, or an optional `useContext` with a no-op fallback).

---

### 6. LOW/MEDIUM — `:snapshot_failed` doc says the opposite of what happens
`lib/lightning/collaboration/session.ex:218`

> `- {:error, :snapshot_failed} - Workflow saved but its snapshot failed`

The snapshot is a step in the same `Ecto.Multi` (`workflows.ex` `handle_save_result` matches `{:error, :snapshot, _, %{workflow: workflow}}` from `Repo.transaction`), so the transaction rolls back and **nothing** is persisted.

A maintainer trusting this doc would write recovery code assuming a durable write — skipping a retry, or telling the user their work is saved. The `@spec` at `:230` and the channel reply (`workflow_channel.ex:1029`) are correct; only the doc is wrong.

---

### 7. LOW — Dead branch left behind by the `isNewWorkflow` header gate
`assets/js/collaborative-editor/CollaborativeEditor.tsx:131`

`BreadcrumbContent` returns `null` when `isNewWorkflow` (`:168`), so the `{!isNewWorkflow && <VersionDropdown …/>}` guard inside `breadcrumbElements` is unreachable, and adding `isNewWorkflow` to the `useMemo` deps (`:165`) only forces a pointless recompute. Two independent gates for the same condition in one component is the sort of thing that gets "fixed" incorrectly later — collapse to one.

---

### 8. LOW — Reconnect wipes the user's in-progress template search
`assets/js/collaborative-editor/components/TemplateBrowserModalWrapper.tsx:41`

The reset effect depends on `[isOpen, channel, …]` and unconditionally runs `setTemplateSearchQuery('')` / `setTemplates(BASE_TEMPLATES)` / `setTemplatesLoading(false)`. If the workflow channel object identity changes while the modal is open (socket reconnect), the effect re-runs and clears whatever the user typed, plus re-fetches.

The "reset on open" intent is only correctly expressed by `isOpen`; the `channel` dep should drive the fetch, not the reset.

---

### 9. LOW — Only one creation path is locked during an in-flight create
`assets/js/collaborative-editor/CollaborativeEditor.tsx:222-225`

`useActionLock` disables the "Build from scratch" card via `isBuildingFromScratch`, but `onBrowseTemplates`, `onImportYAML`, and the AI "Build it" button stay live. While the blank-workflow save is in flight the user can open the template browser and pick a template, which runs a second `createWorkflowFrom` → `importWorkflow` (mutating the Y.Doc mid-save) → `saveWorkflow` against the same not-yet-persisted workflow id.

Given the known `workflows_pkey` duplicate-INSERT class on new-workflow saves, a single shared lock across all four landing-screen paths would be safer than four independent ones.

---

### 10. LOW — `Cmd/Ctrl+\` opens the YAML modal unconditionally
`assets/js/collaborative-editor/components/WorkflowEditor.tsx:306`

No toggle, and no check for the template browser already being open, so both Headless UI dialogs can be stacked with two `Escape` handlers registered at the same priority (100) — `YAMLImportModal.tsx:21` and `TemplateBrowserModalWrapper.tsx:36`. Which one Escape closes becomes registration-order dependent.

---

## Cross-cutting maintainability themes

1. **Invariants encoded in the wrong layer.** The `workflow_kind` promotion (#3) is a server rule protected only by a client comment; the `:snapshot_failed` doc (#6) describes behaviour the code doesn't have. Traps for the next person, not bugs today.
2. **A hook that grew a dependency it doesn't need.** `useWorkflowActions` (#5) taxes ~15 unrelated consumers for one call site. The single change here that most raises the cost of adding new things.
3. **The four creation paths aren't modelled as one thing.** Landing screen, templates, YAML, AI each got their own lock, own reset logic, own escape handling — hence #2, #9, #10. Adding a fifth path means re-deriving all four concerns again.

---

## Checked and cleared

- `Workflows.unique_workflow_name/2` — the docstring's claim that the unique index `[:name, :project_id]` is non-partial is correct (`20220905153252_add_project_id_to_workflows.exs:10`), and the query matches the old `Projects.list_workflows/1` semantics exactly.
- `Permissions.can?(:project_users, :create_workflow, user, project)` in the new `Collaborate` mount guard resolves through the generic `%Project{}` clause (`project_users.ex:71`).
- The `remove_ai_disclaimer_read_preference` migration's `?`/`-` jsonb operators are valid (`preferences` is `:jsonb`), and the empty-string down-migration matches two existing precedents.
- `store.saveWorkflow()` can no longer return `null` (it throws via `ensureConnected()`), so dropping the `if (!response) return` guard in `handleSaveSuccess` is safe.
- Removing the `trigger.id` reset effect in `TriggerInspector` is safe — the component is keyed by `trigger-${currentNode.id}` (`inspector/index.tsx:134`).
- The `MessageList` error-rendering split only affects `role === 'assistant'` messages; the backend only sets `status: error` on *user* messages (`ai_assistant_channel.ex:288-300`), so the "Failed to send / Retry" affordance is not lost for real send failures.

lmac-1:
Hey @midigofrank, thanks so much for your detailed review. Here is where we are at with your findings: 

| # | Status | Finding | Comment |
|---|---|---|---|
| 1 | ✅ **Fixed here** | Eager error formatting turns some channel errors into a permanent hang | You're right, and this one's mine — I introduced it in `40aa7ea6a3` while fixing something else. Fixed at the source, with two tests |
| 2 | ❌ **Won't fix** | "Build with AI" is a one-way trapdoor with no way back | It can't be closed by design: the workflow is committed early so the user's work is never lost, and a "back to options" button undoes that. But you're right that the reason I gave in the description doesn't hold up. Hitting the AI quota was a real gap and is already fixed. I've reworded that part |
| 3 | 📋 **Filed - #5028** | `save_and_sync` doesn't promote `workflow_kind` | Already in the known list, and no user can reach it — Save & Sync can't be the first save. Your point still stands though: the channel is a public API, and right now that rule only exists in a JS comment. Rolling it into the issue for deriving `workflow_kind` rather than caching it |
| 4 | 📋 **Filed — #5023** | Stale validation lets Create save the previous YAML | Correct, and I reproduced it. Not new — it came over unchanged from the old YAML panel on `main`. The issue covers this, the mode toggle, and a spinner that never shows; they're all the same fix |
| 5 | 📋 **Filed - #5029 ** | `useWorkflowActions` hard-requires `LiveViewActionsProvider` | Best find in the review, and nothing in my notes covers it. Around 15 components that never navigate now need the provider mounted — that's why this PR had to change the test helpers. Worth doing properly rather than rushing it in here |
| 6 | ✅ **Fixed here** | `:snapshot_failed` doc is inverted | Correct. The snapshot runs in the same transaction, so if it fails nothing is saved at all. Only the doc is wrong, and it's wrong in the direction that tells a user their work is safe. One line |
| 7 | ✅ **Fixed here** | Dead `VersionDropdown` branch behind the header gate | Correct — the whole component returns `null` for new workflows, so that check can never do anything. This is exactly the pattern I asked reviewers to look for, and you found one. Removed |
| 8 | 🟡 **Partly — filed in #5030** | Reconnect wipes the template search | Not reconnects — a rejoin reuses the same channel object, so nothing re-runs. But there's a real path neither of us spotted: the channel is undefined until it attaches, and the landing cards are clickable before that. Open the modal early, type a search, and it gets wiped when the channel arrives. Your fix is the right one |
| 9 | 📋 **Known — filed in #5031** | Only one creation path is locked while saving | Already in the known list. Your framing around duplicate inserts is clearer than mine, and I'll use it in the issue |
| 10 | ✅ **Mostly fixed here** | `Cmd+\` stacks modals | Not unconditional — there's a guard, so it does nothing on a saved workflow. But you're right that it stacks: the landing screen stays open behind the template browser, so two dialogs ended up fighting over Escape. Opening either modal now closes the other, with a test. The wider "only one creation path at a time" rule goes in the issue with row 9 |
| UI | ✅ **Fixed here** | Upload icon not clickable | Real, and I'd talked myself out of it from the code — I assumed the transparent input on top would take the click. Replaced with a proper label covering the dropzone, so it doesn't depend on stacking at all |

Nothing here blocks the merge, can you confirm it's ok for merging? 

1, 6, 7, 10 and the UI report are on the branch; the rest is either already known or tracked.

midigofrank (approved):
Perfect!!

PR #5009, opened 2026-07-22
AI assistants: Stop appending "1" to workflow names

### LINKED ISSUE #5008: Global chat: The assistant appends a "1" to workflow names
When making changes to workflows with the workflow assistant or the global assistant, "1" is added to the workflow name. This is caused by a bug in the workflow name duplication check in Lightning.

hanna-paasivirta:
## Description

When you applied a workflow update from the assistant (workflow chat or global chat), the name kept gaining a trailing "1". The apply path asks the server to make the name unique, and the server compared the incoming name against every workflow in the project including the one being edited. So a saved workflow always clashed with itself and got "1" appended on each update. The fix excludes the current workflow from that comparison, so a workflow's own name no longer counts as a clash. Genuine clashes with a different workflow of the same name still get a suffix as before, and brand-new workflows are unaffected since their id matches nothing saved. Added two tests: one confirming an edited workflow keeps its own name, and one confirming a real clash still suffixes.

Closes #5008 

## Validation steps

1. Make a change to an existing workflow using the workflow assistant (e.g. add a step).
2. Verify that the name of the workflow did not change.
3. Repeat with the global assistant.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


elias-ba (approved):
Nice one @hanna-paasivirta 👏🏽 clean fix with solid test coverage. Worth flagging this also sorts out YAML import and template-apply, since they share the same handler. Approving ✨

elias-ba:
Hey @hanna-paasivirta, by the way I pushed a commit that fixes a flaky test that was failing here. It's nothing to do with your changes, it's a pre-existing flake, but the fix was small and targeted so I didn't want to let it slip by. Happy to move it into its own PR if you'd prefer.

PR #5038, opened 2026-08-03
Global chat: build workflow YAML server-side when the client sends none

### LINKED ISSUE #5001: Global assistant: Workflow YAML not attached at the start of a conversation
The Global assistant in Apollo often doesn’t receive the workflow YAML from Lightning in production. It’s more likely to receive it on the second conversation turn. I can't reproduce the issue when running main locally. 

This could be because Lightning doesn’t have time to load and send the YAML, and it fails silently.

elias-ba:
## Description

This PR fixes the remaining ways a global chat message can reach Apollo without the workflow YAML, by making the server able to build it when the browser sends nothing.

The main variant of #5001 (first message while a job node is open) was already fixed on main by #4888 and just needs a production deploy. This PR covers what's left:

- The server now resolves the workflow YAML by priority: what the client sent, else the live collaboration doc (includes unsaved edits), else the last saved workflow. A new `Lightning.AiAssistant.WorkflowYAML` mirrors the client's serialization so Apollo receives the same shape regardless of who produced it.
- Empty workflows now serialize (their name and trigger are useful context) instead of silently becoming nothing.
- Every place that used to drop the YAML silently now logs, server- and client-side, so the next issue like this is diagnosable from logs in minutes.

Closes #5001

## Validation steps

1. Open a workflow, open the global assistant, and as the very first message ask "List the steps in my workflow" — it should know them.
2. Repeat with DevTools network throttling (Slow 3G) and send immediately after page load, before the canvas renders: the answer should still know the steps (server fallback), and the server log shows which YAML source was used.
3. Try a brand-new workflow with no steps: the assistant should know the workflow's name rather than nothing.

1. The YAML shape intentionally mirrors `convertWorkflowStateToSpec` (includeIds: true) on the client — the module doc names the files to keep in sync. `ExportUtils` was not reused because its shape diverges (no ids, different edge/credential fields).
2. No new dependencies: the YAML emission is hand-rolled string building in the same style `ExportUtils` already uses (block scalars for job bodies, JSON-quoted fallback for anything unsafe as a plain scalar).
3. The live-doc path only reads docs on the same node (Y.Doc is a NIF resource); anything else falls back to the saved workflow. It's covered by an integration test that renames a workflow in the Y.Doc without saving and asserts Apollo receives the live name. Reading the live doc from the AI processor is a seam we should agree on — see #5039, which proposes making this server-side context gathering the norm; this PR's fallback becomes its first piece.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@ and this project adheres to
 
 ### Fixed
 
+- The global AI chat no longer silently reaches Apollo without workflow context.
+  When the client sends no workflow YAML (e.g. the workflow store hadn't
+  hydrated yet), the server now rebuilds it from the live collaboration doc,
+  falling back to the last saved workflow state; empty (zero-job) workflows now
+  serialize instead of being dropped, and any remaining drop is logged on both
+  client and server. [#5001](https://github.com/OpenFn/lightning/issues/5001)
+
 - The AI assistant no longer appends " 1" to a workflow's name each time it
   edits an already-saved workflow. Name-uniqueness validation now excludes the
   workflow being edited, so its own name isn't treated as a clash.
--- a/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
+++ b/assets/js/collaborative-editor/components/AIAssistantPanelWrapper.tsx
@@ -392,11 +392,15 @@ export function AIAssistantPanelWrapper({
             edges,
             positions
           );
-          if (workflowData) {
-            const workflowYAML = serializeWorkflowToYAML(workflowData);
-            if (workflowYAML) {
-              finalContext = { ...finalContext, code: workflowYAML };
-            }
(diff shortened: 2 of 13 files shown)


elias-ba:
Closing after a reality check on the premise. The bug this grew out of (#5001) is fixed by the 2.18 deploy. Of what's left: the most common no-YAML case is workflow creation, where the workflow is empty and there's genuinely nothing to send — not a bug. The remaining harmful case (a slow connection losing real workflow context on the first message) is real but rare, and the architecture proposed in #5039 eliminates it by design rather than patching it.

Rather than merge insurance for a problem nobody has observed, we're folding this into #5039: the YAML module and its hostile-input tests on this branch become the starting material for its first phase. If users hit the race before then, this branch is ready to revive.

PR #5041, opened 2026-08-04
Template preview modal improvements

lmac-1:
## Description

This PR does the template modal improvements to address the following issues:
1. Improve spacing where the scrollbar was
2. Get rid of weird pixel problem in bottom border of the preview section
3. Improve spacing overall

I also did some further updates:

1. Made the title/description part of the preview section a fixed height so there is no layout shift for shorter descriptions
2. Added a fade at the bottom of the template list, so a card cut off by the fold reads as "there is more" rather than as the end
3. Roomier template cards, and the selected/focus states now use gray-500 so a picked card is easier to spot against an unpicked one

**Point 1 turned out to be the hard one, so it's worth explaining what it took.** The list's left and right spacing could never be made to match, because a native scrollbar takes its width out of the content box and that width isn't knowable. macOS can hide scrollbars by default and charges 0px; turn them on, or use Windows, and it's ~15px. It also disappears entirely whenever the list is too short to scroll. So the cards' right edge moved depending on whose machine it was and how many templates happened to exist, and there is no arrangement of padding that fixes that — every version just moved the discrepancy somewhere else.

The list now hides its scrollbar and draws one in the gutter from measured geometry, so the spacing is a fixed 24px on both sides in every case, matching the 24px in the header. The thumb is draggable and hides itself when there's nothing to scroll. The fade at the bottom reads the same measurements, so it only appears while something is genuinely still below the fold — previously it was always on, bleaching the last card even when you'd reached the end.

While in there I fixed a bug and tidied a couple of seams the original PR left behind:

**The preview could show you a different workflow from the one you got.** Creating from a template applies the `pos:` values in its YAML, but the preview discarded them and auto-laid out instead. So a template published from a hand-arranged workflow previewed as one shape and opened as another — and the new workflow silently started in manual layout mode, since that's derived from having any positions at all. The preview now draws the positions it's given, and only falls back to auto-layout when there aren't any (or when they don't cover every node).

**The preview no longer knows what a template is.** It takes a parsed `WorkflowState`, so the modal does the one parse and owns the failure message. Previously the same YAML was parsed twice for two separate decisions — "is Create worth offering?" and "what do I draw?" — which could in principle disagree. Renamed `TemplatePreview` → `WorkflowPreview` to match; it can be pointed at the YAML import modal later for free.

**`onSelect` → `onCreate`.** The prop creates a workflow, but the cards in the same modal also "select" — they pick what the preview shows.

**The diagram has an accessible name.** It was an unlabelled graphics region, so a screen reader announced nothing for the pane the modal is mostly made of.

Part of #4848 

## Validation steps

1. Open the landing screen → **Templates**. Clicking a card only previews it; **Create** is a separate action. Check the spacing around the list's scrollbar, the preview pane's bottom border, and that the description block doesn't shift as you click between templates.
2. **The spacing fix needs checking under both scrollbar settings.** In System Settings → Appearance → Show scroll bars, try **Automatically** and **Always**, and under each compare a full template list against a filtered-down one. The space to the left of the cards and the space between the cards and the preview should be the same in all four, and the thumb should sit centred between the two panes rather than against either. Scrolling to the very bottom should fade the gradient out.
3. [Bonus - not required] Preview a template whose YAML can't be parsed: the preview says so and **Create** is disabled.

1. Creating parses the YAML a second time on purpose: that parse mints the ids the new workflow is built from, and reusing the preview's would give two workflows the same job ids. Commented in the code.
2. **Follow-on work is coming.** The preview still routes through `layout()`, the canvas's animated layout function, purely to get Dagre positions. A branch extracts that Dagre pass into a pure function, which lets the preview drop its provider, effect and `useReactFlow()`. It edits `layout.ts`, which the live canvas depends on, so it's held for the next release rather than landing late in staging.
3. Known and not fixed here: the preview sits ~8px left of centre, because nodes declare a 120px box around a 102px icon and `fitView` centres the box. Pre-existing and shared with the real canvas.
4. **The scrollbar is two hooks, not one.** `useScrollMetrics` owns the element refs and the listeners; `useOverlayScrollbar` turns that geometry into a thumb and a drag handler. They're split so the fade can read scroll position without pulling a scrollbar in with it. `thumbFor()` is exported as a pure function — it's the whole of the arithmetic and touches no DOM, so it can be tested with plain objects.
5. `WorkflowPreview` is now memoised. The modal re-renders on every scroll event now that the scrollbar is drawn from measured state, and re-running a ReactFlow tree on each one isn't cheap. Its only prop is a `WorkflowState` parsed under `useMemo`, so the reference is stable and the memo holds.
6. `thin-scrollbar` is gone from `tailwind.config.ts` — nothing referenced it once the list stopped painting a native bar.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)
- [ ] I have updated the **changelog**. **- don't think it's needed? as it's fixing the templates modal which is new in this release?**


lmac-1:
@midigofrank this is ready for review :)

midigofrank (approved):
Nicely done

Issue #5039, opened 2026-08-03
AI assistant: let the server gather the context instead of the browser


elias-ba:
## Why

Today, when someone talks to the AI assistant, it's the browser that gathers everything the AI needs to know — the workflow, the job code, the logs — and sends it along with the message. The server just passes it on. The odd thing is that the server already has all of this information; we just never ask it to use it.

```mermaid
sequenceDiagram
    participant U as User
    participant B as Browser
    participant L as Lightning
    participant A as Apollo (AI)
    U->>B: types a message
    B->>B: gathers workflow, code, logs...<br/>(can fail silently)
    B->>L: message + context<br/>(one route for the first message,<br/>another for every message after)
    L->>A: forwards whatever arrived
    Note over L,A: if the context is missing,<br/>nobody notices
```

I think this one decision has quietly created most of the AI assistant bugs we've been chasing lately:

- The workflow travels one way on the first message and another way on every message after that. That's why some bugs only ever happen on the first message — it's what bit us in #5001, and what #4888 fixed before it.
- When the browser fails to send something, nothing tells us. The AI just answers without context, and no log explains why. #5001 took six weeks to diagnose mostly because of this silence.
- The same message field means different things depending on who sent it and from where, and knowing which kind of chat a session is relies on a handful of scattered hints rather than one clear answer.
- Every new feature adds one more thing the browser has to remember to send, and one more way for it to fail quietly. We've paid this tax on almost every AI PR this quarter (#4969, #5038).

## What

I'd like us to flip the responsibility. The browser sends only what the user actually said, plus their choices (like "attach my logs"). When the server processes the message, it gathers the context itself, from what it already knows — the saved workflow, and the live editing session when there are unsaved changes.

```mermaid
sequenceDiagram
    participant U as User
    participant B as Browser
    participant L as Lightning
    participant A as Apollo (AI)
    U->>B: types a message
    B->>L: just the message + the user's choices
    L->>L: gathers the context itself<br/>(saved workflow, live editing session)
    L->>A: message + context
    Note over L: can't find something?<br/>one clear log line says what and why
```

What we'd gain:

- The first message and every later message would take the same path, so the whole family of "only on the first message" bugs disappears rather than getting fixed one at a time.
- Failures become loud: the server either finds the context or tells us exactly what it couldn't find.
- The browser gets simpler — it stops carrying data it was never a reliable courier for.
- One clear agreement with Apollo about what gets sent, instead of a shape both sides maintain by habit.

[...]

PR #5043, opened 2026-08-04
Make `APOLLO_TIMEOUT` govern every Apollo request, including streaming

elias-ba:
## Description

> **✅ Deploy precondition met:** this PR required `APOLLO_TIMEOUT=300000` in staging and prod before merging (with the old 60000, this change would have bounded chat streams below observed AI run times). Rory applied and verified both on Aug 3 — safe to merge once approved.

This PR makes `APOLLO_TIMEOUT` do what everyone already thought it did: bound every request to Apollo, including the streaming requests all AI chats use.

The backstory: when streaming was added in March, the streaming client got its own internal timeout key — clearly meant to be wired to an environment variable later, which never happened. So for five months, every environment ran chat streams with a hardcoded 120s no matter what was configured, while `APOLLO_TIMEOUT` quietly governed only non-streaming calls that no live flow uses anymore. The tests stubbed both keys as if both were real, which is what kept this invisible.

## Implementation Details

- Both Apollo clients now read `APOLLO_TIMEOUT`. The 120s default lives in exactly one place (`config.exs`) — the same value streaming has effectively always run at, so nothing changes for environments that don't set the variable. For streams it bounds time-to-headers and each gap between chunks; the AI job's total-runtime ceiling derives from the same value (+10s).
- The dead internal key is deleted, along with 32 test stubs of it.
- Dev default raised from 30s to 300s so local AI planner runs aren't cut off.

Surfaced while auditing the timeout landscape after #5001; the silent-drift theme is the same one #5039 addresses for context-passing.

## Validation steps

1. Set `APOLLO_TIMEOUT=1` locally and send an AI chat message: the stream should now fail fast (it previously ignored the variable entirely).
2. Unset it: chats behave exactly as before (compiled defaults).

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lib/lightning/config/bootstrap.ex
```diff
@@ -138,6 +138,12 @@ defmodule Lightning.Config.Bootstrap do
           :string,
           Utils.get_env([:lightning, :apollo, :endpoint])
         ),
+      # APOLLO_TIMEOUT (ms) bounds every request to Apollo. For streaming
+      # (all AI chat) it is the time-to-headers and max gap between SSE
+      # chunks; the AI job's total-runtime ceiling is derived from it
+      # (+10s). Unset, it falls back to the per-env compiled config
+      # (config.exs default 120_000; dev.exs overrides to 300_000). Size it
```
josephjclark: Let's not specify the values in comments here - there's really no value to it and one day it'll go stale

lib/lightning/config/bootstrap.ex
```diff
@@ -138,6 +138,12 @@ defmodule Lightning.Config.Bootstrap do
...
+      # above the longest expected AI run.
```
josephjclark: Also the sentence `Size it  above the longest expected AI run.` can't be right if the timeout is the allowed gap between chunks?

lib/lightning/ai_assistant/message_processor.ex
```diff
@@ -66,10 +66,9 @@ defmodule Lightning.AiAssistant.MessageProcessor do
     # The Finch receive_timeout on the streaming client is the primary
     # timeout mechanism. The Oban worker timeout is a safety net that
     # should be slightly longer.
-    streaming_timeout_ms =
-      Lightning.Config.apollo(:streaming_timeout) || 120_000
+    timeout_ms = Lightning.Config.apollo(:timeout)
```
josephjclark, line 69: Ok, this is the actual fix. The env var binds to `timeout`, not `streaming_timeout` - on main the env var is ignored, right?

josephjclark (approved):

lib/lightning/config/bootstrap.ex
elias-ba: Good point, removed the values. The comment just describes the resolution order now.

lib/lightning/config/bootstrap.ex
elias-ba: You're right, the gap alone wouldn't justify that sentence. It works because the job's total ceiling is derived from the same value (+10s), so it ends up bounding the whole run too. I reworded the comment so the reason is right there in the sentence.

lib/lightning/ai_assistant/message_processor.ex
elias-ba: Exactly. On main the env var only reaches the non-streaming client, and every live chat flow streams. So for anything users actually touch it was ignored and streams ran on the hardcoded 120s. That's the whole bug.

PR #5047, opened 2026-08-04
Remove the unreachable non-streaming AI chat path

### LINKED ISSUE #5046: AI assistant: remove the unused non-streaming chat path
## What

The AI assistant has a whole conversation path that nothing uses anymore. Before streaming, chat requests to Apollo were plain request-and-wait calls. When streaming arrived it became the only route we actually dispatch through, but the old path was never removed: the non-streaming query functions for job and workflow chat, the HTTP calls behind them, the HTTP client they share, and an Apollo health check that nothing ever calls.

## Why it matters

The dead path still has a full test suite, so it looks alive in coverage reports and keeps costing maintenance: this week alone we cleaned up its timeout stubs in five test files while fixing the real timeout bug (#5043). It also hides the actual shape of the system from anyone reading the code. This is the third piece of silent drift we've found in this area in two days (see also #5039), and the cheapest one to fix.

## What to do

Delete the non-streaming chat path and everything that becomes orphaned with it, including its tests. Streaming behavior must not change at all; the existing streaming and channel test suites staying green is the acceptance bar.

Surfaced while working on #5043.

elias-ba:
## Description

This removes the AI assistant's non-streaming conversation path, which nothing has used since streaming became the only dispatch route. The old query functions for job and workflow chat, the HTTP calls behind them, the client they shared, and an Apollo health check nothing ever called - all gone, along with their tests.

Closes #5046

## Implementation Details

- Everything deleted was verified to have zero production callers before removal, and the compiler (warnings-as-errors) was used to catch anything newly orphaned.
- Helpers shared with the streaming path were left untouched. A few of them lost their only direct test coverage with the deleted tests, so that coverage was ported to the live streaming entry points (context options, langfuse metadata, including the rag-echo regression guard).
- One extra commit fixes a pre-existing test flake (a timestamp tie) that showed up while validating this branch.
- Streaming behavior is unchanged; the streaming, channel, and API controller suites passing unchanged is the acceptance bar from the issue.

## Validation steps

1. Grep for the deleted function names - no production references existed before, none exist now.
2. Run the AI assistant, apollo client, channel, and API controller suites: 391 tests, 0 failures.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


josephjclark (approved):
Looks great. Ran some tests and all 3 endpoints look stable. Nice one!

CHANGELOG.md
```diff
@@ -25,6 +25,14 @@ and this project adheres to
 
 ### Changed
 
+- Removed the unreachable non-streaming AI chat path. Since streaming became the
```
josephjclark: Oh @elias-ba  we should simplify this changelog. Just a simple note here and let the PR explain the details

PR #5048, opened 2026-08-05
Route the "Build with AI" landing prompt through the global assistant

### LINKED ISSUE #5040: Global assistant: Switch default model in AI-first generation to the global assistant
In the new AI-first layout, there is no "Global assistant" tickbox. As the global assistant is relatively stable for one-shot generation, we should switch to using it by default in the "Build with AI" screen.

elias-ba:
## Description

Creating a workflow with the "Build with AI" box now routes through the global assistant instead of workflow-template chat.

This is purely a routing change: no UI changes at all. The global-assistant checkbox in the chat input is untouched, and follow-up messages in the conversation keep consulting it exactly as before (a conversation can already switch assistants between turns via the checkbox, so this stays consistent with that).

Nice side effect: the streaming status updates from #4969 live on the global endpoint, so users now see what the assistant is doing while their workflow is being created.

Closes #5040

## Implementation Details

- The landing prompt's auto-send carries the global-assistant flag (plus the page context Apollo expects), mirroring what a manual global send does. Only that path is affected.
- The backend join already handled global sessions for not-yet-saved workflows correctly, so no routing changes were needed there.
- The "no workflow YAML" warning would have fired on every single creation (a brand-new workflow legitimately has nothing to attach), so it's now an info line for new workflows and stays a warning for saved ones, where missing YAML genuinely means something went wrong.

## Validation steps

1. Create a workflow via "Build with AI" and watch the server log: the request goes to the global chat service, with an info line (not a warning) about the new workflow having no YAML.
2. Continue the conversation with the checkbox unticked: follow-ups go through workflow-template chat as before.
3. Tick the checkbox on a follow-up: it routes global, as before.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ and this project adheres to
 
 ### Changed
 
+- Creating a workflow with "Build with AI" from the new-workflow landing screen
+  now sends the prompt to the global AI assistant instead of the workflow
+  template chat. Follow-up messages in the conversation behave as before.
+  [#5040](https://github.com/OpenFn/lightning/issues/5040)
+
 - Remove the unreachable, non-streaming code in the AI assistant
   [#5046](https://github.com/OpenFn/lightning/issues/5046)
 
--- a/assets/js/collaborative-editor/hooks/useAIInitialMessage.ts
+++ b/assets/js/collaborative-editor/hooks/useAIInitialMessage.ts
@@ -7,6 +7,12 @@
  * - Prepares the workflow context
  * - Initiates the AI connection with the message
  * - Clears the initial message after sending
+ *
+ * The landing prompt always routes through the GLOBAL assistant (#5040):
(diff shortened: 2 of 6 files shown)


lmac-1:
3 little questions @elias-ba 😇: 
1. Does this need some extra set up like pointing to a certain local branch of apollo? I just checked out this code and didn't see any visible change in the way that the AI assistant responded so suspect something isn't quite right on my side. 
2. What happens if the user has "Experimental features" disabled in the User Profile page? Is global assistant still considered an experimental feature?
3. _"Create a workflow via "Build with AI" and watch the server log."_ Can you give more specific instructions as to what this should look like? I have no experience with the global assistant so some extra context would be helpful

elias-ba:
Not comfortable with the idea this PR implements. Closing it until the team decides !

PR #5057, opened 2026-08-09
Invalidate stale YAML validation after edits

### LINKED ISSUE #5023: Creating from pasted YAML within 300ms of an edit saves the previous version
Depends on #4848. However issue was possibly present before this epic. 

### User story

As someone importing a workflow from YAML, I want the workflow that gets created to be the one I can see in the editor, so that I don't silently end up with something I already replaced.

### Details

In the YAML import modal, paste workflow **A**, wait for it to validate, then paste workflow **B** and press Create within 300ms. **Workflow A is persisted while the editor shows B.** There is no error and nothing visibly wrong — the user finds out later, if at all.

**Reproduced by hand on 2026-07-27**, by temporarily widening the debounce from 300ms to 3000ms to make the window hittable. Workflow A was created while the editor displayed B. At the shipped 300ms the window is about a third of a second, so this is easy to miss when testing and easy to hit when pasting quickly.

**Mechanism.** In `assets/js/collaborative-editor/components/YAMLImportModal.tsx`, pasting new content calls `handleYAMLChange` (`:104`), which sets `yamlContent` and schedules the debounce — but resets neither `importState` nor `validatedState`. The Create button is enabled whenever `importState === 'valid'` (`:133-137`), so it stays live, and `handleSave` (`:116`) reads `validatedState` directly and never re-parses. For the whole debounce window the button offers to create content the editor is no longer showing.

`setImportState('parsing')` — the thing that would disable the button — sits *inside* the debounced body at `:80`, so it doesn't run until 300ms after the last keystroke.

**Related, and fixed by the same change:** `setImportState('parsing')` (`:80`) and `setImportState('valid')` (`:85`) run in the same synchronous callback, so React batches them and `'parsing'` never renders. The spinner at `:205` gated on that state is currently dead code; hoisting the call out of the debounce makes both the disable and the spinner real.

**Provenance:** pre-existing — carried over from `main`'s `YAMLImportPanel`, not introduced by the #4848 epic. Filed now because the import path is more prominent in the new landing screen than it was as a left-hand panel.

**A second bug in the same file has the same shape and is worth fixing in the same sitting.** The mode toggle (`:160`) — `setMode(mode === 'upload' ? 'paste' : 'upload')` — clears none of `yamlContent` / `validatedState` / `importState` / `errors`. Paste valid YAML, then switch to Upload: the editor is unmounted but Create is still enabled and will create from content you can no longer see. Mirror case: paste errors stay rendered under the dropzone. Same root cause as the debounce race — state that survives an input change that should have invalidated it.

### Implementation notes

The main fix is three lines: hoist `setImportState('parsing')` out of the debounced body so the button disables the moment the content changes. Belt and braces would be to have `handleSave` re-parse rather than trust `validatedState`.

The mode toggle wants the same treatment — reset validation state when the input source changes.

### Release notes

Fixed a bug where creating a workflow immediately after editing pasted YAML could create the previous version instead.

### User acceptance criteria

- Paste A, wait for validation, paste B, press Create within 300ms → workflow **B** is created. (Equivalently: Create is disabled until validation of B completes.) Easiest way to verify by hand is to widen the debounce to 3000ms temporarily.
- Switching between Upload and Paste clears content, validation state and errors.
- A regression test covers the debounce race.

jmtdev0:
## Summary

- Invalidate pending YAML validations when the editor content or input mode changes.
- Clear stale parsing, validation, and error state immediately after a relevant edit.
- Ignore debounced validation results that belong to an older editor version.
- Add regression coverage for rapid edits and mode changes.

Fixes #5023

## Validation

- YAML import tests: 12 passing
- Prettier
- Focused ESLint
- git diff --check

The full frontend validation was not run because the repository environment is missing the phoenix_live_view/constants dependency and contains unrelated baseline TypeScript errors.

<details>
<summary>AI Disclaimer</summary>

This pull request was primarily developed with assistance from OpenAI Codex, an AI coding agent. For this PR, Codex analyzed issue #5023, inspected the relevant OpenFn Lightning code, implemented the fix, added the regression tests, and ran the reported validation commands under the supervision of jmtdev0.

Human involvement in this PR was very low.

If you do not agree with the use of AI assistance or with the level of human involvement in this PR, please feel free to disregard it, close it, or request changes. I will fully respect that decision.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@ and this project adheres to
 
 ### Fixed
 
+- YAML imports now invalidate validation state immediately when pasted content
+  changes or the input mode switches, so Create cannot persist stale workflow
+  content.
+  [#5023](https://github.com/OpenFn/lightning/issues/5023)
+
 - `APOLLO_TIMEOUT` now governs every request to Apollo, including the streaming
   requests all AI chats use. Streaming previously read an internal timeout key
   that no environment could set, so it was always 120s regardless of
--- a/assets/js/collaborative-editor/components/YAMLImportModal.tsx
+++ b/assets/js/collaborative-editor/components/YAMLImportModal.tsx
@@ -68,9 +68,12 @@ function YAMLImportContent({ onClose, onSuccess }: YAMLImportContentProps) {
   const [validatedState, setValidatedState] =
     useState<YAMLWorkflowState | null>(null);
   const [mode, setMode] = useState<'upload' | 'paste'>('upload');
+  const validationVersion = useRef(0);
 
   const debouncedValidate = useRef(
-    pDebounce((content: string) => {
+    pDebounce((content: string, version: number) => {
+      if (version !== validationVersion.current) return;
+
       if (!content.trim()) {
         setImportState('initial');
         setErrors([]);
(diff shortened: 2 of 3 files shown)


jmtdev0:
The failed \security-review\ check appears to be a workflow-permission issue rather than a problem with this PR.

The workflow runs with the \jmtdev0\ actor having read-only access to the upstream repository, so the Claude review action cannot publish its result. The follow-up attempt to create the review comment is rejected by GitHub with \403 Resource not accessible by integration\.

The CircleCI compile, lint, Dialyzer, Elixir, and JavaScript checks are passing.

Issue #4260, opened 2025-12-18
AI chat session gets stuck in processing state after server restart


elias-ba:
## Description

When the server restarts while an AI message is being processed, the session gets permanently stuck showing the "..." loading indicator with no way to recover. Users must create a new session to continue.

## Steps to Reproduce

1. Open an AI chat session
2. Send a message to the AI
3. While the AI is processing (showing "..."), restart the server
4. Reconnect/refresh the page
5. The session shows "..." forever with no way to cancel or send another message

## Root Cause

1. When server restarts mid-processing, the message remains in DB with `status: :processing`
2. On reconnect, frontend loads this status via channel join and sets `isLoading = true`
3. No recovery mechanism exists to detect orphaned processing messages

## Why This Is Tricky

We can't simply "auto-fail messages in processing state on join" because in collaborative sessions, another user might have legitimately sent a message that's actively being processed.

### Option 1: Time-based recovery
Add a background job that marks messages as `:error` if stuck in `:processing` for > 5 minutes:
- Check `inserted_at` or add `processing_started_at` timestamp
- Run periodic Oban job to clean up orphaned messages

### Option 2: Verify Oban job exists
On session load, check if the Oban job for processing messages actually exists in the queue. If not, the message is orphaned and should be failed.

### Option 3: Client-side timeout warning
After X minutes of `:processing` status, show a warning with a "Retry" or "Cancel" button.

PR #5069, opened 2026-08-15
Stop losing AI messages when a deploy interrupts them

### LINKED ISSUE #5124: A deploy leaves an AI message stuck processing forever
When a deploy interrupts a running AI chat message, nothing says so. The message stays in processing, the panel stays locked for everyone in that session, and no error is raised anywhere. You only find it by looking at the row.

Oban tells us when it stops a job and we are not listening. There is a handler for it that was written and never attached.

There is a second half. Our shutdown window is shorter than an AI job is allowed to run, so a deploy landing on one kills it after the part that would have reported it has already stopped.

elias-ba:
## Description

A deploy cuts off whatever is running. When that was an AI chat message, nothing said so: the message stayed `:processing` forever, the panel stayed locked for everyone in that session, and no error was raised. You only found it by looking at the row.

Closes #5124

Oban announces a job being stopped and we were not listening. The handler for it had been written and never attached. Attaching it needed three other things to be true.

It gets its own handler id rather than joining the exception one. `:telemetry` drops a handler from every event it registered for the first time it raises, so sharing an id would let one bad `:stop` take Oban exception reporting down with it until the next restart.

The two `[:oban, :circuit, _]` events go with it. Oban has not emitted them since 2.6, and there was no clause for them, so one arriving would have raised and detached everything.

The lookups become `get` rather than `get!`. `:stop` fires for every successful job in every queue, which is the busiest path we have, and a message deleted while its job ran would raise there and trigger exactly the detach above. Both paths now have a test that deletes the message first, since returning at all is what proves the clause is there.

The drain window moves from two minutes to six. It was shorter than an AI job's own ceiling, so a deploy landing on a running job killed it after Oban had already stopped the producer that would have reported it. That is no telemetry at all, and the one case attaching `:stop` cannot rescue. Application start now warns if that inverts again, and the ceiling is available on its own as `job_timeout/0` for the callers that have no job in hand.

Two log lines drop to warning: a timeout on the exception path, and a non-success stop. Both sat next to a Sentry capture deliberately set to warning, so the error level was raising a second and louder event for something the code had already judged not to be a fault.

The new test emits the event rather than calling the handler directly, the way the others do. That is why this was covered and still broken.

Part of #4260 (the other half is [#5071](https://github.com/OpenFn/lightning/pull/5071)).

## Validation steps

1. Open a workflow, open the AI assistant, and send a message. While it is still thinking, stop the server (Ctrl+C twice) and start it again. Reload the page: the message should show as failed with a retry, instead of the panel sitting on "..." forever.
2. In IEx, check the drain window is longer than a run can take, which is what makes the above work rather than relying on the sweep:

    ```elixir
     Application.get_env(:lightning, Oban) |> Keyword.get(:shutdown_grace_period)
     Lightning.AiAssistant.MessageProcessor.job_timeout()
     ```

     The first should be comfortably larger than the second.

3. Set `APOLLO_REQUEST_TIMEOUT_MS` high enough to invert those two and restart. There should be a warning at boot naming both.

The two `[:oban, :circuit, _]` events are removed rather than given a clause. Oban has not emitted them since 2.6, so nothing should arrive on them; say so if you disagree.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/message_processor.ex
+++ b/lib/lightning/ai_assistant/message_processor.ex
@@ -62,13 +62,19 @@ defmodule Lightning.AiAssistant.MessageProcessor do
   """
   @impl Oban.Worker
   @spec timeout(Oban.Job.t()) :: pos_integer()
-  def timeout(_job) do
-    # The Finch receive_timeout on the streaming client is the primary
-    # timeout mechanism. The Oban worker timeout is a safety net that
-    # should be slightly longer.
-    timeout_ms = Lightning.Config.apollo(:timeout)
+  def timeout(_job), do: job_timeout()
 
-    timeout_ms + 10_000
+  @doc """
+  The same ceiling as `timeout/1`, for callers that have no job in hand.
+  """
+  @spec job_timeout() :: pos_integer()
+  def job_timeout do
+    # This is the only bound on a run's total duration. The Finch timeout is
+    # per-chunk, so a stream that keeps sending never trips it - which is why
+    # this must stay below Oban's shutdown_grace_period. If it doesn't, an
+    # interrupted job is killed after its producer has already stopped, and no
+    # telemetry fires at all.
+    Lightning.Config.apollo(:timeout) + 10_000
   end
 
   @doc false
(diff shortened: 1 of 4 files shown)


elias-ba:

PR #5070, opened 2026-08-15
Record why the assistant failed on the message

### LINKED ISSUE #5125: We do not record why an AI message failed
Every AI chat failure looks the same, to the user and to us. A hung Apollo, a dropped connection and a deploy interrupting a run all end as one generic error.

So people retry a service that is hard down, and support cannot tell from a screenshot which of those happened.

The reasons already exist, written as sentences a person can read, and then thrown away. All the browser is ever told is a message id and the word error. They should be kept on the message so they survive a reconnect.

elias-ba:
## Description

Every AI chat failure looks the same, to the user and to us. A hung Apollo, a dropped connection and a deploy interrupting a run all end as one generic error, so people retry a hard-down service forever and support cannot tell from a screenshot which happened.

Closes #5125

The reasons already exist as sentences meant for a person to read, and were logged and thrown away: the channel only ever sent a message id and the word error. They now live on the message, as a category to group by and a sentence to show.

Columns rather than a broadcast, because the failure that matters most is a deploy interrupting a run, and that is exactly when the browser reconnects to a different node with the broadcast long gone.

**This is the plumbing, not the visible change.** Nothing in the frontend reads either field yet. What it buys is that the reason is recorded, survives a reconnect, and reaches the client.

The sentence is user-facing prose only; raw error terms stay in the log, where they cannot leak an internal hostname or a stack trace into the chat panel. It is bounded, since it is written through `change/2`, which validates nothing, and re-sent on every channel join.

## Validation steps

1. Stop Apollo (or point `APOLLO_ENDPOINT` at a port nothing is listening on), then send a message from the assistant panel.
2. The panel shows its usual error, unchanged. The new part is on the row, so read it in IEx:
   ```elixir
   import Ecto.Query
   Lightning.Repo.one(
     from m in Lightning.AiAssistant.ChatMessage,
       order_by: [desc: m.inserted_at], limit: 1
   ) |> Map.take([:status, :failure_category, :failure_message])
   ```
   You should get a category and a sentence written for a person, not an error term.
3. Make a service raise mid-answer and repeat. The exception text should be in the server log and *not* in `failure_message`.

Nothing in the frontend reads the two new columns yet, so no user sees anything different once this merges. The reason has to survive a reconnect before anything can show it.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/ai_assistant.ex
+++ b/lib/lightning/ai_assistant/ai_assistant.ex
@@ -32,7 +32,19 @@ defmodule Lightning.AiAssistant do
   @title_max_length 40
   @success_status_range 200..299
 
+  # What a user sees when the reason is ours and not fit to show them.
+  @internal_failure "Something went wrong. Please try again."
+
   @type opts :: keyword()
+
+  @typedoc """
+  Why a streamed query failed.
+
+  A bare string is already written for a user to read. `{:internal, text}` is
+  ours and is not: the caller logs it and shows something else.
+  """
+  @type stream_error ::
+          String.t() | {:internal, String.t()} | Ecto.Changeset.t()
 
   @doc """
   Checks if the AI assistant feature is enabled via application configuration.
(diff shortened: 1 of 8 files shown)


elias-ba:

PR #5071, opened 2026-08-15
Clear AI messages abandoned in processing

### LINKED ISSUE #4260: AI chat session gets stuck in processing state after server restart

elias-ba:
## Description

`processing_started_at` has been written on every message since it started processing, and read by nothing. This reads it.

A message is left processing forever when its job dies without emitting telemetry. Attaching Oban's stop event in [#5069](https://github.com/OpenFn/lightning/pull/5069) covers most of that, but not the case it cannot reach: when a job outlives the drain window, Oban stops the producer first and kills the task after, so there is nothing left to report it. Until something clears the row the panel stays locked for everyone in that session, and nothing raises.

Two guards decide what to reap, because neither is sound alone. Age is vulnerable to clock skew and to a queue that was paused. A live Oban job for the same message is the cross-node authority. The live set is read after the candidates, never before: a job starting in between then appears in it and is skipped, where the other order could reap a message whose job had only just begun.

The write is a guarded update rather than a read-modify-write, so a job that finishes between the select and the write keeps its own result. The grace period sits well clear of the longest a run can legitimately take, so a slow answer is never mistaken for an abandoned one, and a partial index keeps the every-five-minutes scan off the table. The sweep is batched at 200, since each message reaped loads its whole session to tell the panel and the backlog this exists for is exactly when that would be largest. Oldest go first and the rest wait for the next pass.

Reaping something also meant telling the panel which message it was about, which is the second half of this PR.

The broadcast now carries the message id. Without it a listener has to guess, and it guessed by taking the newest, so clearing a message stranded several exchanges back marked an unrelated later message as failed. `broadcast_status/3` carries the id, `broadcast_message_error/2` takes it, and the channel resolves it. Broadcasts that do not carry an id keep the old guess, so nothing else changes behaviour.

Closes #4260

## Validation steps

1. Leave a message stranded, as if its job died without saying so. In IEx:
   ```elixir
   import Ecto.Query
   msg = Lightning.Repo.one(from m in Lightning.AiAssistant.ChatMessage,
     order_by: [desc: m.inserted_at], limit: 1)
   Lightning.Repo.update_all(
     from(m in Lightning.AiAssistant.ChatMessage, where: m.id == ^msg.id),
     set: [status: :processing,
           processing_started_at: DateTime.add(DateTime.utc_now(), -3600)]
   )
   ```
   With that session open in the browser, the panel sits on "...".
2. Run the sweep by hand: `Lightning.AiAssistant.StuckMessageReaper.perform(%Oban.Job{})` The panel should unlock and the message show as failed.
3. For the second half, do the same in a session that has a *later* successful exchange. Only the stranded message should be marked failed; the newer one should be left alone. That is what the id on the broadcast is for.

The second half changes which message the client is told failed. Broadcasts carrying no id keep the old "newest user message" guess, so only the reaper's path behaves differently.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/message_processor.ex
+++ b/lib/lightning/ai_assistant/message_processor.ex
@@ -312,17 +312,47 @@ defmodule Lightning.AiAssistant.MessageProcessor do
   @doc false
   @spec broadcast_status(
           String.t(),
-          atom() | {atom(), AiAssistant.ChatSession.t()}
+          atom() | {atom(), AiAssistant.ChatSession.t()},
+          Ecto.UUID.t() | nil
         ) :: :ok
-  defp broadcast_status(session_id, status) do
+  # message_id is carried so a listener does not have to guess which message
+  # this is about. It guessed by taking the newest, which is wrong the moment
+  # anything reports on an older one - the reaper clearing a message stranded
+  # several exchanges back would have marked the newest as failed instead.
+  defp broadcast_status(session_id, status, message_id) do
     Lightning.broadcast(
       "ai_session:#{session_id}",
       {:ai_assistant, :message_status_changed,
        %{
          status: status,
-         session_id: session_id
+         session_id: session_id,
+         message_id: message_id
        }}
     )
+  end
+
+  @doc """
+  Tells a session one of its messages has failed.
+
+  For callers that write the status themselves - the reaper uses a guarded
+  update rather than a read-modify-write - and so cannot go through
+  `update_message_status/3`.
+  """
+  @spec broadcast_message_error(Ecto.UUID.t(), Ecto.UUID.t()) :: :ok
+  def broadcast_message_error(session_id, message_id) do
+    # get/1 rather than get!/1: the reaper calls this per message inside an
+    # Enum.each, so a session deleted since it selected its candidates would
+    # raise and abandon the rest of the sweep. Nobody is listening to a
+    # deleted session anyway.
+    case AiAssistant.get_session(session_id) do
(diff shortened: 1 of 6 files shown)


elias-ba:

PR #5072, opened 2026-08-15
Keep what arrived when a stream dies

### LINKED ISSUE #5126: A half-finished AI reply is thrown away when the stream fails
The panel types out an answer and then replaces it with a red box. Everything the user just watched appear is gone.

Nothing is kept as it arrives. The saved reply is built only from the final event, so when that event never comes there is nothing left to save.

What did arrive should be kept and shown as cut off, with a way to try again, rather than discarded.

elias-ba:
## Description

A reply the user watched appear was thrown away the moment the stream failed. Every event except the final one returned the accumulator unchanged, so the saved message was built entirely from a `complete` event that never came: the panel typed out an answer and then replaced it with a red box.

Closes #5126

The accumulator now carries the text as it arrives, and the yaml that workflow and global chat send ahead of the text. If the stream ends without completing, whatever did arrive is saved as an assistant message marked incomplete rather than discarded. It is deliberately not presented as a finished answer, so the panel can show it as cut off and offer a retry. The content is clamped to the column's limit, because a reply that ran past it would fail the changeset and lose the lot.

Saving it is not enough on its own. The channel has to send it, and send it before the error, or the client clears its streaming buffer and the user only finds the reply again by reloading. That is a new `new_message` broadcast on the error path.

Apollo's own words stop reaching the panel and the row. Apollo wraps any unhandled exception as `str(e)`, which carries internal hostnames, upstream URLs and container paths. That it failed is worth recording, because it separates a service giving up from a socket going quiet; its words are not, so they go to the log and the user gets a sentence of ours. Three existing tests changed with it, since they asserted on Apollo's text.

Job chat's suggested code is not recoverable this way and is not attempted. It is assembled from the tool call after the model finishes, so when a stream dies mid-answer the code does not exist yet on either side. There is nothing to save rather than something being dropped.

Chunks are prepended and reversed once at the end rather than appended to a growing binary, since a long reply arrives as thousands of them.

## Validation steps

1. Ask the assistant something that produces a long answer, and watch the text stream in. Kill Apollo part-way through.
2. The text you had already watched appear should stay on screen, marked as cut off, with a retry — rather than being replaced by a red error. Reload the page and it should still be there.
3. Check the server log has Apollo's own error text, and that the panel never showed it.

This is where Apollo's own error text stops reaching the panel. Three existing tests changed with it, since they asserted on Apollo's words.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/lib/lightning/ai_assistant/ai_assistant.ex
+++ b/lib/lightning/ai_assistant/ai_assistant.ex
@@ -1090,17 +1090,20 @@ defmodule Lightning.AiAssistant do
   # the full response payload.
 
   defp process_stream(session, body_stream, message_builder) do
-    complete_payload =
+    acc =
       body_stream
-      |> Enum.reduce(nil, fn sse_event, acc ->
-        handle_sse_event(session.id, sse_event, acc)
-      end)
+      |> Enum.reduce(
+        %{complete: nil, text: [], code: nil, apollo_error: nil},
+        fn sse_event, acc -> handle_sse_event(session.id, sse_event, acc) end
+      )
 
-    if complete_payload do
-      {message_attrs, opts} = message_builder.(complete_payload)
-      save_message(session, message_attrs, opts)
-    else
-      {:error, "Stream ended without complete response"}
+    case acc do
+      %{complete: payload} when is_map(payload) ->
+        {message_attrs, opts} = message_builder.(payload)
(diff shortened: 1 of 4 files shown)


elias-ba:

PR #5073, opened 2026-08-15
Say why a stream stopped, and give each wait its own setting

### LINKED ISSUE #4882: AI assistant: streaming errors are hidden behind a generic message
## What's broken

When an AI assistant streaming request fails, Lightning reports the same generic `"Stream ended without complete response"` no matter what actually went wrong. `process_stream/3` in `lib/lightning/ai_assistant/ai_assistant.ex` only returns a real result when it accumulates a `complete` payload off the SSE stream; everything else falls through to that one string. Three very different failures collapse into it: Apollo sent an explicit `error` event, the `complete` event arrived but couldn't be decoded, or the stream genuinely ended early.

The worst of the three is the `error`-event case. The handler calls `broadcast_streaming_error/2` with Apollo's actual reason (bad API key, prompt too long, provider error) and then returns `acc` (nil), so the real message reaches the frontend but is dropped from the return value. `MessageProcessor` then logs the generic string instead. And when a `complete` event fails to decode, the handler returns `nil` with no log at all, so a truncated or malformed payload is indistinguishable from getting no response.

## How it surfaced

Debugging a P2 where the assistant did nothing for a user. Taylor got `[MessageProcessor] Failed to process message: <id>, reason: "Stream ended without complete response"` on every message in his local setup, while Apollo's own logs said the stream completed fine. I read the streaming path end to end to see what that log could actually tell us, and the answer was nothing: you can't tell an Apollo error from a transport problem from a truncated payload, which is exactly what made the P2 slow to pin down.

## What to fix

- When Apollo sends an `error` event, capture the message and return `{:error, message}` so the real reason lands in the logs, not just the frontend broadcast.
- When a `complete` event fails to decode, log it explicitly rather than silently treating it as "no complete".
- Ideally keep the three cases distinct in the returned error so the next failure like this is diagnosable from logs alone.

elias-ba:
## Description

A hung Apollo, a severed connection and a genuinely short answer all logged "Stream ended without complete response". You could not tell them apart, which is what made the P2 behind #4882 slow to pin down. They now log as three different things, and the single timeout behind them becomes three settings that each bound one thing.

Closes #4882

Tesla's Finch adapter throws the reason away. Its stream returns `nil` for a mid-stream error, a mid-stream timeout and a clean end alike. We use our own copy of that adapter, which leaves the reason in the calling process to be read once the stream has been consumed. Upstream has since fixed this, so the copy goes when we bump Tesla, tracked in #5080.

`APOLLO_TIMEOUT` is renamed `APOLLO_IDLE_TIMEOUT_MS`, and joined by `APOLLO_CONNECT_TIMEOUT_MS` and `APOLLO_REQUEST_TIMEOUT_MS`. The old setting only ever measured silence. Connecting used a fixed five seconds you could not change, and nothing bounded a whole request at the HTTP layer, so a slow but steady stream ran until Oban killed the job. The old name now warns at boot that it is ignored, and so does a connect timeout set above the idle one, which can never expire.

The idle default drops from 120s to 30s. That assumes Apollo v3.1.1 or later and its 15s keepalive, so anyone on an older Apollo has to raise it. DEPLOYMENT.md says so.

Apollo also gets its own Finch pool. An AI stream holds a connection for the length of an answer, and sharing the default pool tied up connections other outbound calls were waiting on.

## Validation steps

1. Point `APOLLO_ENDPOINT` at something that accepts a connection and then goes quiet (`nc -l 3000` will do), set `APOLLO_IDLE_TIMEOUT_MS=5000`, and send the assistant a message. It gives up after about five seconds and the `[MessageProcessor] Failed to process message` log says the assistant stopped responding partway through.
2. Do it again, but kill the listener while it is waiting. The same log now says the connection was lost. Those two used to read the same.
3. Set `APOLLO_TIMEOUT` to anything and restart. The boot log warns that it is ignored and names the three settings that replaced it.

`lib/lightning/tesla/adapter/finch.ex` is a copy of upstream's adapter and is most of the diff. It is not a faithful copy: it drops upstream's `build/4` clauses, which Apollo does not need, and the clauses that cannot fire against the Finch we pin. It is excluded from coverage, since only the lines we changed are ours to test.

The authorization box is unticked because nothing here is reachable by a user action with a role attached.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,15 +57,16 @@ and this project adheres to
   complete response"; they now read as three different things.
   [#4882](https://github.com/OpenFn/lightning/issues/4882)
 
-- `APOLLO_TIMEOUT` is replaced by `APOLLO_CONNECT_TIMEOUT_MS`,
-  `APOLLO_IDLE_TIMEOUT_MS` and `APOLLO_REQUEST_TIMEOUT_MS`, which bound reaching
-  Apollo, a silence part-way through an answer, and a whole request. One number
-  had to be sized for the longest of those, so a broken path took as long to
-  notice as a slow answer. Setting the old name now logs a warning at boot
-  saying it is ignored. The idle default of 30s assumes Apollo v3.1.1 or later,
-  which sends a keepalive every 15s. On an older Apollo a working stream can go
-  quiet for longer than that, so raise `APOLLO_IDLE_TIMEOUT_MS` or upgrade
-  Apollo. [#4882](https://github.com/OpenFn/lightning/issues/4882)
+- `APOLLO_TIMEOUT` is renamed `APOLLO_IDLE_TIMEOUT_MS`, and joined by
+  `APOLLO_CONNECT_TIMEOUT_MS` and `APOLLO_REQUEST_TIMEOUT_MS`. The old setting
+  only ever measured silence. Connecting used a fixed five seconds you could not
+  change, and nothing bounded a whole request at the HTTP layer, so a slow but
+  steady stream ran until Oban killed the job. Setting the old name now logs a
+  warning at boot saying it is ignored. The idle default drops to 30s, which
+  assumes Apollo v3.1.1 or later and its 15s keepalive; on an older Apollo a
+  working stream can go quiet for longer than that, so raise
+  `APOLLO_IDLE_TIMEOUT_MS` or upgrade Apollo.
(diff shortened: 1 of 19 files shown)


elias-ba:

PR #5074, opened 2026-08-15
Stop an AI chat failure from losing both the answer and the reason

### LINKED ISSUE #4260: AI chat session gets stuck in processing state after server restart

### LINKED ISSUE #4882: AI assistant: streaming errors are hidden behind a generic message

### LINKED ISSUE #5123: Two AI tests fail only when the whole suite runs
Two tests pass on their own and fail in a full run. That kind of red lands on whoever pushes next rather than on whoever caused it, so it costs time from someone who did nothing wrong.

One picks the assistant's reply by taking the last message. Messages are ordered by when they were written, and that is stored only to the second, so a question and its answer written in the same moment tie and come back either way round.

The other asserts that nothing at all was logged, which any warning from a test running alongside it will break.

Neither is a product bug. They just need to stop depending on things they do not control.

### LINKED ISSUE #5124: A deploy leaves an AI message stuck processing forever

### LINKED ISSUE #5125: We do not record why an AI message failed

### LINKED ISSUE #5126: A half-finished AI reply is thrown away when the stream fails

### LINKED ISSUE #5127: A failed workflow chat loses the steps the user watched
When a workflow chat fails part way and the page is reloaded, the reply comes back missing the status updates the user had watched go by. The text and the workflow are there, the steps are not.

It reads as though less happened than actually did, which makes the failure harder to describe to us than it needs to be.

elias-ba:
## Description

An AI chat that failed left the user with nothing and left us with nothing to go on. The answer they had watched appear was discarded, the message could sit in "processing" for good, and every different failure logged the same sentence. This was seven stacked PRs; they are folded into this one because reviewing them in order was more work than reviewing them together.

A deploy cut off whatever was running and nothing said so. The message stayed in processing, the panel stayed locked for everyone in that session, and no error was raised anywhere. Oban announces a job being stopped and we were not listening: the handler had been written and never attached. The shutdown window was also shorter than an AI job's own ceiling, so a deploy landing on one killed it after the part that would have reported it had already stopped. That window moves from two minutes to six, boot warns if it inverts again, and a cron sweep clears anything still stranded.

A reply the user watched appear is kept instead of discarded. Every event except the final one was thrown away, so the saved message was built entirely from an event that never came, and the panel typed out an answer and then replaced it with a red box. What arrived is now saved as an incomplete reply, along with the workflow YAML and the status updates, in the order they were shown.

A hung Apollo, a severed connection and a genuinely short answer all logged one sentence, which is what made the incident behind #4882 slow to pin down. Tesla's Finch adapter discards the reason, so we use our own copy that keeps it. Upstream has since fixed this, so the copy goes when we bump Tesla, tracked in #5080.

Why a chat failed is now recorded on the message and shown. Apollo's own words reach the reader only when Apollo named the failure as one of the kinds it writes for a person: `entry.py` rewraps every unhandled exception as `ApolloError(500, str(e), type="INTERNAL_ERROR")`, and `str(e)` carries internal hostnames and container paths, so a type on its own is not enough to trust the text.

The panel shows a failure once, where the answer belongs. It used to put two red things on screen for one failure: the prompt said "Failed to send", which is untrue since it was sent and often half answered, and the half answer sat in a red box with an error icon, as though the text were the error. There is now one grey notice under whatever arrived, carrying the reason, with one Try again that re-runs the prompt. The prompt takes the notice only when nothing came back at all.

`APOLLO_TIMEOUT` is renamed `APOLLO_IDLE_TIMEOUT_MS` and joined by `APOLLO_CONNECT_TIMEOUT_MS` and `APOLLO_REQUEST_TIMEOUT_MS`. The old setting only ever measured silence: connecting used a fixed five seconds you could not change, and nothing bounded a whole request at the HTTP layer. Boot warns if the old name is still set, and if a connect timeout is set above the idle one, which can never expire. The idle default drops to 30s, which assumes Apollo v3.1.1 or later and its 15s keepalive, so anyone on an older Apollo has to raise it. Apollo also gets its own Finch pool, since an AI stream holds a connection for the length of an answer and was tying up connections other outbound calls were waiting on.

Two tests that passed alone and failed in a full run are fixed on the way through. Neither was a product bug.

Closes #4260
Closes #4882
Closes #5123
Closes #5124
Closes #5125
Closes #5126
Closes #5127

## Validation steps

Point `APOLLO_ENDPOINT` at something you can break on demand. `nc -l 3000` is enough for the first two.

1. Have it accept the connection and then go quiet. Set `APOLLO_IDLE_TIMEOUT_MS=5000` and send a message. The reply gives up after about five seconds, whatever text arrived stays as normal text, and one grey line under it says the assistant stopped responding partway through, with Try again.
2. Do it again but kill the listener while it is streaming. Same shape, but the line now says the connection was lost. Those two used to be the same sentence, and neither used to reach the panel at all.
3. Reload. The half answer and the reason are both still there.
4. Send a message and restart the server while it is answering. The message sits in "processing" with the panel locked, and a restart does not heal it. Within five minutes the sweep turns it into "The assistant was interrupted before it finished" with a Try again, without a reload.
5. Set `APOLLO_TIMEOUT`, or set `APOLLO_CONNECT_TIMEOUT_MS` above `APOLLO_IDLE_TIMEOUT_MS`, and restart. Boot warns about each.

This was #5079, #5069, #5070, #5071, #5072, #5073 and #5074 stacked in that order. They are closed and every commit is here.

`lib/lightning/tesla/adapter/finch.ex` is a copy of upstream's adapter and is a large part of the diff. It is not a faithful copy: it drops upstream's `build/4` clauses, which Apollo does not need, and the clauses that cannot fire against the Finch we pin. It is excluded from coverage, since only the deviations are ours to test.

Two limitations worth knowing rather than discovering. Finch reports the idle timeout and the whole-request cap as the same value, so an answer cut off for running long reads as a stall; the log names both ceilings so an incident can be told apart by how long the request lasted. And a retry appends a new reply rather than replacing the failed one, so repeated failures stack in the transcript. Both are called out in the code.

The authorization box is unticked because nothing here is reachable by a user action with a role attached.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


elias-ba:
I am going to merge this. I have tested it extensively both the code and the UI. I am happy with it.

PR #5076, opened 2026-08-16
feat: normalise datetime filters for non utc timezones

### LINKED ISSUE #4983: Existing input filters are broken in non UTC timezones
I am in timezone UTC+1 right now

Look closely at this image:

Image

- The filter is set to only show dataclips created before 16:39
- The list shows several dataclips created after 16:40

if I update the criteria to show dataclips created before  15:40 it shows this:

Image

Note that there is one dataclip created at 16:38 - which should be filtered out!

What's happening here is that my browser is showing times in my local timezone, UTC+1. So when it says 16:40 on my UI, it actually means 15:40 in the database.

When I pick 16:40 in the date picker, I _think_ we need to internally adjust that time to match server time

Found while on prod 2.16.8

Promist-Moon:
## Description

This PR closes #4983.  This PR adds end-to-end coverage for the timezone handling bug in History filters, where URL datetime values (e.g. 2026-07-22T10:05) filtered out workflows only in UTC timezones.

Changes made:
1. Filters in Workflow History filter out workflows by the browser's local timezone, detected by the browser's websocket offset minutes, both from editing the URL directly in the URL bar and through the page's filters

### Validation

Go to the History tab.

Workflow run at 3:10PM UTC+8 (7:10AM in GMT) is filtered out when the filter 3:51PM is applied.

image

image

image

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have used another model (Codex 4.1)
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


assets/js/app.js
```diff
@@ -63,7 +63,10 @@ const csrfToken = document
   .getAttribute('content');
 
 const liveSocket = new LiveSocket('/live', Socket, {
-  params: { _csrf_token: csrfToken },
+  params: {
```
elias-ba, line 66: We read the offset here once, when the socket connects, and the server then reuses that single number for every date anyone filters on. For anyone in a zone that shifts for summer, that number is only right for half the year. Someone in the UK opening History in September has an offset of -60, so filtering for a run back in January takes an hour off a time that was already GMT. January comes out as 09:05Z when it should be 10:05Z.

Your dataclip side doesn't have this, because handing the local string to the browser lets it apply the rules for that particular date. Same zone, the browser gives 10:05Z for January and 09:05Z for July, both right.

So I think we want the zone name here rather than the offset, Intl.DateTimeFormat().resolvedOptions().timeZone, and the same conversion on both sides. One thing that'll bite you if you don't know it, we ship tzdata but we've never pointed Elixir at it, so a named zone currently comes back as :utc_only_time_zone_database. It needs config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase adding. Small, but it will stop you dead.

test/lightning_web/live/run_live/index_test.exs
```diff
@@ -928,6 +928,30 @@ defmodule LightningWeb.RunLive.IndexTest do
       assert html =~ "after"
     end
 
+    test "datetime-local filter fields keep URL wall-time value (no visual timezone shift)",
```
elias-ba: I don't think this one can fail. The value in that input renders from the filters assign, and the conversion only ever runs on a copy inside the search, so the assertion holds either way. Flipping the sign of the conversion, changing the declared offset to +05:30, or taking the connect params out entirely all leave it green.

What I'd love instead is a work order at a known instant, an offset that isn't zero, and an assertion that it lands on the right side of the boundary. That's the thing Joe reported, and right now it's the one thing none of us can see break.

test/lightning_web/controllers/dataclip_controller_test.exs
```diff
@@ -351,6 +351,34 @@ defmodule LightningWeb.DataclipControllerTest do
       assert %{"data" => _dataclips} = json_response(conn, 200)
     end
 
+    test "accepts direct API datetime without timezone in before filter", %{
```
elias-ba: Same thing here. This passes as long as the response carries a data key, whatever ended up in the list, so it stays green however the filter behaves. Two dataclips either side of the boundary and a note on which one should come back would earn its keep.

lib/lightning_web/live/run_live/index.ex
```diff
@@ -894,6 +924,83 @@ defmodule LightningWeb.RunLive.Index do
...
+       when value in [nil, ""],
+       do: value
+
+  # Convert browser-local wall time to UTC using the browser offset from connect params.
+  defp normalize_datetime_filter_value(value, offset_minutes)
+       when is_binary(value) and is_integer(offset_minutes) do
+    if String.match?(value, ~r/(Z|[+-]\d{2}:?\d{2})$/) do
+      value
+    else
+      with {:ok, naive} <- parse_local_naive_datetime(value) do
```
elias-ba: This is what's turning lint red. Credo wants a case when there's a single arrow and an else. Main is green, so it came in with this branch rather than being something you walked into.

lib/lightning_web/live/run_live/index.ex
```diff
@@ -894,6 +924,83 @@ defmodule LightningWeb.RunLive.Index do
...
+        _ ->
+          nil
+      end
+    end
+  end
+
+  defp normalize_history_datetime_filters(filters, nil), do: filters
+
+  defp normalize_history_datetime_filters(filters, offset_minutes)
+       when is_integer(offset_minutes) do
```
elias-ba: Small one. With nil as the default this adds the key when it wasn't there, so filter maps come out carrying empty date entries they didn't go in with. Nothing downstream minds, nil and missing land in the same place. Could you use Map.replace instead, so it says what we actually meant.

lib/lightning_web/live/run_live/index.ex
```diff
@@ -894,6 +924,83 @@ defmodule LightningWeb.RunLive.Index do
...
   end
 
+  @history_datetime_filter_keys ~w(
+    date_after
+    date_before
+    wo_date_after
+    wo_date_before
+  )
+
+  defp browser_tz_offset_minutes(socket) do
```
elias-ba: The missing else means the nil on the dead render is accidental rather than chosen. Could you add an explicit clause so the next reader knows it's deliberate.

lib/lightning_web/live/run_live/index.ex
```diff
@@ -756,7 +783,10 @@ defmodule LightningWeb.RunLive.Index do
   end
 
   defp handle_bulk_cancel(socket, %{"type" => "all"}) do
-    filter = SearchParams.new(socket.assigns.filters)
+    filter =
+      socket.assigns.filters
+      |> normalize_history_datetime_filters(socket.assigns.tz_offset_minutes)
```
elias-ba: Worth noting the template still builds the search params from the raw filters for the bulk modals, while this now uses the converted ones. Nothing's broken, and the raw one is arguably right there since what people want to see is the time they typed. But the same call meaning two things will catch someone out later, so please add a line of comment saying the display wants wall time on purpose.

lib/lightning_web/controllers/dataclip_controller.ex
```diff
@@ -124,47 +124,34 @@ defmodule LightningWeb.DataclipController do
...
-                    project
-                  )
-
-                json(conn, %{
-                  data: dataclips,
-                  next_cron_run_dataclip_id: next_cron_run_dataclip_id,
-                  can_edit_dataclip: can_edit_dataclip
-                })
-
+            with {:ok, filters} <- NewManualRun.get_dataclips_filters(params),
```
elias-ba: One thing worth knowing about moving the parsing up here. It now runs before the project check, and the comment sitting just above that check says this should never hand back an error the client can tell apart from an empty list. A bad date against a job in someone else's project used to come back empty and now comes back 422. I don't think it gives anything away, since the 422 depends only on the date and not on the job. It does contradict the comment though, so either the order or the comment wants adjusting.

lib/lightning_web/controllers/dataclip_controller.ex
```diff
@@ -124,47 +124,34 @@ defmodule LightningWeb.DataclipController do
...
+                  project
+                )
+
+              json(conn, %{
+                data: dataclips,
+                next_cron_run_dataclip_id: next_cron_run_dataclip_id,
+                can_edit_dataclip: can_edit_dataclip
+              })
+            else
               {:error, changeset} ->
```
elias-ba: I think this can come off. The only thing that reaches it is the error from the parse above, and returning it is what would happen anyway.

assets/js/collaborative-editor/components/ManualRunPanel.tsx
```diff
@@ -394,8 +404,10 @@ export function ManualRunPanel({
     const timeoutId = setTimeout(() => {
       const filters: Record<string, string> = {};
```
elias-ba: This block is identical to buildFilters just above, which is why the change needed making twice. Please call buildFilters here so next time it only needs doing once.

elias-ba (changes requested):
Thanks @Promist-Moon, I enjoyed reading this and I'd like to land it. You've spotted the real thing, which is that the list and the filter were speaking two different clocks, and the dataclip half in particular is the right shape.

I've put the detail on the lines rather than pile it up here. The two that matter most are the one on app.js and the one on the History test. The others are smaller, but I'd like them all picked up while you're in here.

Give me a shout when it's had another pass and I'll come straight back to it.

assets/js/app.js
Promist-Moon: [refactor: use browser timezone name over offset](https://github.com/OpenFn/lightning/pull/5076/commits/ee9e512ae9b39f8ce1bb2bc72a343bb45854cdee)

test/lightning_web/live/run_live/index_test.exs
Promist-Moon: Part 1: [chore: remove redundant tests](https://github.com/OpenFn/lightning/pull/5076/commits/4e21d29d553d130fc80973205afcf3a7642b4b0c)

test/lightning_web/controllers/dataclip_controller_test.exs
Promist-Moon: Part 1: [chore: remove redundant tests](https://github.com/OpenFn/lightning/pull/5076/commits/4e21d29d553d130fc80973205afcf3a7642b4b0c)

lib/lightning_web/live/run_live/index.ex
Promist-Moon: [fix: address lint issue with case](https://github.com/OpenFn/lightning/pull/5076/commits/f5683cd7ba37ee80f68d1ac452b772c58ec7eaa5)

lib/lightning_web/live/run_live/index.ex
Promist-Moon: [refactor: use replace instead of update in datetime filter](https://github.com/OpenFn/lightning/pull/5076/commits/66d516602807e38b8b74ccdfafa9cd7faa71a6e8)

lib/lightning_web/live/run_live/index.ex
Promist-Moon: [refactor: use browser timezone name over offset](https://github.com/OpenFn/lightning/pull/5076/commits/ee9e512ae9b39f8ce1bb2bc72a343bb45854cdee)

lib/lightning_web/live/run_live/index.ex
Promist-Moon: [chore: add comment for updated filter definition](https://github.com/OpenFn/lightning/pull/5076/commits/930c47f099ddc0bfcae76c047a488bcc7420d755)

assets/js/collaborative-editor/components/ManualRunPanel.tsx
Promist-Moon: [chore: lint manualrunpanel.tsx](https://github.com/OpenFn/lightning/pull/5076/commits/7eea1283aded6f23b2ca7bb6f4f5828f961bbc8e)

lib/lightning_web/controllers/dataclip_controller.ex
Promist-Moon: i'm not sure if what you mean is to remove the else clause returning the error, so that's what I did here. [chore: remove redundant else in dataclip controller](https://github.com/OpenFn/lightning/pull/5076/commits/9eada3449916e73407c1a61dda1468ac5ae07b77)

lib/lightning_web/controllers/dataclip_controller.ex
Promist-Moon: Thanks for the comment! I've adjusted it such that parsing happens after the job check, by searching dataclips via raw params instead of the filtering, and normalising the params inside the `search_selectable_dataclips/5` function

[fix: parse filters after job-in-project to preserve error behaviour](https://github.com/OpenFn/lightning/pull/5076/commits/7e57a1667c22cddffec67d2b05b6b269747cd267)

test/lightning_web/live/run_live/index_test.exs
Promist-Moon: [test: add test for summer-winter offsets](https://github.com/OpenFn/lightning/pull/5076/commits/5d99530e31a3de7b6cd9bcd51adf116672fc629c)

[test: add test for summer offset](https://github.com/OpenFn/lightning/pull/5076/commits/2869a5c2c3cc14421ffbf0ecd3ea510f8613f883)

test/lightning_web/controllers/dataclip_controller_test.exs
Promist-Moon: [test: add timezone boundary coverage for dataclip before filter](https://github.com/OpenFn/lightning/pull/5076/commits/0e1e15e0113ab8177e6c7b3c85a2c55dfcc90856)

Promist-Moon:
@elias-ba hello I've made the changes to the PRs are requested :) Please take the time to review them when you can, but no rush!

PR #5079, opened 2026-08-17
Stop two AI tests depending on timing they do not control

### LINKED ISSUE #5123: Two AI tests fail only when the whole suite runs

elias-ba:
## Description

Two tests that passed alone and failed in a full run, which is the worst kind of red: it lands on whoever pushes next rather than on whoever caused it.

Closes #5123

The chat tests picked a reply with `List.last/1`, but messages are ordered by `inserted_at` and that is stored to the second, so a question and its answer written in the same millisecond tie and come back either way round. They now pick by role, which is what each of them meant. The ordering itself is left alone; that needs a migration and is filed as #5081.

The credentials schema test asserted that nothing at all was logged, which any warning from a neighbouring async test would break. It now asserts the absence of the one message it actually cares about.

## Validation steps

1. Nothing here is user-visible, so there is no scenario to click. The check is the suite: `mix test test/lightning/ai_assistant/ai_assistant_test.exs test/lightning/credentials/schema_test.exs`
2. To see the flake it fixes, put `List.last/1` back in place of `assistant_message/1` and run the full suite a few times. The global chat assertions fail on some runs and not others.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [ ] I have updated the **changelog**

### DIFF
--- a/test/lightning/ai_assistant/ai_assistant_test.exs
+++ b/test/lightning/ai_assistant/ai_assistant_test.exs
@@ -749,7 +749,7 @@ defmodule Lightning.AiAssistantTest do
           code: workflow_yaml
         )
 
-      saved_message = List.last(updated_session.messages)
+      saved_message = user_message(updated_session)
       assert saved_message.code == workflow_yaml
       assert saved_message.role == :user
     end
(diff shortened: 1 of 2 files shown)


elias-ba:

PR #5084, opened 2026-08-18
Remove unreachable streamed AI error-body branch

### LINKED ISSUE #5083: Delete the unreachable branch that prefers Apollo's error text over ours
`handle_error_response/2` in `lib/lightning/ai_assistant/ai_assistant.ex` still prefers Apollo's own `body["message"]` over a sentence of ours when a request comes back with a non-success status. `error_message_from_body/1` is what reads it.

That is left over from when the assistant was not streaming. Showing Apollo's message to the user goes back to #2787 in December 2024, where it was read straight off the body with no guard and no fallback. #4969 later narrowed it, adding both, because a streamed body is a lazy stream rather than a decoded map and could not be indexed.

Now that every call streams, the clause that would index the body never matches, so the user always gets our own "AI server returned an error (HTTP ...)" instead. In other words we stopped showing Apollo's text by accident rather than on purpose, and what is left is a branch that cannot run.

Worth deleting rather than leaving. Two reasons.

It reads as if it is live, so anyone working in here has to trace all three callers to convince themselves it is not, which is a small tax every time.

And if a non-streaming call is ever added back, it quietly comes to life. Apollo wraps an unhandled exception as `str(e)`, so its `message` is not written for a person to read. The rest of the assistant now deliberately keeps Apollo's words in the log and shows the user a sentence of ours; this branch would do the opposite. It would also stick, rather than pass by, since failure reasons are stored on the message now and re-sent to the client on every channel join.

So: drop `error_message_from_body/1`, always use our own status sentence for the user, and keep logging Apollo's text as we already do.

Noticed while reviewing the AI streaming stack (#5079 and the PRs above it). Deliberately not changed there, since that stack does not otherwise touch these lines.

jmtdev0:
<details>

<summary>AI Disclaimer</summary>

This pull request was primarily developed with assistance from OpenAI Codex (GPT-5.6 Luna Max), an AI coding agent. For this PR, Codex analyzed issue #5083, inspected the relevant OpenFn Lightning code and tests, implemented the focused fix, and ran the affected test suite under the supervision of jmtdev0.

Human involvement in this PR was very low.

If you do not agree with the use of AI assistance or with the level of human involvement in this PR, please feel free to disregard it, close it, or request changes. I will fully respect that decision.

</details>

## Description

This PR removes the unreachable error_message_from_body/1 branch from AI assistant error handling. Assistant calls now stream responses, so the response body is lazy and cannot be indexed as the legacy decoded map. Failed requests now use the local HTTP status message consistently.

Closes #5083

## Validation steps

1. Run MIX_ENV=test mix deps.get && mix deps.compile && mix test test/lightning/ai_assistant/ai_assistant_test.exs in the Docker development environment.
   - 112 tests, 0 failures.
2. Run git diff --check.
3. mix format --check-formatted was checked; it reports pre-existing formatting differences in other parts of these existing files, so no unrelated formatting changes were included.

1. The change is intentionally scoped to deleting the unreachable body-message fallback and updating the three affected expectations.
2. This change does not alter authorization behavior.
3. The PR is ready for maintainer review.

## AI Usage

Please disclose whether AI was used anywhere in this PR:

- [ ] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

- [x] I have performed an AI review of my code.
- [x] Related authorization policies are not applicable to this error-message-only change.
- [ ] I have updated the changelog.
- [x] I have ticked a box in AI usage in this PR.

### DIFF
--- a/lib/lightning/ai_assistant/ai_assistant.ex
+++ b/lib/lightning/ai_assistant/ai_assistant.ex
@@ -1259,11 +1259,9 @@ defmodule Lightning.AiAssistant do
 
   defp handle_error_response(error_response, session) do
     case error_response do
-      {:ok, %Tesla.Env{status: status, body: body}}
+      {:ok, %Tesla.Env{status: status}}
       when status not in @success_status_range ->
-        error_message =
-          error_message_from_body(body) ||
-            "AI server returned an error (HTTP #{status})."
+        error_message = "AI server returned an error (HTTP #{status})."
 
         Logger.error(
           "AI query failed for session #{session.id}: #{error_message}"
@@ -1287,13 +1285,6 @@ defmodule Lightning.AiAssistant do
         {:error, "Oops! Something went wrong. Please try again."}
     end
   end
-
-  # Streaming requests carry a lazy Stream (a fun or %Stream{} struct) as the
-  # body, so error responses can't be indexed like decoded JSON maps.
-  defp error_message_from_body(body) when is_map(body) and not is_struct(body),
-    do: body["message"]
-
-  defp error_message_from_body(_body), do: nil
 
   defp build_job_message(body) do
     message = body["history"] |> Enum.reverse() |> hd()
--- a/test/lightning/ai_assistant/ai_assistant_test.exs
+++ b/test/lightning/ai_assistant/ai_assistant_test.exs
@@ -2579,7 +2579,7 @@ defmodule Lightning.AiAssistantTest do
          }}
       end)
 
-      assert {:error, "Bad request"} =
+      assert {:error, "AI server returned an error (HTTP 400)."} =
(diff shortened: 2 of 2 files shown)


jmtdev0:
CI note:

The CircleCI lint job is currently failing only at hex.audit because the repository dependency cowlib 2.19.0 is flagged for CVE-2026-43971.

This PR changes only lib/lightning/ai_assistant/ai_assistant.ex and test/lightning/ai_assistant/ai_assistant_test.exs; it does not modify mix.lock or dependency versions. The other lint stages (format, Credo, Sobelow, and deps.audit) passed, and the compile, Dialyzer, Elixir, and JavaScript checks are green.

I have left the dependency advisory out of this focused PR because resolving it would be an unrelated dependency change.

PR #5087, opened 2026-08-21
fix: Recover GitHub installation lookup from invalid OAuth tokens

### LINKED ISSUE #4972: Improve GitHub Sync error handling when dropdown disabled

Users see this and can't select GitHub installations dropdown:
Image

It's not clear what they need to do but is normally because their GitHub token is expired. 
The steps to fix this is to go into User Profile, disconnect GitHub and reconnect GitHub and it works again. 
This isn't a great experience and should be improved. 

**Version number** The version of Lightning you have seen the bug on: `v2.16.8 (1.5.89-pre)`

- [x] Yes
- [ ] No

1. Go to a project
2. Settings > Sync to GitHub

**Expected behavior** User can select from the dropdown. If they can't, it should be clear what to do to resolve and not leave the user stuck.

**Additional context** Not sure why this message appears but we should handle it more gracefully.

mvanhorn:
## Description

Extend the existing token lifecycle in `Lightning.VersionControl` so an installations request rejected with `401` gets one recovery attempt when the stored grant has a usable refresh token: refresh and persist the OAuth token, rebuild the bearer client, and retry the installations request exactly once. If refresh is unavailable or fails, or the retry is also unauthorized, return the existing `GithubError.invalid_oauth_token/2` contract so `GithubSyncComponent` renders its current profile/reconnect guidance; preserve the existing raw error behavior for non-authentication GitHub failures. Keep this logic in the production installation-fetch path rather than introducing a test-only seam, and record the user-visible fix in the changelog.

The project GitHub Sync settings load a user's installations through `Lightning.VersionControl.fetch_user_installations/1`, and the installation selector remains disabled when that asynchronous request fails. Locally expired access tokens are already refreshed, but a `401` returned by GitHub is reduced to an untyped response body, so the LiveComponent treats an authentication failure like a generic API outage. Users therefore receive refresh-or-contact-support guidance instead of an automatic recovery attempt or the existing invalid-token reconnect guidance.

Closes #4972

## Validation steps

This PR adds unit and LiveView tests covering the cases below. They are included in the diff but were not executed in my local environment, so CI is the first run.

1. A stored access token receives `401`, its valid refresh token produces a replacement token, and the single retried installations request succeeds; asserts the replacement is persisted and installations are returned.
2. A `401` cannot be refreshed because the grant is non-refreshable or GitHub rejects the refresh; asserts the result is an `:invalid_oauth_token` `GithubError` without repeatedly calling the installations endpoint.
3. The retried installations request also returns `401`; asserts the retry stops after one attempt and returns the typed invalid-token error.
4. A non-authentication response such as `500` remains a generic GitHub API failure and does not trigger token refresh.
5. Project Settings receives an unrecoverable installation authentication failure and renders the expired/invalid authentication banner with the profile reconnect link while leaving the installation selector disabled.

The retry is deliberately capped at exactly one attempt so a persistently rejected grant cannot loop against the installations endpoint. Non-authentication failures are left on their existing path so this change does not alter behavior for GitHub outages.

## AI Usage

AI was used for assistance.

## Pre-submission checklist

- [ ] I have implemented and tested all related **authorization policies**. (e.g., `:owner`, `:admin`, `:editor`, `:viewer`) - not applicable: this change does not touch authorization policies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ and this project adheres to
 
 ### Fixed
 
+- GitHub Sync now refreshes and retries a rejected OAuth token once when loading
+  app installations, and directs users to reconnect their account when recovery
+  is not possible. [#4972](https://github.com/OpenFn/lightning/issues/4972)
 - The AI assistant no longer appends " 1" to a workflow's name each time it
   edits an already-saved workflow. Name-uniqueness validation now excludes the
   workflow being edited, so its own name isn't treated as a clash.
--- a/lib/lightning/version_control/version_control.ex
+++ b/lib/lightning/version_control/version_control.ex
@@ -196,14 +196,80 @@ defmodule Lightning.VersionControl do
   def fetch_user_installations(user) do
     with {:ok, access_token} <- fetch_user_access_token(user),
          {:ok, client} <- GithubClient.build_bearer_client(access_token) do
-      case GithubClient.get_installations(client) do
-        {:ok, %{body: body}} ->
-          {:ok, body}
+      case request_user_installations(client) do
+        {:unauthorized, body} ->
+          refresh_and_retry_user_installations(user, body)
 
-        {:error, %{body: body}} ->
-          {:error, body}
+        result ->
+          result
       end
     end
+  end
+
+  defp request_user_installations(client) do
+    case GithubClient.get_installations(client) do
+      {:ok, %{body: body}} ->
+        {:ok, body}
+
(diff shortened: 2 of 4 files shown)


mvanhorn:
Pushed 8c79dbe. The test asserted on a lowercase `"authorization"` header, but `GithubClient` sets `{"Authorization", "Bearer ..."}` capitalised (github_client.ex:145), so those `assert ... in env.headers` checks could never match. Casing corrected in the four places, plus one line reflowed for the formatter.

Issue #1694, opened 2024-02-01
Add "on failure" option to i/o data storage policy


taylordowns2000:
Right now, input/output data can either be retained for all runs, or _never retained_ for runs (this is what we call "zero-persistence").

This issue is to enable a third option, which _waits_ until a run completes and then removes the i/o dataclips for successful runs but keeps them around for failed runs.

<img width="750" alt="image" src="https://github.com/OpenFn/Lightning/assets/8732845/5b547051-2296-46fc-8481-5e28ff6cb6c2">

This might be helpful for an admin that wants to reduce their data footprint (for storage costs or privacy concerns) but keep i/o data for retries only in the event of a failure.

The frontend code has already been built, this issue to to uncomment the frontend options and build the `wipe_dataclip_after_success` feature.

taylordowns2000:
hey is this done or "not planned"?

<img width="447" height="272" alt="Image" src="https://github.com/user-attachments/assets/967e9774-11d3-436d-8312-7f0e0bd754a5" />

PR #5089, opened 2026-08-22
Fix notification & audit trail for dataclip retention policy changes

### LINKED ISSUE #5088: Email not sent when I/O data retention policy changes
Lightning is meant to notify project owners and admins when the dataclip retention policy is changed. Currently, it only notifies if the policy _and_ the period changes.

When turning on/off retention policies, no email is sent.

This is reproduced on `v2.18.0`

### LINKED ISSUE #5090: Audit trail not updated when I/O retention policy changes
Similar to the email bug in #5088 , changes to _only_ the policy and not the retention _periods_ do not trigger an audit trail event.

Image

taylordowns2000:
## Description

This PR fixes #5088 by checking for a change in the `retention_policy` when deciding whether to send the policy change alert. The new functionality in this PR:

1. should properly send the email (not stay silent on critical changes!)
2. should not send misleading information about i/o retention periods when the i/o policy is set to `:erase_all`

The PR also fixes #5090 (the audit trial) in exactly the same fashion.

## Validation steps

1. Change only the retention policy, don't touch the retention periods.
3. See the email.
4. Log in as a super user.
5. Check the audit trail.
6. See the appended audit trail entry.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/lib/lightning/accounts/user_notifier.ex
+++ b/lib/lightning/accounts/user_notifier.ex
@@ -147,6 +147,21 @@ defmodule Lightning.Accounts.UserNotifier do
 
     io_data_saved = updated_project.retention_policy != :erase_all
 
+    # No I/O retention bullet under :erase_all — I/O data is never saved,
+    # so quoting a retention window for it would contradict the line above.
+    details =
+      [
+        "- #{history_retention_period} #{pluralize_with_s(history_retention_period, "day")} history retention",
+        "- input/output (I/O) data #{if io_data_saved, do: "is", else: "is not"} saved for reprocessing"
+      ] ++
+        if io_data_saved do
+          [
+            "- #{io_data_retention_period} #{pluralize_with_s(io_data_retention_period, "day")} I/O data retention"
+          ]
+        else
+          []
+        end
+
     deliver(
       user,
       "The data retention policy for #{updated_project.name} has been modified",
@@ -155,9 +170,7 @@ defmodule Lightning.Accounts.UserNotifier do
 
       The data retention policy for your project, #{updated_project.name}, has been updated. Here are the new details:
 
-      - #{history_retention_period} #{pluralize_with_s(history_retention_period, "day")} history retention
-      - input/output (I/O) data #{if io_data_saved, do: "is", else: "is not"} saved for reprocessing
-      - #{io_data_retention_period} #{pluralize_with_s(io_data_retention_period, "day")} I/O data retention
+      #{Enum.join(details, "\n")}
 
       This policy can be changed by owners and administrators. If you haven't approved this change, please reset the policy by visiting the URL below:
 
--- a/lib/lightning/projects.ex
+++ b/lib/lightning/projects.ex
@@ -641,7 +641,8 @@ defmodule Lightning.Projects do
   end
 
   defp retention_setting_updated?(changeset) do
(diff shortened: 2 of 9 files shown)


lib/lightning/projects.ex
```diff
@@ -641,7 +641,8 @@ defmodule Lightning.Projects do
   end
 
   defp retention_setting_updated?(changeset) do
-    Map.has_key?(changeset.changes, :history_retention_period) or
+    Map.has_key?(changeset.changes, :retention_policy) or
```
taylordowns2000, line 644: we didn't count changing the `policy` as part of the `retention_setting_updated?` function

lib/lightning/projects/project.ex
```diff
@@ -228,7 +228,8 @@ defmodule Lightning.Projects.Project do
 
   defp maybe_validate_dataclip_retention_period(changeset) do
     if get_change(changeset, :history_retention_period) ||
-         get_change(changeset, :dataclip_retention_period) do
+         get_change(changeset, :dataclip_retention_period) ||
+         get_change(changeset, :retention_policy) do
```
elias-ba, line 232: Quick question about this line. I was poking at what happens when someone changes only the policy and leaves the periods alone, and noticed the dataclip retention period gets cleared. That got me wondering what the nightly wipe does afterwards, since that period is the only thing it keys off.

So I set up a project with a month of history and a week of dataclip retention, dropped in a record from ten days ago, and flipped just the policy to `erase_all`. On main the record gets blanked, as you'd expect. On this branch it's still sitting there with its body intact.

Then I went looking to see whether this was new, and mostly it isn't. If you change the policy at the same time as a period, or if a sandbox inherits `erase_all` from its parent, the old data already stays put on main. So I think this line is really just making things consistent, which I'd guess was the point. It happens to be the one route where the wipe still ran.

Which makes me wonder whether we could leave it out of this PR and deal with the underlying thing on its own in a follow up issue, maybe having `erase_all` actually clear what's already stored rather than leaning on that period at all. The email side doesn't seem to need it, since the bullets already key off the policy. 

Could easily be missing something though, let me know what you think.

lib/lightning/accounts/user_notifier.ex
```diff
@@ -147,6 +147,21 @@ defmodule Lightning.Accounts.UserNotifier do
 
     io_data_saved = updated_project.retention_policy != :erase_all
```
elias-ba, line 148: This treats `retain_with_errors` the same as `retain_all`, so now that a policy-only change sends the email, someone switching to `retain_with_errors` would get a mail reading exactly like the `retain_all` one. Nothing to act on today, since that option is still commented out in the settings radio group.

The reason I mention it at all is that the same gap shows up in the `save_dataclips?` check in `projects.ex`, which has no clause for `retain_with_errors` and would raise on it. The TODO next to the commented-out option points at #1694, which is closed, so I'm not sure if that option is still on the cards. Either way it seemed worth knowing before anyone switches it on.

taylordowns2000:
Absolutely open to it. Please go ahead and do what you need to fix the
security notification/audit trail bugs, pushing other fixes into subsequent
PRs as you see fit.


Taylor Downs
Founder & CEO
OpenFn <https://www.openfn.org> // the DPG for AI orchestration & public
service automation

Issue #5037, opened 2026-08-03
Global Assistant: How to allow users to attach full YAML, Logs and I/O Data


hanna-paasivirta:
The tickboxes for allowing users to control whether the job code, logs and I/O data will be sent to the new **global assistant** are now outdated.

The following areas need to be fixed:
- Job code is currently attached regardless of whether the user has ticked to attach code
- The scope is unclear: are we asking the user about one step, or the whole workflow?
- We need to actually scope it correctly. We should send the whole workflow YAML instead of steps; decide how to send logs and I/O data for a specific step (we may need different solutions for a v0 and a v1)
- The data helper text needs adjusting, as the user might already have a step open: "Select a step to include I/O data"
- In the workflow canvas and the history view, job code is attached automatically and there is no tickbox asking the user for permission

Confusing helper text issue:
<img width="1395" height="849" alt="Image" src="https://github.com/user-attachments/assets/995eb8f0-91a9-4323-b1ab-1ba1df987c2d" />

Missing tickboxes in history view:
<img width="925" height="654" alt="Image" src="https://github.com/user-attachments/assets/9efb6c63-c63c-4a8c-bf60-98765755cff4" />

PR #5096, opened 2026-08-26
Global assistant: Forward attachments to Apollo

### LINKED ISSUE #5093: Global assistant: Forward logs and I/O data to Apollo
Lightning does not currently forward any logs or I/O data to Apollo when using the Global Assistant, even if the user ticks to attach them to the conversation.

This backend bug is adjacent to the issue around fixing the UI side for users to control what attachments are passed on: https://github.com/OpenFn/lightning/issues/5037

hanna-paasivirta:
## Description

Global chat ignored the "Send logs" and "Send scrubbed I/O" checkboxes. The UI said the data was attached; nothing was sent to Apollo. This wires it up.

Global chat sends an `attachments` list to Apollo: run logs, scrubbed step input and scrubbed step output. Log lines carry `job_id`, `step_id` and `level` so Apollo can tell which job a line came from, and logs are fetched project-scoped the way step I/O already was. The selected run id survives session creation, so the attachment resolves on the first message as well as later ones, and an oversized payload comes back as a message naming the checkbox to untick rather than raw Apollo text.

Closes #5093

## Validation steps

1. Open the assistant in a workflow with past runs. Tick Global assistant, tick attach logs and I/O data, and ask whether it can see your logs. Check it describes them correctly.
2. Repeat from different steps, with new conversations and with conversations that continue across views.
3. Edit the workflow, come back, and check it still sees the logs.

Lightning sets no size limit of its own. Apollo refuses an oversized payload and we turn that into a message naming the checkbox to untick, which is deliberate: shortening the logs would answer from evidence the user thinks we read in full.

No role gating. The assistant has never told a viewer from an editor, so this follows job chat rather than changing that. Run-level dataclips are not sent, for the same parity reason.

Job chat and workflow chat are untouched. Only `global_chat_stream` changed in the Apollo client, and the one shared change is a new branch in the SSE error handler, so nothing breaks against an Apollo that predates it.

Five fixes came out of review and are on this branch in one commit, each with a test: the log read is now bounded, a non-uuid `step_id` no longer crashes the job, the too-large message no longer raises on an unexpected payload shape, the run id comes from the checkbox that gated it, and an unresolved attachment is logged instead of silently dropped.

Attaching logs from the canvas is https://github.com/OpenFn/lightning/issues/5037. Scrubbing rewrites values but not map keys, which is worth a product decision rather than a change here.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


elias-ba (approved):
Nice work @hanna-paasivirta 👏🏽

PR #5098, opened 2026-08-26
Enable optional support for dataclips to be sent as binary

josephjclark:
## Overview

When the worker sends dataclips, allow the dataclip to be encoded as a string (for back compat), or inlined as a binary object (to prevent lost runs)

## Description

This is part of the fix for https://github.com/OpenFn/kit/issues/1504

Today, when the worker sends a dataclip via  `output_dataclip`, the data is encoded into a JSON string.

I don't know why we do this. I think it's an error from the initial implementation that wasn't noticed or wasn't thought through 🤔

But there are two problems with it:

1. It's double encoded at both ends. The worker takes a JSON object and stringifies it into the payload envelope,,, which then gets JSONified in the websocket.... and lightning then parses the envelope, and then goes an parses the dataclip specially. Waste of processing.
2. It bloats the payload size - and this is where the lost runs come from. The worker refuses to send dataclips >10mb. But after the worker measures them, when it stringifies the dataclip is bloated (because of escaped quotes) and can grow by a significant %. So mauybe now a 9mb dataclip is 12mb on the wire. And cowboy rejects frames that are greater than 11mb (worse than that, it also disconnects the whole socket!)

So! There's a corresponding fix in the worker which removes this stringiy and just sends the dataclip as a normal object, which I'm sure is what is always should have done.

Now, that's a breaking change worker-side, and we can't have that. So there are a few mitigations to enable a safe release:
1. The worker uses an option (env var) to control how to serialize dataclips
2. This PR here is compatible with both legacy and new strategies, so Lightning is happy either way.

The plan is we patch lightning, update the worker, and flick the switch in the worker env. And those lost runs will go away.

The only possible advantage I see in the production architecture is that we have the option to not parse the dataclip in lightning at all. We could dump it in the DB as a JSON string. But then you need to parse it when you retrieve it, and anywhere we'er not (and I don't think we plan to) do anything like that right now.

## Validation steps

In both worker modes, ensure that output dataclips from a run are properly represented.

You know it's wrong if the dataclip looks like this:
```
{
  "value": "{\"x\":1234}"
}
```
## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
The worker changes (1.29.2) are now live on staging, with the fix disabled, and seem to be working just fine with prod lightning.

lib/lightning/runs/handlers.ex
```diff
@@ -550,12 +555,20 @@ defmodule Lightning.Runs.Handlers do
...
+        body: output_dataclip |> maybe_decode_dataclip() |> ensure_map(),
         type: :step_result
       })
       |> Repo.insert()
     end
 
+    # For back compat: older workers JSON-encode output_dataclip into a string before sending
+    # it (Lightning then decodes it); newer workers send the value already
+    # decoded.
+    defp maybe_decode_dataclip(value) when is_binary(value),
```
elias-ba: Am I right that this can't tell a JSON-encoded string from a state that just is a string? I tried `fn(state => state.data.id)` against the runtime and the final state comes back as the plain string `"abc-123"`, and with `noStringifyState` on the worker sends it unquoted. Pushing that here gave me a `Jason.DecodeError` and the channel died, so the step never got recorded.

Would something explicit be safer, like the worker telling us the encoding? Or having it wrap non-object state as `{value: x}` so a binary always means an old worker? I couldn't measure how often jobs actually end on a non-object, so I may be overweighting this.

lib/lightning/runs/handlers.ex
```diff
@@ -550,12 +555,20 @@ defmodule Lightning.Runs.Handlers do
...
+      do: Jason.decode!(value)
+
+    defp maybe_decode_dataclip(value), do: value
```
elias-ba, line 580: Smaller version of the same question. A job returning the string `"42"` decodes to the integer 42 and gets stored as a number. Is that fine, or does it fall out of whatever we do above?

lib/lightning/runs/handlers.ex
```diff
@@ -550,12 +555,20 @@ defmodule Lightning.Runs.Handlers do
       Dataclip.new(%{
         id: dataclip_id,
         project_id: project_id,
-        body: output_dataclip |> Jason.decode!() |> ensure_map(),
+        body: output_dataclip |> maybe_decode_dataclip() |> ensure_map(),
```
elias-ba, line 558: Heads up rather than a request. #4910 rewrites this same line. I merged the two to check and it conflicts here, so whoever lands second has to redo the hunk.

test/lightning/runs_test.exs
```diff
@@ -401,6 +401,35 @@ defmodule Lightning.RunsTest do
...
+
+      step =
+        step
+        |> Repo.preload(output_dataclip: Invocation.Query.dataclip_with_body())
+
+      assert step.exit_reason == "success"
+      assert Jason.decode!(step.output_dataclip.body) == %{"foo" => "bar"}
+    end
+
+    test "accepts a JSON-encoded string output_dataclip, for backward compatibility with older workers" do
```
elias-ba, line 429: `ensure_map/1` covers scalars and lists, but I don't think anything exercises those through the decoded path now. Worth a case for each?

elias-ba (changes requested):
A few questions inline, mainly about the binary case. The lint failure looks like the cowlib hex.audit advisory and main fails the same way, so I don't think that one's yours.

lib/lightning/runs/handlers.ex
josephjclark: Ok - this is going to be patched out so we gotta do the work here first.

btw that other PR has been open a long time and we can't keep asking the poor guy to keep rebasing. We should merge to a branch in lightning and take it over properly. That's also quite an important fix.

lib/lightning/runs/handlers.ex
josephjclark: Nono, this is nothing to do that that. You can't affect this in job code.

This is all about how the worker embeds dataclips in the `step:complete` payload.

In prod right now, the dataclip will ALWAYS be a string type, which needs decoding into JSON here to be processed downstream.

The new worker will use a different encoding if an option is enabled.

I did think about sending some kind of option in the claim request or as part of the run. But this is a change we only need to make once: we can deprecated `maybe_decode_dataclpi` later and skip the explicit decoding.

The maybe is only here to make upgrading deployments safer. With zero config any worker can speak to any lightning. This is good :)

test/lightning/runs_test.exs
josephjclark: Ok - extra test incoming

lib/lightning/runs/handlers.ex
elias-ba: You're right about today. The old worker stringifies unconditionally, so job code can only change what is inside the string. I framed my last comment around the runtime, which made it sound like a claim about job semantics. The claim is about the payload.

With `noStringifyState` on, `evt.output_dataclip = outputState`, so the type travels too, and the runtime has never promised an object. From a worker against a local Lightning on this branch:

```
event: "step:complete", payload: %{
  "output_dataclip" => "abc-123",
  "worker_version" => "1.29.2", ...}

[error] GenServer #PID<0.1355.0> terminating
** (Jason.DecodeError) unexpected byte at position 0: 0x61 ("a")
    lib/lightning/runs/handlers.ex:558: CompleteStep.maybe_save_dataclip/2
    lib/lightning_web/channels/run_channel.ex:211: RunChannel.handle_in/3
```

The job is two operations ending in `fn(state => state.data.id)`. I triggered it from the UI and the run hangs forever, but it looks fine while it does. The worker succeeds and its log is stored, so the run page reads "Run complete with status: success" while the status sits on Running and the duration climbs. `step:complete` died before inserting the output dataclip, so the `run:complete` behind it fails `resolve_final_dataclip/2` with "does not exist" and the run never finishes.

`Runs.complete_run/2` already handles this. `wrap_final_state/1` wraps a non map `final_state` as `%{"value" => state}`, and you can see it working on `"final_state" => "abc-123"` above. Only `step:complete` has no equivalent.

Screenshot 2026-08-27 at 16 11 39

josephjclark: Ok, got it. I see the issue. Working on a fix.

elias-ba: Thanks man

josephjclark: Ok, hold up. I've just pushed a fix but actually, this exact same issue is on main. But worse. When we move to "don't parse the dataclips" with hte new option enabled, this goes away entirely.

I think this is a red herring. I'll leave the "fix" I've just pushed because its fine. But it will need reverting when we remove `maybe_decode_dataclip`

i'm gonna raise a draft PR now to handle the removal of this.

josephjclark: Ok - all sorted.

elias-ba (approved):

midigofrank (approved):

PR #5100, opened 2026-08-27
Global assistant: visualise workflow diffs as they stream

### LINKED ISSUE #5036: Global assistant: Visualise diffs across different parts of a workflow
The global assistant can now change workflow code for several steps at a time. However, the user has no way of checking what the assistant changed before hitting "accept changes". It is essential that we communicate to the user in some way what parts of the workflow and what parts of the job code were edited. Both **job code changes** (edits to other steps) and **structural changes** (add a step, remove a step, re-order, conditional flow etc) need to be surfaced.

This will require more involved design work eventually, but for a v0, we just need to find the simplest possible implementation that will make it safer for users to use the assistant.

**a) Job code changes**
Displaying snippets of code diffs in red and green (along with the step names where those changes were made) within the assistant conversation may be the easiest approach. 

This implementation should live in Lightning instead of Apollo, as it pertains to visualisation and we already handle diffing in Lightning.

**b) Workflow structure changes**
Showing the full YAML (e.g. with job code redacted) with similar diffs could solve the problem for v0. 

If that's too difficult for the user to follow (e.g. for a multi-step workflow), is there any other text-based, diffable, simplified workflow structure representation that could work?

**c) Optional: Additionally, allow the user to navigate between diffs**
Not sure if this is required for v0, but the user should be able to navigate between job code steps to view the full code changes in context for each step. Currently, the user can only see the red & green diff visualisation for a single step, if they happened to have it open when they asked for the changes. This means that if they navigate to another step, the diff visualisation will not be generated there even if that step was edited by the assistant.

---

+If diffing snippets don't work well, another minimal option would be to do c) (i.e. make sure diffs are visible for all steps) and surface links to those steps in the AI chat instead. Structural changes would still need to be noted somehow and linked to as well. I started doing this some time ago, but it seemed like the more time-consuming option to me.

+Also, a related inconsistency to keep in mind and possibly fix along the way, is how we apply AI-generated changes. We auto-apply all changes (code + structure) if the user is viewing the workflow canvas. However, if the user is viewing job code for any step, no changes will be auto-applied, and structural changes are completely invisible until they hit "apply".

elias-ba:
## Description

The diff blocks landed all at once when a reply finished, which made them feel like an afterthought rather than a record of what happened. They now render as the assistant works. In a real reply the first block appears about 27 seconds before the message settles.

Apollo already sends the workflow every time it mutates it, immediately followed by the status describing that change, so each pair is one action. The client kept only the newest of those and refused to draw anything until the message finalised. Snapshots are now kept in wire order, pinned to the status they belong under, and ride the same character drain as the statuses so nothing overtakes the prose it followed. They carry over to the finalised message, so nothing moves at the transition.

Attribution no longer reads English. Apollo reports the steps it acted on as data, Lightning persists them on the segment, and the regex that matched write verbs and substring-matched step names is gone. A reload replays the same records, so history renders the way the live stream did. No migration, `response_segments` is already jsonb.

The blocks are rebuilt on Primer's light palette. Syntax highlighting through Prism, separate old and new line gutters, ligatures off so `=>` is not drawn as an arrow, long lines wrapped instead of hidden behind a scrollbar, and a link to open a changed step in the editor.

Two bugs turned up while testing against a real Apollo. It re-serialises the YAML on every mutation, turning a `body: |` block scalar into an inline one, so an unchanged step differed by a trailing newline and hung an empty diff block under a status on almost every reply. And the panel could not be scrolled during a reply: the auto-scroll checked whether you had scrolled away when it queued a frame rather than when that frame ran, and text arrives every 15ms so there was almost always one queued.

Closes #5036

Needs OpenFn/apollo#653 for the `steps` field. Without it the blocks still render, they just fall to the end of the message instead of weaving in, which is also what happens for replies persisted before this change.

## Validation steps

1. Open a workflow with no Kafka trigger, open the assistant and tick "Global assistant (experimental)".
2. Ask it to write code for a couple of steps. Blocks should appear under each status while the reply is still streaming, not in one lump at the end.
3. Scroll up while it is still streaming. It should stay where you put it.
4. Reload the page and reopen the conversation. It should render exactly as it did live.
5. `cd assets && npm test` and `mix test test/lightning/ai_assistant`.

1. The AI assistant is unusable on any workflow with a Kafka trigger, on `main` and unrelated to this branch. `Trigger`'s `@derive` omits `kafka_configuration`, the client's `kafkaTriggerSchema` requires it, so the whole session context fails Zod validation and the panel sits on "Loading conversations...". Worth its own issue.
2. `planner.py` on the Apollo side builds a `StreamManager` and then overwrites it a few lines later, discarding the one the router passed in. Looks like a merge leftover. Also not this PR.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


hanna-paasivirta:
@elias-ba How would you feel if I changed the mixed language at the top of diffs:

image


...to natural language like this:

Added step   Gmail notification      +1   Open
Updated      Transform data       +12 -3   Open
Removed step Googlesheets              -8

elias-ba:
Hey @hanna-paasivirta I hesitated a little bit to do that especially because of the "Removed" one which reads like the steps is removed. But honestly I think natural language would feel better. Feel free to change but just bear in my mind about that

hanna-paasivirta:
@elias-ba Is there a specific reason you removed the "apply" button for code changes? I know it had drifted to being inconsistent (structural changes auto-applied vs code changes presented a diff to accept), but there are other options as well for fixing this. I think the auto-apply implemented here is quite radical, and I'm having a look at re-implementing an apply option (a consistent version) and having an undo button. What's your thinking on this?

hanna-paasivirta:
Since the "Apply" and "Copy" buttons were removed, I have added:
- Undo/redo on the last reply, with the confirmation only when the workflow has actually changed since the apply. Having an apply button is more complex now, as I can't think of an obvious way to visualise a diff preview for structural changes. If we want to bring back an "apply" button, we need to do some design work. 
- A copy button on each diff block.

fa Assistant Global (experimentet)

Done! I&#39;ve removed the write to gSheet after alert step

Overwrite changes to the workflow

@elias-ba Could you please look over my changes? Everything else looks ok to me.

hanna-paasivirta (approved):

Issue #2966, opened 2025-02-24
Code indentation breaks OpenFn -> GitHub Sync


christad92:
A user's job code while not pretty runs without failure but when they initiate a GitHub sync from OpenFn to Lightning, it fails with [this error](https://github.com/OpenFn/solar-sister/actions/runs/13459954191/job/37612628367). 

This reads more like an issue between how Yaml reads files and the indentation of the fn() block but again I know we escape the code with the | operator. 
 
See the error below: 

<img width="826" alt="Image" src="https://github.com/user-attachments/assets/97187b4c-78bd-427d-8b29-4570ac5e5fa2" />

See the job code causing the error here: 

![Image](https://github.com/user-attachments/assets/dd5cd858-8f12-4bfc-a5af-83c39bdee560)

midigofrank:
I'm unable to reproduce this.

midigofrank:
Given that this bug is caused by leading spaces in the job code, a quick fix for this can be to trim the `job.body`, i.e. `update_change(job_changeset, :body, &String.trim/1)`. 

Alternatively we can just use a library to the yaml encoding. Unfortunately we don't have a lot of options in elixirland. I've found this https://github.com/ufirstgroup/ymlr , which supports keyword lists for ordering, but it prefixes the yaml doc with `----`.

Given the work already done in https://github.com/OpenFn/lightning/pull/3084, I'm strongly leaning towards having the yaml encoding in js land.  Both the cli and the ui already use https://eemeli.org/yaml for encoding/decoding  yaml code therefore making this change should be trivial and we never have to worry about special characters ever again. On the downside, we would lose native encoding in the server, of which I don't see any problem. The project export/backup doesn't even include the .yaml

midigofrank:
@theroinaochieng @stuartc please note that I have moved this to the backlog

lmac-1:
Fixed in #5106

PR #5106, opened 2026-08-30
Support special characters in workflow, step and credential names

### LINKED ISSUE #2808: Sync to github fails when special characters exist in credential names
When syncing a project that uses a credential called "Taylor Mailchimp June '24" the sync fails.

Image

After exporting the project yaml, I can see that the yaml looks like this:

```yaml
credentials:
  "taylor@openfn.org-MailChimp-June'24":
    name: 'MailChimp June'24'
...
        add-to-drip-campaign:
        name: add to drip campaign
        adaptor: '@openfn/language-mailchimp@1.0.0'
        credential: 'taylor@openfn.org-MailChimp-June'24'
```

And VS code knows that this is invalid.

Image

We should fix this so that credential names are made safe for yaml on export/sync, and we should extend the error message in the sync action to give the user more information about _WHY_ their sync fails.

### LINKED ISSUE #4577: [P3] Support special characters in workflow step names

elias-ba:
## Description

Step names were ASCII-only, so a team working in French, Spanish, Arabic or Japanese could not name a step the way they say it. Names may now hold anything except a control character.

Stacked on #5109, which moves us to Erlang 28. That matters here: Erlang 27 normalises names incorrectly, and this is the PR that starts normalising them.

That broke the YAML export, which concatenated strings against a hand-rolled regex. `ExportUtils.Scalar` replaces it under one constraint: anything the old encoder wrote correctly comes out byte-identical, because customers keep their specs in git and a quoting change is a diff in every synced repo. Also fixes #2808 and #2966.

An unrelated XSS fix rides along because it is the same code path. `wrapper_tooltip` hard-coded `data-allow-html="true"`, so a workflow named with an `` fired for anyone viewing the project list.

Closes #4577

## Validation steps

1. Name a step `Vérifier l'état` or `患者確認`. It saves and survives a reload.
2. Export the project YAML, re-upload it, then push through GitHub sync. Names survive, and job bodies are byte-identical for any name that was already legal.
3. Name two workflows `My Flow` and `My-Flow` in one project, then try sync, the reconnect form and "Export project". Each fails naming both.
4. Name a workflow `` and open the project list. You should see that text sitting in the page as a name, and no alert box. Before this PR the browser ran it, for anyone who opened the page rather than just whoever typed it.

1. I queried production for the one existing-data condition this changes, two names in a project that become the same spec key, and got zero rows. An instance that does have one is told at sync, at reconnect or at export, with both names and renaming as the fix.
2. Step names need no backfill. They were ASCII-only, and ASCII is spelled the same way under every normaliser, so nothing stored can be in the wrong form. Workflow names are different: they never had a format rule, so an existing one may hold anything, and this is the first thing that normalises them. A workflow name not already in normal form changes shape on its next save. That is also why this is stacked on #5109 rather than shipping first, since normalising them on Erlang 27 would write the wrong form.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


assets/js/collaborative-editor/components/inspector/CodeViewPanel.tsx
```diff
@@ -46,11 +46,14 @@ export function CodeViewPanel() {
   // Generate sanitized filename from workflow name
   const fileName = useMemo(() => {
     if (!workflow) return 'workflow.yaml';
-    // Remove special characters, replace spaces with hyphens
+    // Names can be any script now, and stripping everything outside ASCII
```
josephjclark: `now` should not be in the comment here. AI should not be commenting the diff, only the current state of the code.

assets/js/collaborative-editor/components/inspector/CodeViewPanel.tsx
```diff
@@ -46,11 +46,14 @@ export function CodeViewPanel() {
...
+    // leaves a CJK or Arabic name as the empty string and the download as
+    // `.yaml`. Fall back to a usable name when nothing survives.
     const sanitized = workflow.name
       .replace(/[^a-zA-Z0-9-_\s]/g, '')
```
josephjclark, line 53: I can't help but  wonder if we should be using some kind of library for this, consistently across all platforms.

For the workflow download it's not super important to santize the name correctly. But I do think we should have a single "sanitize" function shared or duplicated across repos and we just call that

josephjclark:
@elias-ba my biggest initial concern is this

Users can now input basically any step name, with emoji and non ASCII characters.

The CLI and runtime are mostly agnostic to this (they don't care), EXCEPT that the CLI will convert a step name into a step ID. Basically by making it URL safe with a slugify function.

That gives us local yamls like this:
```
id: special-chars
name: special chars! 🎉
steps:
  - id: step-1
    name: step number 1!%& 🤭
```
Notice special characters are dropped . So if you have `step 👍️` and `step 👎️`  in your workflow, they'll both be saved locally as `step`, which will cause problems.

I also think we should consider: are we making changes here that would prevent us build semantic URLs like `projects/my-project/my-workflow/step-1`? We should really support this one day, which means making a URL safe id. Now you have the same problem as the CLI - step names with special characters might compress badly and conflict.

What I think we should do in the short term is:
- Make sure there's a single name-to-url-safe-id conversion algorithm, shared with kit or duplicated (and covered in integration tests). 
- In lightning, ensure that every step name encodes to a UNIQUE value. So not a unique display name, but a unique encoded id. That's kind of a complicated error to explain to users, but it would just catch subtle errors with similar step names separated only by special characters

josephjclark:
Btw the CLI side is looking absolutely fine. Workflow yamls are loaded safely as utf8, display names are respected and used, and step ids are safelt generated. So no work there that I can see. Havne't checked the worker but it should be the same (and easy to test from this branch in QA!)

assets/js/utils/nameValidation.ts
stuartc, line 1: I'm trying to find a reason why we don't use the browser builtins for this.

The `Intl.Segmenter` interface has been standard for years now in browsers.

```ts
const segmenter = new Intl.Segmenter(undefined, { granularity: 'grapheme' });

export const graphemeLength = (value: string) =>
  [...segmenter.segment(value)].length;

export const isNameTooLong = (value: string) =>
  graphemeLength(value) > NAME_MAX_LENGTH;
```

Like thats (afaict) the same purpose as this module.

lib/lightning_web/live/components/common.ex
stuartc, line 1: I think we can do this a little more elegantly, instead of having to conditionally sanitise and not escape (which gets risky if someone forgets). This isn't a new problem, just the need to sanitise names makes it more obvious that the way were doing it wasn't ideal.

We can lean on heex and browser templates and I think we wouldn't need to escape at all.

```heex
attr :id, :string, required: true
attr :tooltip, :string, default: nil
attr :subtitle, :string, default: nil
slot :inner_block, required: true

def wrapper_tooltip(%{tooltip: tooltip} = assigns) when not is_nil(tooltip) do
  ~H"""
  <span
    id={"#{@id}-tooltip"}
    phx-hook="Tooltip"
    aria-label={@tooltip}
    data-hide-on-click="false"
  >
    <template data-tooltip-content>
      {@tooltip}
      <span :if={@subtitle} class="block text-xs text-gray-500">{@subtitle}</span>
    </template>
    {render_slot(@inner_block)}
  </span>
  """
end
```

and in the Tooltip hook:

```ts
mounted() {
  const template = this.el.querySelector<HTMLTemplateElement>('template[data-tooltip-content]');
  const content = template
    ? template.content.cloneNode(true)
    : this.el.ariaLabel;   // legacy callers that only set aria-label

  this._tippyInstance = tippy(this.el, {
    placement: this.el.dataset.placement ?? 'top',
    animation: false,
    interactive: this.el.dataset.interactive === 'true',
    hideOnClick: this.el.dataset.hideOnClick !== 'false',
  });
  this._tippyInstance.setContent(content);
  this._originalContent = content;
  ...
}
```

I see that Tippy's `setContent` also accepts a DOM node as well as a string (not sure how we missed that back when we implemented it), but with this approach we can have phoenix sanitise everything, safely and without conditionals.

lib/lightning/export_utils/scalar.ex
stuartc, line 1: Dude, I won't lie. I barely understand this kinda stuff 😅 . I've read it twice and can't hold it in my head. 

So we're going for a single printable line and so for that single quotes with doubled apostrophes I think are the right way to go (i.e. `'O''Brien'` over `"O'Brien"`) that way you don't need any backslash processing at all. So stoked we do that. But I think the double quoted fallback can't actually be reached, because the name validator stops that before we get here.

So theres a bit of tension between us writing a generic scalar encoder for yaml, and us making our exporter work properly. For the former we need to be really careful because this is what libaries with big test suites are for (i haven't checked if something like this exists inside any existing libraries). So we have a choice, either make doubly sure we are making this module bulletproof or relax/delete some of the unreachable (because of the code we have before this is called) and add something like:

```ex
def encode_value(value) do
  if String.match?(value, @control_chars) do
    raise ArgumentError, "value reached the YAML encoder unvalidated: #{inspect(value)}"
  end
  ...
end
```

That way if we add a new field to the export and we forget to validate it, we get a bang vs silently broken yaml in someones repo.

elias-ba: The double quoted branch is not unreachable, so I would rather not delete it. Ordinary names go through it, a Japanese name or a credential called MailChimp June'24. encode_key sends every key that is not bare-legal there, and that is every name in a non-Latin script and every name with a colon, an apostrophe or an emoji.

On the raise, you are right that the validator stops control characters before this point. I am not sure it earns its place though. A field we forgot to validate still comes out as correct escaped YAML today, so the raise would turn a working export into a crash. Happy to be talked round if you think the loud failure is worth more.

assets/js/utils/nameValidation.ts
elias-ba: Done, thanks. Added es2022.intl to the lib list so we get the real types, and dropped the hand written ones.

lib/lightning_web/live/components/common.ex
elias-ba: Done, thanks. The content sits in a template now and the hook clones it into tippy, so nothing has to be escaped by hand.

josephjclark:
@elias-ba this is still working great on CLI:  I can pull (v1 and v2) and execute steps with emojis and special characters

I've still got open concerns about validation in the app and ensuring consistent name encoding. Do we have any time to look into this?

assets/js/collaborative-editor/components/inspector/CodeViewPanel.tsx
elias-ba: Thanks for the catch. Fixed

assets/js/collaborative-editor/components/inspector/CodeViewPanel.tsx
elias-ba: This is only a download filename, so it doesn't need to match anything. Where matching did matter was the step id, and OpenFn/kit#1535 removes that by recording the id instead of deriving it.

elias-ba:
@josephjclark you were right about the step ids colliding. If you have two steps called `step 👍` and `step 👎`, the CLI writes one file and the second step's code is lost. Workflow names do the same thing, except there the pull crashes instead. I've opened OpenFn/kit#1535 to fix it by having the pull keep the id a project already has, rather than working it out from the name every time, which is why I haven't added the uniqueness check in Lightning you suggested.

josephjclark (approved):
Looks good to me.

I've spun out my validation edge case into https://github.com/OpenFn/lightning/issues/5156 - but I don't think it's worth the time to fix now.

PR #5108, opened 2026-08-31
Add a workflow health page

midigofrank:
## Description

A per-workflow page at `/projects/:project_id/w/:workflow_id/health` summarising one workflow over a selectable window (last 24 hours, 7 days or 30 days, defaulting to 30): an outcomes donut over **work orders**, a breakdown of the failing ones by state, a stacked bar chart of **run** volume over time, and a triage table grouping failures by error signature, heaviest first.

The donuts count work orders rather than runs, deliberately — a run's state is immutable, so a retried failure would sit on this page forever. Retry the work order and the numbers actually move. The volume chart counts runs, because it is about traffic and load rather than outstanding work, so the two disagree on purpose and the bar card carries no total.

**Data path.** A LiveView (`WorkflowLive.Health`) renders the shell and mounts a React component via `phx-hook="ReactComponent"`; the charts fetch their own data from a cookie-authenticated JSON API (`API.WorkflowHealthController`), one action per chart — `/health/outcomes`, `/health/failures`, `/health/runs` — so a cheap chart isn't held up by an expensive one. The two donuts share the single `outcomes` reply, so their totals can't disagree.

**Refresh.** The browser polls every 30s (`useHealthQuery.ts`), and also re-reads when the tab regains focus. There is no timer while a read is in flight, so a read slower than the interval isn't aborted by the tick behind it, and the gap is measured from the answer rather than from the request. A poll that finds nothing changed costs one cheap marker query, not a recompute. The frontend keeps the last numbers on screen while new ones land, and a failed poll keeps the last answer rather than blanking the page.

**Query design** (`Lightning.Workflows.Stats`):

- Windows anchor on `last_activity`, not `inserted_at`: an old work order retried today counts as the work it currently is. The volume chart is the exception — its buckets count on `inserted_at`, so a run stays in the bar the traffic arrived in.
- Failures are attributed to *every* failing step of the work order's latest run (lateral join), so a run that broke in two branches shows as two things to fix. That means triage rows can sum past the donut's failure total — intended.
- Job names and adaptors come from the run's own snapshot, so a rename or an adaptor bump doesn't relabel history and a deleted job is still nameable.
- Run volume is bucketed server-side — 2-hourly over a day, AM/PM over a week, daily over a month — rather than shipping every run in the window to draw thirty bars. Buckets are zero-filled and flat (one key per state), which is the row shape Recharts takes as `data`. The window anchors on the bucket `now` falls in, so the axis labels aren't lies.
- Kept separate from `DashboardStats`: that one batches across many workflows to dodge an N+1 and collapses every failure state into one `:failed` bucket — the granularity both donuts need apart.
- Results are cached in a Cachex instance (5 minute TTL), `window` included so it does not roll per request. The cache key carries a change marker (the workflow's latest settled `last_activity`) rather than the cache being invalidated on settle: a poll that finds nothing moved is a hit on every pod, because the marker is read from Postgres and not from one node's ETS. See the known gap below for the trade.

**Supporting changes:**

- New composite index on `work_orders (workflow_id, last_activity)`, created concurrently. The existing single-column indexes made the planner pick one and heap-filter the rest.
- `Run.state_reasons/0` — one table mapping worker reasons to run states, read inbound by `Handlers.CompleteRun` and back out by `Stats` so the two directions can't drift. Replaces the inline `case` in the handler.
- `WorkOrder.final_states/0` and `failure_states/0`.
- `SearchParams` gains `error_signature_exit_reason`, `error_signature_error_type` and `error_signature_job_id`, so a triage row's "View" button opens the history page filtered to exactly the work orders behind it, where the existing "retry all" can retry the group. `Invocation.filter_by_error_signature/2` applies them. Note this also reworks `from_uri/1`: a URL carrying a search term but none of the four search-field flags used to match nothing and now falls back to the schema defaults. Traced and believed safe — the form ships a hidden `false` per checkbox, and app-built links always carry all four — but it changes history filtering for everyone, not just this page, so it deserves a second pair of eyes.
- `get_workflow_for_project/3` now also excludes workflows marked for deletion (it routes through `Query.workflows_for/1`); `workflow_exists_in_project?/2` reuses it.
- `export_load.exs` seeds every state the page draws with a spread of real error signatures, so it doubles as fixture data.
- `recharts` added for the donuts and the bar chart.
- Stacked on this branch and already reviewed and merged into it: #5114 (range picker and `?days=` validation), #5151 (grouped triaged work orders in the history page), #5157 (browser polling in place of the pushed refresh), #5162 (bucketed run counts behind the window).

**Authorization.** The controller is its own trust boundary — the LiveView's `on_mount` guards don't protect it, since a client can request any path. It re-checks `:access_project` via the same policy as the rest of the app, and returns 404 (not 403) on every failure so it can't confirm a workflow exists to someone who can't see the project. Covered by tests: anonymous, non-member, cross-project, support user, project scheduled for deletion, MFA not enrolled, workflow marked for deletion, malformed id — on every endpoint.

## Known gaps

Refresh refinements raised in review of #5157 are tracked under CON-183 — not release blocking, all low priority:

- CON-184 — a page that has stopped updating (failed reads, expired session) looks like one with nothing to report.
- CON-185 — "Last Updated" is the server's compute time, not the browser's read time, so it can be stale and can jump backwards across pods.
- CON-186 — returning to the tab skips the refresh if a read is already in flight.
- CON-187 — the cache can do the same work twice across a settle boundary, and has no size limit.

From review of #5162, on the volume chart:

- CON-188 — buckets are in UTC, not the reader's timezone.
- CON-189 — clicking a bar segment should open the filtered history, the way a triage row's "View" does.

Open question, not yet filed: `Stats.runs/2` may never serve from cache on a busy workflow. The key carries the latest settled `last_activity`, so a workflow settling faster than the 30s poll recomputes every time — and that's the workflow with the most rows to aggregate. Related to CON-187, but a different edge of the same design. Unmeasured; it needs production-shaped data.

The migration builds its index concurrently. If that half-fails during deploy, Postgres leaves an invalid index behind and `create_if_not_exists` skips it on a retry, so the queries quietly fall back to the old single-column index. Worth confirming `indisvalid` after deploy.

Closes CON-106

## Validation steps

1. Seed a project with health-page history (much smaller than the export test needs, since this page reads no logs or dataclips):

   ```bash
   WORK_ORDERS=400 LOGS_PER_RUN=2 DATACLIP_POOL=50 DATACLIP_BYTES=500 \
     SPREAD_DAYS=45 PROJECT_NAME=health-test \
     mix run tooling/seeds/export_load.exs
   ```

   `SPREAD_DAYS=45` puts some history outside the 30-day window, so you can confirm the window actually excludes it.

2. Visit `/projects/<project_id>/w/<workflow_id>/health`. Check the two donut totals agree (Outcomes' red wedge = the breakdown's total), and that the triage table is sorted heaviest-first with a tip on every row. Switch the range picker between 24 hours, 7 days and 30 days and confirm the bar chart's buckets change shape (2-hourly, AM/PM, daily) and that empty buckets draw as gaps rather than vanishing.

3. Click "View" on a triage row: the history page should open filtered to just the work orders behind that signature, and "retry all" should offer to retry the group.

4. Run the workflow (or retry a failed work order) and watch the numbers move within 30 seconds without a page reload. Switch to another tab and back: it should re-read immediately on focus.

5. Authorization: as a user who isn't a member of the project, request `/api/projects/<project_id>/workflows/<workflow_id>/health/outcomes` — 404, not 403.

6. Tests:

   ```bash
   mix test test/lightning/workflows/stats_test.exs \
     test/lightning_web/controllers/api/workflow_health_controller_test.exs \
     test/lightning_web/live/workflow_live/health_test.exs \
     test/lightning/workorders/search_params_test.exs
   cd assets && npm test -- health
   ```

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist


assets/js/workflow-health/WorkflowHealth.tsx
```diff
@@ -0,0 +1,74 @@
...
+  if (error) {
+    return <p className="text-sm text-red-700">{error}</p>;
+  }
+
+  if (!outcomes) {
+    return <p className="text-sm text-gray-500">Loading…</p>;
+  }
+
+  const days = windowDays(outcomes.window);
+  const total = Object.values(outcomes.counts).reduce((a, b) => a + b, 0);
```
lmac-1: The header total includes pending but the donut only draws success and failed, so the two numbers on screen don't add up.

The design has them matching (the subtitle count equals the outcomes legend), so the header should count finished work orders only:

```suggestion
  const total = outcomes.counts.success + outcomes.counts.failed;
```

lmac-1: I think you'll need to also update WorkflowHealth.test.tsx after this change. It will expect the old total (including pending runs).

lib/lightning_web/channels/workflow_health_channel.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  @impl true
+  def join(
+        "workflow_health:" <> workflow_id,
+        %{"project_id" => project_id},
+        socket
+      ) do
+    with %_{} = user <- socket.assigns[:current_user],
+         %_{} = project <- Projects.get_project(project_id),
+         %_{} <- Projects.get_project_user(project, user),
```
lmac-1: Support users have no `project_users` row, so this refuses them where other channels admit them. Worth going through `Permissions.can(ProjectUsers, :access_project, ...)` instead, it also covers projects with `scheduled_deletion` set.

lib/lightning_web/channels/workflow_health_channel.ex
```diff
@@ -0,0 +1,42 @@
...
+  alias Lightning.Workflows
+
+  @impl true
+  def join(
+        "workflow_health:" <> workflow_id,
+        %{"project_id" => project_id},
+        socket
+      ) do
+    with %_{} = user <- socket.assigns[:current_user],
+         %_{} = project <- Projects.get_project(project_id),
```
lmac-1: A malformed `project_id` crashes this instead of erroring cleanly. `Repo.get` raises `Ecto.Query.CastError` on a bad UUID, it only returns `nil` for a valid one that doesn't exist. 

Other lookups guard with `valid_uuid?/Ecto.UUID.cast` first (`job.ex`, `channel_proxy_plug.ex`, `workflow_channel.ex`), should do the same here.

lib/lightning_web/channels/workflow_health_channel.ex
```diff
@@ -0,0 +1,42 @@
...
+         %_{} <- Projects.get_project_user(project, user),
+         true <- Workflows.workflow_exists_in_project?(project.id, workflow_id) do
```
lmac-1: I noticed that this join does quite a few separate DB lookups, including one to check if the workflow exists and then a separate one to fetch it. Is there a reason those aren't combined into a single query? Also the project fetch pulls in `:parent`, which doesn't look like it's used here.

lib/lightning/workflows/stats.ex
```diff
@@ -0,0 +1,61 @@
...
+  # else — cancelled, killed, crashed, failed, exception, lost — is a failure.
+  # Same bucketing the workflow list view shows, so the two pages agree.
+  defp count_workorders(workflow_id, since) do
+    from(wo in WorkOrder,
+      where: wo.workflow_id == ^workflow_id and wo.inserted_at > ^since,
+      group_by: wo.state,
+      select: {wo.state, count(wo.id)}
+    )
+    |> Repo.all()
+    |> Enum.reduce(%{success: 0, failed: 0, pending: 0}, fn
```
lmac-1: This is the same rule as `DashboardStats.count_workorders/1` (`dashboard_stats.ex:239-248`). CON-112 is considering changing this - cancelled will not be counted as failed. If that happens, the rule would need updating in both places, and there is nothing here to tell you the other copy exists.

Would it be worth putting the rule in one place, in `Lightning.WorkOrder` next to `active_states/0`?

```elixir
def outcome(:success), do: :success
def outcome(state) when state in @active_states, do: :pending
def outcome(_state), do: :failed
```

Then both reducers would just be `Map.update!(acc, WorkOrder.outcome(state), &(&1 + count))`.

lmac-1 (commented):
Hey Frank, nice work! I have directly committed some updates to help improve the React side: 

1. **A socket that never connects now shows an error instead of "Loading…" forever.** The hook only handled channel errors, so if the socket itself never came up the page just sat there. It only shows before the first load, so a brief drop doesn't blank a page that already has numbers on it.

2. **Loading and error states moved into the card.** Loading and error states were early returns in `HealthContent`, so a failed request took the workflow name and card heading with it. It's also set things up for the next cards, since each one loads separately.

3. **A rejected join no longer prints "unauthorized" as the page body.** That's the channel guard talking, and non-members get redirected before the page renders anyway. It goes to the console now. I left the `catch` alone — those messages are already written for people.

I have left some inline comments on other areas that I found. I know you asked for a review focusing on the React architecture, but you got some Elixir bits for free 🤭

I also have some design questions:

1. **Should stats update live, or only on page refresh?** CON-107 covers refreshing when you switch range. I'm not sure whether they should also update while you're sat on the page. (Guessing no, but wanted to confirm.)

2. **Are we counting work orders or runs?** CON-106 says the subtitle should read "# runs" but this counts work orders, and CON-110 is a failed *run* breakdown. Different numbers on the same page will confuse people.

midigofrank:
Thanks @lmac-1 . Looks like you forgot to push your changes 👀

lib/lightning_web/channels/workflow_health_channel.ex
midigofrank: Great catch

lib/lightning/workflows/stats.ex
midigofrank: I decided to completely isolate it. I din't want it sharing anything with the `DashboardStats` so that we can modify it as we like

midigofrank:
Thanks @lmac-1 , I have implemented the changes. Regarding your two questions:

1. Runs vs WorkOrders: The workflow index page tracks workorders and it kinda made sense to stick with that. You can retry runs multiple times within a work order until it succeeds. Once the run turns green, that workorder is marked as successful
2. Range switching wasn't part of the scope here. This was to set this up with the utmost minimum which can kick off the other issues

lib/lightning_web/channels/workflow_health_channel.ex
lmac-1: Did this get fixed Frank? I think yes, please mark as resolved if so.

priv/repo/migrations/20260903172213_add_work_orders_workflow_activity_index.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Lightning.Repo.Migrations.AddWorkOrdersWorkflowActivityIndex do
```
lmac-1, line 1: Nice! This is the right index for these queries. 

One thing: we have had `create index(:work_orders, [:workflow_id])` since the original work orders migration in 2022. 

This new one starts with `workflow_id` as well, and Postgres will happily use a composite index for a query that only filters on its leading column. So the old single-column one can't answer anything this one can't. 

Do you think we still need the original index? Postgres updates the index on every insert and update, so on a table as busy as `work_orders` we'd be paying for two where one does the job. What do you think?

assets/js/health/WorkflowHealth.tsx
```diff
@@ -0,0 +1,185 @@
...
+              <FailureBreakdownDonut
+                counts={counts}
+                emptyMessage={emptyMessage(window, 'failures')}
+              />
+            )}
+          </Panel>
+        </Card>
+      </div>
+
+      <Card title="Triage" meta="grouped by failure type">
```
lmac-1: perhaps a note here to say something about how a work order may appear in more than one row? 
(as we are now counting each failed step for multi-branch workflows)

lib/lightning_web/live/workflow_live/health.ex
```diff
@@ -0,0 +1,120 @@
...
+  @impl true
+  def mount(_params, _session, socket) do
+    {:ok, assign(socket, active_menu_item: :overview, refresh: :idle)}
+  end
+
+  @impl true
+  def handle_params(%{"id" => id}, _url, socket) do
+    workflow = Workflows.get_workflow!(id)
+
+    if connected?(socket), do: WorkOrders.subscribe(workflow)
```
lmac-1: Should `WorkOrders.subscribe` be in `handle_params` or `mount`?

This came up in a Claude review and I don't completely understand if I am truly honest. But it mentions that subscriptions might stack up here as nothing unsubscribes. 

> Should this be in `mount` rather than `handle_params`? `handle_params` runs again on patch and there is no unsubscribe, so it looks like it could end up subscribed more than once. Nothing patches to this route today so I don't think it bites, but it feels like the wrong place for it.

test/lightning_web/controllers/api/workflow_health_controller_test.exs
```diff
@@ -0,0 +1,330 @@
+defmodule LightningWeb.API.WorkflowHealthControllerTest do
+  use LightningWeb.ConnCase, async: true
+
+  import Lightning.Factories
+
+  alias Lightning.Workflows.Snapshot
+
+  setup %{conn: conn} do
```
lmac-1, line 8: Is it worth clearing the cache between tests? `:workflow_stats` is a global named cache and nothing resets it. It works at the moment because every test builds its own workflow, but a test that called `outcomes` twice around an insert would get the cached answer the second time and fail for a reason that's hard to spot.

lib/lightning/workflows/stats.ex
```diff
@@ -0,0 +1,300 @@
...
+  # outcome, not the red wedge, which is why this narrows the schema's list
+  # rather than changing it.
+  @failure_states WorkOrder.failure_states() -- [:cancelled]
+
+  # The signature grammar is written in the worker's words, so a
+  # run-level failure has to be mapped back out of its state.
+  @state_reasons Run.state_reasons()
+
+  @doc """
+  Work order counts by final state over the last `days_back` days.
```
lmac-1, line 47: we should try and get the date picker in soon cause i think we need it all working nicely together with caching etc. to regroup on monday to make a plan!

lmac-1 (commented):
This is getting quite hard to review. I'm worried we're going to miss things, and that the back and forth on these points is going to hold up work downstream. I think we need to regroup in sync on Monday about how to make this work better. The refresh changes are quite big.. I'm wondering if they should have been in their own branch while we finalise the details.

Can you confirm whether you've run `/code-review` on the most recent changes? It's hard to review at the moment and the PR description is now stale.

Here's what's come up so far.

**Probably needs a conversation rather than a comment thread (raised on Slack to discuss but I am documenting here):**

- The refresh works by the server signalling and the browser then asking for the numbers. Wondering whether the browser just polling would be simpler.
- We've added a second PubSub topic. WorkflowChannel has the same "just this one workflow" need and filters the project topic instead of adding one
- The stats cache lives in each server's own memory, so the invalidation only clears the server holding the websocket. The browser's next request could land elsewhere.

**Smaller, probably just fixes (came back from reviewing with Claude)**

- The triage counts can now total more than the failed number the donut shows, and nothing on the page says why.
- The 2022 index on `work_orders(workflow_id)` looks redundant now the composite one exists - see inline comment
- `WorkOrders.subscribe` is in `handle_params` rather than `mount`, so it can re-subscribe on patch.
- The `:workflow_stats` cache is shared across the async test suite with nothing resetting it between tests.
- A few nits I'll leave inline.

test/lightning_web/controllers/api/workflow_health_controller_test.exs
midigofrank: Good idea, but I don't think it's worth it. A test that calls `outcomes` twice will need to clear the cache because that is how it is meant to work

lib/lightning_web/live/workflow_live/health.ex
midigofrank: Good point, let me add it to mount

assets/js/health/WorkflowHealth.tsx
midigofrank: Err, yeah. I'm just not sure which words to use, let me try something

priv/repo/migrations/20260903172213_add_work_orders_workflow_activity_index.exs
midigofrank: Good point, I would say we do this as a follow up once the new index has settled in

midigofrank:
Thanks @lmac-1 . I have resolved your change requests, really great feedback 👌 . By the way, I have removed the extra broadcast I introduced. It introduced some failing tests which I didn't want to spend more time reviewing — I'm now filtering the specific workflow events in the Liveview process

priv/repo/migrations/20260903172213_add_work_orders_workflow_activity_index.exs
lmac-1: How can we track this?

lmac-1:
@midigofrank @stuartc I merged #5114 into this branch, so it now carries the range picker. 

I also edited 2 lines of the PR description to match this new behaviour. I did this directly in the PR body rather than a comment in case Stu picks up the review in the morning.  

Stu, apologies if you already started your review. I think whatever you had before this was in is still valid. Feel free to glance at the merge commit if you wish.

lmac-1:
FYI Frank I have raised https://github.com/OpenFn/lightning/pull/5153 to unify the definition of "failed work order" across the site and also the dates we use to filter 30 days. No action needed in this PR but in case it comes up in any reviewing.

lmac-1:
Pr desc updated to reflect merged PRs in the parent branch. Running ultra code review now.

lmac-1 (approved):
Hey Frank, 

We're almost there. What's left for you to do: 
1. Approve and merge #5153 (if there's anything release blocking, please fix directly)
2. Do one final pass on this PR and approve and merge into main
3. Please read [my comment on 5162](https://github.com/OpenFn/lightning/pull/5162#pullrequestreview-5161845041), particularly the point about "Stats.runs/2 never serves from cache on a busy workflow". I am pretty sure this is not release blocking.. but there might be a small thing to tidy up as the final refresh changes came in separately from the runs query. If not release blocking and we need to do something, please add something brief in Linear to track. 

I've tried to make number 2 easier for you. I ran ultra code review and it came back with one nit (yay!): 

```
lib/lightning/workflows.ex:1224 — workflow_exists_in_project?/2 now does get_workflow_for_project(...) != nil, which hydrates a full %Workflow{} (all columns, including positions) just to test for nil. It runs on every LiveView mount carrying a workflow id (hooks.ex:145).

Confirmed against the code — get_workflow_for_project/3 at workflows.ex:148-154 is a Repo.one on SELECT *.

Fix is a one-liner if you want it:

def workflow_exists_in_project?(project_id, workflow_id) do
  Lightning.Validators.valid_uuid?(workflow_id) &&
    from(w in Query.workflows_for(%Project{id: project_id}), where: w.id == ^workflow_id)
    |> Repo.exists?()
end
```

I would push back on this. The reason we used the function the way we did here was to make sure we went through the same functions with the proper auth guard. Can you verify?

Btw my most recent commits were ponytail refactors. I made sure to verify them before committing.

midigofrank:
Thanks @lmac-1 , I have gone through the changes and resolved the feedback

Issue #5042, opened 2026-08-04
Sunset Non Global AI Assistant


elias-ba:
Lightning should have one AI assistant, with no "global", "job chat" or "workflow chat" anywhere in the code, the wire, or the database.

Most of this is already done. What is left is four pieces of work, one product decision, and four traps that will each cost a day if you meet them by surprise.

## Already done

Job chat has been unjoinable from the UI since #4393 in February. `useAIMode` returns a hardcoded `workflow_template`, every channel topic is built from that, and the server rejects a topic that disagrees with the stored row. `job_code` sessions are legacy data with no way in.

Old sessions also heal themselves. The client sets `use_global_assistant: true` on every send and session meta is merged rather than replaced, so an old session becomes a global one as soon as someone types in it.

So the old routing is nearly dead. Do not plan for a migration of live traffic.

## What is left

**1. Delete the job and workflow paths.** `AiAssistant.query_stream/3`, `query_workflow_stream/3`, the `job_chat?/2` branch in `MessageProcessor`, and the two matching wrappers in `ApolloClient`. Also dead and deletable: `AiAssistant.create_session/4` (no callers in `lib`, 54 in tests), `enrich_session_with_job_context/1` and everything only it reaches, and the `step_id` I/O branch of `process_job_message/2`.

This needs one decision first. Retry is the only remaining way to reach those branches: it re-enqueues a message without writing session meta, so a pre-#5121 session still dispatches down the old path. Deleting the branches means those retries go global instead. That is almost certainly what we want, but it is a behaviour change and should be stated rather than arrive as a side effect.

**2. Drop the two meta keys.** `use_global_assistant` in session meta and `from_global` in message meta. Drop them and read a default rather than renaming them — see the traps.

**3. Shorten the channel topic** from `ai_assistant:<session_type>:<id>` to `ai_assistant:<id>`. Server and client together, in this order: teach the server the new form while it still parses the old one, ship, then move the client. The client must also stop reading `session_type` from the join reply before the server stops sending it.

**4. Rename the vocabulary.** Twelve Elixir functions, each with one caller. Around 180 JavaScript identifiers, none of which cross the wire. The module docs in `ApolloClient` and `AiAssistantChannel` that still describe two session types. Two factory names. Three test filenames nothing imports. And the two unreleased changelog entries that describe this sunset in the words it retires.

Piece 4 is independent of everything else and safe to do first if you want momentum.

## The real cost is the tests

[...]

PR #5121, opened 2026-09-06
Global assistant: attach the run you are looking at

### LINKED ISSUE #5037: Global Assistant: How to allow users to attach full YAML, Logs and I/O Data

### LINKED ISSUE #5040: Global assistant: Switch default model in AI-first generation to the global assistant

elias-ba:
## Description

This PR changes what the assistant's attachment tickboxes mean, and makes the global assistant the only one the UI reaches.

1. "Send code" is gone. The assistant sends the whole workflow on every message regardless of that box, so it did nothing.
2. "Send run logs" and "Send run data" now sit above the message box on any page and appear once a run is loaded. "Send run data" covers the whole run rather than one highlighted step. Values are replaced by their types, but field names go as they are, so a job that keys state by patient name or national id sends those. Their number and length are capped.
3. The experimental gate, the opt-in tickbox and the assistant badge are gone. The older paths stay and still serve existing sessions; removing them is #5042.
4. "Build with AI" went to workflow chat, which leaves every step empty. It goes to the global assistant now.
5. A session's first message lost its attachments when asked from the canvas: `buildJoinParams` forwarded the flags only when a job was in context. Same bug `code` had, fixed the same way in #4888.

Closes #5037
Closes #5040

## Validation steps

1. Open a workflow and open the assistant with no run selected. There are no attachment tickboxes, and no "Send code" box, global assistant toggle or assistant badge anywhere.
6. Select a run in the history panel. "Send run logs" and "Send run data" appear above the message box, both on the canvas and with a step open in the editor.
7. Tick both, start a **new** conversation, and ask "Summarize the log and data you received". The reply describes the run's log lines and the input and output of every step in the run, not just the highlighted one.
8. Read the data it quotes back. Every value is its type, so a name reads `"string"` and a count reads `"number"`.
9. Untick both and ask again in a new conversation. The assistant answers that nothing was attached.
10. Go to a project's new workflow screen and use "Build with AI". The prompt opens the assistant as the first message of a new session, the workflow appears on the canvas, and there are no attachment tickboxes, since a new workflow has no run.

Step 3 asks for a new conversation on purpose. Every message after the first goes through a path that always carried the flags, so a follow-up passes even with the join bug in place.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,35 @@ and this project adheres to
 
 ### Changed
 
+- The AI assistant is the global assistant for everyone. It was behind the
+  experimental features setting and an opt-in tickbox on the chat input, and
+  both are gone: every message goes to it, and the badge naming which assistant
+  answered goes with them. Existing workflow conversations still open and read
+  as they always did, and replying in one moves it to the global assistant from
+  that message on.
+  [#5042](https://github.com/OpenFn/lightning/issues/5042)
+
+- The AI assistant's attachment tickboxes are now about the run you are looking
+  at, and they appear wherever you are. "Send run logs" and "Send run data" sit
+  above the message box on the canvas and in the run history as well as in the
+  step editor, and appear once a run is loaded rather than sitting greyed out.
+  "Send run data" now covers every step in the run rather than the one that
+  happened to be highlighted, and still sends the shape of the data with the
+  values removed. The "Press Enter to send" hint below the box is gone, since
+  the notice and the send button now share that row.
+  [#5037](https://github.com/OpenFn/lightning/issues/5037)
+
 - Runs on Erlang/OTP 28 and Elixir 1.18.4. OTP 27 only finishes normalising
   the first character of a string, which breaks names in many languages.
   Lightning does not normalise anything today, but #4577 adds it on every name,
   so the runtime moves first.
+
+### Removed
+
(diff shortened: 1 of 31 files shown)


hanna-paasivirta (approved):
Attaching/removing attachments and I/O data works reliably. Looks great!

elias-ba:
Thanks for the review and for the design brainstorming @hanna-paasivirta

PR #5151, opened 2026-09-08
Workflow health: Ability to view grouped triaged work orders in history page

lmac-1:
## Description

Adds a **View** button to each row of the workflow health page's triage table. It links to the history page filtered to exactly the work orders behind that row, where the existing "retry all" can act on the group.

image

Underneath the button:

- A shared signature query, so the health page and the history filter agree on what counts as "the same failure."
- Triage rows key on `job_id` now, not the resolved job name, so a rename or adaptor bump mid-window merges into one row instead of two.
- Three new filter keys on the history page (`exit_reason`, `error_type`, `job_id`), which also scope bulk retry, bulk cancel, and CSV export since they share one query.
- A filter chip, so an active signature filter is visible and clearable, not just applied silently.

Also fixes a bug where an empty error type split one failure into two identical-looking triage rows.

Builds on #5108 (still open); this branch rebases onto it once it merges.

Closes CON-113

## Validation steps

1. On a workflow with failed work orders, open its health page and click **View** on a triage row.
2. Confirm history opens filtered to the workflow, the picked date range, and that row's signature. Confirm the filter chip is showing too.
3. Confirm the work orders listed match the row's count, and that "retry all" only retries that group.
4. Type a term into the search box and confirm it narrows that list rather than emptying it.
5. Clear the chip and confirm the filter drops back to just the date range.
6. Click **View** on a rejected row and confirm it filters on `filters[rejected]=true` rather than a signature.

- **Rolling deploy risk on CSV export.** It runs as a background job that any node can pick up. If an old node (pre-this-PR) runs it, it silently drops the three new filter keys and exports more rows than the user asked for. Self-resolves once the deploy finishes.
- **A bug in the shared query would widen retry/cancel/export, not just search.** `SearchParams.new/1` raising on a malformed field limits how far that can reach.
- **The adaptor no longer shows its version**, since a merged row can span more than one. Showing versions again later needs us to record which version actually ran a work order, which we don't today.
- **The UI needs a design pass** (follow-on PR). Right now the View button sits too far from its row.
- **The View link builds its query string in TypeScript**, not through `SearchParams.to_uri_params/1` like every server-rendered link. So it has to carry history's four search-field flags itself, or the search box on the page it lands on matches nothing. Any filter key added later needs the same treatment.
- **It's normal for the count to drift by one or two right after clicking View.** Health numbers are cached for up to 30 seconds.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,9 @@ and this project adheres to
 - A workflow health page at `/projects/:project_id/w/:workflow_id/health`,
   summarising one workflow over a selectable window (last 24 hours, 7 days, or
   30 days): a donut of work order outcomes, a breakdown of the failing ones, and
-  a triage table grouping failures by error signature, heaviest first. The page
+  a triage table grouping failures by error signature, heaviest first. Each row
+  has a View button linking to the history page filtered to just the work orders
+  behind it, where the existing "retry all" can retry the group. The page
   refreshes itself as that workflow's work orders settle, at most once every 30
   seconds. Reachable from the workflows list via a "Health" link in each row's
   Actions column.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,16 @@ and this project adheres to
 
 ### Changed
 
-- Runs on Erlang/OTP 28 and Elixir 1.18.4. OTP 27 only finishes normalising
-  the first character of a string, which breaks names in many languages.
-  Lightning does not normalise anything today, but #4577 adds it on every name,
-  so the runtime moves first.
+- Runs on Erlang/OTP 28 and Elixir 1.18.4. OTP 27 only finishes normalising the
+  first character of a string, which breaks names in many languages. Lightning
+  does not normalise anything today, but #4577 adds it on every name, so the
+  runtime moves first.
+- The workflow health page's triage table now merges two rows that describe the
+  same failure into one, correcting a split count on a shipped page: a step
+  reporting an empty error type instead of none, and a job renamed or
+  adaptor-bumped mid-window, both used to draw two identical-looking rows with
+  the count divided between them. Merged rows are labelled from the most recent
+  snapshot to fail.
 
 ### Added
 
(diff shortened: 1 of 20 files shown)


midigofrank (changes requested):
Hey @lmac-1 , really great work. A few things I've noticed:

1. The view link opens in the same tab. I wish it could open on another tab so that I can go back to view others
2. The `status` is not selected on the history page. We have the value, the signature starts with the `status:....` right?
3. The job id is displayed as a uuid, but I think we can show the name

Codewise:

1. Because the newly added fields to `SearchParams` are filtered as one, I think we should add nested field in there to group them, like:
 ```elixir
embedded_schema do
...
embeds_one :error_signature, ErrorSignature do
    field(:exit_reason, :string)
    field(:error_type, :string)
    field(:job_id, :binary_id)
end 
end
```

This way the `filter_by_signature` would pick them as a whole. Alternatively, I would add the fields in full:

```elixir
    field(:step_exit_reason, :string)
    field(:run_error_type, :string)
    field(:failed_job_id, :binary_id)
```

2. I don't see the need of adding `Query.failure_states()`. Why can't we fix it directly from `WorkOrder.failure_states` ?

lmac-1:
Hey @midigofrank, thanks for the review.

1. Done, opens in a new tab now (with `rel="noopener"`).

2. Not quite. The status on a row is the _step's_ exit reason, not the work order's state. The worker reports the run state and each step's reason separately, so a run whose step failed with `fail` can still end up `crashed` or `killed` further down a branch, and since we attribute a work order to every failing step of its latest run, a `fail:` row legitimately counts work orders in several states. There's a test for exactly this in `invocation_test.exs`, where one signature matches both a failed and a crashed work order. So ticking "Failed" would hide rows the triage row counted, and "retry all" would then miss them. The filter already limits results to failure states server-side, so the status boxes would only ever subtract. I've left a comment in `historyUrl` saying so.

3. Done, it resolves the name now, scoped to the project so an id from somewhere else can't name a job. Falls back to the short id if the job is gone.

4. Went with your second option, the fields in full, named `error_signature_exit_reason` / `_error_type` / `_job_id`. The embed reads better, but the history filters travel as a flat `filters[...]` query string cast through a flat type map in the LiveView, and the chip's clear button pushes a flat map back through `apply_filters`, so `cast_embed` means teaching all three about nesting. Say the word if you'd still prefer it.

5. Done. `WorkOrder.failure_states/0` takes `:cancelled` off itself now, and `Query.failure_states/0` is gone. It had no other callers.

One extra thing: your `filter_by_error_signature` rename turned up that we had three names for one concept, `filter_by_signature`, the `signature_*` fields, and `Stats.failure_signatures/2` with the `FailureSignature` type. It all says error signature now, since that's what we say out loud anyway. Nothing crosses the wire: `/health/failures` and the response keys are unchanged.

Commits are split one per point if that's easier to read through.

PR #5152, opened 2026-09-08
Record which credential values a project may read

elias-ba:
### Validation steps

A project now records which set of a credential's values it may read, on the sharing row itself.

Until now that was decided by matching the project's environment *name* against the credential's value sets at run time. The name is typed on the project settings screen; the values are typed on the credential screen, by a possibly different person. Nothing tied the two together, so typing the right name was enough. And the whole chain is reachable by an editor: creating a sandbox makes you its owner, a sandbox owner can edit that project's environment, and a sandbox already holds a reference to every one of its parent's credentials. So an editor on a production project could manufacture owner rights for themselves on a project holding references to that project's credentials, type the parent's environment name into a text box, and decrypt its production secrets. No admin approves anything, and the change is not audited.

It was not hypothetical. 27 sandboxes across 13 customer projects were sitting on an ancestor's environment when I checked.

Five commits, in this order, so the tree is green at each one.

The first adds the column and the constraint. The foreign key is composite on `(credential_id, credential_body_id)`, so a share can only ever grant a body of the credential it was given; a single-column reference would be the same hole in a new place. It is `NO ACTION` rather than `RESTRICT` because deleting a credential relies on two cascades firing inside one statement, and `RESTRICT` checks too early to allow that.

The second sets the grant wherever a share is created. A credential with one body grants it, so the credential works where it was made; with more than one the choice is a project admin's. Three paths deliberately grant nothing and now say so: a sandbox clone, a merge carrying a sandbox-only credential up, and adding a credential to a parent, which reaches every sandbox beneath it.

The third backfills and gives someone a way to choose. Every root project's share is granted the values it resolves today, so nothing changes for a root. Sandboxes are left ungranted, including the ones whose environment already matches an ancestor's, because granting those would record the accident as a decision. A project's credentials tab gains a "Values used here" column, which an admin can change and an editor can only read.

The fourth switches every reader over at once. Two readers disagreeing about how a credential resolves is the bug rather than a step towards fixing it, so the run path, the channel proxy's destination auth and the editor's adaptor metadata all move together. `Projects.Environment` goes with them: it existed to answer a question nobody asks any more. No grant is a refusal, not a fallback, and the error names the credential and points at where the choice is made. It no longer suggests changing the project's environment to match, which was the exposure shipped as product copy.

Two things to know before this ships.

**58 sandboxes stop resolving credentials.** That is every sandbox with a credential wired to a job, not only the 27 that were exposed. Some of those are working setups. There is deliberately no override: someone with admin on the project picks the values on the credentials tab, and that is the only route. The backfill prints the count on the way past.

**The scrubber now masks every value set of a credential** rather than the one an environment named. Nothing records which values a step actually spent, so narrowing it would be guessing, and a wrong guess puts a secret on screen. That also fixes a live bug where the first credential's environment was applied to all of them.

I have left the private tracker's numbers out of this deliberately.

1. On a root project, check credentials resolve exactly as before and runs are unaffected.
2. Clone a sandbox, turn it on, and run a job. Check it refuses with a message naming the credential, and check the message does not tell you to change the project's environment.
3. On the sandbox's credentials tab, pick a value set under "Values used here". Run again and check it uses those values.
4. Set the sandbox's environment to the parent's. Check it still reads only what it was granted.
5. As an editor, check the column is readable and not changeable.
6. Check a channel with destination auth in an ungranted project returns 502 with `credential_not_granted` recorded, and works once granted.
7. Check adaptor metadata in the editor uses the granted values.
8. Try to delete a value set that a project is granted, and check it is refused rather than silently breaking that project.

### AI usage

Implementation and PR body drafted with Claude Code, reviewed and corrected by me.


elias-ba:
Closing this. The work is on `sandbox-devx` and ships with #5137. Nothing here should have targeted `main` in the first place, so there is nothing left for this PR to merge.

PR #5157, opened 2026-09-09
Workflow Health: poll from the browser instead of pushing refreshes

midigofrank:
## Why

The health page refreshed by push: the LiveView subscribed to the project's work order events, filtered them to this workflow, throttled them, dropped the stats cache, then told the page to re-read. The invalidation is `Cachex.del` against a node-local ETS table, so it only holds on one node — a viewer on another pod kept being served the value computed before the change.

## How

**The cache key carries the change marker.** `Workflows.Stats` appends the workflow's `max(work_orders.last_activity)` to the key instead of invalidating on settle. A read that finds nothing has moved hits on every pod, because the marker comes from Postgres rather than one node's ETS; a read after something settled lands on a new key. The lookup is an index-only scan on `(workflow_id, last_activity)`. TTL goes 30s → 5 minutes, since it now only caps what the marker can't see. The trade is that a continuously busy workflow recomputes on every poll. `invalidate/1` is gone, along with the LiveView's subscription and throttle.

**The page polls.** `useHealthQuery` swaps the `phx:health:changed` listener for a 30s interval, skipped while the tab is hidden and bumped on `visibilitychange`. A failed poll keeps the numbers already on screen — stale by an interval, not wrong. "Last Updated" now reads `window.to` off the response, so it stamps when the numbers were true rather than when the browser asked.

**Cache errors reraise.** `Cachex.fetch` rescues whatever the fallback raised and hands it back as an error tuple, which used to fall through to a `CaseClauseError` — Sentry saw the tuple mismatch instead of the DB failure underneath.

## Also here

Second commit, separable: steps and runs store both `""` and `NULL` for "we were never told an error type". The triage table renders both as "unknown", so they drew as two rows under one React key, and `""` being truthy meant an empty step type masked the run's real one in the signature. Both selects now wrap the column in `NULLIF(?, '')`.

## Testing

`health_test.exs` loses the push/throttle tests with the code they covered. `stats_test.exs` covers marker-keyed caching, the reraise, and the error-type grouping. `WorkflowHealth.test.tsx` covers the poll interval, the hidden-tab skip, and a failed poll keeping its data.


lmac-1 (approved):
Three commits from a review pass on this branch.

**Don't let a poll abort the read it is waiting on.** The tick sat in the fetch effect's deps, so that effect's cleanup aborted any read slower than the 30s interval, and the tick behind it started another. On an endpoint consistently that slow the panel never left "Loading…" while the server recomputed for a reader who was no longer there. The tick now waits for the read to land, which also measures the gap from the answer rather than from the request.

**Stop claiming two things the health page doesn't do.** Comment and CHANGELOG only, no behaviour change.

**Only let a settled work order move the change marker.** The marker took `max(last_activity)` over all work orders, but `last_activity` moves on every run state change while the page counts only work orders that have settled. So a workflow on a cron got a fresh cache key on every 30s poll, and both slices recomputed in full for every open tab, while the numbers on screen never moved. Only work orders in a final state move the marker now.

The gap this leaves, also in the code comment: a work order is not final while a retry or a sibling run is still in flight, so a settled run inside it stops counting until that work order settles or the entry expires.

**Not in this PR**

Five smaller findings from the same review are deliberately left out and will be filed separately: a page that stops updating silently after repeated failed reads, "Last Updated" showing the server's compute time rather than the read time, the tab-return refresh being skipped while a read is in flight, cache duplication with no size limit on the cache, and one Sentry error that cannot be grouped.

Logged in CON-183

**Note for the #5162 merge**

That branch predates the change marker, so its `cached/2` has no marker in the key and it still has `Stats.invalidate/1`. Resolving that hunk in its favour would revert the polling model silently, with no failing test.

PR #5160, opened 2026-09-09
Stop anyone changing a project's environment

elias-ba:
## Description

This PR stops anyone changing a project's environment after the project exists.

The environment decides which of a credential's value sets the project reads, and
it is a text field on project settings that any project admin can type. So an
admin of a project a credential is shared with can read any set of values on that
credential. It is worse in a sandbox. A sandbox holds a reference to every
credential its parent holds, and creating a sandbox makes you its owner, so a
sandbox admin can name the environment after the parent's and read the parent's
production values. Nobody approves it and nothing records it.

The field stays on the page, because it tells you which values resolve there, but
it is read-only, and `env` comes out of both places that took it from a user: the
settings form's allowlist, so a crafted submit is ignored as well, and
`update_sandbox`. Our own code still sets it, so a new root project gets `main`
and a new sandbox gets `dev`.

## Validation steps

1. As a project admin, open project settings. The environment is shown and cannot
   be typed in.
2. Submit the settings form with an `env` value anyway, by hand. The project
   saves and the environment does not change.
3. Create a sandbox, become its owner, and try the same on its settings. Same
   result.
4. Check a credential still resolves as before on a project whose environment
   already matches one of its value sets.

1. This does not undo access anyone already has. Sandboxes whose environment was
   changed before this keep reading the parent's values until those are reset,
   which is a data fix. 27 sandboxes across 13 projects are in that state.
2. Nobody can set an environment now, including support. If a customer needs a
   project mapped to a credential's non-default value set, that has to go through
   a console or a follow-up that gives someone the ability deliberately.
3. This is the stopgap. It defends a name rather than recording which values a
   project may read, so the rule has to hold at every future write path. The
   fuller fix records the grant on the share itself.
4. Also in here: the `.gitignore` pattern for `priv/schemas` loses its trailing
   slash, so the symlink a worktree gets is ignored rather than committed by
   accident.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,8 +45,9 @@ adaptor_registry_cache.json
 # Ignore digested assets cache.
 /priv/static/cache_manifest.json
 
-# Ignore credential schemas.
-/priv/schemas/
+# Ignore credential schemas. No trailing slash: it is often a symlink to a
+# shared checkout, and a trailing slash matches only directories.
+/priv/schemas
 
 # In case you use Node.js/npm, you want to ignore these.
 npm-debug.log
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Security
+
+- A project's environment can no longer be changed once the project exists. It
+  decides which of a credential's value sets the project reads, so anyone who
+  could type it could read any set on any credential shared with that project.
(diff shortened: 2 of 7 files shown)


elias-ba:
Closing this. The same change is on `sandbox-devx` and ships with #5137, so there is no reason for a second route to main.

PR #5162, opened 2026-09-09
Workflow health: bucketed run counts behind the window

midigofrank:
## Description

The outcomes and failures panels both aggregate over the whole window, so neither can answer *when* the traffic arrived. Add `Stats.runs/2` and `GET /health/runs`: final run counts per state, bucketed over time — 2-hourly over a day, AM/PM over a week, daily over a month.

The bucketing is done in Postgres, not the browser. The alternative is shipping every run in the window — six figures of rows on a busy workflow, cached whole and JSON-encoded — to draw thirty bars.

Runs are counted on `inserted_at` rather than the state transition, so a run stays in the bar its attempt started in. The redundant `last_activity` condition on the work order is there for the planner: it lets the existing `work_orders(workflow_id, last_activity)` index narrow the work orders before the nested loop into `runs(work_order_id, inserted_at)`.

### Response

```
GET /api/projects/:project_id/workflows/:workflow_id/health/runs?days=1
```

```json
{
  "window": {
    "from": "2026-09-08T12:00:00Z",
    "to": "2026-09-09T13:47:09.512345Z"
  },
  "buckets": [
    { "at": "2026-09-08T12:00:00Z", "success": 41, "failed": 2, "crashed": 0,
      "cancelled": 0, "killed": 0, "exception": 0, "lost": 0 },
    { "at": "2026-09-08T14:00:00Z", "success": 38, "failed": 0, "crashed": 1,
      "cancelled": 0, "killed": 0, "exception": 0, "lost": 0 },
    "… 11 more, through 2026-09-09T12:00:00Z"
  ]
}
```

Three things about that shape, all of them for the chart:

**A bucket is flat, not nested.** `{ at, success, failed, … }` is the row shape Recharts takes as `data`, so the list goes to the chart untouched and each `Bar` names the state it draws:

```tsx
<ResponsiveContainer width="100%" height={220}>
  <BarChart data={data.buckets} accessibilityLayer>
    <XAxis dataKey="at" tickFormatter={tick} />
    <YAxis allowDecimals={false} />
    <Tooltip labelFormatter={tick} />
    {RUN_STATES.map(state => (
      <Bar key={state} dataKey={state} stackId="runs" fill={COLORS[state]} />
    ))}
  </BarChart>
</ResponsiveContainer>
```

No `reduce` to pivot counts into rows, no `groupBy` in the client, and a state added to `Run.final_states/0` shows up as a key rather than as a silently dropped row.

**Every bucket carries every state, zero-filled**, so the chart draws a flat window without reasoning about which bars are missing — an idle day is 13 bars of zero, not an empty box. (This is the case the donut needed an explicit `emptyMessage` branch for; a bar chart doesn't.)

**Boundaries sit on the clock**, so `at` can be labelled honestly and the labels don't shift with whatever minute the request landed on:

```tsx
const tick = (at: string) => {
  const d = new Date(at);

  switch (days) {
    case 1:  return d.toLocaleTimeString([], { hour: 'numeric' });   // "2 PM"
    case 7:  return `${d.toLocaleDateString([], { weekday: 'short' })} ` +
                    `${d.getHours() < 12 ? 'AM' : 'PM'}`;            // "Tue AM"
    case 30: return d.toLocaleDateString([], { month: 'short', day: 'numeric' });
  }
};
```

Every bucket width divides a day evenly, and `window.from` is aligned to the width, so integer division by the width is the whole of the bucketing in SQL — no `date_trunc` special case per range.

### Two details worth a look in review

`extract(epoch from …)` yields `numeric`, and `numeric::bigint` rounds. Without the `floor` a run at 01:59:59.7 is counted in the 02:00 bar; there's a test pinned a hair under a boundary for exactly this.

Each window carries one bucket *more* than its width divides into (13 / 15 / 31, not 12 / 14 / 30). `now` sits mid-bucket, so a grid of exactly `days_back / width` bars would start *after* `now - days_back` and leave the oldest hours of the window undrawn while the donuts beside it counted them. The oldest bar reaches back past `from` instead, and `window` reports the range the bars actually cover — the last bar is the one `now` falls in, so it's still filling.

Closes CON-182

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)

### DIFF
--- a/assets/js/health/WorkflowHealth.tsx
+++ b/assets/js/health/WorkflowHealth.tsx
@@ -71,12 +71,9 @@ export const HealthContent = ({
   return (
     <div className="flex flex-col gap-6">
       <div className="flex items-start justify-between gap-4">
-        <div>
-          <h1 className="text-2xl font-semibold text-gray-900">
-            {workflowName}
-          </h1>
-          <Subtitle outcomes={outcomes.data} error={outcomes.error} />
-        </div>
+        <h1 className="min-w-0 text-2xl font-semibold break-words text-gray-900">
+          {workflowName}
+        </h1>
         {/* The picker and the freshness stamp both belong to the whole page,
             so they stack in the header rather than sitting on any one card. */}
         <div className="flex shrink-0 flex-col items-end gap-1">
@@ -201,7 +198,7 @@ const Panel = <T,>({
   children: (data: T) => ReactNode;
 }) => {
   // `alert` is the assertive live region: a failure is read out at once,
-  // where the polite subtitle would only fall silent.
+  // where the polite stamp above would only fall silent.
   if (error) {
     return (
       <p role="alert" className="text-sm text-red-700">
@@ -216,8 +213,7 @@ const Panel = <T,>({
 
 // Reached on a first load and again after a failure, so it holds the chart's
 // frame either way and the card doesn't jump when the data lands. Not a live
-// region: the subtitle announces loading once for the page, this text is only
-// for a reader who lands inside the card.
+// region: this text is only for a reader who lands inside the card.
 const ChartLoading = () => (
   <div className={FRAME}>
     <span className="sr-only">Loading…</span>
@@ -226,39 +222,16 @@ const ChartLoading = () => (
 
 // The stamp is the server's compute time, not the moment the browser asked —
 // `window.to` is when the numbers were true, however long the round trip took.
-const UpdatedAt = ({ at }: { at: string | null }) => {
-  if (!at) return null;
-
-  return (
-    <span className="text-xs text-gray-500">
-      Last Updated {new Date(at).toLocaleTimeString()}
-    </span>
-  );
-};
-
-// Holds its line while empty (`min-h-5` is one `text-sm` line). It names the
-// window the numbers beside it came from, and goes back to "Loading…" on a
-// range switch rather than naming a window no panel is showing yet. The page's
-// one polite live region: a first load and a range switch are each read out
-// here once, rather than by every card in turn. Falls silent on a failure
-// rather than announcing a load that isn't happening — the cards raise that as
-// an alert, which is read out at once where this region would have to wait.
-const Subtitle = ({
-  outcomes,
-  error,
-}: {
-  outcomes: Outcomes | null;
-  error: string | null;
-}) => (
-  <p role="status" className="min-h-5 text-sm text-gray-500">
-    {outcomes
-      ? `Last ${windowLabel(outcomes.window)} · ${workOrders(outcomes.counts)}`
-      : !error && 'Loading…'}
+//
+// Holds its line while empty (`min-h-4` is one `text-xs` line) so the picker
+// above it doesn't move.
+const UpdatedAt = ({ at }: { at: string | null }) => (
+  <p className="min-h-4 text-xs text-gray-500">
+    {at && `Last Updated ${new Date(at).toLocaleTimeString()}`}
   </p>
 );
 
-// "1 work order", "1,287 work orders": the subtitle and the Outcomes card both
-// say it, so it is spelled once.
+// "1 work order", "1,287 work orders".
 const workOrders = (counts: Outcomes['counts']) => {
   const total = Object.values(counts).reduce((sum, count) => sum + count, 0);
 
--- a/assets/js/health/WorkflowHealth.tsx
+++ b/assets/js/health/WorkflowHealth.tsx
@@ -83,8 +83,8 @@ export const HealthContent = ({
       </div>
 
       {/* Four columns: a narrow donut beside a wide time axis, then an even
-          split. Stacks in source order below `md`. */}
-      <div className="grid gap-6 md:grid-cols-4">
+          split. Stacks in source order below `lg`. */}
+      <div className="grid gap-6 lg:grid-cols-4">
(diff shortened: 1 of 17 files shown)


lmac-1:
## Reviewed, plus the chart on top

I've reviewed this and then built the frontend for it on the same branch, so this comment is both halves. Six commits on top of the `frank/con-106` merge, all of them in `assets/` — `stats.ex`, the controller and the migration are untouched, so the backend you're asking about is exactly as you left it.

### What's on top

- **`VolumeBars.tsx`** — the stacked bar chart this endpoint exists for.
- **`ChartTooltip.tsx`** — extracted, and both donuts moved onto it, so the three charts hover the same way instead of two Recharts defaults and one custom.
- **Page tidy** — header layout, the empty-panel branches, and the grid holding one column until `lg` rather than `md`.
- **Card meta lines** cut back to a unit and a count.

### Where the code diverges from the sketches above

The frontend snippets in the description were written before the chart existed and three of them didn't survive contact. Flagging so the mismatch doesn't read as a bug during review:

- **The `tick` sketch reads local time** (`d.getHours()`) and labels the week `Tue AM`. Shipped: everything renders UTC, and the 12-hour view puts the date on each day's morning bar and blanks the afternoon. `Tue AM` across fifteen half-day buckets is a label per bar, which is too dense to read.
- **`accessibilityLayer`** is `false`, not on. Turning it on for one chart adds a tab stop the donuts don't have. The chart is `aria-hidden` and its accessible representation is one `sr-only` line of window totals, named as totals — a sum over the window says nothing about *when*, and the honest per-bucket reading is ninety-odd cells.
- **Stacking is failures-on-the-axis**, not success. In a stacked bar only the total and the bottom series are accurately readable; everything between floats. The total is free either way, so the real question is which single series you want comparable over time, and on a page whose job is driving failures down, that's failures.

### Timezones

The buckets are UTC and the chart says so — the axis, the tooltip's explicit range, and now the card meta line (`daily buckets · UTC`).

That's honest, not right. Almost all our users are outside UTC, so a "day" bar is 3am–3am in Nairobi and 5:30am–5:30am in Delhi, and it's the exception to how the rest of Lightning behaves — History, audit, tokens and the dashboard all render local via `Common.datetime` and the `LocalTimeConverter` hook.

I looked at fixing it today and decided against it this close to the release. It's genuinely small — the whole grid hangs off `bucket_window/2`, so flooring on `now + offset` instead of `now` moves every boundary, label and zero-fill with it — but it has three quiet failure modes that a same-day change wouldn't catch: the offset has to go in the stats cache key or the first viewer pins their grid on every other timezone for five minutes; the Vitest suite needs `TZ` pinned once the labels stop reading UTC directly; and half-hour zones need minutes in the axis formatter or a 19:30 bar prints as `19:00`. Filing it as a fast follow with those written up.

One live symptom to know about, since it's the version a user will actually hit: the "Last Updated" stamp in the page header is in the reader's local clock and unlabelled, directly above bars that are UTC. In Nairobi that reads as "Last Updated 17:23" over a last bar labelled 14:00 — a chart that looks three hours behind. Naming UTC on the meta line is what mitigates it for now; it goes away with the fix above.

### Worth eyes rather than tests

The tooltip and the axis label positioning have no tests — Recharts renders both with real layout and jsdom does neither, so they need looking at rather than asserting. Everything else is covered: `npx vitest run test/health/`.

Two open items I left deliberately, both noted for follow-up rather than blocking:

- **The 7-day axis.** Fifteen half-day buckets is an awkward shape for a week and the labels are working around it. The alternative is daily buckets (8 bars, one label each), which costs morning-vs-afternoon everywhere — a one-line change to `@buckets` if we'd rather have the cleaner axis.
- **The trailing bar** reads as a collapse in volume. It's the currently-filling bucket plus in-flight runs being excluded. Options if we revisit: dim it, or end the axis at the last complete bucket.

lmac-1 (approved):
Hey Frank! Thank you for your amazing work on getting this out so quickly and also doing the buckets stuff. Really great work. I've since done some work to get this done on the frontend and also added some UI polishing stuff in here too. My previous comment gives a summary of what was added and then used as a basis for my self review. 

Two known gaps deferred and filed:
- CON-188: chart buckets are UTC, not the reader's timezone
- CON-189: clicking a bar segment through to history

Three things worth an eye:

**`search_params.ex` changes history filtering for everyone**, not just this page. A URL carrying a search term but none of the four search-field flags used to match nothing; it now falls back to the schema defaults. I've traced it — the form always ships a hidden `false` per checkbox, so "all toggles off" still means off, and app-built links always carry all four. Safe, but flagging it because the diff's centre of gravity is elsewhere and it's easy to skim past.

**The migration builds its index concurrently.** If that half-fails during deploy, Postgres leaves an invalid index behind and `create_if_not_exists` skips it on a retry, so the queries quietly fall back to the old single-column index. Worth confirming `indisvalid` after deploy.

**`Stats.runs/2` never serves from cache on a busy workflow** — the key carries the latest settled `last_activity`, so a workflow settling faster than the 30s poll recomputes every time, and that's the workflow with the most rows to aggregate. Unmeasured; needs production-shaped data. @midigofrank this came up in self review and I didn't really want to make cache changes so late in the day. I'll post this in a Slack comment to you too. 

Approving and merging! 🎉

PR #5189, opened 2026-09-15
Adds links to workflow health page to emails

lmac-1:
## Description

This PR points the two "something is wrong" emails at the workflow health page, and shortens the history URLs they print.

Both emails sent readers to **history**, which lists work orders but never says whether failures repeat. **Health** groups them by error signature. That is the question both emails leave the reader holding.

### Digest

**Before** — one link, unlabelled destination:

```
Workflow A:
• 12 workorders were successful today
• 3 workorders were not (crashed, exception, failed, killed, lost, rejected etc.)

Click this link to review: .../history?filters[body]=true&filters[dataclip_name]=true&filters[date_after]=2026-09-14+17%3A55%3A36.587141Z&filters[date_before]=2026-09-15+17%3A55%3A36.587141Z&filters[id]=true&filters[log]=true&filters[wo_date_after]=&filters[wo_date_before]=&filters[workflow_id]=4a0af176-...
```

**After** — labels say where they go, and the URL carries only what it filters on:

```
Workflow A:
• 12 workorders were successful today
• 3 workorders were not (crashed, exception, failed, killed, lost, rejected etc.)

See these runs in history: .../history?filters[date_after]=2026-09-14+17%3A55%3A36.587141Z&filters[date_before]=2026-09-15+17%3A55%3A36.587141Z&filters[log]=true&filters[workflow_id]=4a0af176-...

See this workflow's failures grouped by error: .../w/4a0af176-.../health
```

**A workflow with no failures keeps only the history link.** Nothing to diagnose, no second URL.

```
Workflow B:
• 10 workorders were successful today
• 0 workorders were not (crashed, exception, failed, killed, lost, rejected etc.)

See these runs in history: .../history?filters[date_after]=...
```

### Failure alert

The only HTML email, so this one is an actual anchor. Added under the existing run and work order links:

> See [this workflow's health](#) for failures grouped by error.

### Why history stays

The history URL carries the **absolute** window the counts came from, so it stays accurate however long the mail sits unread. Health is **relative** (`Stats` takes `days_back`, the page defaults to 30), so it cannot answer for a past window. Hence: history for "show me those runs", health for "group these by error".

The health link is worded to promise only that, not the digest's own set — a daily digest saying "3 failures today" would otherwise land the reader on 30 days of them.

### Shorter history URLs

`to_uri_params/1` was spelling out all four `filters[...]=true` search-field flags and two empty `filters[wo_date_*]=` params on every link it built. The empty dates change nothing. The four flags drop to one: `log`, which is what a bare visit to history sets anyway.

| | chars |
| --- | --- |
| history, before | 369 |
| history, after | 256 |
| health | 113 |

This shortened the in-app dashboard links too, not just the email.

Closes CON-198

## Validation steps

1. **Tests.** `mix test test/lightning/accounts/user_notifier_test.exs test/lightning/failure_alert_test.exs test/lightning/workorders/search_params_test.exs test/lightning/digest_email_worker_test.exs test/lightning_web/live/workflow_live/index_test.exs test/lightning_web/live/run_live`
2. **To actually read the emails**, start `iex -S mix phx.server`, send one from the console, open `/dev/mailbox`. This only works from inside the running server — `Swoosh.Adapters.Local` keeps mail in a GenServer in the sending node, so `mix run` sends into its own instance and the mailbox stays empty.
3. **Digest:** `Lightning.DigestEmailWorker.perform(%Oban.Job{args: %{"type" => "daily_project_digest"}})`. A workflow with failures should show both links; one without should show only history.
4. **Failure alert:** fail a workflow in a project where your user has alerts enabled. Check the health link renders as an anchor under the existing two.
5. **History links still work:** click the digest's history link, and the failed-count links on the project dashboard. Each should land filtered to that workflow and window, with the **log** search toggle lit and the other three dark. Then type a search term and hit Search — it should return matches, filtered to logs.

1. **Why one flag and not zero.** Dropping all four looks safe from the schema side — `SearchParams.new/1` reads a URL naming none of them as "search all four", same struct as before. But the raw params are also cast into `RunLive.Index.filters_changeset/1`, and the toggle chips are painted from `checked?/2` on that changeset, which only reads `{:changes, true}`. So a flagless URL renders four dark chips over a query that searches everything, and the checkbox values the form ships back are `false` — meaning the reader's first search from that page would have matched **nothing**. Emitting `log` alone makes both readings agree, keeps most of the length saving, and matches `init_filters/0`. The search fields are inert on arrival either way: `filter_by_body_or_log_or_id/3` short-circuits on a nil search term.
2. **The `Enum.any?` guard stays.** A link naming *some* of the four (say `id: false` alone) is making a choice, so the rest are still filled in with `true`. That path is unchanged.
3. **It exposed a latent test bug.** `build_digest_url/3`'s test ran `URI.decode_query` over the whole URL instead of the query string, so whichever param sorted first absorbed the path and went unchecked. `filters[body]` used to take that hit; with the flags gone `filters[date_after]` inherited it and failed. Now parses the query properly — which means all the params are genuinely asserted for the first time.
4. **No new authorization surface.** Both emails already gate on `MailRecipients.may_receive?/2`, so every recipient already holds `:access_project` — the same standing the health page requires. The link exposes nothing they could not already open.
5. **Deliberately not done: a `?days=` param on health.** A relative window drifts once the mail has sat for a day, so it would not make the link agree with the email's numbers. Doing it properly means absolute windows in `Stats.outcomes/2` and `Stats.error_signatures/2` — an API contract change, out of scope. The link's wording no longer implies it.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


lmac-1:
@midigofrank I tried to keep this as simple as possible for now. There's a lot to improve on these emails in follow-on work! I got as far as I could with my self review. Things that are coming back now feel more like nits. 

However, this work involves quite a bit of Elixir updates written by Claude and I can't be sure that they are all best practices. Feel free to edit or change anything you see fit. Thanks and hope that the PR is clear. I tried to make it as human-readable as possible.

midigofrank (approved):

PR #5191, opened 2026-09-17
Bucket the runs volume chart on the reader's clock, and click through it

midigofrank:
## Description

The grid was laid on the raw epoch, so a "daily" bar was a UTC day — 03:00 to 03:00 in Nairobi, 05:30 in Delhi — and the day a clock changed split across two bars. Buckets are now cut with date_bin on the local wall clock, in the zone the browser sends as an x-timezone header: a boundary lands on local midnight, London's 25-hour October day is one bar, and every axis and tooltip label is read in that same zone rather than in UTC. Bar widths are unchanged — 2-hourly over a day, half-days over a week, daily over a month — but they are now wall-clock hours, so a daily bar is 23 or 25 hours on the day the clocks move.

The response carries bucket_hours and timezone alongside the buckets. The chart used to measure its own bar width by subtracting the first two timestamps, which reads 25 on a clock-change day and mislabels the axis; the server cut the grid, so the server names the width and the clock.

The timezone joins the cache key, because it moves every boundary and Cachex keys an in-flight fetch by key alone — without it the first reader to load the page would pin their grid on everyone else's for the TTL. It is validated against the tz database before it gets there, since :workflow_stats has no size limit and an arbitrary header value would let one authenticated reader mint unbounded entries. An unknown or missing zone falls back to UTC rather than 400ing: a bad window cannot be drawn, but a bad zone still has a drawable answer.

Tzdata and pg_timezone_names are versioned apart — :tzdata autoupdates at runtime, Postgres' zone set is fixed at the server's build — so a zone the controller validated can still be one the binning query has never heard of. America/Coyhaique against PG 15 is exactly that, and it is what Chrome reports for readers in Coyhaique. The query rescues invalid_parameter_value and redraws the window on UTC rather than 500ing, and the result caches under the requested zone, so the failing query runs once per TTL rather than once per poll.

Configures Tzdata as the time zone database, without which every DateTime.shift_zone/2 by name returns :utc_only_time_zone_database.

### Click through the chart to the history page

Each band of each bar now links to the work orders behind it: the bar's slot, and the states that band stacked. The red band folds five run states, so it carries all five — a column-wide link would land the reader on the answer they came for mixed back in with the successes.

The history page had no run-level filter to link to. run_date_after, run_date_before and run_status bound a run's inserted_at and its state, matched with an EXISTS over the work order's runs. The existing status filter is the work order's own state, which is its latest run: a failure since retried to success is a success there, and the band that counted the failure still has to reach it. run_status takes final states only, since an unsettled run has no outcome to have been counted.

A bar's slot ends at the next bar's start rather than its own plus a width. The grid is cut on the reader's clock, so a bar spanning a clock change is 11 or 13 real hours wide, and a run landing on a boundary belongs to the same bar date_bin counted it in. The newest bar goes over open-ended — it is still filling. The filter chips read the range in UTC and say so: nothing in Lightning records what clock the bars were cut on, and a bare "14:00" under a bar labelled "16:00" is worse than a suffix.

Closes CON-188

## Validation steps

1. Open the workflow health page
2. The runs volume chart should be showing in your local timezone
3. Click a band of a bar — the history page opens filtered to that bar's window and that band's run states, with the range shown in UTC on the chips
4. Click the newest (right-most) bar — its filter is open-ended at the top, since that bucket is still filling

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

## Pre-submission checklist

      with Claude Code)


stuartc (approved):
Nice work dude. I made a few changes:

1. The bucketing now works out the edges with Tzdata, and then passes that to postgres. So we don't have to deal with the mismatch between tzdata and postgres. Since tzdata is updated all the time I picked that as the source of truth.
2. Moved the default timezone into one place (in the controller plug). No header means UTC, browsers can say "Etc/Unknown" and we make that UTC. Anything else is a 400; I might regret that hardening but yeah thought I'd rather see an error than quietly have users get UTC when they shouldn't.
3. Added `Vary: x-timezone` so that any caching layer we might end up having would know that it's part of the cache keys and it will consider it a different request.

